PR #6, opened 2019-02-18
simplify naming convention

dhellmann:
We don't expect to have a lot of competition in the namespace for
design documents, so let's just use the titles


dhellmann:
This change got lost in the host profile branch.

PR #8, opened 2019-02-20
add nodes-machines-and-hosts design document

dhellmann:


knowncitizen:
You can also do it this way, I think. @jtomasek @russellb ... etc

dhellmann:
> You can also do it this way, I think. @jtomasek @russellb ... etc

Yeah. I think not everyone is a member of the MetalKube org, yet, because some of the names aren't showing up here. Hence the ping on slack.

knowncitizen:
Fair :) Just trying to help!

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+- Discussion of user interface designs related to the resources.
+
+## Proposal
+
+The MetalKube project is building a hardware provisioning system with
+a Kubernetes-native API. It includes a provider implementation that
+fits into the API defined by the cluster-api SIG. It differs from
+other similar implementations by being self-hosted. Where the AWS or
+OpenStack providers run within Kubernetes, they allocate resources
+from a pool *outside* of the Kubernetes cluster. MetalKube will use
```
russellb: I may expand on "allocate resources from a pool outside of the Kubernetes cluster" to also clarify that this is done by talking to an API external to the cluster.  Maybe that's clear, but I want to hammer the point about the difference between the API being within Kubernetes or outside of it.

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+### Implementation Details/Notes/Constraints
+
+#### Data Model
+
+`Node` objects represent a running instance of kubelet. Their status
+fields include basic information about the health of the environment
+in which kubelet is running. A user manipulates the Node to change the
+cluster, such as draining the node for maintenance. The definition of
+Node is owned by the Kubernetes core, and offers few opportunities to
+attach custom data.
```
russellb: Let's link to definitive sources of definitions where we can find them, as well.  For Nodes, I see https://kubernetes.io/docs/concepts/architecture/nodes/

I'm not sure the first few sentences there really help, but it does look like a good resource for diving into a lot more detail on the API.

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+parallel design is the physical volume claim and physical volume in
+the storage APIs. The Machine CRD is owned by the cluster-api
+SIG. Machine objects are managed by the cluster-api controller using
+an *actuator* which works as a plugin driver for the
+controller. Machine objects include a "provider spec" field to allow
+the actuator to store custom data. The lifecycle of a Machine is based
+on the desired size of the cluster. That is, a Machine object does not
+typically exist in the database when it does not represent a request
+to expand the cluster (see caveats below in the
+[Alternatives](#alternatives-optional) section.
```
russellb: The cluster-api project publishes some docs here: https://kubernetes-sigs.github.io

This page is particularly relevant: https://kubernetes-sigs.github.io/cluster-api/common_code/machine_controller.html

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Growing the Cluster
+
+The admin user interacts with MetalKube by first registering hosts to
+be part of the inventory available to the provisioning system. Each
+host is then matched against known hardware profiles (comparing
+information such as NICs, storage, CPUs, RAM, etc.). Hosts matching
+known profiles are marked as being ready to be used. The admin (or
+autoscaler) can then create a new Machine object as a request to grow
+the cluster.
```
russellb: Or increments the count in a MachineSet, but maybe that detail isn't really important here ...

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+from the MetalKube project. The CRD is made up of completely custom
+data, needed to fully realize a Kubernetes-native API for a
+self-hosted baremetal provisioning tool. The lifecycle of a host is
+based on the availability of hardware, rather than its use as part of
+the cluster.
+
+#### Growing the Cluster
+
+The admin user interacts with MetalKube by first registering hosts to
+be part of the inventory available to the provisioning system. Each
```
russellb: Can we add something about discovery here?  The admin may manually create the host, or it may be created automatically as a result of the baremetal-operator automatically discovering a new host on the network.

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+host is then matched against known hardware profiles (comparing
+information such as NICs, storage, CPUs, RAM, etc.). Hosts matching
+known profiles are marked as being ready to be used. The admin (or
+autoscaler) can then create a new Machine object as a request to grow
+the cluster.
+
+The actuator in the
+[cluster-api-provider-baremetal](https://github.com/metalkube/cluster-api-provider-baremetal)
```
russellb: "in the cluster-api-provider" *repo* ?  component?

or remove "the" ?

Something reads weird for me here

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+
+`Node` objects represent a running instance of kubelet. Their status
+fields include basic information about the health of the environment
+in which kubelet is running. A user manipulates the Node to change the
+cluster, such as draining the node for maintenance. The definition of
+Node is owned by the Kubernetes core, and offers few opportunities to
+attach custom data.
+
+`Machine` objects represent a *request* for an instance of kubelet. A
+parallel design is the physical volume claim and physical volume in
```
lizsurette: 'persistent volume claim' and 'persistent volume' rather than 'physical'.

design/nodes-machines-and-hosts.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Version Skew Strategy
+
+N/A
+
+## Drawbacks [optional]
+
+N/A
+
+## Alternatives
```
russellb: Should we add a second alternative discussion covering what it would take to make the Machine object work for us?

To me, the key thing is that the Machine needs to be decoupled from Kubernetes.  A valid Machine would need to be able to represent a host that is intentionally off, or intentionally not being provisioned because of some policy reason (hardware profiles, the current desired size of the cluster, as examples).

We can link to this discussion: https://github.com/kubernetes-sigs/cluster-api/issues/721

That describes a direction that makes it less necessary to have BareMetalHost, but it also doesn't look like it's happening any time soon, and we're looking to implement this ASAP.

We should acknowledge that once we create this, it will be very, very hard to drop it.  We could further acknowledge that at some future point, the separate CRD may make less sense depending on the direction of the cluster-api project, but I also don't imagine it will be harmful to have it.

lizsurette: > Should we add a second alternative discussion covering what it would take to make the Machine object work for us?

Considering a lot of the feedback on the UX side so far, I'd love to see something here if it's possible. Even if we consider the designs towards something like this being future future.

design/nodes-machines-and-hosts.md
dhellmann: d'oh, thanks!

design/nodes-machines-and-hosts.md
dhellmann: I expanded that sentence to emphasize this point.

design/nodes-machines-and-hosts.md
dhellmann: Done.

design/nodes-machines-and-hosts.md
dhellmann: That first link is a 404, but I did link to the second page which works fine.

design/nodes-machines-and-hosts.md
dhellmann: That's a bit of a detail, but I went ahead and included it.

design/nodes-machines-and-hosts.md
dhellmann: Done

design/nodes-machines-and-hosts.md
dhellmann: Yeah, I left out "repo". Added.

design/nodes-machines-and-hosts.md
dhellmann: I added this. I'm not sure it would actually be that hard to deal with an upgrade. The new controller could just import the existing host objects into Machines and then delete them, right? I described the case without getting into the upgrade question.

dhellmann:
I think I have addressed all outstanding comments in the "address comments" patch, which I will squash into the base patch before we merge this.

russellb:
LGTM.  Feel free to squash and merge.

I feel this accurately represents the current state, so we should merge it as is.  Any follow-up proposals should be handled through subsequent PRs.

PR #10, opened 2019-02-22
Iterate on ironic doc

dhellmann:


design/how-ironic-works.md
```diff
@@ -40,56 +40,40 @@ have to become an expert in Ironic in order to learn to leverage it.
...
-available with vendor neutral IPMI. Newer protocols such as Redfish
+available with the vendor neutral IPMI driver. Newer protocols such as Redfish
 may support virtual media, but not as of the time of this document
-having been composed. Use of virtual media should be considered
-useful, but not well enough supported at this time to be considered useful.
+having been composed. Virtual media should be considered
+not well enough supported at this time to be considered useful.
 Other methods such as booting directly from iSCSI should be considered
-out-of-scope in this use case as they are centered around an external block
+out-of-scope in this use case as they are require an external block
```
juliakreger: Actually, you can tell ironic to make static configurations to a known iscsi san..... but require works.

design/how-ironic-works.md
```diff
@@ -120,22 +104,30 @@ will deploy successfully without issues.
 
 ## What connectivity is required?
 
-To boot the discovery and deployment image on the baremetal hardware:
+Access to the BMC can be via a routed IP network, however this may be
+less desirable than having it on the same L2 network as Ironic from a
+security standpoint.
```
juliakreger: ++

design/how-ironic-works.md
```diff
@@ -120,22 +104,30 @@ will deploy successfully without issues.
...
 * HTTP in order to download kernel/ramdisk images via HTTP over a TCP
   connection.
 
-The discovery and deployment ramdisk image needs to be able to:
+The discovery and deployment ramdisk image needs to:
 
-* DHCP
-* Resolve DNS
+* DHCP (via the ironic host, for IP assignment and PXE instructions)
+* Resolve DNS (FIXME - also via the ironic host?)
```
juliakreger: It doesn't need to hand out any DNS address, it can point to another DNS server. Dnsmasq is not exactly a fan of doing this, but it can be done.

design/how-ironic-works.md
dhellmann: I was trying to emphasize that it needs something that won't be provided by MetalKube or a typical kubernetes cluster.

design/how-ironic-works.md
dhellmann: If the network isn't routed, the only DNS server we'll see is the one on the ironic host, right?

russellb:
@juliakreger Can you ACK that you're ready for this to merge?

russellb:
@dhellmann want to squash and merge?

design/how-ironic-works.md
hardys: I don't think DNS is strictly required for the IPA ramdisk to work?  We just need to be able to tell it what the ironic endpoint is?

We do need some DNS records to exist for the OCP cluster to come up, but that's not really related to ironic?

Not blocking but we might want to drop the DNS references from this doc in a future commit.

dhellmann: IPA needs to be able to resolve the name in the URL for the image being deployed, right?

hardys: Only if you use a name in the URL, currently we just use an IP in dev-scripts and it works just fine

hardys: https://github.com/metalkube/dev-scripts/blob/master/07_deploy_masters.sh#L32 for example

dhellmann: OK, I guess if we already have the RHCOS image downloaded locally we can use the IP.

hardys: > OK, I guess if we already have the RHCOS image downloaded locally we can use the IP.

Yeah, my assumption is we'll want to either download or package the image so it's cached on the provisioning host/appliance, so any name resolution would only apply to the initial configuration of that host.

PR #12, opened 2019-02-27
Logo updates

maryshak1996:
Updated the logo with both SVG and PNG formats


maryshak1996:
@russellb all set with those black and white versions

russellb:
Thank you!

PR #13, opened 2019-03-01
Adding Ironic-inspector info

elfosardo:
Expanding the Ironic document adding some information on
ironic-inspector and the introspection process/rules.

### DIFF
--- a/design/how-ironic-works.md
+++ b/design/how-ironic-works.md
@@ -16,6 +16,7 @@
       * [How ironic boots hardware](#how-ironic-boots-hardware)
       * [How ironic writes an operating system image to baremetal?](#how-ironic-writes-an-operating-system-image-to-baremetal)
       * [What connectivity is required?](#what-connectivity-is-required)
+      * [What is Ironic Inspector?](#what-is-ironic-inspector)
       * [How-to](#how-to)
          * [How to discover hardware?](#how-to-discover-hardware)
          * [How to add hardware to ironic?](#how-to-add-hardware-to-ironic)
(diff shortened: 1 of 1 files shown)


dhellmann:
This is a good start, but by itself it isn't quite enough for me to write any code to interact with the inspector. Some of the ironic sections include details about the APIs to be called. Could you add a bit of that here for inspector, too? For example, what endpoints should I look at to trigger inspection and to get the results? How do I know when the inspection is done and the results are available?

design/how-ironic-works.md
```diff
@@ -146,6 +146,34 @@ Between ironic and ironic-inspector:
 * Each service must be able to reach the API endpoint for
   the the other service.
 
+## What is ironic-inspector?
+Ironic-inspector is an auxiliary service that provides separated API to
```
dtantsur: nit: s/separated/separate/

design/how-ironic-works.md
```diff
@@ -146,6 +146,34 @@ Between ironic and ironic-inspector:
...
+inspect and register hardware properties for a bare metal node.
+This process may be triggered during discovery and allow automatic node
+registration in the ironic DB.
```
dtantsur: nit: s/the ironic DB/ironic/ (DB is a detail here)

design/how-ironic-works.md
```diff
@@ -146,6 +146,34 @@ Between ironic and ironic-inspector:
...
+The process of "hardware introspection" allows also to get the hardware
+parameters from a bare metal node and ready it for scheduling.
+Power management credentials should be already set to allow node
+introspection.
+
+Ironic-inspector relies on plugins for data processing even for standard
+functionality; basic hardware introspection plugins are enabled by
+default.
```
dtantsur: Do we need these two paragraphs here? Seems a bit too detailed.

design/how-ironic-works.md
```diff
@@ -146,6 +146,34 @@ Between ironic and ironic-inspector:
...
+
+Ironic-inspector supports the creation of custom introspection rules to
+run during the introspection process.
+
+### What are introspection rules?
```
dtantsur: Are we planning to use introspection rules? If no, let's probably skip them?

elfosardo: I will leave this out for now, we can always add this later if we need it

design/how-ironic-works.md
elfosardo: Maybe you're right, I'll stick to an overview of the service without too many under the hood details that can be retrieved from the official docs.
We can always add this to a different section later.

elfosardo:
@dhellmann that makes sense, my doubt was more about where to add all the info. I've opted to include the API reference in the mini ironic-inspector section.

design/how-ironic-works.md
```diff
@@ -146,6 +147,66 @@ Between ironic and ironic-inspector:
...
+This process may be triggered during discovery and allow automatic node
+registration in ironic.
+The process of "hardware introspection" allows also to get the hardware
+parameters from a bare metal node and ready it for scheduling.
+Power management credentials should be already set to allow node
+introspection.
+
+To start the introspection process using ironic-inspector API
+
+    POST /v1/introspection/node-id
```
dhellmann: Does the body need any information, or is the URL enough to start the process? I assume the node needs to be registered in ironic already so they inspector will fetch anything it needs from there?

elfosardo: I expanded the description trying to give more info on the status of the node in respect to ironic and added information on the body in request/response.

dhellmann: Thanks!

dhellmann:
I'm happy with this version. I'll leave it open for a bit for @dtantsur to have a chance to review it before merging.

design/how-ironic-works.md
```diff
@@ -146,6 +147,71 @@ Between ironic and ironic-inspector:
 * Each service must be able to reach the API endpoint for
   the the other service.
 
+## What is ironic-inspector?
+Ironic-inspector is an auxiliary service that provides separate API to
+inspect and register hardware properties for a bare metal node (node
+introspection process).
+Ideally the node uuid must be already known by ironic, that means the node
```
dtantsur: nit: UUID

PR #15, opened 2019-03-04
add design for handling addresses for BMCs

dhellmann:

### DIFF
--- a/design/bmc-address.md
+++ b/design/bmc-address.md
@@ -1,0 +1,132 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# bmc-address
+
+## Status
+
+implementable
+
+## Table of Contents
+
+<!--ts-->
+   * [bmc-address](#bmc-address)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Alternatives](#alternatives)
+      * [References](#references)
+
+<!-- Added by: dhellmann, at: 2019-03-04T09:33-0500 -->
+
+<!--te-->
+
+## Summary
+
+This document explains the way users will provide the network location
+of the bare metal management controller (BMC) on a host.
+
+## Motivation
+
+We need to document how we are going to specify the location of a BMC,
+including how to tell its type, IP address, port number, and
+potentially the path.
+
+### Goals
+
+1. To agree on an address specification system, including assumptions
+   that can be made when parts of an address are left out.
+
+### Non-Goals
+
+1. To list the scheme to use for every type of controller.
+2. To specify the user interface for entering address information.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+For each BMC, we need to know the type, IP, port, and optionally the
+path to use to communicate with the controller.
+
+We can collect all of this information using a single "address" field,
+if we allow that field to contain partial or full URLs.
+
+For each type of controller, we can often assume we know the protocol
+used to communicate with it (HTTP, HTTPS, etc.). Therefore the scheme
+in a URL provided by the user is redundant, and we can use that
+portion of the URL to specify the controller type. For example:
+
+    ipmi://192.168.111.1
+
+In cases where we cannot assume the correct communication protocol, we
+will need to combine the type and protocol. For example:
+
+    redfish+https://IP/redfish/v1/Systems/42
+
+Initially, we will only support IPMI controllers, so we do not need
+users to specify the type or protocol.  If the field only contains a
+network address, we can assume that the controller uses IPMI on the
+standard port, 623. Therefore this would be equivalent to the previous
+example:
+
+    192.168.111.1
+
+### Risks and Mitigations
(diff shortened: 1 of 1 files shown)


design/bmc-address.md
```diff
@@ -0,0 +1,126 @@
...
+
+We can collect all of this information using a single "address" field,
+if we allow that field to contain partial or full URLs.
+
+For each type of controller, we can assume we know the URL scheme used
+to communicate with it (HTTP, HTTPS, etc.). Therefore the scheme in a
+URL provided by the user is redundant, and we can use that portion of
+the URL to specify the controller type. For example:
+
+    ipmi://192.168.111.1
```
dtantsur: This looks nice and consistent. I'm a bit worried about splitting the redfish URL. The unfortunate bit of redfish is that it takes two parameters. E.g. `https://192.168.111.1/redfish/v1/Systems/42` is split into:
```
redfish_address = https://192.168.111.1/redfish/v1
redfish_system_id = /redfish/v1/Systems/42
```
I'm not sure what to do about it. Probably just look for `/redfish/v1` bit, but it's not guaranteed to be there.. I'd like to summon @etingof to comment on it.

dhellmann: This is definitely not something we want to require users to do. Is that prefix consistent? Even if there are a few different ones, we could encode all of them. Is "redfish" a "type" in the sense that IPMI is a type here? If so, I think we'd want users to enter something like redfish://IP/redfish/v1/Systems/42 and then we would turn that into the values Ironic needs.

dtantsur: I think it's not unreasonable to expect this, but I'm not sure that the Redfish standard actually dictates the `/redfish/v1/` bit. We can probably ignore it for now (like we ignore IPMI bridging and other craziness).

design/bmc-address.md
```diff
@@ -0,0 +1,126 @@
...
+      * [References](#references)
+
+<!-- Added by: dhellmann, at: 2019-03-04T09:33-0500 -->
+
+<!--te-->
+
+## Summary
+
+This document explains the way users will provide the network location
+of the management controller on a host.
```
etingof: nit: bare metal management controller (BMC)

design/bmc-address.md
etingof: Redfish standard calls `/redfish` part as the *root service*, the `v1` part - *protocol version*. The standard does not dictate BMC to expose the root service and the protocol version in exactly `/redfish/v1` form.

I think `redfish://IP/redfish/v1/Systems/42` form should suffice, though may be I'd also include the original schema because https is not strictly required. For example, can it be `redfish+https://host/redfish/v1/Systems/42'?

design/bmc-address.md
```diff
@@ -0,0 +1,126 @@
...
+### Version Skew Strategy
+
+N/A
+
+## Alternatives
+
+The primary alternative is to expand the data structure to have
+separate fields for each value that would go into the URL. This
+complicates the UI, even for the simplest cases, since we have either
+have to show all of the fields all of the time or include logic in the
```
andybraren: Nit: Remove the first "have" in "since we have either have to show..."

design/bmc-address.md
```diff
@@ -0,0 +1,126 @@
...
+the URL to specify the controller type. For example:
+
+    ipmi://192.168.111.1
+
+Initially, we will only support IPMI controllers, so we do not need
+users to specify the type.  If the field only contains a network
+address, we can assume that the controller uses IPMI on the standard
+port, 623. Therefore this would be equivalent to the previous example:
+
+    192.168.111.1
```
andybraren: This assumption makes sense to me, and requiring the user to specify a `type` in the future when more controllers are supported (either via the same field or a dropdown in the UI) seems natural.

Do BMCs typically expose multiple controller types at once? Are certain types better than others? I assume so, but if not, would we be able to make our backend logic even smarter and automatically check across multiple controller types to avoid asking the user to specify the `type` in addition to the IP? Are there risks to doing that, like the user getting blocked by the BMC?

dhellmann: It seems likely that a given controller would support a standard and vendor-specific interfaces. I think for our purposes, though, we would pick one and tell the user to always select that one based on the test matrix and what we find most reliable. I don't know if we have any way of detecting the type, since we need the user to tell us where it is and that location may be more complex than just an IP and port (see the redfish path discussion in other comments), but perhaps one of the folks more familiar with Ironic can give us more info.

dtantsur: Yes, one controller can support several protocols (usually on different ports). Also worth knowing that one controller can support several nodes (hence the problem with Redfish).

etingof: > select that one based on the test matrix and what we find most reliable

May be the other criterion could be based on the feature set required.

>  if we have any way of detecting the type

We could perhaps do a port scan. For IPMI I recall something like that being in TripleO.

Also, I think the MaaS project has quite developed BMC discovery capabilities.

design/bmc-address.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Risks and Mitigations
+
+One risk in this approach is that we would need to tell users how to
```
lizsurette: Is the concern for risk when we get to the UI piece? We could have a smart text input that prepends the box with "ipmi://" and appends ":623" at the end allowing the user to simply provide the BMC IP. Eventually, it might make sense to all the user to build that out more freely as we support more types and different ports?

dhellmann: I was concerned with the user thinking "URL" and putting "https://" when they should put IPMI, so having the text field stick in the extra parts might make sense. It won't need to do that for the backend, though (we have to protect against bad data coming in through the API anyway), so only implement it if you think it helps with usability in the UI.

design/bmc-address.md
dhellmann: I added the redfish example, thanks for pointing that out as a special case.

design/bmc-address.md
dhellmann: Done

design/bmc-address.md
dhellmann: Done

design/bmc-address.md
dhellmann: I don't think we want the user having to choose which driver to use for a given host. We're going to have a limited set of hardware to support, so we should be able to tell them the best setting to use in each case.

dhellmann: Discovery of hosts is something we've put off for future releases, but if we can identify the right BMC that would be a good addition at that point.

dhellmann:
I've updated the draft to address all of your comments so far.

PR #16, opened 2019-03-08
add image-ownership design document

dhellmann:
Write down some thoughts about how we need to manage the versions of
the provisioning and target images.

### DIFF
--- a/design/image-ownership.md
+++ b/design/image-ownership.md
@@ -1,0 +1,154 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# image-ownership
+
+## Status
+
+implementable
+
+## Table of Contents
+
+<!--ts-->
+   * [image-ownership](#image-ownership)
+      * [Status](#status)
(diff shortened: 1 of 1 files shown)


dhellmann:
> LGTM. I'm not sure I agree with lacking configuration, but since we're focusing on fairly specific slices of supported hardware, we'll just have to wade through issues like hardware y requires package z in the ramdisk later.

Yeah, if that comes up we can certainly look at making it configurable but to start I think we want to shoot for an all-in-one provisioning agent rather than asking users to build their own.

design/image-ownership.md
```diff
@@ -0,0 +1,154 @@
...
+do not need to make it configurable. We can either build the name and
+version into the source code for the operator, or the operator can use
+a versionless name/URL when passing the data to Ironic and we can have
+our build system install the image into the container using that same
+name. The latter should make updating Ironic simpler over time, but
+may require extra work in the short term that we would not prioritize
+highly.
+
+The version of the image being written to the host (the "target
+image") will change with each update of OpenShift, and may ultimiately
```
jmolmo: typo: s/ultimiately/ultimately

dhellmann: Thanks, fixed.

russellb:
lgtm

PR #17, opened 2019-03-11
MetalKube Facet logo variation

maryshak1996:
Added MetalKube Facet logos with both black and white text in PNG and SVG formats


knowncitizen:
A few things - can you open this against https://github.com/metalkube/facet instead of here? Also, I thought you were going to use Rubik font? Lastly, it needs to be white text due to the black patternfly header.

maryshak1996:
@knowncitizen sure, I can open the PR against that instead. The font used for "Facet" is Rubik, and the MetalKube font is Open Sans - a thinner font to reduce overall weight of the logotype. There should be a white-text logo in there as well :) 
I'll open that PR tomorrow morning for ya

PR #24, opened 2019-04-12
Update slack channel in README.md

longkb:
It's seem that we already change our home from #cluster-api to #cluster-api-baremetal. This PR aims to update slack channel to the new one.

Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>


hardys:
Thanks!

PR #26, opened 2019-04-15
Adding design for hardware devices info collection

elfosardo:
Adding design to implement automated and manual hardware devices info collection, including health data.

### DIFF
--- a/design/hardware-status.md
+++ b/design/hardware-status.md
@@ -1,0 +1,159 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# hardware-status
+
+## Status
+
+provisional
+
+## Table of Contents
+
+   * [Title](#title)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints-optional)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [References](#references)
+
+## Summary
+
+This document explains how information about hardware components (drives, power
+supplies, fans, etc.) including health data, will be retrieved, collected and
+monitored to provide constant up-to-date status on hardware devices present in
+a server.
+The base idea behind this is having a service running in each active host that
+will regularly collect the data and post it to ironic-inspector, from where it
+can be consumed as part of the regular polling for status.
+
+## Motivation
+
+We need to be able to collect and update enough information about hardware
+devices in a server to monitor their status and identify them for
+maintenance and replacement.
+Valuable information are for example the current status, type, serial number,
+physical location, part number.
+We need to be able to refresh the information about the status of the hardware
+devices and, if a device has been replaced, we need to be able to retrieve the
+new information and update the hardware inventory to provide a constant
+up-to-date vision on the status of the devices.
+
+### Goals
+
+1. Decide which information on the devices to retrieve.
+2. Agree on how to retrieve such information.
+3. Provide a way to manually/automatically collect and update the devices
+information.
+
+### Non-Goals
+
+1. Any update on device data should not trigger any reaction from ironic.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+- Depending on the device, we need to collect the following information, when
+applicable:
+    * Capacity
+    * Location
+    * Manufacturer
+    * Model
+    * Part Number
+    * Serial Number
+    * Status
+    * SMART details
+
+- NVMe drives require special tools to get SMART data (e.g. nvme-cli),
+although [smartmontools](https://www.smartmontools.org/wiki/NVMe_Support)
+supports NVMe devices since version 6.5.
+
+- [Hdparm](https://en.wikipedia.org/wiki/Hdparm) can provide SMART data for
+most of drives. It doesn't work with NVMe devices.
+
+- [python-hardware](https://github.com/redhat-cip/hardware) can collect SMART
+data, although it requires code changes to detect and provide info on NVMe
+devices, for example relying directly on nvme tools instead of smartmontools.
+
+- [Ironic Python Agent (IPA)](https://github.com/openstack/ironic-python-agent)
+is already able to collect most of this information.
+Using a subset of the IPA code running inside a container in the hosts where we
+want to collect the drives data from can also allow to provide periodic
+updates on all the required info.
+The container may require privileged permissions to be able to access and
+collect all the needed information.
+The new containerized service will talk to Ironic Inspector.
+
+- [Ironic Inspector](https://docs.openstack.org/ironic-inspector/latest/)
+needs to be able to publish up-to-date data at any time, including from active
+and available nodes.
+This feature requires changes in the inspector code that currently accept data
+updates only from nodes in manageable state.
+
+### Risks and Mitigations
+
+1. Upstream Ironic community may reject changes in Ironic Inspector code.
+
+## Design Details
+
+### Work Items
+
+- Add option to detect and provide info on NVMe drives to python-hardware.
+- Extract and modify part of the IPA code to be able to run it in a container
+as a new service.
+- Modify Ironic Inspector code to accept data from active nodes.
+- Package (containerize) the new service based on IPA and the necessary tools,
+specifically for NVMe compatibility.
+- Distribute the new service to the hosts.
+  This point requires to take decisions on different operations and might
+  need a separate discussion on possible approaches, and includes at least the
+  following aspects:
+   - Ensure correct deployment of the new service to all the members of a
+   cluster.
+   - Correct configuration of the service, that needs to be aware of the
+   ironic inspector api.
+   - Verify the service is correctly running and regularly reporting up-to-date
+   data to ironic inspector.
+
+   One possible approach would be building a new element (as part of this or a
+   different design) that is able to both coordinate deployment and
+   configuration using current components, such as the baremetal-operator.
+
+### Dependencies
+
+* python-hardware
+* SMART tools
+* NVMe utils
+
+### Test Plan
+
+TBD
+
+### Upgrade / Downgrade Strategy
+
+N/A
+
+### Version Skew Strategy
+
+N/A
+
+## References
+
+- [Ironic Python Agent Documentation](https://docs.openstack.org/ironic-python-agent/latest/)
+
+- [Ironic Inspector Documentation](https://docs.openstack.org/ironic-inspector/latest/)


design/drive-status.md
```diff
@@ -0,0 +1,129 @@
...
+data
+
+[Ironic Python Agent (IPA)](https://github.com/openstack/ironic-python-agent)
+is already able to collect most of this information.
+Using a subset of the IPA code running inside a container in the hosts where we
+want to collect the drives data from can also allow to provide periodic
+updates on all the required info.
+This new service will talk to inspector that needs to be able to publish
+up-to-date data at any time.
+
```
dtantsur: I think we should talk about enabling ironic-inspector to accept data for active and available nodes

design/drive-status.md
```diff
@@ -0,0 +1,129 @@
...
+    * Manufacturer
+    * Model
+    * Part Number
+    * Serial Number
+    * Status
+    * SMART details
+    
+- NVMe drives require special tools to get SMART data, although
+[smartmontools](https://www.smartmontools.org/wiki/NVMe_Support) supports NVMe
+devices since version 6.5
```
dtantsur: Do we need to update python-hardware and/or IPA to support it?

design/drive-status.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Non-Goals
+
+N/A
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+- For each drive we need to collect at least the following information:
```
dtantsur: This is a part of "Goals", not "Proposal".

design/drive-status.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Goals
+
+1. Decide which information on the drives to retrieve
+2. Agree on how to retrieve such information
+3. Provide a way to manually/automatically update the information
+
+### Non-Goals
+
+N/A
```
dtantsur: Non-goal: provide any reaction on data changes inside of ironic

design/drive-status.md
```diff
@@ -0,0 +1,129 @@
...
+### Risks and Mitigations
+
+N/A
+
+## Design Details
+
+### Work Items
+
+- Extract and modify part of the IPA code to be able to run it in a container
+as a new service.
```
dtantsur: Missing item: modify ironic-inspector to accept data for active nodes

design/drive-status.md
dhellmann: Yes, from our earlier discussion that was a key prerequisite for any of this working.

design/drive-status.md
```diff
@@ -0,0 +1,129 @@
...
+is already able to collect most of this information.
+Using a subset of the IPA code running inside a container in the hosts where we
+want to collect the drives data from can also allow to provide periodic
+updates on all the required info.
+This new service will talk to inspector that needs to be able to publish
+up-to-date data at any time.
+
+### Risks and Mitigations
+
+N/A
```
dhellmann: There is some risk in having this approach rejected by the upstream Ironic team, isn't there?

design/drive-status.md
dhellmann: What sort of information can we get on libvirt VMs? We're still primarily testing with VMs

elfosardo: @dtantsur: python-hardware as it is now basically ignores nvme devices:
https://github.com/redhat-cip/hardware/blob/master/hardware/diskinfo.py#L33
I tested it and the disk are not reported by the hardware-detect command.
One solution could be to modify that module and maybe even add support for nvme tools (adding a new parser in this case).

@dhellmann: if you mean running smartmontools on vd* devices, I don't think that's possible as you can only use smartmontools on actual physical drives.
But maybe I got this wrong?

design/drive-status.md
```diff
@@ -0,0 +1,141 @@
...
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [References](#references)
+
+## Summary
+
+This document explains how information about hardware components (drives, power
+supplies, fans, etc.) including health data, will be retrieved, collected and
+monitored to provide constant up-to-date status on hardware devices present in
+a server.
```
dhellmann: For reviewers of this document who were not involved in the conference call where we talked about a design, it would be useful to have a brief description in this section. It doesn't need to go into all of the details, but should explain that we intend to run a service on each host in the cluster, that the service will collect data and post it to ironic-inspector, and that the baremetal operator will consume it as part of the regular polling for status that it is already performing.

design/drive-status.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Work Items
+
+- Add option to detect and provide info on NVMe drives to python-hardware.
+- Extract and modify part of the IPA code to be able to run it in a container
+as a new service.
+- Modify Ironic Inspector code to accept data from active nodes.
+- Package (containerize) the new service based on IPA and the necessary tools,
+specifically for NVMe compatibility.
+- Distribute the new service to the hosts.
```
dhellmann: I think we need to add more detail around this, because it's going to involve making some decisions about which component is responsible for ensuring that it happens.

I could see the baremetal operator setting up the DaemonSet to ensure the agent is deployed. However, the operator does not currently assume that it is deploying members of any cluster, or that members of the current cluster were deployed on baremetal.

The baremetal provider for the machine-api could set up the DaemonSet, but it does not currently know anything about the fact that we are using Ironic so it would be a little odd for it to do so.

It seems like we need something that knows it is managing a baremetal cluster using those other 2 components, and that it should set up monitoring on that cluster by creating the DaemonSet for the agent as one of the components.

design/drive-status.md
elfosardo: I will add a short paragraph explaining the base idea.

design/drive-status.md
elfosardo: I went ahead and started modifying python-hardware to get info on nvme disks, getting some first results on my laptop:
```
[('disk', 'logical', 'count', '1'),
 ('disk', 'nvme0n1', 'size', '512'),
 ('disk', 'nvme0n1', 'model', 'SAMSUNG MZVLB512HAJQ-000L7'),
 ('disk', 'nvme0n1', 'optimal_io_size', '0'),
 ('disk', 'nvme0n1', 'physical_block_size', '512'),
 ('disk', 'nvme0n1', 'rotational', '0'),
 ('disk', 'nvme0n1', 'nr_requests', '1023'),
 ('disk', 'nvme0n1', 'scheduler', 'none'),
 ('disk', 'nvme0n1', 'id', 'nvme-eui.0025388981b211a2'),
 ('disk', 'nvme0n1', 'id', 'nvme-SAMSUNG_MZVLB512HAJQ-000L7_S3TNNF0K939750'),
```

design/hardware-status.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Work Items
+
+- Add option to detect and provide info on NVMe drives to python-hardware.
+- Extract and modify part of the IPA code to be able to run it in a container
+as a new service.
+- Modify Ironic Inspector code to accept data from active nodes.
+- Package (containerize) the new service based on IPA and the necessary tools,
+specifically for NVMe compatibility.
+- Distribute the new service to the hosts.
```
dhellmann: I'm still looking for more detail about this. See my earlier comment.

juliakreger: We should keep in mind this would be a container with one of the special flags set to operate as a service on each container. I don't remember the way this is done off-hand but it is what is done for things like the existing system utilities like the node-exporter.

dhellmann: I think it needs to be part of a DaemonSet. There are several open questions about what owns managing that DaemonSet, though. We have similar questions about how Ironic is set up in the first place, which @sadasu has been working on.

design/hardware-status.md
```diff
@@ -0,0 +1,144 @@
...
+1. Upstream Ironic community may reject changes in Ironic Inspector code.
+
+## Design Details
+
+### Work Items
+
+- Add option to detect and provide info on NVMe drives to python-hardware.
+- Extract and modify part of the IPA code to be able to run it in a container
+as a new service.
+- Modify Ironic Inspector code to accept data from active nodes.
```
dhellmann: The baremetal operator uses gophercloud to get the inspection data. Will the operator need any of the new values being collected, or are all of those just going to go through the prometheus exporter as metrics? If the operator does need any of the new values, do we have to modify gophercloud to support them?

elfosardo: @dhellmann we won't collect new values, we're going to extend support for a new type of device, nvme, but it won't change the data collected, the fields will be part of the disk info.

design/hardware-status.md
elfosardo: As discussed with @dhellmann, and as he mentioned in his comment, there's already an effort to integrate different services on the nodes, and ensure that the pods are deployed, configured and run correctly.
I will add some details about the problem this point tries to solve and a possible approach.

design/hardware-status.md
dhellmann: @elfosardo do we already collect SMART data somewhere? I thought part of this was collecting runtime stats?

juliakreger: @dhellmann The operator will likely need to identify when new disks appear as that information will need to be updated in the UI. Regarding smart, nvme is different and is just going to end up being yet another type of storage device.

dhellmann: @juliakreger When you say "another type of storage device" I start to worry about things like whether the keys in the JSON payload coming from ironic-inspector are the same. Because if they're different, then we have to update gophercloud to look for those new keys explicitly in order to populate the data structure it defines and that the operator uses to identify something as a storage device.

juliakreger: @dhellmann That is a very valid concern because it is true, although I thought we left the high level structures for introspection data such that it was not on the low level keys.

Ultimately, @elfosardo needs time with python-hardware and an nvme storage device to provide a clearer picture.

dhellmann: OK. It would be best if we could use the same names for those keys, unless they're just "wrong" somehow.

elfosardo: @dhellmann currently python-hardware exports information on smart data using the `SMART` keyword. This is parsed by ironic-python-agent, jsonized and transmitted to ironic inspector.
The keys all have the `SMART` prefix, for example (taken from an actual device):

{
  "sda{megaraid,0}": {
    "SMART/product": "ST600MM0006",
    "SMART/start_stop_cycle_count": 739,
    "SMART/current_drive_temperature": 27,
    "SMART/blocks_sent": 2456935760,
    [.....]
    "SMART/health": "OK",
    "SMART/blocks_received": 2118230176,
    "SMART/non_medium_errors_count": 31,
    "SMART/specified_start_stop_cycle_count_over_lifetime": 10000,
 
This behaviour won't change for nvme devices.

design/hardware-status.md
```diff
@@ -0,0 +1,157 @@
...
+- [Hdparm](https://en.wikipedia.org/wiki/Hdparm) can provide SMART data for
+most of drives. It doesn't work with NVMe devices.
+
+- [python-hardware](https://github.com/redhat-cip/hardware) can collect SMART
+data, although it requires code changes to detect and provide info on NVMe
+devices, for example relying directly on nvme tools instead of smartmontools.
+
+- [Ironic Python Agent (IPA)](https://github.com/openstack/ironic-python-agent)
+is already able to collect most of this information.
+Using a subset of the IPA code running inside a container in the hosts where we
```
zaneb: Do we need to talk about what sort of privileges the container will have to run with to collect the information we need? (net=host at a minimum)

elfosardo: @zaneb we can mention that without going into too much details

elfosardo:
@dhellmann please let me know if there's anything more to add when you get a chance

design/hardware-status.md
dhellmann: I think we should add at least a little detail here to say that the host may need escalated privileges.

design/hardware-status.md
dhellmann: Sounds good.

dhellmann:
> @dhellmann please let me know if there's anything more to add when you get a chance

There are 2 comments open on the review that ask for more details, and I think we should answer both before proceeding.

design/drive-status.md
elfosardo: I think this was already addressed in a more recent commit.
Please re-open it if it's not the case.

design/hardware-status.md
elfosardo: I've added a separate line at L98 to add this detail

design/hardware-status.md
```diff
@@ -0,0 +1,159 @@
...
+
+# hardware-status
+
+## Status
+
+provisional
+
+## Table of Contents
+
+   * [Title](#title)
```
dhellmann: It looks like the inline stuff for gh-md-toc is gone and this may be out of date. We can fix that with a follow-up patch.

PR #29, opened 2019-04-24
Update the deprecated URL

longkb:
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>


russellb:
Good catch, thank you!

PR #37, opened 2019-06-12
add design for managing hardware service indicator lights

dhellmann:


dtantsur:
cc @etingof

design/service-indicator-lights.md
```diff
@@ -0,0 +1,201 @@
...
+<!-- Added by: dhellmann, at: Wed Jun 12 16:19:59 EDT 2019 -->
+
+<!--te-->
+
+## Summary
+
+Many hardware platforms have "indicator" lights in the chassis front
+(or rear) panel and sometimes have other lights associated with
+individual components like drives inside the chassis. We need to
+provide a way to turn those lights on when service is needed for a
```
etingof: The light can be turned ON by hardware itself to indicate a failure. In that case we may want to turn it OFF for unit identification purposes.

design/service-indicator-lights.md
```diff
@@ -0,0 +1,201 @@
...
+
+1. Polling too frequently will cause more than the usual level of
+   issues with the BMC or disk controllers. We mitigate these problems
+   by reducing the number of devices we actually poll, and reducing
+   the polling frequency.
+
+2. Because less polling is performed, the state of the CR is more
+   likely to become out of sync with the hardware if some other
+   component is also manipulating the indicator light. We can mitigate
+   the effect of that by ensuring that the code that turns the light
```
etingof: What is our goal here? To avoid turning ON the light if it's already ON? Would that actually increase the load on BMC (two operations might be required in some/most cases)?

design/service-indicator-lights.md
```diff
@@ -0,0 +1,201 @@
...
+The controller for the indicator CRD will use the same Ironic URL
+configuration settings and communicate directly with Ironic to ask
+about and manipulate the status of an indicator light.
+
+The controller for the indicator CRD will only communicate with Ironic
+when its `Spec.On` value does no match its `Status.On` value. As soon
+as that condition is met without an error, the controller will assume
+that the state is "sticky" and does not need to be changed. This may
+lead to the resource settings being out of sync with the hardware if
+the BMC is rebooted.
```
etingof: Perhaps many other reasons exist for BMC to actuate the LED.

design/service-indicator-lights.md
```diff
@@ -0,0 +1,201 @@
...
+We will define a new CRD `BareMetalHostServiceIndicator` (referred to
+as "indicator CRD" elsewhere in this document) with the following spec
+and status structures.
+
+```go
+type BareMetalHostServiceIndicatorSpec struct {
+	// HostRef is a reference to the bare metal host
+	HostRef corev1.ObjectReference `json:"hostRef"`
+
+	// Indicator ID
```
etingof: How do we initially discover these IDs? Currently, ironic uses indicator names as manifested by the Redfish agent.

design/service-indicator-lights.md
dhellmann: How would turning off the light help? Or do you just mean after the problem is resolved?

design/service-indicator-lights.md
dhellmann: To avoid asking about the state of the light more than necessary.

design/service-indicator-lights.md
dhellmann: There will be a list on the host object. We will either have the raw list as strings, or if we can map them to devices we can weave them into the existing data structures.

dhellmann: I added a few sentences to explain that at the start of this section.

design/service-indicator-lights.md
dhellmann: We could also just try to turn the light on/off as needed, and if we get an error then we could interpret that to decide whether it's really an error. Does the API give enough information for us to know that a light is already on if we try to turn it on? Or is the Ironic API itself already idempotent?

design/service-indicator-lights.md
etingof: No, I mean if we have many drives lighting up their LEDs at the same time to indicate failures. The operator wants to tell one failed drive from the other. The only (?) way to do that is to turn **off** the LEDs one by one to find out which drive is where.

design/service-indicator-lights.md
etingof: ironic API allows for GET'ing current LED state as well as PUT'ing the new desired state. It is not an error if the desired state is the same as the current one.

As of the current implementation, ironic does not have any intermediate logic or maintained  state when it comes to GET/PUT the indicators. Conceptually, ironic client query traverses ironic and hits the BMC.

design/service-indicator-lights.md
etingof: Possible instability of the enumeration bothers me....

Imagine we have many drives, Redfish BMC enumerates them somehow. At some point the operator replaces some drives (pulls old and plugs new). What will happen to the initial enumeration? I fear that is vendor-specific.

Arguably, the bullet-proof approach would be to "calibrate" the indicators meaning lighting up each of the discovered LEDs and manually assigning it to corresponding hardware unit...

design/service-indicator-lights.md
dhellmann: Ah, I see. Well, they will be able to do that by creating the resource and setting the flag to turn the light off.

design/service-indicator-lights.md
dhellmann: If ironic is not going to consider it an error to be told to turn on a light that is already on, we don't need to query first from the operator.

design/service-indicator-lights.md
dhellmann: Refreshing the known devices in a host is an orthogonal problem and there's a separate design to discuss that. I would expect the indicator list to be refreshed when the rest of the hardware is.

I don't think it's realistic to expect a user to configure the indicator lights in the way you describe, so we're going to have to do the best we can.

Do we have any idea what an indicator ID will look like for a drive? The examples in the spec at http://specs.openstack.org/openstack/ironic-specs/specs/not-implemented/expose-hardware-indicators.html don't include drives.

etingof: > Do we have any idea what an indicator ID will look like for a drive?

Ironic take on indicator naming is that it's up to the driver talking to the BMC (AKA *hardware type* in ironic parlance).

When it comes to Redfish, in the schema, the indicator LED resource does not carry any dedicated name. Thus, `redfish` driver in ironic [reuses](https://review.opendev.org/gitweb?p=openstack/ironic.git;f=ironic/drivers/modules/redfish/management.py;hb=refs/changes/40/652740/5#l350) parent resource's UUID for the indicator sitting on it. Well, for as long as it's a single LED per addressable unit.

dhellmann: @etingof if you could give me a real example, we could include it in this doc. Do we have hardware that has indicators that we can use to build examples?

design/service-indicator-lights.md
```diff
@@ -0,0 +1,209 @@
...
+Because of these 2 constraints, we do not want to add indicator
+management to the existing BareMetalHost controller or CRD. Instead,
+we will define a new CRD to represent a service indicator light that
+is in use.
+
+Only lights associated with an instance of the new CRD will be managed
+at all, and that management is limited to checking and toggling the
+state of the light when the resource in k8s changes. No active
+management is performed to ensure that state continues to match the
+resource.
```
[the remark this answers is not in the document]
dhellmann: Yeah, I'm not entirely comfortable with "never" either but I couldn't come up with a better answer to "how often is it OK to interrupt production I/O?".

design/service-indicator-lights.md
```diff
@@ -0,0 +1,209 @@
...
+operations. For this reason, Ironic does not regularly poll the status
+of indicator lights and does not cache their status when an outside
+user queries it.
+
+We do not want to disturb production data flow any more than
+necessary, so we want to limit the amount of polling we do related to
+indicator lights from metal3.
+
+Because of these 2 constraints, we do not want to add indicator
+management to the existing BareMetalHost controller or CRD. Instead,
```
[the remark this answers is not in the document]
dhellmann: The host CR is reconciled every minute. It doesn't know which of the indicator fields is different from the last time it was reconciled. By moving the indicator control to a different CR, we can have its controller behave differently so that it does not reconcile periodically and can assume if some change was made to the CR it should update the device.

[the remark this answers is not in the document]
zaneb: > It doesn't know which of the indicator fields is different from the last time it was reconciled.

Are you suggesting that we'll write the indicator status whenever we get an event, even if the Spec and the Status already match?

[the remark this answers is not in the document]
dhellmann: > > It doesn't know which of the indicator fields is different from the last time it was reconciled.
> 
> Are you suggesting that we'll write the indicator status whenever we get an event, even if the Spec and the Status already match?

I'm trying to do everything possible to avoid any calls related to the lights, especially on disks, so I don't want to ask for their status in the first place because that would involve too much interrogation of the host. And if we don't ask, we won't have the 2 values to compare against like we normally would.

The chassis and system indicators shouldn't be a big deal to manage more often, but for consistency it seems reasonable to just treat them all the same way.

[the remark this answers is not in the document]
zaneb: Right, but if we do:
```
  if Indicator.Spec.On != Indicator.Status.On {
      set_hardware_indicator(Indicator.Spec.On)
      Indicator.Status.On = Indicator.Spec.On
  }
```

then it doesn't matter how often that gets called.
Anyway, this is all a bit academic since we agree that there are other reasons to structure it like this :)

dhellmann:
Where do we stand on this? Do we need to add any more details?

PR #39, opened 2019-06-17
add design for managing IPs on secondary networks

dhellmann:


design/secondary-network-ipam.md
```diff
@@ -0,0 +1,175 @@
...
+## Alternatives
+
+We could require an external DHCP and IPAM solution for the secondary
+networks, as we do for the primary network. This complicates
+deployments and requires more services running outside of the cluster
+to know about implementation details of the cluster in order to have
+the external DHCP server pass PXE requests to the dnsmasq instance
+that is part of the metal3 deployment, and which might change hosts
+and IPs if the pod is restarted.
+
```
imain: Well, I guess we could do this, but I think it would require writing a new dhcp option backend for ironic and having some kind of agent on the dhcp server to update the configurations.  Currently dnsmasq and ironic need to have access to the same filesystem.

design/secondary-network-ipam.md
```diff
@@ -0,0 +1,175 @@
...
+We could monitor the DHCP reservations given by dnsmasq and ensure
+they are configured to be persistent, then also use those IPs to set
+static addresses on the hosts during provisioning. This would leave a
+reservation to be cleaned up when a host is removed, which might be
+tricky for a discovered host that is never actually provisioned.
+
```
imain: Or we set a single static IP on the provisioning host with the pod with a lifetime.. :)  This could be made more robust than the version I currently have set up.

design/secondary-network-ipam.md
dhellmann: Yeah, this is an alternative so the details won't matter as much because we aren't going to do it.

design/secondary-network-ipam.md
dhellmann: Yeah, I should link over to the other proposal here.

design/secondary-network-ipam.md
juliakreger: Ironic is working with static configuration in this case, and as long as it points back to a vip that ironic can be on, then the world is a happy place. If not, then... we'll need to be able to set configuration.

design/secondary-network-ipam.md
juliakreger: The downside with dhcp is if someone moves a cable, the port will still dhcp and the persistency is only as long as the dhcp server responds and has the same data set.

design/secondary-network-ipam.md
```diff
@@ -0,0 +1,181 @@
...
+### Work Items
+
+1. Ensure the IP ranges for secondary networks are captured by the
+   installer and saved to the kubernetes database as `Pool` resources.
+1. Ensure the installer registers the IP allocations for masters.
+1. Ensure the IPAM service is deployed along with the other metal3
+   components.
+1. Update the metal3 machine controller to allocate IPs and create
+   host-specific ignition configurations containing the IPs.
```
russellb: A few of these work items are OpenShift / CoreOS specific, the "installer" and "ignition" references.

The machine controller is not involved today in creating the ignition config (or cloud-init config, or whatever user data is in use), so maybe this should go somewhere else.

I expect management of secondary interfaces should be done by something else, like https://github.com/nmstate/kubernetes-nmstate

an IPAM component allocating addresses for secondary network interfaces could then create the CRs that specify that the interface should be configured with that IP, and applying the configuration would be done by kubernetes-nmstate

design/secondary-network-ipam.md
hardys: AFAIK ironic only manages the pxe configuration - the dnsmasq configuration is static and in a different container to the ironic conductor?

dtantsur: `the dnsmasq configuration is static` - yes, `in a different container to the ironic conductor` - yes (although built from ironic's Dockerfile for some reason). What ironic does modify is PXE configuration, but this is not related to the discussion.

imain: Right I see, I was confused.  It's really just the TFTP config we are updating.  So we could separate those if we wanted.

design/secondary-network-ipam.md
dhellmann: Yeah, I was trying to make sure I didn't miss anything. Should I move this doc to an internal location?

russellb: no need for it to be internal.  It just depends if we come up with something that's more generally useful, or is an OpenShift specific integration thing.  If it's just for OpenShift integration, the openshift-metal3 github org has a docs repo as well that hasn't been used much yet.

Speaking of OpenShift specific solutions and how to do network config ... I wonder if the existing MachineConfig resource is enough, provided by the machine-config-operator.  We can create those resources to drop new config files on nodes, or to replace existing config files.

PR #40, opened 2019-06-18
Proposal to add Helm chart for deploying Metal3 and Ironic

ogelbukh:
The proposal is to add Helm chart (or multiple charts) that will allow to deploy Metal3 and Ironic using Helm v2.

### DIFF
--- a/design/single-pod-helm-chart.md
+++ b/design/single-pod-helm-chart.md
@@ -1,0 +1,137 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Single-pod Helm chart
+
+## Status
+
+provisional
+
+## Table of Contents
+
+<!--ts-->
+   * [Title](#title)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [User Stories [optional]](#user-stories-optional)
+            * [Story 1](#story-1)
+            * [Story 2](#story-2)
+         * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Drawbacks [optional]](#drawbacks-optional)
+      * [Alternatives [optional]](#alternatives-optional)
+      * [References](#references)
+
+<!-- Added by: stack, at: 2019-02-15T11:41-05:00 -->
+
+<!--te-->
+
+## Summary
+
+Provide a Helm chart to deploy Metal3 and Ironic as a single
+pod in a Kubernetes cluster.
+
+## Motivation
+
+### Goals
+
+The goal is to support a popular way to deploy Kubernetes applications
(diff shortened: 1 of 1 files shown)


design/single-pod-helm-chart.md
```diff
@@ -0,0 +1,146 @@
...
+### Work Items
+
+1. Create a Helm chart for Ironic and its components:
+   * ``ironic``
+   * ``ironic-inspector``
+   * ``ironic-dnsmasq``
+   * ``ironic-httpd``
+   * ``mariadb``
+2. Create a Helm chart for Metal3 components:
+   * ``baremetal-operator``.
```
russellb: Minor thing, right now I think we expect Ironic to be in the same pod as baremetal-operator

ogelbukh: Thank you, Russel. This is a part of our attempt to evaluate scaling strategies for the future. Probably such separation, if needed at all, will make more sense a few iterations down the road.

design/single-pod-helm-chart.md
```diff
@@ -0,0 +1,146 @@
...
+
+Potentially, the Helm charts will require changes in the way the
+Docker images for components of Ironic and Metal3 are built. The
+changes will include additional parameters that will be exposed
+through the charts metadata (values.yaml files).
+
+### Work Items
+
+1. Create a Helm chart for Ironic and its components:
+   * ``ironic``
```
dtantsur: Ideally, you should split ironic-api and ironic-conductor since running two processes in one container is not the best practice.

design/single-pod-helm-chart.md
alanmeadows: I agree with @russellb that the initial target should a single self-contained pod.  Scaling will also require looking at how well the operator itself deals with scaling (just in terms of being able to Reconcile a very large number of hosts) before determining the right architecture is splitting out and independently scaling ironic versus (just as an example) sharding across self-contained pods.

design/single-pod-helm-chart.md
alanmeadows: Agreed.   I think this is also an opportunity to standardize the logging practices across all single-process pods as another example of proper production behavior.  All pods should be logging to stdout/stderr.

russellb: It occurs to me that this point may not have been captured very well in any design docs, but a design goal we had discussed before building this was that we really wanted the baremetal management component to be ephemeral worker pods someday that we could scale out as necessary.

First we just need a self contained Ironic that we know can work without any real persistent storage.  The baremetal-operator needs to be able to re-populate Ironic with information as needed at any point.

Eventually thought, it would be nice to be able to spin up bare metal management pods to perform management tasks as needed.  That's a ways off probably ... I like the idea of just a single unit while we make things work at all.

design/single-pod-helm-chart.md
ogelbukh: Given that we will need multiple instances of Ironic to handle multiple group of nodes (for example, multiple racks), baremetal-operator will need to handle groups of BareMetalHost objects based on labels or something else. So it could be either one instance of baremetal-operator per group, or single instance managing all groups through multiple Ironic API endpoints. I have no immediate answer what it should be, and it seems to be part of another discussion anyways.

Self-contained pod should the current goal anyways. I will reflect this in the design.

russellb: and by all means feel free to experiment with alternative deployment models.  I don't think anything is particularly enforced, except baremetal-operator assumes the ironic API is reachable via localhost at the moment.  That's easily fixable, but if it's remote we'll have to put some proper security in place.

All of this aside, I think the important thing with this whole PR is "should the metal3-io project should host a repo where people collaborate on helm charts for metal3-io" and IMO, it's yes

ogelbukh:
@russellb @dhellmann please, note that the bootstrap repo was created here: https://github.com/Mirantis/metal3-helm-chart

Please, let me know if it needs to be improved before you could import it. Thanks!

russellb:
@ogelbukh I've invited you to be a member of the metal3-io github org.  I think that will let you transfer ownership of the repo to the metal3-io github org.

As for the contents of the REPO, I'd request adding a copy of CONTRIBUTING.md and DCO.

PR #41, opened 2019-06-20
design: Propose prototyping a physical network API.

russellb:
This commit includes a design document that describes the creation of
a prototype of a physical network API that would work as a complement
to the API for managing physical hosts.

### DIFF
--- a/design/physical-network-api-prototype.md
+++ b/design/physical-network-api-prototype.md
@@ -1,0 +1,184 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# prototype-a-physical-network-api
+
+## Status
+
+implementable
+
+## Table of Contents
+
+<!--ts-->
+   * [Title](#title)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [User Stories [optional]](#user-stories-optional)
+            * [Story 1](#story-1)
+            * [Story 2](#story-2)
+         * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Drawbacks [optional]](#drawbacks-optional)
+      * [Alternatives [optional]](#alternatives-optional)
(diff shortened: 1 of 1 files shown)


design/physical-network-api-prototype.md
```diff
@@ -0,0 +1,176 @@
...
+Networking](https://docs.ansible.com/ansible/latest/network/index.html) to
+perform device configuration.  Ansible includes modules for managing the
+configuration of many different network devices.  Creation of CRDs in this new
+API can use modeling from Ansible as inspiration, particularly if there is any
+vendor neutral modeling already done there.
+
+Another project that could be used as inspiration is the [Ansible Networking
+Neutron ML2 Driver](https://networking-ansible.readthedocs.io/en/latest/) which
+created a Python API to abstract generic switch configuration on top of Anisble
+networking modules.
```
zaneb: Note that the relevant part of this driver was split out into https://github.com/ansible-network/network-runner.

The capabilities exposed by the API at the moment are confined to creating and deleting VLANs, and assigning them to switch ports (either as access ports or using 802.1Q VLAN trunking). So far there's no L3 support (adding IP addresses to VLANs, setting up routes, configuring DHCP Proxying, &c.), although in principle anything that's supported by ansible networking should be easy enough to add.

design/physical-network-api-prototype.md
```diff
@@ -0,0 +1,176 @@
...
+
+The [operator-sdk](https://github.com/operator-framework/operator-sdk) project
+includes some support for Ansible operators.  This should be considered as one
+approach for re-use of Ansible networking modules.  Other approaches to Ansible
+networking re-use can be considered and discussed in the read-out from the
+prototype.
```
zaneb: What network-runner provides is a series of Ansible roles with cross-platform implementations, plus a Python library with an API that internally creates a playbook to run a role. So the two most likely options are probably:

* generate playbooks for each role and run them using an Ansible Operator; or
* add a gRPC API to network-runner and call it from a Golang Operator.

design/physical-network-api-prototype.md
```diff
@@ -0,0 +1,176 @@
...
+corresponding physical network changes that must be made, such as reconfiguring
+a ToR switch port.  If the provisioning of the physical host is managed through
+a Kubernetes API, it would be convenient to be able to reconfigure related
+network devices using a similar API.
+
+### Goals
+
+Produce a `physical-network-api` prototype, which includes the following:
+
+* A `Switch` CRD with one or more fields that corresponding to switch
```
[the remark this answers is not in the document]
russellb: Yeah, I was thinking of a trivial CRD just for prototype purposes.  I could add a note that the data model is likely to be revisited post-PoC.

design/physical-network-api-prototype.md
russellb: Thanks for the tip.  I'll add this to the doc as a reference.

design/physical-network-api-prototype.md
russellb: Thanks!  I'll incorporate the ideas.

design/physical-network-api-prototype.md
```diff
@@ -0,0 +1,184 @@
...
```
dhellmann: I think for this prototype approaching it as a Switch is fine. I think for future work it would be good to consider whether we can abstract that further to provide an API that focuses on connectivity, and not implementation details of switches.

PR #42, opened 2019-06-21
README.md: Expand on bare metal management description.

russellb:
The high level architecture diagram includes a "Bare Metal Management
Pods" box, but this was not described in its own section like other
boxes in the diagram.  Create a new diagram and add a bit more text,
in particular the important design goal of making Ironic an ephemeral
worker with state that can be safely rebuilt at any time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,10 +51,6 @@ following:
   provisioning actions as necessary to reach the desired state.  It will do so
   by managing and driving a set of underlying bare metal provisioning
   components.
-* The implementation will focus on using Ironic as its first implementation of
-  the Bare Metal Management Pods, but aims to keep this as an implementation
-  detail under the hood such that alternatives could be added in the future if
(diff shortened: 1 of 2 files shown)


markmc:
Is "The `BareMetalHost` controller must assume that Ironic can be restarted at any time" something that's covered in use-ironic.md?

I see 

```
We want to eliminate the need for a MySQL database. Ironic needs a database, but it is not the source of truth for what Hosts exist or what their status should be, so there is no need for the complexity of a MySQL database.
```

but maybe it needs to be expanded on.

Point is that maybe README.md should call out some design highlights and point to metal3-docs for more info, rather than being the best source of some design info

russellb:
Thanks, @markmc.  I agree that `use-ironic.md` is a better place for this.  I'll update this to move that design info down into the ironic doc.

PR #46, opened 2019-06-30
Create machine remediation request proposal

cynepco3hahue:
This proposal explains what it is `machine-remediation-request` controller and how you can use it to apply different remediation strategies on the unhealthy machine.

### DIFF
--- a/design/machine-remediation-request.md
+++ b/design/machine-remediation-request.md
@@ -1,0 +1,151 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# machine-remediation
+
+## Status
+
+in-progress
+
+## Table of Contents
+
+<!--ts-->
+* [machine-remediation](#machine-remediation)
+  * [Status](#status)
+  * [Table of Contents](#table-of-contents)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+  * [Proposal](#proposal)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      * [MachineRemediation spec API](#MachineRemediation-spec-api)
+      * [MachineRemediation status API](#MachineRemediation-status-api)
+    * [Risks and Mitigations](#risks-and-mitigations)
+  * [Design Details](#design-details)
+    * [Work Items](#work-items)
+    * [Dependencies](#dependencies)
+    * [Test Plan](#test-plan)
+    * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    * [Version Skew Strategy](#version-skew-strategy)
+  * [Alternatives](#alternatives)
+  * [References](#references)
+
+<!-- Added by: dhellmann, at: Wed Jun 26 16:08:57 EDT 2019 -->
+
+<!--te-->
+
+## Summary
+
+This document explains the implementation of the Machine Remediation Request controller and
+how it will work together with Machine Health Check controller to restore unhealthy machines.
+
+## Motivation
+
+To provide a mechanism to apply remediation strategy on an unhealthy node, that will be different by type and by a provider.
+
+### Goals
+
+1. To provide the interface that will make possible to apply different remediation strategy for each cloud or bare metal providers.
+2. To provide details regarding the progress of the remediation operation.
+
+### Non-Goals
+
+1. To provide a remediation strategy for each cloud or bare metal providers.
+2. To provide the integration with the **MachineHealthCheck** controller.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+It's common that high available environment should have a mechanism to restore unhealthy nodes and when delete and create a new node looks feasible solution for a cloud provider, in case of bare metal hosts with additional infrastructure configuration it looks like too destructive approach.
+To make possible to specify the remediation strategy for different failure cases and for nodes that run on different providers, this proposal introduces **MachineRemediation** controller that should monitor all **MachineRemediation** CR's and apply the remediation strategy according to details specified under the CR.
+When the **MachineRemediation** will apply the remediation strategy, one who is initiating it will be the **MachineHealthCheck** controller. It will monitor nodes, and for each unhealthy node will create **MachineRemediation** CR, if after the specified timeout the **MachineRemediation** will not have the `remediationState` equal to `Succeeded`, the **MachineHealthCheck** controller will notify about the remediation failure, and can try to use different type of the remediation strategy.
+
+#### MachineRemediation spec API
+
+This proposal introduces a new API type: **MachineRemediation**, that makes possible to create simple **MachineRemediation** object with the type of the remediation strategy for the specific machine.
+
+##### Example
+
+```yaml
+apiVersion: MachineRemediation.metal3.io/v1alpha1
+kind: MachineRemediation
+metadata:
+  name: mr-test
+  namespace: openshift-machine-api
+spec:
+  type: reboot
+  machineName: test-machine
+```
+
+#### MachineRemediation status API
+
+**MachineRemediation** status will show what the state the remediation operation has and the time when the remediation operation started.
+
+```yaml
+status:
+  state: InProgress
+  reason: The fencing operation still in progress
+  startTime: Thu, 20 Jun 2019 03:38:39 -0400
+  endTime: Thu, 20 Jun 2019 03:40:39 -0400
+```
+
+### Risks and Mitigations
+
+It can introduce some integration complexity between **MachineHealthCheck** and **MachineRemediation** controllers,
+that will make harder to debug and to test things.
+
+## Design Details
+
+### Work Items
+
+* Create new **MachineRemediation** CRD.
+* Write a new controller that will watch **MachineRemediation** CR's.
+* Create an operator to deploy the **MachineRemediation** controller.
+* Create integration with the OLM.
+
+### Dependencies
+
+N/A
+
+### Test Plan
+
+As basic tests, it will be enough to create **MachineRemediation** CR's with the reboot type and verify that the machine really rebooted. In the future, we will need to test integration between **MachineHealthCheck** and **MachineRemediation** controllers to verify that our final fencing solution works as expected.
+
+### Upgrade / Downgrade Strategy
+
+Should be managed by OLM.
+
+### Version Skew Strategy
+
+The **MachineRemediation** controller version should be the same as **MachineHealthCheck** version.
+
+## Alternatives
+
+### Implement the remediation strategies under the actuator
+
+We implement in the remediation strategy under the machine actuator. It's already watching Machine resources of course, and it already knows how to interact with whatever the local management API is that would carry out remediation.
+
+In this case, the reboot request could be reflected as an annotation on the Machine. Status would be a little trickier, but it's not unheard of to reflect status in an annotation. So you could have a reboot/request annotation and a reboot/last annotation for example, whose values might be timestamps.
+
+Without direct support in the machine API, this would not be as robust or clear as the **MachineRemediation**, but it would be simple to implement and would not require a new controller. It also seems like a good behavioral fit for the machine controller.
+
+### Implement remediation strategies under the MachineHealthCheck controller
+
+We can implement the remediation strategy under the **MachineHealthCheck** controller,
+in this case, we will have fewer controllers but also less granularity, because **MachineHealthCheck** works on a group of machines.
+
+### Implement remediation strategies under the machine-api-operator repository
+
+We can implement the remediation strategy under the `machine-api-operator` repository, as separate controllers, it not a bad idea in the long term, because we will not need to implement our own operator and integration with OLM.
+Also, it will be easier to provide components with the same version.
+But it does not look real to merge this stuff quickly under the `machine-api-operator` repository, but we can move it in the future.
+
+## References
+
+* [machine-health-check](https://github.com/openshift/machine-api-operator/tree/master/pkg/controller/machinehealthcheck)


design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# machine-remediation-request
```
mhrivnak: I'm conflicted about including the word "request". Creating a Machine resource for example is equally a "request", as is the case for most things in a declarative API. What do you think about just calling this "Machine Remediation"?

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+It can introduce some integration complexity between **MachineHealthCheck** and **MachineRemediationRequest** controllers,
+that will make harder to debug and to test things.
+
+## Design Details
+
+### Work Items
+
+* Create new **MachineRemediationRequest** CRD.
+* Write a new controller that will watch **MachineRemediationRequest** CR's.
+* Write functional tests under <https://github.com/openshift/cluster-api-actuator-pkg> repository.
```
dhellmann: This bit is openshift-specific, so we should track that work elsewhere.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+  machineName: test-machine
+```
+
+#### MachineRemediationRequest status API
+
+**MachineRemediationRequest** status will show what the state the remediation operation has and the time when the remediation operation started.
+
+```yaml
+status:
+  remediationState: InProgress
```
dhellmann: Having "remediation" in these names makes them feel a bit redundant. What do you think about State and StartTime?

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+
+```yaml
+apiVersion: machineremediationrequest.metal3.io/v1alpha1
+kind: MachineRemediationRequest
+metadata:
+  name: mrr-test
+  namespace: openshift-machine-api
+spec:
+  type: reboot
+  machineName: test-machine
```
dhellmann: Do we need this to be a full object reference? Or at least to include an optional namespace field?

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Design Details
+
+### Work Items
+
+* Create new **MachineRemediationRequest** CRD.
+* Write a new controller that will watch **MachineRemediationRequest** CR's.
+* Write functional tests under <https://github.com/openshift/cluster-api-actuator-pkg> repository.
+* Create an operator to deploy the **MachineRemediationRequest** controller.
+* Create integration with the OLM.
```
dhellmann: This step is also specific to openshift.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+<!--te-->
+
+## Summary
+
+This document explains the implementation of the Machine Remediation Request controller and
+how it will work together with Machine Health Check controller to restore unhealthy machines.
+
+## Motivation
+
+To provide a mechanism to apply remediation strategy on an unhealthy node, that will be different by type and by a provider.
```
mhrivnak: What do you mean by "type" versus "provider"?

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+### Implement remediation strategies under the MachineHealthCheck controller
+
+We can implement the remediation strategy under the **MachineHealthCheck** controller,
+in this case, we will have fewer controllers but also less granularity, because **MachineHealthCheck** works on a group of machines.
+
+### Implement remediation strategies under the machine-api-operator repository
+
+We can implement the remediation strategy under the `machine-api-operator` repository, as separate controllers, it not a bad idea in the long term, because we will not need to implement our own operator and integration with OLM.
+Also, it will be easier to provide components with the same version.
+But it does not look real to merge this stuff quickly under the `machine-api-operator` repository, but we can move it in the future.
```
dhellmann: It would also not make it easy for us to move this new API upstream.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Upgrade / Downgrade Strategy
+
+Should be managed by OLM.
+
+### Version Skew Strategy
+
+The **MachineRemediationRequest** controller version should be the same as **MachineHealthCheck** version.
+
+## Alternatives
```
mhrivnak: Another alternative worth capturing and evaluating here: this could be implemented in the machine actuator as part of the machine controller. It's already watching Machine resources of course, and it already knows how to interact with whatever the local management API is that would carry out remediation.

In this case, the reboot request could be reflected as an annotation on the Machine. Status would be a little trickier, but it's not unheard of to reflect status in an annotation. So you could have a `reboot/request` annotation and a `reboot/last` annotation for example, whose values might be timestamps.

Without direct support in the machine API, this would not be as robust or clear of a solution as your proposal. But it would be simple to implement and would not require a new controller. It also seems like a good behavioral fit for the machine controller.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+2. To provide details regarding the progress of the remediation operation.
+
+### Non-Goals
+
+1. To provide a remediation strategy for each cloud or bare metal providers.
+2. To provide the integration with the **MachineHealthCheck** controller.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
```
mhrivnak: How will the controller implement reboot? Particularly how will it do it for different providers?

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+
+It's common that high available environment should have a mechanism to restore unhealthy nodes and when delete and create a new node looks feasible solution for a cloud provider, in case of bare metal hosts with additional infrastructure configuration it looks like too destructive approach.
+To make possible to specify the remediation strategy for different failure cases and for nodes that run on different providers, this proposal introduces **MachineRemediationRequest** controller that should monitor all **MachineRemediationRequest** CR's and apply the remediation strategy according to details specified under the CR.
+When the **MachineRemediationRequest** will apply the remediation strategy, one who is initiating it will be the **MachineHealthCheck** controller. It will monitor nodes, and for each unhealthy node will create **MachineRemediationRequest** CR, if after the specified timeout the **MachineRemediationRequest** will not have the `remediationState` equal to `Succeeded`, the **MachineHealthCheck** controller will notify about the remediation failure, and can try to use different type of the remediation strategy.
```
mhrivnak: Where is the timeout specified? Rebooting bare metal will take orders of magnitude more time than rebooting a VM on most public clouds, so we need to be sure that's accounted for.

design/machine-remediation-request.md
beekhof: i dont think machines can live outside of the openshift-machine-api namespace

design/machine-remediation-request.md
beekhof: Possibly master vs. worker

design/machine-remediation-request.md
beekhof: IIUC, the only provider we will create a controller for initially is metal3.
Exactly how each controller "fixes" the affected machine is intended to be opaque to MachineHealthCheck.

This allows the metal3 flavor of the MachineRemediationRequest controller to look up and use Spec.Online and Status.PoweredOn from the BareMetalHost to implement reboot.

design/machine-remediation-request.md
beekhof: +1

design/machine-remediation-request.md
beekhof: We also need an `endTime`, probably a `reason` field (for additional information on failure), as well as a way to differentiate between "node is unhealthy but in a safe state to continue recovery" failed and a "we've lost any and all communication to / control over the node" failure.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+```yaml
+apiVersion: machineremediationrequest.metal3.io/v1alpha1
+kind: MachineRemediationRequest
+metadata:
+  name: mrr-test
+  namespace: openshift-machine-api
+spec:
+  type: reboot
+  machineName: test-machine
+```
```
beekhof: I'd like to see a timeout value that puts a cap on how long the controller has to try and make the node healthy again (possibly before taking more dramatic action).
Additionally, I suspect the controller would like to know which condition/timeout led to the creation of this object. Possibly in order to help decide on the correct recovery proceedure.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+
+As basic tests, it will be enough to create **MachineRemediationRequest** CR's with the reboot type and verify that the machine really rebooted. In the future, we will need to test integration between **MachineHealthCheck** and **MachineRemediationRequest** controllers to verify that our final fencing solution works as expected.
+
+### Upgrade / Downgrade Strategy
+
+Should be managed by OLM.
+
+### Version Skew Strategy
+
+The **MachineRemediationRequest** controller version should be the same as **MachineHealthCheck** version.
```
beekhof: I think this is going to be hard to achieve unless either the implementation is part of the official machine-api-operator codebase (unlikely) or we fork the machine-api-operator codebase (which there is a very strong push to avoid)

design/machine-remediation-request.md
beekhof: Actually more than the above, the current intention for 4.2 is to only create the CR if the bare metal host annotation is found on the Machine object.  We would then expand to other providers as appropriate in later versions.

design/machine-remediation-request.md
beekhof: This was one of the original implementation ideas, but there wasn't a clear way to tell the actuator whether reboot or deletion was required.  Applying the same signalling idea proposed here to this concept would work, however I suspect mixing the reboot state machine into the actuator will get messy

design/machine-remediation-request.md
mhrivnak: So each provider will need its own machine-remediation-controller implementation?

design/machine-remediation-request.md
mhrivnak: I wouldn't worry about the complexity of the machine controller. There's not a whole lot to it today, and adding reboot logic would not be a burden by any means. People are making far more complex controllers as a normal practice these days.

design/machine-remediation-request.md
```diff
@@ -0,0 +1,142 @@
...
+This document explains the implementation of the Machine Remediation Request controller and
+how it will work together with Machine Health Check controller to restore unhealthy machines.
+
+## Motivation
+
+To provide a mechanism to apply remediation strategy on an unhealthy node, that will be different by type and by a provider.
+
+### Goals
+
+1. To provide the interface that will make possible to apply different remediation strategy for each cloud or bare metal providers.
```
markmc: The way I see it, "remediation" is a machine health check concept, and the machine health check controller should be the one responsible for the remediation strategy. It seems very strange to me that each machine api provider would figure out its own remediation strategy.

The machine API should provide low-level actions for implementing a remediation strategy, and perhaps per-machine metadata to help it make better remediation choices.

i.e. something like

```
machinehealthcheck_controller:

  remediate() {
      if !machine.hasAnnotation("doNotRemediateWithReboot") && machine->reboot  != null {
          machine->reboot()
      } else {
          machine->delete()
      }   
  }
```

i.e. perhaps not all provider will implement reboot, and perhaps we can allow some indication to not try reboot even if the provider supports it ... but otherwise we should try reboot first

(Sorry for the pseudo-code - I know it omits a lot of complications about tracking progress and falling back to delete if reboot doesn't work)

But IMO ... the conclusion for metal3 is that the Machine API should have a reboot interface

cynepco3hahue: IMHO the one who is implementing actuator should be more aware regarding the needed action for the remediation, MHC should only monitor node conditions and apply the remediate logic in the case when it needed, but it should not be aware regarding the implementation of the remediation logic.
It similar to the cluster API and actuators relations, cluster API only applies that actuator logic.

IIRC the community has some proposals to introduce the reboot method and the actuator interface under the actuator interface, but without luck, because reboot and re-provision for the cloud provider is quite the same, also after the interface change you will need to change all controllers that use this interface

design/machine-remediation-request.md
cynepco3hahue: yep, it was the idea, for the short term we will implement only bare metal remediator, in general remediation can apply additional actions depends on a provider and the infrastructure configure on the node.

design/machine-remediation-request.md
cynepco3hahue: for the short term, we can set some default timeout from 5 to 10 minutes, but I think it pretty easy to configure the timeout for each host separately via machine(or node) annotation

design/machine-remediation-request.md
cynepco3hahue: `MachineRemediationRequest` should be created in the machine namespace, so the name should be enough to get the relevant machine object

design/machine-remediation-request.md
cynepco3hahue: I think the remediator the one who will be responsible to fail the operation on timeout, so the timeout value will be part of the MachineRemediate controller that can use some default value or get the value from the machine(node) annotation.
Yep, it a good idea to add a reason for the state to have a better explanation on the state.

design/machine-remediation-request.md
cynepco3hahue: done

design/machine-remediation-request.md
cynepco3hahue: done

design/machine-remediation-request.md
cynepco3hahue: You can deploy OLM on top of k8s cluster so I prefer to leave it here

design/machine-remediation-request.md
cynepco3hahue: I did not say it will be easy:) But if we will implement bare-metal-remediator machine-api operator can deploy, similar to other images.

design/machine-remediation-request.md
cynepco3hahue: I will add this as an option.

design/machine-remediation-request.md
beekhof: We're in basic agreement, the challenge has been that upstream has been slow moving (to the point that OpenShift forked it) and primarily focused on cloud use cases which are less likely to face the challenges of a hyper-converged bare metal environment.

While there may be more cause for optimism now under VMware's leadership, getting the API changes through and backported in time for OpenShift feature freeze is unrealistic IMHO. Originating changes in the OpenShift Machine API fork first is also not an option.

Hence the idea to have $SomethingElse that fills the same role in the short term, can be used to demonstrate value in upstream conversations, and deprecated in the longer term.

On the subject though, I see value in adding the following to Machine API:
- reboot()
- on()
- off(), if rebooting the node doesn't result in it becoming healthy again, then we'd want to turn it off and move on
- remediate(), an opportunity to hide even the details of how to return the node to health from MHC.

design/machine-remediation-request.md
dhellmann: That assertion about namespacing is enforced by OpenShift, but not by kubernetes or the upstream machine-api. I don't think it's something we want to back into this API if we're going to try to upstream it.

design/machine-remediation-request.md
dhellmann: Regardless of where we might store it, how does any component know a reasonable timeout for rebooting a host in the first place? It depends on the hardware, image, DHCP responses, and a host of other factors outside of the control of the remediation software.

design/machine-remediation-request.md
cynepco3hahue: I can change the namespace value to something not openshift specific, but I still think that this resource should be created under the machine namespace.

design/machine-remediation-request.md
cynepco3hahue: The one who will responsible to set annotation on the node with the timeout is cluster administrator so he should be aware regarding reasonable timeout for each node, in case if the annotation does not exist on the node, we will use the default of 5 minutes that looks like the reasonable one

dhellmann:
I think this is superseded by #48. Please re-open the PR if I'm wrong.

PR #48, opened 2019-07-25
Add a design doc for a declarative reboot interface

zaneb:
Design document for the proposal in the draft PR metal3-io/baremetal-operator#239.

### DIFF
--- a/design/reboot-interface.md
+++ b/design/reboot-interface.md
@@ -44,8 +44,8 @@ running processes are stopped to avoid a split-brain scenario while still
 allowing the node to rejoin the cluster with its data intact (albeit stale)
 after the reboot.
 
-The expected implementation of this on the fencing side is a Custom Resource
-Definition that requests a Machine be rebooted. The actual reboot will be
+The expected implementation of this on the fencing side is an annotation on the
+Machine resource that requests that it be remediated. The actual reboot will be
 effected either by the Machine Actuator itself (if the Cluster-API SIG can be
 persuaded that this should be part of the API in the long term), or some
 equivalent of the Machine Actuator limited to just this purpose.
@@ -154,18 +154,19 @@ The controller automatically removes all annotations with the
 
 ## Drawbacks
 
+This requires clients to be somewhat well-behaved - for example, only deleting
+their own annotations and never deleting others' annotations.
+
+It would be difficult to apply RBAC to the reboot operation specifically, since
+it is triggered by simply adding an annotation (that doesn't even have a fixed
+key).
+
 ## Future Enhancements
 
 ### Defining a Formal User Interface
 
-
-### Soft Reboots
-
-The resource could accept a "soft" reboot option, in which we would attempt to
-reboot the Host without switching the power off. Since these requests may fail,
-more error reporting would likely be required. In the event that a mix of hard
-and soft reboot requests were pending, the host would be hard-rebooted and all
-of the pending requests marked complete.
+Once we have experience with using this, we will have more information about
+how to design a formal interface for it (rather than using annotations).
 
 ## Alternatives
 
@@ -175,16 +176,24 @@ sequence to ``false`` and then back to ``true`` again once the ``poweredOn``
 status was seen to be ``false``. However, in the presence of multiple actors
 this approach is prone to race conditions or controller fights.
 
+We could create a ``HostRebootRequest`` CRD and have the existing host
+controller check for pending reboot requests, perform reboots when necessary,
+and update the state of the request. This would be easier to apply RBAC to, and
+indeed does not require clients to co-operate in any way. It would be a
+flexible jumping off point for adding more advanced features should they become
+required. However, it adds significant complexity (another CRD to deploy) and
+would be a permanent, formal interface. It's also not clear how outdated
+requests should be cleaned up.
+
 The request could simply be a timestamp field in the Host spec. The Host would
 coalesce multiple requests into one at the time they are received (rather than
 at the point of action). This would be simpler, but requires the client to
 provide a meaningful timestamp (i.e. in the recent past), forcing us to make
-awkward decisions about what to do with bogus data. In contrast, with a CRD we
-can use the creation timestamp provided by the system. This option also makes
-it more difficult to add further reboot-related API features in the future.
+awkward decisions about what to do with bogus data. In contrast, allowing
+multiple annotations allows the system to record the time at which it first
+noticed them. Selecting this option would also make it more difficult to add
+further reboot-related API features in the future.
 
 The request could be made by adding to a list of client-chosen strings in the
-Host spec. This would be again be simpler than adding a new CRD at the cost of
-future flexibility, but requires clients to co-operate (by only adding to the
-list, and never replacing it), so one misbehaving client could cause another to
-act erratically.
+Host spec. This effectively formalises the annotation-based system proposed
+here and makes it a permanent part of the interface.
--- a/design/reboot-interface.md
+++ b/design/reboot-interface.md
@@ -1,0 +1,193 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# reboot-interface
+
+## Status
+
+implementable
+
+## Table of Contents
+
+<!--ts-->
+   * [reboot-interface](#reboot-interface)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+      * [Drawbacks](#drawbacks)
+      * [Future Enhancements](#future-enhancements)
+      * [Alternatives](#alternatives)
+
+<!--te-->
+
+## Summary
+
+A declarative API is proposed to request the baremetal operator to reboot a
+provisioned Host in an atomic step (i.e. without the need to make multiple
+sequential changes to the spec).
+
+## Motivation
+
+We require a non-destructive way to fence a Kubernetes node. Some nodes cannot
+be replaced (it is currently not possible for a new master to join a cluster),
+or are expensive to replace (e.g. if this would require rebalancing Ceph data).
+A solution to this is for fencing to reboot the Host, thus ensuring that
+running processes are stopped to avoid a split-brain scenario while still
+allowing the node to rejoin the cluster with its data intact (albeit stale)
+after the reboot.
+
+The expected implementation of this on the fencing side is a Custom Resource
+Definition that requests a Machine be rebooted. The actual reboot will be
+effected either by the Machine Actuator itself (if the Cluster-API SIG can be
+persuaded that this should be part of the API in the long term), or some
+equivalent of the Machine Actuator limited to just this purpose.
+
+### Goals
+
+The Machine (Reboot) Actuator will require from the BareMetalHost:
+
+* A declarative API to perform the reboot
+* The ability to determine a point where all processes running on the Host at
+  the time of the fencing decision are guaranteed to have been stopped
+* Rapid convergence to a state where all running processes are stopped,
+  independent of other happenings in the system
+
+### Non-Goals
+
+There is no requirement to implement a scheduled reboot. In the Kubernetes
+context, reboot decisions should generally be made in realtime by some higher
+level of the Node/Machine hierarchy, to take into account such questions as the
+overall health of the cluster and the effect of a reboot on that. The best
+implementation for this would be a RebootSchedule CRD that waits until the
+appointed time before issuing an immediate reboot request to the BareMetalHost.
+This allows multiple reboots to be scheduled, scheduled reboots to be
+manipulated or cancelled, and a record to be left behind of past scheduled
+reboots. The proposed design could easily be extended to accomodate this
+requirement should it arise in future.
+
+There is no requirement to hold the power off for any particular length of
+time. For fencing purposes we only need to be sure when the node is powered
+off.
+
+This API is not responsible for managing unprovisioned hosts, e.g. to recover
+from state inconsistencies in Ironic prior to provisioning. Any such
+inconsistencies that are not handled internally by the BareMetalHost will
+likely need manual intervention to recover anyway.
+
+## Proposal
+
+A new date-time field, ``lastPoweredOn``, will be added to the ``provisioning``
+section of the BareMetalHost status. This records a time after which the Host
+was last booted using the current image. Processes running prior to this time
+may be assumed to have been stopped.
+
+To provide a declarative interface for performing a reboot, a new
+BareMetalHostRebootRequest CRD will be created. The spec of this custom
+resource will contain only a reference to a BareMetalHost. The status will
+contain timestamps for when the Host was powered off and on again. Once the
+power off time is present in the status, the host has been stopped. Each
+RebootRequest will be owned by the Host it refers to, so that if the Host is
+deleted all associated RebootRequest CRs will be cleaned up.
+
+The actual power management will be performed by the Host controller. This is
+necessary to avoid race conditions by ensuring that the ``Online`` flag and any
+reboot requests are managed in the same place. The Host controller
+will watch RebootRequest objects and trigger a reconciliation of the linked
+Host when one appears or changes.
+
+For simplicity, we will try to avoid creating a separate controller for the
+RebootRequest at all, but instead perform all operations on it in the course of
+reconciling the BareMetalHost object that it references. However, it may yet
+prove necessary to also create a trivial controller for the purposes of basic
+housekeeping (e.g. setting the owner of the resource).
+
+The creation timestamp of the request resource will be interpreted as the time
+of the reboot request. Requests that predate the last booted time will be
+immediately marked as complete without taking any further action. The Host
+controller will timestamp all pending reboot requests as having completed the
+power off stage whenever any of the following occur:
+
+* The Host has been deprovisioned (in this case the ``lastPoweredOn`` field
+  will be empty)
+* The ``lastPoweredOn`` time is after the time of the reboot request
+* The ``poweredOn`` status is ``false``
+
+Whenever there are pending reboot requests, the Host controller will respond by
+powering off the Host (if it is not already) and timestamping all pending
+requests as having completed the power off stage. Power will then be restored
+only if/when the ``Online`` spec of the Host is true. Once Hosts are powered up
+again, any satisfied reboot requests will be timestamped as having completed
+the power on stage, with this timestamp matching the ``lastPoweredOn`` time of
+the Host.
+
+## Drawbacks
+
+This approach entails a whole new custom resource definition to maintain, and
+the instances of this resource must be managed by the Host controller, which
+adds a degree of complexity.
+
+If large numbers of requests are created but never deleted, searching for
+pending requests may start to consume significant resources.
+
+## Future Enhancements
+
+### Labelling
+
+Adding labels to partially completed and completed requests would allow us to
+exclude them when querying for pending requests. That would eliminate the
+performance concerns associated with allowing old requests to accumulate. This
+would be best performed by a RebootRequest-specific controller.
+
+### Automatic deletion
+
+The operator could automatically delete requests once they have been fulfilled
+(either immediately or after a predefined timeout). This would both eliminate
+performance concerns and reduce clutter. Clients wishing to observe data from
+the resource could set a finalizer that keeps the resource around only until
+they are done with it. This might best be performed by a RebootRequest-specific
+controller.
+
+### Custom Timestamps
+
+The resource could allow the user to specify a timestamp, which may of course
+be in the future. The operator would ignore these requests until the appointed
+time, and only then effect the reboot. The request could safely be deleted at
+any point prior to the requested time.
+
+### Soft Reboots
+
+The resource could accept a "soft" reboot option, in which we would attempt to
+reboot the Host without switching the power off. Since these requests may fail,
+more error reporting would likely be required. In the event that a mix of hard
+and soft reboot requests were pending, the host would be hard-rebooted and all
+of the pending requests marked complete.
+
+## Alternatives
+
+The Machine (Reboot) Actuator could perform the reboot in an imperative, rather
+than declarative, manner by manipulating the ``Online`` spec of the Host in
+sequence to ``false`` and then back to ``true`` again once the ``poweredOn``
+status was seen to be ``false``. However, in the presence of multiple actors
+this approach is prone to race conditions or controller fights.
+
+The request could simply be a timestamp field in the Host spec. The Host would
+coalesce multiple requests into one at the time they are received (rather than
+at the point of action). This would be simpler, but requires the client to
+provide a meaningful timestamp (i.e. in the recent past), forcing us to make
+awkward decisions about what to do with bogus data. In contrast, with a CRD we
+can use the creation timestamp provided by the system. This option also makes
+it more difficult to add further reboot-related API features in the future.
+
+The request could be made by adding to a list of client-chosen strings in the
+Host spec. This would be again be simpler than adding a new CRD at the cost of
+future flexibility, but requires clients to co-operate (by only adding to the
+list, and never replacing it), so one misbehaving client could cause another to
+act erratically.
--- a/design/reboot-interface.md
+++ b/design/reboot-interface.md
@@ -88,30 +88,32 @@ section of the BareMetalHost status. This records a time after which the Host
 was last booted using the current image. Processes running prior to this time
 may be assumed to have been stopped.
 
-A new date-time field, ``pendingRebootSince``, will be added to the ``provisioning``
-section of the BareMetalHost status. This records a time before which the Host
-was last requested to reboot (because we cannot trust any value from the user,
-who even if well intentioned, may have created the timestamp on a machine that 
-was not synchronised with the cluster or has a different timezone).
+A new date-time field, ``pendingRebootSince``, will be added to the
+``provisioning`` section of the BareMetalHost status. This records a time
+before which the Host was last requested to reboot (because we cannot trust any
+value from the user, who even if well intentioned, may have created the
+timestamp on a machine that was not synchronised with the cluster or has a
+different timezone).
 
-Since the user interface requirements are still unclear, we will follow 
-standard practices of using an annotation to trigger reboots.  
+Since the user interface requirements are still unclear, we will follow
+standard practices of using an annotation to trigger reboots.
 
-The basic annotation form ( ``reboot.metal3.io`` ) triggers the controller to
-power cycle the Host.  This form has set-and-forget semantics and the 
-controller removes the annotation once it restores power to the Host. 
+The basic annotation form (``reboot.metal3.io``) triggers the controller to
+power cycle the Host. This form has set-and-forget semantics and the controller
+removes the annotation once it restores power to the Host.
 
-An advanced form ( ``reboot.metal3.io/{key}`` ) instructs the controller hold 
-the Host in a ``PoweredOff`` state so that the caller can perform any required
-actions while the node is in a known safe state.   Callers indicate to the 
-controller that they are ready to continue by removing the annotation with their 
-unique ``{key}`` suffix.
+An advanced form (``reboot.metal3.io/{key}``) instructs the controller hold the
+Host in a ``PoweredOff`` state so that the caller can perform any required
+actions while the node is in a known safe state. Callers indicate to the
+controller that they are ready to continue by removing the annotation with
+their unique ``{key}`` suffix.
 
-In the case of multiple clients, the controller will wait for all annotations 
-of the form ``reboot.metal3.io/{key}`` to be removed before powering on the Host.
+In the case of multiple clients, the controller will wait for all annotations
+of the form ``reboot.metal3.io/{key}`` to be removed before powering on the
+Host.
 
-If both ``reboot.metal3.io`` and ``reboot.metal3.io/{key}`` forms are in use, 
-the  ``reboot.metal3.io/{key}`` form will take priority.
+If both ``reboot.metal3.io`` and ``reboot.metal3.io/{key}`` forms are in use,
+the ``reboot.metal3.io/{key}`` form will take priority.
 
 In all cases, the content of the annotation is ignored but preserved. This
 ensures that whatever placed the annotation can have a way of tracing it back
@@ -120,7 +122,7 @@ Machine resource being remediated.
 
 The actual power management will be performed by the Host controller. This is
 necessary to avoid race conditions by ensuring that the ``Online`` flag and any
-reboot requests are managed in the same place. 
+reboot requests are managed in the same place.
 
 If:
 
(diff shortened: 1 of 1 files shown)


design/reboot-interface.md
```diff
@@ -0,0 +1,155 @@
...
+BareMetalHost would search for pending reboot requests naming it as the target
+and initiate a reboot if any are found. This would mean that multiple requests
+would be consolidated into a single reboot at the point of action rather than
+at the point of receiving each request. This would maintain a Host-level record
+of reboot requests. It would also allow for future-dated requests if we still
+allow the user to provide a timestamp, or could prevent the ambiguity by using
+the creation timestamp of the resource (which is not under the control of the
+user) instead. However, this would be considerably more involved to implement,
+and may become inefficient (requiring large amounts of data to be transferred
+from etcd) if stale requests are allowed to accumulate.
```
zaneb: @mhrivnak WDYT about this last paragraph as a compromise? It's more work to write/maintain, but it should maintain correctness without sacrificing the UX.

design/reboot-interface.md
```diff
@@ -0,0 +1,155 @@
...
+from state inconsistencies in Ironic prior to provisioning. Any such
+inconsistencies that are not handled internally by the BareMetalHost will
+likely need manual intervention to recover anyway.
+
+## Proposal
+
+A new date-time field, ``lastBooted``, will be added to the ``provisioning``
+section of the BareMetalHost status. This records a time after which the Host
+was last booted using the current image. Processes running prior to this time
+may be assumed to have been fenced.
```
dhellmann: It's a bit pedantic, but the operator doesn't actually know when the host is booted. We know when it was powered on, but even when it is powered on it might fail to boot. That's why the existing status field is called PoweredOn rather than Booted, for example. So this field should be lastPoweredOn to be consistent.

design/reboot-interface.md
```diff
@@ -0,0 +1,155 @@
...
+
+To provide a declarative interface for performing a reboot, a new
+``rebootRequestTimestamp`` field will to be added to the BareMetalHost spec.
+This field could have any type, so long as clients are able to provide a unique
+value to trigger another reboot, but the current timestamp at the time of the
+request is the logical value to use.
```
dhellmann: I think the fact that it's a timestamp is what confused me about the scheduling aspects in the code review.

design/reboot-interface.md
dhellmann: Presumably all matching reboot request CRs would be updated with status, so that the clients that created them could watch them?

design/reboot-interface.md
zaneb: +1 that's a more accurate name

design/reboot-interface.md
zaneb: Yes. I'm assuming that the controller for the RebootRequest would watch the BareMetalHost and update its status once it saw the Host reboot.

dhellmann: If the baremetal operator is supposed to respond to the separate reboot request, I would expect it to be the one updating the CR with status.

zaneb: Is anything allowed to update the status (as opposed to the spec) of another resource? I'm not clear on what the rules are there.

dhellmann: Good question. Technically it's possible. It does seem to go against the usual pattern of having the "owning" controller responsible for the status fields, though. It also seems like mixing management of the request CR and host CR in one controller is going to complicate that controller.

What exactly are the conditions under which we have a race when there are multiple requests to reboot the same host?

If 2 reboot CRs are created at the same time, their controller will have a race to turn the host off. But when the second CR is processed I would expect the controller to see that the host is already in a state where the BMO is trying to turn it off, and immediately update its status to reflect that the reboot is in progress. Then when the host state continues to change, we may also have a race to turn it back on with the same sort of results. So, that's not a problem.

If one reboot CR is already being processed with the host in an off state and being turned back on when a second reboot CR is created, I would expect the controller to recognize that, interpret the state as not needing a reboot (because the host was powered off to begin with), and update its status to indicate that the host is coming back up. So that's also not a problem.

If one reboot CR is being processed with the host in an on state after a reboot but the reboot CR does not yet reflect that when a second CR is created, I would expect a second reboot to be triggered by the new CR. The first CR may, or may not, be set to indicate that the host was powered back on before the second reboot, but I'm not sure it matters very much. The controller should update it as the second CR is processed. At worst it's slightly inaccurate information about when the host was powered on and if the reboots come back-to-back like that it's not _that_ inaccurate.

Am I missing a case?

mhrivnak: Seems reasonable to me. Usually for any resource, there should be one "owning" controller that makes changes to the world based on the spec and writes to the status. I don't see a problem in this case with having one controller owning both resources, since the HRR is basically a child of the BMH.

design/reboot-interface.md
```diff
@@ -0,0 +1,155 @@
...
+## Proposal
+
+A new date-time field, ``lastBooted``, will be added to the ``provisioning``
+section of the BareMetalHost status. This records a time after which the Host
+was last booted using the current image. Processes running prior to this time
+may be assumed to have been fenced.
+
+To provide a declarative interface for performing a reboot, a new
+``rebootRequestTimestamp`` field will to be added to the BareMetalHost spec.
+This field could have any type, so long as clients are able to provide a unique
```
mhrivnak: If I provide unique strings as my values, that are not timestamps, how does it know when I've modified it to a different value?

design/reboot-interface.md
mhrivnak: Agree. The next two paragraphs talk about how exactly the `rebootRequestTimestamp` will be parsed and compared to either the current time or a saved timestamp. But this paragraph says the `rebootRequestTimestamp` can be any unique value, not necessarily a timestamp.

I think we should pick one and commit to it. Either it has to be a timestamp and the controller will do time comparisons as described below, or it can be any unique string and nothing will ever parse it as more than a unique string.

design/reboot-interface.md
zaneb: What this is intending to say is that we could have designed it to take any unique string, but instead we designed it to take a timestamp (which is a unique string with some other nice properties).
Validation will automatically enforce that the user provides something resembling a timestamp here.

mhrivnak:
Relevant discussion that I think is worth referencing: https://github.com/kubernetes/kubernetes/issues/72637

design/reboot-interface.md
```diff
@@ -0,0 +1,155 @@
...
+  fairly severe clock skew is possible)
+
+## Drawbacks
+
+The need for the consumer to supply a recent timestamp means that the design
+requires some co-operation from the client that is not naturally enforced by
+the API. Without good documentation this could lead to confusion and incorrect
+usage. It is also vulnerable to clock skew in severe cases.
+
+## Alternatives
```
mhrivnak: Here's an option that's growing on me and that I think is worth serious consideration. It takes inspiration from how finalizers work. @jarrpa this is relevant to your use case in ocs-operator; it's another option for inducing "factory reset" of your data.

In the host's Spec, we add a field called `rebootRequests`. It is an array of strings. If you want a reboot, you put a string of your choice in that array.

When the controller sees one or more strings in the array, it begins a reboot.

If additional strings are added while the reboot is in-progress, the controller ignores them.

When the reboot is complete, the controller clears the array.

It allows multiple actors to safely request reboots at close to the same time. It allows each actor to know if its reboot request has been completed. And it enables the controller to know for sure if anyone has requested a reboot since the previous reboot.

The only part that's a bit unconventional is that the controller is modifying the spec for its primary resource. But in this case I think that remains within the spirit of the declarative API.

To get around that, we could theoretically make a separate controller that watches for entries in `status.rebootRequestsCompleted` and deletes them from `spec.rebootRequests`, and then have our main controller remove those requests from `status.rebootRequestsCompleted` as they disappear from the spec. That would fit the pattern of only writing to your primary resource's status and letting others write to your spec, but it would be more code and complexity for no practical gain. It's easier to let the one control loop just clear out the entries from the spec; the API experience is the same ether way for the actors who want reboots.

WDYT?

dhellmann: This document asserts that we need an atomic reboot operation, but doesn't really explain why. I would like to understand exactly where the race condition (or other issue) is in the design that just uses the existing power switch with a separate CRD to indicate the request for a reboot. I went through a few cases in another comment and only came up with one race (the back-to-back reboot), which wouldn't be solved by any of the proposed approaches and which I don't consider all that bad (we're not going to impose a minimum-time-between-reboots, are we?).

Is there actually a race condition if we have a properly written reboot controller that examines the host state before deciding what action to take?

zaneb: @dhellmann I would say that first and foremost, having a declarative interface is its own reward. You certainly can work around design failures by sequentially updating declarative interfaces, but it gets un-fun pretty quickly and a system where such hacks are allowed to proliferate freely turns into a big ball of mud. As far as specific races go, there's the case where an admin attempts to power off the server during the period where the fencing code is trying to reboot it, which would result in the server being unintentionally booted again. In the opposite case, where an operator exists that is forcing the power state to be on, we might never be able to fence the node (the CAPBM doesn't currently do this, but if we can't convince upstream cluster-api to add a reboot interface then our Machine reboot operator should be completely independent of CAPBM).

@mhrivnak my reservation about doing it as a list in the main Host spec is that IIUC it pushes some of the responsibility for correctness onto the clients - a client wanting to request a reboot must append to the list without removing any other entries, otherwise the source of the other entries could legitimately conclude that its reboot request had been completed. Copying entries to the status as they're received could help, but I think there's still a race. We shouldn't rely on clients being well-behaved in terms of implementing logic correctly to ensure they play nice with others. ``kubectl apply`` has no such logic, for a start.

beekhof: What about having the `rebootRequestTimestamp` field, but having the BMH controller create a corresponding `HostRebootRequest`, or update it if it already exists (depending on how the value of `rebootRequestTimestamp` changes)?

mhrivnak: @zaneb re: your concern about correct array handling, that's already how finalizers work. We depend on clients to correctly add and remove their finalizer without affecting other finalizers in the array. It's simple, safe and common.

beekhof:
The only other thing I would add is machine-healthcheck's desire/need to know when the `off` part of a reboot completes, as this is the step that allows workload recovery to commence (safely)

design/reboot-interface.md
dhellmann: I don't see anything in the design that addresses the issue of conflicting requests to reboot and set the power state -- they're separate fields and can be changed in separate calls (or the same call) in ways that conflict.

If we want to prevent simultaneous action, then either we need to use a web hook to validate the inputs or we need to redesign the power interface at the same time to combine the two things.

That said, I don't feel like I've seen a lot of other kubernetes APIs that worried about conflicting requests. Isn't the point that the operator reconciles the state of the world to match the resource based on its most current values precisely so it isn't necessary to design APIs that understand all of the transitions between different states?

dhellmann:
> The only other thing I would add is machine-healthcheck's desire/need to know when the `off` part of a reboot completes, as this is the step that allows workload recovery to commence (safely)

Does the thing rebooting the host need to pause at that point to allow something else to happen before turning the host on? Or does the health check just need to know when the host is actually powered off?

design/reboot-interface.md
zaneb: The design makes the reboot request independent of the long-term desired power state. So `oc patch -type patch {"spec": {"online": false}}` always results in the power remaining off, whereas if the fencing code were twiddling the power state around the same time then the power might come back on even after issuing that command.
Generally that should only happen in the case where it's controlled by another operator, and the other operator provides the preferred user interface for setting the long-term desired state. That isn't the case here, because there isn't another interface for controlling the power state; users should expect that things they set here will stick.

zaneb:
The important thing for fencing is that the old processes have gone away. As soon as they are gone it is safe to allow their hot backups to become masters, because a split-brain event can no longer occur. How long they've been away is irrelevant.

zaneb:
After discussing with @dhellmann and @mhrivnak over video, we're leaning toward the alternative described in the last paragraph - this is the gold-plated version that will give us a flexible interface upon which we can implement future features, while still keeping the implementation in a single place inside the Host operator (which is necessary to avoid races). There are some minor technical challenges but we believe we can work around them.
I'll rework this design doc to describe that alternative in detail instead.

dhellmann:
Note for my future self: @zaneb, @mhrivnak, and I talked about this today and I am now convinced that we do need an API that allows us to ensure that a reboot is completed successfully without interference from another actor simultaneously fiddling with the power switch.

We agreed that the baremetal-operator should manage that reboot process, rather than a third-party. We also discussed the appeal of the final alternative approach described here, which uses a separate CRD to represent the request for a reboot in case there are multiple simultaneous requests (where new instances would trigger the host reconciliation loop to take action, and all active instances would be updated as progress is made). @zaneb agreed to update this design with more detail about what a minimal version of that CRD would look like.

I do also think we need to explicitly address the case where the Online flag for a host is turned off while it is being rebooted. IIRC, @zaneb said that the current implementation would leave the host powered off, which seems reasonable, but let's document that.

design/reboot-interface.md
```diff
@@ -0,0 +1,155 @@
...
+equivalent of the Machine Actuator limited to just this purpose.
+
+### Goals
+
+The Machine (Reboot) Actuator will require from the BareMetalHost:
+
+* A declarative API to perform the reboot
+* The ability to determine a point where all processes running on the Host at
+  the time of the fencing decision are guaranteed to have been stopped
+* Rapid convergence to the fenced state independent of other happenings in the
```
markmc: Not obvious to me what "fenced state" means here - e.g. is "fenced state" the "point where  all process running on the Host are guaranteed to have been stopped"?

markmc: If the term fencing was avoided everywhere except the "Motivation" section, it would help with clarity - I find the term often brings confusion, and replacing it with a more specific term doesn't cause any meaning to be lost

zaneb: Done

design/reboot-interface.md
```diff
@@ -0,0 +1,184 @@
...
+may be assumed to have been stopped.
+
+To provide a declarative interface for performing a reboot, a new
+BareMetalHostRebootRequest CRD will be created. The spec of this custom
+resource will contain only a reference to a BareMetalHost. The status will
+contain timestamps for when the Host was powered off and on again. Once the
+power off time is present in the status, the host has been stopped.
+
+If possible, we will not create a separate controller for the RebootRequest,
+but instead perform all operations on it in the course of reconciling the
```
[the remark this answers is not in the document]
zaneb: Good question, without having tried to implement it yet I don't know enough about it to be sure (I expect @mhrivnak will know off the top of his head :). I think we'll avoid creating another controller if possible at the beginning, but if it's necessary to get the notifications working right then we'll do it.

design/reboot-interface.md
```diff
@@ -0,0 +1,184 @@
...
+BareMetalHost object that it references. The Host controller will watch
+RebootRequest objects and trigger a reconciliation of the linked Host when one
+appears or changes.
+
+The creation timestamp of the request resource will be interpreted as the time
+of the reboot request. Requests that predate the last booted time will be
+ignored. The Host controller will timestamp all pending reboot requests as
```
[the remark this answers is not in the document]
zaneb: Yep, I think this wording was left over from the previous version. I will fix that up.

design/reboot-interface.md
```diff
@@ -0,0 +1,184 @@
...
+
+* The Host has been deprovisioned (in this case the ``lastPoweredOn`` field
+  will be empty)
+* The ``lastPoweredOn`` time is after the time of the reboot request
+* The ``poweredOn`` status is ``false``
+
+Whenever there are pending reboot requests, the Host controller will respond by
+powering off the Host (if it is not already) and timestamping all pending
+requests as having completed the power off stage. Power will then be restored
+only if/when the ``Online`` spec of the Host is true. Once Hosts are powered up
```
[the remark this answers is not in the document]
zaneb: I think you meant to say Online=true in the second line, but that's correct.

design/reboot-interface.md
```diff
@@ -0,0 +1,184 @@
...
+time, and only then effect the reboot. The request could safely be deleted at
+any point prior to the requested time.
+
+### Soft Reboots
+
+The resource could accept a "soft" reboot option, in which we would attempt to
+reboot the Host without switching the power off. Since these requests may fail,
+more error reporting would likely be required. Hard reboots would take
+precedence over soft reboots, with both being marked complete if they were
+pending.
```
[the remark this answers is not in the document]
zaneb: That's what the last sentence is trying to convey (possibly poorly).

design/reboot-interface.md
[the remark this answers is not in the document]
mhrivnak: We can do it all in the existing controller. We'll have to setup a slightly customized Watch handler that reads the host reference and then queues the corresponding BareMetalHost, but that's no big deal. We're [already doing that pattern](https://github.com/metal3-io/cluster-api-provider-baremetal/blob/4841e88/pkg/manager/wrapper/wrapper.go#L53) in the cluster-api-provider-baremetal.

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+necessary to avoid race conditions by ensuring that the ``Online`` flag and any
+reboot requests are managed in the same place. The Host controller
+will watch RebootRequest objects and trigger a reconciliation of the linked
+Host when one appears or changes.
+
+For simplicity, we will try to avoid creating a separate controller for the
+RebootRequest at all, but instead perform all operations on it in the course of
+reconciling the BareMetalHost object that it references. However, it may yet
+prove necessary to also create a trivial controller for the purposes of basic
+housekeeping (e.g. setting the owner of the resource).
```
mhrivnak: That won't be necessary. See my other comment that's "outdated". It doesn't hurt to say we might need this, but I can't think of a reason why we would.

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+If large numbers of requests are created but never deleted, searching for
+pending requests may start to consume significant resources.
+
+## Future Enhancements
+
+### Labelling
+
+Adding labels to partially completed and completed requests would allow us to
+exclude them when querying for pending requests. That would eliminate the
+performance concerns associated with allowing old requests to accumulate. This
```
mhrivnak: Keep in mind that the records will be held in the in-memory cache of the operator. Any performance concerns will be around how quickly the operator can scan those results and retrieve a desired resource. Go is fast, and I think you'd need A LOT of records before scanning is going to have a performance impact. That said, labels are still a fine idea, particularly if any other tool (maybe kubectl) is going to search in the same way.

RAM usage may be more impactful than any performance impact.

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+would be best performed by a RebootRequest-specific controller.
```
mhrivnak: Why would this be best?

I like targeted controllers, but there's also a real benefit to having just one controller that "owns" a resource. In this case, I'd default to letting the BMH controller own the RebootRequest resource as a sort of child, which includes any writes that reflect its status, unless there's a clear reason to have a second controller.

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Automatic deletion
+
+The operator could automatically delete requests once they have been fulfilled
+(either immediately or after a predefined timeout). This would both eliminate
```
mhrivnak: A TTL-based approach could be useful. Something similar [is in the works](https://github.com/kubernetes/enhancements/issues/592) for the Job resource.

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+performance concerns and reduce clutter. Clients wishing to observe data from
+the resource could set a finalizer that keeps the resource around only until
+they are done with it. This might best be performed by a RebootRequest-specific
+controller.
+
+### Custom Timestamps
+
+The resource could allow the user to specify a timestamp, which may of course
```
mhrivnak: It would be a good idea to more fully define the use cases for this before spending much time considering implementation.

design/reboot-interface.md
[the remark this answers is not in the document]
dhellmann: Yes, I pasted and missed that edit.

design/reboot-interface.md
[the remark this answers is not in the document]
dhellmann: No, just my poor reading comprehension due to distraction.

design/reboot-interface.md
dhellmann: That works for me, then.

design/reboot-interface.md
dhellmann: I think it's safe to leave this as an item for future consideration.

design/reboot-interface.md
zaneb: Yeah, I included this "future enhancements" section more as a way to record stuff we have talked about for later recollection, not as a proposal to implement it any time soon.

design/reboot-interface.md
zaneb: It seemed like it would be easier to have an independent thing that just checks the status and adds the label if it's missing and the request is complete. Then the Host controller doesn't have to worry about the label, or about something else removing the label.
I guess the Host controller is going to get notified anyway though, so maybe it doesn't actually add any value.

design/reboot-interface.md
zaneb: Yeah, it's likely not a problem in practice. But every Host is going to have to do this scan at least once every 60s, so you can imagine that there'll be a certain scale/longevity beyond which we'll want to do something like this or the automatic deletion.

design/reboot-interface.md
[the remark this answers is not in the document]
zaneb: Fixed it anyway :)

design/reboot-interface.md
mhrivnak: We can code whatever reconciliation we want for the reboot request and put that logic in either controller. I don't see an advantage to putting the label part in a dedicated controller.

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Proposal
+
+A new date-time field, ``lastPoweredOn``, will be added to the ``provisioning``
+section of the BareMetalHost status. This records a time after which the Host
+was last booted using the current image. Processes running prior to this time
+may be assumed to have been stopped.
+
+To provide a declarative interface for performing a reboot, a new
+BareMetalHostRebootRequest CRD will be created. The spec of this custom
```
mhrivnak: What do you think about just calling this `BareMetalHostReboot`? Since k8s has a declarative API, every resource is merely a "request".

A more common example of an imperative action is "backup". Operators such as etc-operator tend to have resources called `EtcdBackup` or similar.

design/reboot-interface.md
zaneb: The only thing I can think of is if something adds the label to request that is not actually complete - a dedicated reboot request controller would simply remove the label again on reconcile, but the Host controller would ignore all resources with the label (it doesn't have to, but that is the whole point of having a label).
We're getting into extreme edge case territory with borderline malicious actors there though.

mhrivnak: Ah, I see, good point. We can keep that in mind if this became a problem we want to address. But I think we'll be better off by (eventually) implementing some kind of cleanup for old reboots and not having the operator use labels to identify which reboot resources have a particular state.

design/reboot-interface.md
zaneb: I was going to ask you about the name, since it's so long... do we need the BareMetal part either, given that the resource will already be under `metal3.io/`?
I'm not a fan of naming declarative APIs with verbs, but I'd go along with it if that's the pattern. 'Backup' can either be a verb or a noun though, and I suspect it's being used as a noun in that example (come to think of it, "back up" as a verb is two words - you can "back up some data" or "make a backup of some data" - so I'd expect to see `BackUpEtcd` if it were being used as a verb).

design/reboot-interface.md
```diff
@@ -0,0 +1,193 @@
...
+resource will contain only a reference to a BareMetalHost. The status will
```
dhellmann: We should add a Reason field, too.

design/reboot-interface.md
mhrivnak: > I was going to ask you about the name, since it's so long... do we need the BareMetal part either, given that the resource will already be under `metal3.io/`?

This is an area of ambiguity. You would think the namespacing would be enough. But the group and version can be omitted when using a client like `kubectl`, and that can introduce ambiguity. What if some other operator, like MCO, also create a `Reboot` resource, and then I do `kubectl get reboots`? kubectl seems to pick one of them somewhat arbitrarily.

In practice, many operators address this with a prefix on their resource, such as `EtcdBackup` does. I suggest we do that here.

> I'm not a fan of naming declarative APIs with verbs, but I'd go along with it if that's the pattern. 'Backup' can either be a verb or a noun though, and I suspect it's being used as a noun in that example (come to think of it, "back up" as a verb is two words - you can "back up some data" or "make a backup of some data" - so I'd expect to see `BackUpEtcd` if it were being used as a verb).

Reboot can also [be a noun](https://www.merriam-webster.com/dictionary/reboot).

zaneb: True! But for some reason you would never hear anybody say "let's create a reboot" or "let's delete that reboot" in the way that you might "create an etcd backup" or "delete an etcd backup".
Do you think `HostRebootRequest` would be sufficiently unique?
I guess we always have the option of specifying multiple short names that we can use in practice even if the official one is painfully long.

mhrivnak: > True! But for some reason you would never hear anybody say "let's create a reboot" or "let's delete that reboot" in the way that you might "create an etcd backup" or "delete an etcd backup".

It's normal to talk about creating a backup because the act of performing a backup produces something. Performing a reboot does not normally produce anything.

But it's perfectly normal to ask someone to perform a reboot.

As another example, here's an interesting feature of journalctl:

```
$ journalctl --list-boots
 -2 04daa48a0ba249089166eed69fa913b8 Wed 2019-07-31 23:03:59 EDT—Fri 2019-08-02 18:23:19 EDT
 -1 25660fb316124c0ca6de986ee0fc76ba Fri 2019-08-02 18:23:37 EDT—Tue 2019-08-06 13:54:33 EDT
  0 ece6a74488224946a59960a2675d4aed Tue 2019-08-06 13:54:54 EDT—Tue 2019-08-06 15:59:16 EDT
```

Nobody created or deleted those "boots", which may have been "reboots", but they sure happened. We can list them and see their properties.

> Do you think HostRebootRequest would be sufficiently unique?

I would go with BareMetalHostReboot, and likely include a short name of `bmhreboot`. HostReboot would probably be unique enough to avoid direct collision, but it could still cause confusion if there are any other "host"-related APIs from some other api group.

design/reboot-interface.md
```diff
@@ -0,0 +1,190 @@
...
+
+The basic annotation form ( ``reboot.metal3.io`` ) triggers the controller to
+power cycle the Host.  This form has set-and-forget semantics and the 
+controller removes the annotation once it restores power to the Host. 
+
+An advanced form ( ``reboot.metal3.io/{key}`` ) instructs the controller hold 
+the Host in a ``PoweredOff`` state so that the caller can perform any required
+actions while the node is in a known safe state.   Callers indicate to the 
+controller that they are ready to continue by removing the annotation with their 
+unique ``{key}}`` suffix.
```
beekhof:
```suggestion
unique ``{key}`` suffix.
```

design/reboot-interface.md
```diff
@@ -0,0 +1,190 @@
...
+
+We require a non-destructive way to fence a Kubernetes node. Some nodes cannot
+be replaced (it is currently not possible for a new master to join a cluster),
+or are expensive to replace (e.g. if this would require rebalancing Ceph data).
+A solution to this is for fencing to reboot the Host, thus ensuring that
+running processes are stopped to avoid a split-brain scenario while still
+allowing the node to rejoin the cluster with its data intact (albeit stale)
+after the reboot.
+
+The expected implementation of this on the fencing side is a Custom Resource
```
dhellmann: This paragraph feels out of date now?

design/reboot-interface.md
```diff
@@ -0,0 +1,190 @@
...
+status was seen to be ``false``. However, in the presence of multiple actors
+this approach is prone to race conditions or controller fights.
+
+The request could simply be a timestamp field in the Host spec. The Host would
+coalesce multiple requests into one at the time they are received (rather than
+at the point of action). This would be simpler, but requires the client to
+provide a meaningful timestamp (i.e. in the recent past), forcing us to make
+awkward decisions about what to do with bogus data. In contrast, with a CRD we
+can use the creation timestamp provided by the system. This option also makes
+it more difficult to add further reboot-related API features in the future.
```
beekhof:
```suggestion
awkward decisions about what to do with bogus data. In contrast, we can 
populate ``.Status.pendingRebootSince`` with a timestamp provided by the 
system. Exposing a timestamp on the `.Spec` also makes it more difficult to 
add further reboot-related API features in the future.
```

design/reboot-interface.md
```diff
@@ -0,0 +1,190 @@
...
+
+The request could be made by adding to a list of client-chosen strings in the
+Host spec. This would be again be simpler than adding a new CRD at the cost of
+future flexibility, but requires clients to co-operate (by only adding to the
+list, and never replacing it), so one misbehaving client could cause another to
+act erratically.
```
beekhof:
```suggestion
The request could be made by adding to a map of client-chosen strings in the
Host spec. This would be cleaner than using annotations, but requires clients to 
co-operate (by only adding/removing their entry in the map), so one misbehaving 
client could cause another to act erratically.
```

design/reboot-interface.md
```diff
@@ -0,0 +1,190 @@
...
+of the pending requests marked complete.
+
+## Alternatives
+
+The Machine (Reboot) Actuator could perform the reboot in an imperative, rather
+than declarative, manner by manipulating the ``Online`` spec of the Host in
+sequence to ``false`` and then back to ``true`` again once the ``poweredOn``
+status was seen to be ``false``. However, in the presence of multiple actors
+this approach is prone to race conditions or controller fights.
+
```
beekhof:
```suggestion

We could implement a ``HostOn`` and/or ``HostOff`` CRD with the same name as
the target Machine.  This would:
* ensure uniqueness
* move all power management into a new operator,
* require the baremetal-operator to create and delete HostOff/HostOn objects
  based on ``.Spec.Online`
* require HostOff objects to take precedence over HostOn,
* Hosts to be powered off if neither a HostOn or HostOff object exists,
* the operator to remove HostOff objects once the Host reaches a powered off
  state,
* machine-remediation to create/update HostOff objects and register a
  finaliser (in order to be able to delete the Node once the Host is off)
This would work, but is highly invasive, complex, and whether it provides any
tangible benefits is debatable.

```

beekhof:
@dhellmann @zaneb I've made some suggestions to update the alternatives section.

I still firmly believe the existing approach is the right one :)

design/reboot-interface.md
```diff
@@ -0,0 +1,190 @@
...
+We require a non-destructive way to fence a Kubernetes node. Some nodes cannot
+be replaced (it is currently not possible for a new master to join a cluster),
+or are expensive to replace (e.g. if this would require rebalancing Ceph data).
+A solution to this is for fencing to reboot the Host, thus ensuring that
+running processes are stopped to avoid a split-brain scenario while still
+allowing the node to rejoin the cluster with its data intact (albeit stale)
+after the reboot.
+
+The expected implementation of this on the fencing side is a Custom Resource
+Definition that requests a Machine be rebooted. The actual reboot will be
```
beekhof:
```suggestion
The expected implementation of this on the fencing side is an annotation
that requests a Machine be rebooted. The actual reboot will be
```

design/reboot-interface.md
beekhof: Suggestion made

design/reboot-interface.md
zaneb: I don't think this does actually work - if multiple things need to create a HostOff resource and the resource name is unique then none of them know when to remove it. I guess you could always create it and delete it right away, and have everything just add finalisers (:screaming:), but I think that's just a more complicated way of doing the same thing as annotations.

In any event, this doesn't feel particularly declarative and I hate it enough that I don't feel the need to include it in the list of alternatives ;)

beekhof: The thinking was that the new controller (not machine-remediation) would delete the object once shutdown was complete. So m-r would create and add finizaliser

beekhof: And yes, i am not a fan either, but this is what alex came up with

PR #55, opened 2019-10-10
Start documenting a set of API design principles

dhellmann:
We've had several design discussions recently in which I have needed
to reiterate these principles, so I think it's time to write them
down in one place.

### DIFF
--- a/design/api-design-principles.md
+++ b/design/api-design-principles.md
@@ -1,0 +1,88 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# api-design-principles
+
+## Status
+
+provisional
+
+## Table of Contents
+
+<!--ts-->
+   * [api-design-principles](#api-design-principles)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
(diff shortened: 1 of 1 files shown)


design/api-design-principles.md
```diff
@@ -0,0 +1,96 @@
...
+
+provisional
+
+## Table of Contents
+
+A table of contents is helpful for quickly jumping to sections of a
+design and for highlighting any additional information provided beyond
+the standard template.
+
+[Tools for generating][] a table of contents from markdown are available.
```
russellb: want to remove this boilerplate?

russellb:
I agree with what you've written so far.-

design/api-design-principles.md
dhellmann: Done

dhellmann:
> I like this so far.
> 
> Another one we could add is: don't make assumptions about what the BaremetalHost will be used for. The Machine API is not the only consumer and running a Kubernetes node is not the only thing users may want to do with the Host.

I like that. Do you want me to add it to this, or should we consider this list something we maintain as we go along?

zaneb:
I'm happy either way. If you want to merge now I'm happy to submit another PR for it, or you can update this PR.

dhellmann:
Let's do it separately. It's a good way to reinforce that the list is meant to be updated over time.

dhellmann:
See #75 for the follow-up.

PR #56, opened 2019-10-22
Add a design proposal for userdata handling

maelk:
This is for discussion on merging different sources of userdata

### DIFF
--- a/design/userdata-merging.md
+++ b/design/userdata-merging.md
@@ -1,0 +1,234 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Userdata handling in Metal3-io
+
+In a baremetal deployment, some of the configuration needs to be tightly tied
+to the physical node that is being provisioned (for example for networking).
+Hence we foresee a need to enable the user to give node-specific configuration.
+
+However, in CAPI v1alpha2, the userdata on the machine for provisioning is
+created by CABPK, that is unaware of such constraint. Cluster-API implements
+deployments, where all the BaremetalMachine and CABPK output are generated
+from a template, that does not allow for node-specific configuration.
+
+Hence we would need to merge the userdata coming from the Machine, the
+BaremetalMachine and the BaremetalHost.
+
+
+## Status
+
+provisional
+
+## Table of Contents
+
+<!--ts-->
+   * [Title](#title)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [User Stories](#user-stories)
+            * [Story 1](#story-1)
+            * [Story 2](#story-2)
+         * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Drawbacks](#drawbacks-optional)
+      * [Alternatives](#alternatives-optional)
+      * [References](#references)
+
+<!-- Added by: stack, at: 2019-02-15T11:41-05:00 -->
+
+<!--te-->
+
+## Summary
+
+With v1alpha2, Cluster-API goal is to split the bootstrap provider from the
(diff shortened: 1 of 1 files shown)


maelk:
Hi @hardys @dhellmann @russellb ! Do you have any comments on this ?

design/userdata-merging.md
```diff
@@ -0,0 +1,234 @@
...
+independently. The deployment could then use physical nodes that are not
+strictly identical.
+
+### Implementation Details/Notes/Constraints
+
+This does not require to modify the current CAPI workflow.
+When selecting a baremetalHost for the BaremetalMachine, the controller would
+merge the userdata from the machine with the user-provided input and the
+load-balancer related input. The output would be written as userdata in the
+BaremetalHost for deployment.
```
hardys: I think I mentioned this previously, but having some hard-coded merge implementation makes me nervous, there are multiple approaches to merging and cloud-init already supports merging multiple configs (e.g multiple cloud-config parts in a mime archive), including specifying the strategy to apply when merging:

https://cloudinit.readthedocs.io/en/latest/topics/merging.html

I get that one of the objectives here is validation, but I wonder if we'll end up reimplementing all that cloud-init merge logic incrementally - maybe it would be better to just consider appending user-data and let the tool take care of the merging?

design/userdata-merging.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Risks and Mitigations
+
+There are several types of userData, such as cloud-init or ignition. This
```
hardys: It's more complex that just cloud-init vs ignition vs ... - cloud-init supports multiple formats not only cloud-config (and this includes custom part-handlers etc):

https://cloudinit.readthedocs.io/en/latest/topics/format.html

I worry that by focussing on some hard-coded merge implementation that only works for one strategy (e.g probably some kind of exclusive deep merge which doesn't allow conflicts or appending data?) we'll end up with a fragile and/or inflexible solution which blocks users from accessing all the interfaces of the underlying tools?

design/userdata-merging.md
```diff
@@ -0,0 +1,234 @@
...
+proposal should not implement a feature for one that would be detrimental or not
+consider another one.
+
+The user could also give some input that directly conflict with the output of
+the bootstrap provider. This needs to be handled carefully.
```
hardys: One question is how much can we realistically validate this beyond looking for conflicting cloud-config keys (which are actually allowed if you use the cloud-init merge strategies implementation), there could be subtle incompatibility which doesn't explicitly overlap (such as modifying storage or networking in a way that breaks the cluster configuration)?

design/userdata-merging.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Design Details
+
+- A new field in the BaremetalHost spec would be added :
+
+```yaml
+  spec:
+    userDataInput:
+      type: cloud-init
```
hardys: Note as mentioned above cloud-init supports multiple formats, I wonder if it would be simpler to consider this a mime archive and append parts to it?

design/userdata-merging.md
```diff
@@ -0,0 +1,234 @@
...
+- For all fields that are lists, the content of the list, if it exists, in
+  userDataAppend will be appended to the content in this list from the
+  bootstrap provider. The content of the list, if it exists, in userDataPrepend
+  will be added at the beginning of the list from the bootstrap provider.
+- For maps, they will be merged and a duplicated key would trigger an error.
+- For other types, a duplicated key in the userdata would trigger an error too.
+- Not specifically to cloud-init, another place where conflict could occur is
+  in the files created by the bootstrap provider. If both the user input and the
+  bootstrap provider output are modifying the same file, an error should be
+  raised.
```
hardys: So if I'm reading this correctly, the proposal to reimplement a cloud-config merger inside the cluster-api-provider, with merge behavior which differs from any of the current cloud-init merge strategies?

design/userdata-merging.md
```diff
@@ -0,0 +1,234 @@
...
+output directly into the baremetalHost.
+
+This option goes against the decoupling of bootstrap and infrastructure provider
+done by CAPI community. The bootstrap provider would be relying on
+Infrastructure provider objects and not be independent.
+
+### Use cloud-init merging capabilities.
+
+This is problematic as we do not have any fine-grained control on how the
+merging is actually performed and can’t easily get errors back.
```
hardys: Honestly I think this will be the case regardless - as mentioned above there's a bunch of ways folks could potentially break this even with some strict hard-coded merge implementation, I wonder if it'd be better to accept that and simplify the interface from the outset, then look into ways to improve the behavior when runtime errors to occur?

hardys:
Just to clarify, I'm just looking to have a discussion on how we might make this more flexible, simpler and less tied to a single tool, not blocking but wanted to add some thoughts to hopefully trigger further discussions.

maelk:
Hi @hardys! Thank you for the very good comments. We have been discussing this quite much and the issues are outweighing the benefits. We will investigate forking CABPK to fit our purpose instead of going this way. We should be able to achieve the same thing in a much more streamlined way. We can always link everything together with the host selectors. Once again, thank you for the feedback, and we'll come back with another way to tackle the issue.

PR #59, opened 2019-12-05
Create annotation-for-power-cycling-and-deleting-failed-nodes.md

beekhof:
We need some new functionality to allow for automated recovery of some cluster failures.

### DIFF
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -1,0 +1,149 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Annotation for Power Cycling and Deleting Failed Nodes
+
+## Status
+
+implemented
+
+## Table of Contents
+
+<!--ts-->
+   * [Title](#title)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [User Stories](#user-stories)
+            * [Story 1](#story-1)
+         * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Drawbacks [optional]](#drawbacks-optional)
+      * [Alternatives [optional]](#alternatives-optional)
+      * [References](#references)
+
+<!-- Added by: stack, at: 2019-02-15T11:41-05:00 -->
+
+<!--te-->
+
+[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
+
+## Summary
+
+It is not always practical to require admin intervention once a node has been identified as having reached a bad or unknown state.
+
+In order to automate the recovery of exclusive workloads (eg. RWO volumes and StatefulSets), we need a way to put failed nodes into a safe state, indicate to the scheduler that affected workloads can be started elsewhere, and then attempt to recover capacity.
+
+## Motivation
+
+Hardware is imperfect, and software contains bugs. When node level failures such as kernel hangs or dead NICs occur, the work required from the cluster does not decrease - workloads from affected nodes need to be restarted somewhere. 
+
+However some workloads may require at-most-one semantics.  Failures affecting these kind of workloads risk data loss and/or corruption if "lost" nodes are assumed to be dead when they are still running.  For this reason it is important to know that the node has reached a safe state before initiating recovery of the workload.
+
+Powering off the affected node via IPMI or related technologies achieves this, but must be paired with deletion of the Node object to signal to the scheduler that no Pods or PersistentVolumes are present there.
+
+Ideally customers would over-provision the cluster so that a node failure (or several) does not put additional stress on surviving peers, however budget constraints mean that this is often not the case, particularly in Edge deployments which may consist of as few as three nodes of commodity hardware.  Even when deployments start off over-provisioned, there is a tendency for the extra capacity to become permanently utilised.  It is therefore usually important to recover the lost capacity quickly.
+
+For this reason it is desirable to power the machine back on again, in the hope that the problem was transient and the node can return to a healthy state.  Upon restarting, kubelet automatically contacts the masters to re-register itself, allowing it to host workloads.
+
+### Goals
+
+- Enable the safe automated recovery of exclusive workloads
+- Provide the possibility of restoring cluster capacity automatically
+
+### Non-Goals
+
+- Creating a public API for rebooting a node
+- Identifying when automated recovery should be triggered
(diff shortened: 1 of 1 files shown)


design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -0,0 +1,180 @@
...
+broadly.  For example, consider both security and how this will impact
+the larger kubernetes ecosystem.
+
+## Design Details
+
+See [PoC code](https://github.com/kubevirt/machine-remediation/)
+
+- A new [controller] 
+- A new [CRD](https://github.com/kubevirt/machine-remediation/blob/master/pkg/apis/machineremediation/v1alpha1/machineremediation_types.go)
+- A new annotation (namespace and name is open for discussion)
```
zaneb: How do you ensure that each annotation results in only one reboot being initiated? Does the data in the annotation contain some unique value that is then used deterministically as part of the name of the MachineRemediation CR?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -0,0 +1,180 @@
...
+
+This proposal calls for a new controller which watches for the presence of a
+specific Node annotation.  If present, the controller will locate the Machine
+and BareMetalHost host objects via their annotations, and (in serial) use the
+BareMetalHost API to power off the machine, delete the Node, and then power the
+machine back on again.
+
+In order to track the state of the recovery workflow, a [new
+CRD](https://github.com/kubevirt/machine-remediation/blob/master/pkg/apis/machineremediation/v1alpha1/machineremediation_types.go)
+is proposed.
```
zaneb: Is it the same controller that both watches for the annotation and reconciles the new CRD?
I ask because if we implement the proposal in #48 then we likely won't need this CRD any longer, but we would still need the controller to watch the Node, so it would be good to make them easily separable.

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -0,0 +1,180 @@
...
+
+While it is likely that the work proposed here may make use of such
+functionality once it exists, it should not be considered a pre-requisite for
+the purposes of evaluating this proposal.
+
+### Risks and Mitigations
+
+What are the risks of this proposal and how do we mitigate.  Think
+broadly.  For example, consider both security and how this will impact
+the larger kubernetes ecosystem.
```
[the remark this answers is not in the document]
beekhof: Nod, I planned to come back to it but I'm not sure there is much to say here.

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
beekhof: There are two separate controllers: nodereboot and machineremediation
So I think we're prepared for that

beekhof: Updated proposal with details

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
beekhof: As-is, the annotation triggers the creation of a machine remediation CR, which triggers a the reboot controller to power the node off.  Once the node is off, the reboot controller deletes the Node object which also erases annotation ensuring that it does not trigger any additional power off events.

I've included this information in the proposal

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -0,0 +1,180 @@
...
+No changes are required to preserve the existing behaviour.
+
+### Version Skew Strategy
+
+By shipping the new controller with the baremetal-operator that it consumes, we
+can prevent any version mismatches.
+
+## Drawbacks [optional]
+
+At some point we may design a more formal signaling method than an annotation.
```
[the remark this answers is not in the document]
beekhof: updated to include

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
zaneb: OK, that sounds like it won't be able to prevent multiple MachineRemediation CRs getting created *before* the Node is removed though. Specifically, [this](https://github.com/kubevirt/machine-remediation/blob/master/pkg/controllers/nodereboot/nodereboot_controller.go#L124-L138) is not sufficient due to [caching](https://youtu.be/wMqzAOp15wo?t=290).
We'll want to address this at some point, but it doesn't need to be in this doc.

zaneb:
This seems like a reasonable short-term plan to me.
/approve

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
beekhof: ack, tracked as MGMT-324

beekhof:
Anything we need to push this forward?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Work Items
+
+- Make any requested modifications
+- Create a PR from  https://github.com/kubevirt/machine-remediation/ into https://github.com/metal3-io/baremetal-operator/
+
+### Dependencies
+
+This design is intended to integrate with OpenShift’s [Machine Healthcheck
+implementation](https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L407)
```
dhellmann: So the machine health check controller notices a problem with a node and adds the annotation. Then the new node reboot controller sees the annotation and creates a machine remediation CR. Then the machine remediation controller uses that CR to power off the host, delete the node, and power the host back on.

Is that right?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -0,0 +1,188 @@
...
+- A new [Machine Remediation CRD](https://github.com/kubevirt/machine-remediation/blob/master/pkg/apis/machineremediation/v1alpha1/machineremediation_types.go)
+- Two new controllers:
+  - [node reboot](https://github.com/kubevirt/machine-remediation/tree/master/pkg/controllers/nodereboot) which looks for the annoation and creates Machine Remediation CRs
+  - [machine remediation](https://github.com/kubevirt/machine-remediation/tree/master/pkg/controllers/machineremediation) which reboots the machine and deletes the Node object (which also erases the signalling annotation)
+- A new annotation (namespace and name is open for discussion)
+
+### Work Items
+
+- Make any requested modifications
+- Create a PR from  https://github.com/kubevirt/machine-remediation/ into https://github.com/metal3-io/baremetal-operator/
```
dhellmann: Why do we need to move any of the code into the baremetal-operator repository? Shouldn't these new controllers have their own repositories?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
beekhof: correct

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
dhellmann: For the other reviewers: Andrew and I discussed this a bit on slack. I pointed out that the BMO currently doesn't know anything about nodes or machines, and this new code will need to. I would prefer to keep the BMO isolated at the bottom of the stack, so we should consider whether these new controllers should go into the cluster-api-provider-baremetal repo or their own repo.

dhellmann: Between the two options, adding the controllers to CAPBM is going to result in fewer changes to OpenShift overall because we don't need to add a new image. I'm not sure whether it makes sense from a community perspective, but I think we can make the argument that rebooting a node is a feature of bare metal and should be part of the machine-api implementation.

PR #60, opened 2019-12-10
Add support for expected hardware configuration validation

digambar15:
- This spec is submitted against https://github.com/metal3-io/baremetal-operator/issues/351 issue.
- Add new section ExpectedHardwareConfiguration in baremetalhost_crd.yaml
- Implement validation and comparison of ExpectedHardwareConfiguration against introspection data.
- Add annotation and ERROR codes


digambar15:
@dhellmann @russellb @arkadykanevsky @pramchan Please review

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,263 @@
...
+     ExpectedNics          ExpectedNics         `json:"expectedNics"`
+     ExpectedRam           int                  `json:"expectedRam"`
+     ExpectedSystemVendor  ExpectedSystemVendor `json:"expectedsystemVendor"`
+    }
+    
+    type ExpectedCpu struct {
+     Count  int `json:"count"`
+    }
+    
+    type ExpectedDisk struct {
```
dhellmann: This seems to assume that all of the storage devices are the same size. Is that a safe assumption?

digambar15:
> I support the idea of this feature, but not this implementation. It adds complexity to the existing baremetal-operator code, API, and user experience for all users even though only some users need the feature. We can accomplish the same thing using a different controller, annotations, and labels in a way that is more flexible and allows the users who need the feature to add it on without introducing the burden for all users.
> 
> One way to do that is to define a new CRD to hold the expected hardware details, and write a controller that reconciles those by looking for host CRs that match and adding labels to them. Machines (and MachineSets) could then use the existing label selector field to find available hosts. If there are no matching hosts, reconciliation would wait until there are. Configuration errors would show up as Machine CRs that cannot be matched to a host after some period of time.
> 
> A variation on that approach is to use an annotation on each host to indicate what profile the admin thinks it should have, and have the new controller annotate the host with an error if the expectations do not match reality.
> 
> Both of those approaches place the responsibility for enforcing policy in the new controller and keep the existing baremetal-operator focused on provisioning and hardware management, which means that users who do not want strict hardware enforcement do not need to install it. Using labels takes advantage of an existing feature to match hosts and machines based on their hardware capabilities, and allows us to extend the system to support other types of policies for selecting hosts in a consistent way.

I got your point @dhellmann. Having seperate CRD will certainly help provide more detailed matching and validation mechanism. Thanks for reviewing this. I will submit another patch for New CRD approach.

design/expected-hardware-configuration-validation.md
digambar15: Here our thinking is to add minumum disk requirement. Not all disks should have same sizes because root disks sizes are different from data disks. I will add two sizes for disks 1 for root disks and 2 for remaining data disks that way we can set required configuration for disks

dhellmann:
> I got your point @dhellmann. Having seperate CRD will certainly help provide more detailed matching and validation mechanism. Thanks for reviewing this. I will submit another patch for New CRD approach.

You may want to take a look at https://github.com/metal3-io/metal3-docs/pull/4 for some background. I think that approach isn't fully viable as written, either, but you could combine the two ideas.

digambar15:
I will address given comments in next patch. Thanks @dhellmann @zaneb for your review

digambar15:
@dhellmann @zaneb We have updated PR as per your suggestion.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,263 @@
...
+     SizeBytesGB    int  `json:"sizeBytesGB"`
+     NumberOfDisks  int  `json:"numberOfDisks"`
+    }
+    
+    type ExpectedNics struct {
+     NumberOfNics   int  `json:"numberOfNics"`
+    }
```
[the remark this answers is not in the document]
digambar15: I think number of NICs will be needed. we can extend support for other attributes in the future like VLAN's etc

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,263 @@
...
+## Summary
+
+Code with the ability to perform the comparison in the baremetal-operator and put the host in an error state if the
+requirements do not match what is discovered.
+
+This should be a unique error state that indicates the inventory failed to match requirements, as opposed to, for instance, an IPMI Authentication Error, or a Memory Test Error (future).
+
+## Motivation
+
+We need to validate and compare expected hardware configuration against ironic introspection data.
```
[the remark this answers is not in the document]
digambar15: Yes, This is exactly what we are going to implement as part of this CRD.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Proposal
+
+We are creating proposal based on the discussion with metal3 community on issue #351:
+https://github.com/metal3-io/baremetal-operator/issues/351
+
+We compared introspection data of Ironic with metal3 schema and we found that default minimum expected hardware
+configuration needs to be added by introducing a new CRD `HardwareValidator`.
+
+We will write a controller that checks provisioned baremetal hosts against expected hardware configurtion and add label
```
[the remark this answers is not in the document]
digambar15: Yes, will correct it

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,199 @@
...
+        required:
+        - cpu
+        - nics
+        - ram
+        - disk
+    ```
+* Write a new API as baremetalhost/v1beta1 and new kind(CRD) HardwareValidator.
+    e.g.
+        
+       kubebuilder create api --group baremetalhost --version v1beta1 --kind HardwareValidator
```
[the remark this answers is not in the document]
digambar15: this controller will be created in new repo only. Definately Not in baremetel-operator

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,199 @@
...
+    
+        Create a new Validator.go file to write comparison and validation logic for provisioned baremetal hosts.
+        - Write a function which will have the expected hardware details and all the baremetal host list.
+        - Will pass above two inputs to validator function defined in validator.go file.
+	    - Write an algorithm to loop over all the hosts and check for comparison and validation of
+	    specs against the expected hardware details.
+	    - If host match found after execution of above algorithm, matched host will append to list.
+        - Return list to caller function.
+   
+    - According to list returned by validator function, will update the label for all hosts.
```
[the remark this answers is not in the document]
digambar15: Yes, it it possible. We are not going add annotation but certainly labels for every profile.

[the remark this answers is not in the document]
digambar15: We are finding a way how we can create multiple label for host

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,199 @@
...
+
```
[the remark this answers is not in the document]
digambar15: Yes, @dhellmann. For user point of view, adding multiple profils labels in host will be nice idea. We will change our implementation as we will check one host against multiple validator profile at a time. Also we will add a watch to the host list for the hardware setting changes and will update the labels accordingly.

design/expected-hardware-configuration-validation.md
digambar15: Yes, if provided CPU in profiles is 24 and if host has 48 CPU, then condition here is CPU >= 24, so that's the baseline criteria for this match.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,263 @@
...
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+      * [References](#references)
+
+## Summary
+
+Code with the ability to perform the comparison in the baremetal-operator and put the host in an error state if the
+requirements do not match what is discovered.
```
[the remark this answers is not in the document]
digambar15: Done

design/expected-hardware-configuration-validation.md
digambar15: It checks for minimums

dhellmann: "Expected" could mean "exact". Let's name the fields to make it clear they are minimums. "MinimumCPUCount" would be an integer with the count. (Maybe we also need MinimumCPUSpeed?) MinimumStorageGB, MinimumStorageDeviceCount, MinimumNICCount, etc.

dhellmann: On the other hand, @zaneb pointed out that if these values are always minimums we're likely to have a lot of overlapping classifications. A host that has > 100GB of storage also has > 10GB of storage, for example.

Would it be useful to have that host labeled twice?

How hard would it be to express these rules in a way that includes both min and max?

```
- property: storage
  min: 10
  max: 100
- property: nic
  count: 4
- property: ram
  min: 32
```

Ashughorla:
@dhellmann 
Please review PR for HWCC (hardware-classification-controller).
Added base framework for HWCC CRD.
Link : https://github.com/metal3-io/hardware-classification-controller/pull/2

design/expected-hardware-configuration-validation.md
digambar15: We can accomodate both parameters in yaml and implement the classification and comparison logic. It's not that hard for us to implement it but it cassifies host with both min and max values. 
I am not able to think scenario for max values ?

digambar15: > "Expected" could mean "exact". Let's name the fields to make it clear they are minimums. "MinimumCPUCount" would be an integer with the count. (Maybe we also need MinimumCPUSpeed?) MinimumStorageGB, MinimumStorageDeviceCount, MinimumNICCount, etc.

Agree, looks good to me.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,263 @@
...
+    }
+    
+    type ExpectedSystemVendor struct {
+     Name  string  `json:"name"`
+    }
+    ```
+
+* Changes In host_state_machine.go: 
+
+    Add "handleMatchDesiredHardwareProfile handler for''StateMatchDesiredProfile after "StateInspecting", where
```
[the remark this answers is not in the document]
digambar15: ok

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,263 @@
...
+                     IDRAC: 
+                          type: String
+            type: Object
+        required:
+        - cpu
+        - nics
+        - ram
+        - disk
+        - systemVendor
+        - firmware
```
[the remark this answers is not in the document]
digambar15: Removed some params

design/expected-hardware-configuration-validation.md
dhellmann: > We can accomodate both parameters in yaml and implement the classification and comparison logic. It's not that hard for us to implement it but it cassifies host with both min and max values.
> Max and MIin value scenario will hit where one profile needs min and max disk, cpu and ram etc, and this kind of request will certainly come in the future.

Right, I think this is going to be important immediately. If I want to classify my hosts based on their size, then I might say hosts with at least 8GB RAM are one type, 16GB RAM another, etc. If I can't express that as a range with both min and max, then every host that has 16GB will also be labeled as being of the type of host that has 8GB RAM, which will make it impossible to distinguish them and place them into separate pools.

Maybe we can build on the matchExpression format used in label selectors? https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

zaneb: It doesn't make it impossible, it just means that when you select Hosts for pools using labels you'll need to select for atLeast8GB and not atLeast16GB instead of just selecting for has8to15GB.
It's clear that the ergonomics of this are not good. It's less clear that the ergonomics of any alternatives are better.
Even then it's only true when you are using an autoscaler thing to make sure your hardware is 100% utilised. There may in fact be times when you just want a Host with at least 8GB of RAM, and if none with exactly 8GB are available but some with 16GB are then you just want to pick one of those. Although perhaps that's getting into the territory of a fully-fledged scheduler that might not make use of these labels.

digambar15: > > We can accomodate both parameters in yaml and implement the classification and comparison logic. It's not that hard for us to implement it but it cassifies host with both min and max values.
> > Max and MIin value scenario will hit where one profile needs min and max disk, cpu and ram etc, and this kind of request will certainly come in the future.
> 
> Right, I think this is going to be important immediately. If I want to classify my hosts based on their size, then I might say hosts with at least 8GB RAM are one type, 16GB RAM another, etc. If I can't express that as a range with both min and max, then every host that has 16GB will also be labeled as being of the type of host that has 8GB RAM, which will make it impossible to distinguish them and place them into separate pools.
> 
> Maybe we can build on the matchExpression format used in label selectors? https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

This looks fine, as it will be scalable also for future requirements. 
I think we should finalize this approach if all aggree with this.

digambar15: > It doesn't make it impossible, it just means that when you select Hosts for pools using labels you'll need to select for atLeast8GB and not atLeast16GB instead of just selecting for has8to15GB.
> It's clear that the ergonomics of this are not good. It's less clear that the ergonomics of any alternatives are better.
> Even then it's only true when you are using an autoscaler thing to make sure your hardware is 100% utilised. There may in fact be times when you just want a Host with at least 8GB of RAM, and if none with exactly 8GB are available but some with 16GB are then you just want to pick one of those. Although perhaps that's getting into the territory of a fully-fledged scheduler that might not make use of these labels.

Understood but there are many factors like this, I think to start with we should limit to min and max, we can add more of such rules in the future.

digambar15: I will talk to you both on slack today to finilize the approach for this classifier and what should go in into first phase and what should be planned for future requirements. I think this will help us defining further future requirements also.

dhellmann:
I've been considering this idea in the context of some of the other planned work, and I think if we work on how to express the rules more flexibly we can get a lot of useful benefits beyond validating hardware configurations. It could actually serve to drive provisioning in the correct way. Specifically, we have #76, which talks about specifying which drive should receive an image during provisioning and https://github.com/openshift/cluster-api-provider-baremetal/pull/52, which will eventually be brought upstream to metal3 and give us automatic provisioning when hosts are added to clusters.

If the rules for classifying a host allow us to label it so that it will be selected by the correct MachineSet, then we can solve the problem of how a user, who does not interact with the BareMetalHost much directly, would set the root device to receive the image. If the classifier rules and the MachineSet are pre-configured, then adding the host CR to the cluster would result in it being provisioned correctly.

What do you think, @mikkosest & @mhrivnak ?

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,209 @@
...
+
+    - Create a classification_manager.go file. Create a function manager() which will have inputs namespace, filter and BMH host_list. This function will call appropriate comparision logic written in classification_filter.go based on 'filter' specified by user in CRD_yaml.
+
+    - The classification_filter.go file will contain multiple comparision algorithms based on filters(i.e minimum and maximum). Create a ClassificationComparison() function which will have inputs fetched baremetalhost list and extracted hardware profile. Comparision function will check each host against profile and update the label if match found.
+
+
+* Create the Schema struct for `ExpectedHardwareConfiguration` inside `HardwareClassificationControllerSpec`,
+in file /api/v1alpha1/hardwareClassificationController_types.go.
+
+    ```yaml
```
dhellmann: I still think this is too limited. I would like for us to explore options for using more dynamic expressions in the CRD, so we don't have to add fields to it later in order to support other types of comparisons or tests. Here are a few expression parsing libraries that look interesting:

- https://github.com/antonmedv/expr 
- https://github.com/google/cel-go
- https://github.com/PaesslerAG/gval

The first one looks like it has support for reflection-based expressions.

We could also build something based on JSONPath syntax, since that is common elsewhere in Kubernetes, but it would mean building our own code to evaluate the operators and that might be more work than adopting one of the syntaxes supported by the existing libraries.

digambar15: Okay. I have gone through the expression concept, it looks interesting technique. I would like to keep customFilters also to implement more specific scenarios. I will go with https://github.com/antonmedv/expr for implementing expression based filter.
Here is a thing - 
We can add both filters in CR  - 
ClassificationFilters - 
   - CustomFilters: MinimumHardware
   - Expression: 'RAM >= 20' or  'RAM == 50' or 'RAM <= 96'

dhellmann: Before you do a bunch of work with that expression library, the feedback I've had from some of our internal experts is that we don't want to build a DSL as part of the API. It's better to do something like

```
- rules:
  - field: someNameHere
    operation: Equals
    value: 5
```

They gave as reference the way resources are exposed on a node https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4440-L4447 or "extended resources" on a node https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#extended-resources, which are used for things like device plugins https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/

digambar15: We will update the proposal based on above inputs.

digambar15: @dhellmann  We have updated proposal by adding two option for user to use anyof classification controller or expression rules filters. Also added representional and tasks in the proposal.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+
+    - Create a classification_manager.go file. If user provieds value for expectedHardwareConfiguration then write a function manager() which will have inputs namespace, customFilter and BMH host_list. This function will call appropriate filter written in classification_filter.go based on 'customFilter' specified by user in CRD_yaml. If user have set different expression rules instead of custom classification filters for filtering BMH then filtering will be done based on 'operation' value provided.
+
+    - The ClassificationFilters.go file will contain multiple comparision algorithms based on multiple custom classification filters(i.e minimum and maximum) or expression filters. Create a FactoryClassificationManager() function which will have inputs baremetalhost list and extracted hardware profile. FactoryClassificationManager's job is to call respective filtering algorithm and parse the inputs accordingly so respective filtering algorithm will check each host against profile and update the label like below if match found. 
+
+        ```yaml
+        "metadata": {
+            "labels": {
+                "profile.metal3.io/profile1" : "BMO-OK",
```
dhellmann, line 151: It would be nice if the label name and this CRD name matched.
```suggestion
                "hardwareclassification.metal3.io/profile1" : "matches",
```

design/expected-hardware-configuration-validation.md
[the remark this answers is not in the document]
dhellmann: This section is still talking about working on lists of hosts. I think that's going to lead to a lot of reconcile conflicts.

Instead of a prose version of the algorithm, let's try to write down what conditions will lead the controller to make different types of changes.

For example, what causes the controller to update a host (which CRDs does it watch)? I think it needs to watch both, and I think Reconcile() should work on one BareMetalHost at a time.

What happens if a host used to match a classification profile but no longer does? Do we remove the label or leave it there? I think it's better to remove the label, but will it have any side-effects that consumers of the labels need to be aware of?

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+                    dataDisk:
+                        sizeBytesGB:
+                            type: Integer
+                    numberOfDisks:
+                        type: Integer
+                    type: array
+            type: Object
+            nics:
+                properties:
+                    numberOfNICS:
```
dhellmann, line 88:
```suggestion
                    count:
```

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+            profileName:
+                    type: String
+            cpu:
+                properties:
+                    count:
+                        type: Integer
+            type: Object
+            disk:
+                properties:
+                    rootDisk:
```
dhellmann, line 75: How do we know which disk is the root disk? There's a lot of discussion about root disk hints in some other PRs that makes it clear to me that software not running on the host will have a hard time determining that, and even software running on the host will under some circumstances.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+                        sizeBytesGB:
+                            type: Integer
+                    type: Object
+                    dataDisk:
```
dhellmann, line 79: What if my host has 4 disks. Which is the data disk? Or is this a total for the size of non-root disk storage?

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+
+We are creating proposal based on the discussion with metal3 community on issue #351:
+https://github.com/metal3-io/baremetal-operator/issues/351
+
+We compared introspection data of Ironic with metal3 schema and we found that default  hardware
+configuration needs to be added by introducing a new CRD `HardwareClassificationController`.
+
+We will write a CRD that checks inspected baremetal hosts against hardware configuration and add label as "BMO-OK" for matched profile into baremetal host baremetalHost_crd.yaml.
+  
+### Implementation Details/Notes/Constraints
```
dhellmann, line 49: Generally for this section, I don't think we need to include the level of detail about which commands to run or what files will be created. We can stay at a higher level, and talk about creating the controller. We should be specific about the fields in the type, since that's the API we're defining.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+                        BIOS:
+                            type: String
+                        IDRAC:
+                            type: String
+                type: Object
+            required:
+            - customFilter
+            - namespace
+            - profileName
+        rules:
```
dhellmann, line 119: We don't need both the `expectedHardwareConfiguration` section and the `rules` section. The rules were meant to allow us to describe everything in the hardware configuration section more generally. We should pick one approach.

I proposed the general rules form, but I'm not set on that as a requirement if we can make the other form reasonably expressive. Let's see if we can work on that by writing out some example profiles.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+            disk:
+                properties:
+                    rootDisk:
+                        sizeBytesGB:
+                            type: Integer
+                    type: Object
+                    dataDisk:
+                        sizeBytesGB:
+                            type: Integer
+                    numberOfDisks:
```
dhellmann, line 82: Let's be consistent with CPUs.

```suggestion
                    count:
```

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+                        type: Integer
+            type: Object
+            ram:
+                properties:
+                    sizeBytesGB:
+                        type: Integer
+            type: Object
+            systemVendor:
+                properties:
+                    productName:
```
dhellmann, line 98: Are the string fields going to require exact matches or will substrings match?

design/expected-hardware-configuration-validation.md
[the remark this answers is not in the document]
dhellmann: Yes, let's leave the more complex stuff for now.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+## Design Details
+
+The following diagram shows the implementation details of Hardware Classification Controller.
+
+![Hardware Classification Controller Implementation Details](hwcc_sequence_diagram.png)
+
+
+### Work Items
+
+1. Implement CRD for `HardwareClassificationController`.
```
dhellmann, line 232: This list of items is also still talking about working on lists of hosts instead of one host at a time.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+
+    type Version struct {
+        RAID               string `json:"RAID"`
+        BasebandManagement string `json:"baseBandManagement"`
+        BIOS               string `json:"BIOS"`
+        IDRAC              string `json:"IDRAC"`
+    }
+
+    type Rules struct {
+        Feild       string  `json:"feild"`
```
mhrivnak, line 213: s/Feild/Field/

design/expected-hardware-configuration-validation.md
digambar15: ok

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+  
+### Implementation Details/Notes/Constraints
+
+Link for Existing Metal3 Specs
+Please refer metal3 spec for bare-metal:
+https://github.com/metal3-io/baremetal-operator/blob/master/deploy/crds/metal3.io_baremetalhosts_crd.yaml
+
+* Write a below schema for new CRD under folder deploy/crds for Kind HardwareClassificationController.
+
+    ```yaml
```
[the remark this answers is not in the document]
digambar15: ok

design/expected-hardware-configuration-validation.md
digambar15: okay. Usually we have separate requirement for root disks and separate for data disks. that's the reason we defined like that but I got the point. We will keep only disk sizeGB and numbers of disks field in the section.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+                    type: String
+            cpu:
+                properties:
+                    count:
+                        type: Integer
+            type: Object
+            disk:
+                properties:
+                    rootDisk:
+                        sizeBytesGB:
```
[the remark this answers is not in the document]
digambar15: sizeGB looks ok

design/expected-hardware-configuration-validation.md
digambar15: in this case, we are defining every disk should have minimum this size.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+                            type: Integer
+                    type: Object
+                    dataDisk:
+                        sizeBytesGB:
```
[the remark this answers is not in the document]
digambar15: All the params defined in HCC profile are minimum baseline params. If disks defined in HCC profiles - 
Disks: 8
SizeGB : 200G
Which means 8 disks should be present on the hosts and every disk should have minimum 200G size, greater than that will be acceptable criteria.

design/expected-hardware-configuration-validation.md
digambar15: ok

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+            systemVendor:
+                properties:
+                    productName:
+                        type: String
+                    manufacturer:
+                        type: String
+            type: Object
+            firmware:
+                properties:
+                    version:
```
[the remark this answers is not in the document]
digambar15: This section is currenly put in as an optional. If user has specific firmware requirement he can define it but most of the time that's not the case.
If you replace the host, then when you classify the host, HCC will catch it if firmware specified in profile

design/expected-hardware-configuration-validation.md
digambar15: Okay. We can discuss this thing in our call.

design/expected-hardware-configuration-validation.md
digambar15: We can discuss about it slack or on call.

design/expected-hardware-configuration-validation.md
digambar15: ok

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,262 @@
...
+
+    - Implement a new function fetchHost() which will fetch all baremetal hosts from Baremetal-Operator of namespace 'metal3' provided by user in CRD_yaml. This function will return a list of filtered hosts(hosts in status 'ready' or 'inspecting'). 
+
+    - In hardwareClassificationController_controller.go, reconcile function will call fetchHost()        function to fetch all baremetal hosts and also extract expectedHardwareConfiguration or expression rules from `metal3.io_HardwareClassificationController_crd.yaml`.
+
+    - Create a classification_manager.go file. If user provieds value for expectedHardwareConfiguration then write a function manager() which will have inputs namespace, customFilter and BMH host_list. This function will call appropriate filter written in classification_filter.go based on 'customFilter' specified by user in CRD_yaml. If user have set different expression rules instead of custom classification filters for filtering BMH then filtering will be done based on 'operation' value provided.
```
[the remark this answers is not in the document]
digambar15: we are going to implement comparison based on provided profile parameters. This idea came in before the expression filter. I had discussion with Doug on this in our last meeting, we kind of decided we should use either of these.

design/expected-hardware-configuration-validation.md
digambar15: Yes

design/expected-hardware-configuration-validation.md
digambar15: ok

design/expected-hardware-configuration-validation.md
digambar15: ok

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,138 @@
...
+   metadata:
+     name: profile1
+   spec:
+   expectedValidationConfiguration:
+      namespace: metal3
+      CPU:
+         minimumCount: 4
+         minimumSpeed: 4.3
+      Disk:
+         minimumSizeGB: 2200
```
dhellmann, line 72: We should make clear whether this needs to be the total amount of storage or the per-disk amount. Maybe call the field `minimumTotalSizeGB` and if we need to do something about per-disk size later we can add more fields. What do you think?

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,138 @@
...
+         minimumCount: 2
+      NICS:
```
dhellmann, line 74: This is the only example that is plural. Maybe just `NIC`?

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,138 @@
...
+         minimumSizeGB: 32
+   ```
+   Similarly, user can mention maximum requirements for CPU, Disk, NICS and RAM.
+
+* Write a new API as hardware-classification/api/v1alpha1 and new kind(CRD) HardwareClassification.
+    
+    - Create the Schema struct for `ExpectedHardwareConfiguration` inside `HardwareClassificationSpec`,
+    in file /api/v1alpha1/hardwareClassification_types.go.
+
+    - Implement a new function fetchHost() which will fetch all baremetal hosts from Baremetal-Operator of namespace 'metal3' provided by user in CRD_yaml. We will add a watch on the baremetalhost kind for the hardware setting changes.
```
dhellmann, line 86: There is work going on (maybe it has landed) to make the baremetal-operator optionally able to look at all namespaces. I think we will need this controller to work the same way.

design/expected-hardware-configuration-validation.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Risks and Mitigations
+
+None
+
+## Design Details
+
+The following diagram shows the implementation details of Hardware Classification Controller.
+
+![Hardware Classification Controller Implementation Details](hwcc_sequence_diagram.png)
```
dhellmann, line 103: The diagram shows the operation to fetch the list of hosts communicating with the baremetal-operator. That call will be a regular API call, and won't need the operator to be involved.

design/expected-hardware-configuration-validation.md
digambar15: Understand your point. 
We specified it as per disk size considering each disk should have minimum size which user will expect to have, so can we put like minumumPerDiskSizeGB or minimumPerSizeGB?

design/expected-hardware-configuration-validation.md
digambar15: ok

design/expected-hardware-configuration-validation.md
digambar15: okay. I will discuss about it with you to get more details on it.

design/expected-hardware-configuration-validation.md
digambar15: okay. Yeah.

design/expected-hardware-configuration-validation.md
dhellmann: How about minimumDiskSizeGB and minimumTotalSizeGB?

dhellmann: Or minimumIndividualSizeGB and minimumTotalSizeGB

design/expected-hardware-configuration-validation.md
dhellmann: https://github.com/metal3-io/baremetal-operator/pull/471

dhellmann: When we discussed this, we decided to start that the classification resource would be namespace scoped and only apply to hosts in the same namespace.

dhellmann:
I'm happy with this latest draft. Let's give a few days for other community members to review.

@digambar15 do you want to add this to the community meeting to ask for more reviews?

digambar15:
> > I'm happy with this latest draft. Let's give a few days for other community members to review.
> > @digambar15 do you want to add this to the community meeting to ask for more reviews?
> 
> Thanks @dhellmann for your help.
> 
> Yes, I will add it in today's metal3 meeting.
Added in

digambar15:
> > I'm happy with this latest draft. Let's give a few days for other community members to review.
> > @digambar15 do you want to add this to the community meeting to ask for more reviews?
> 
> Thanks @dhellmann for your help.
> 
> Yes, I will add it in today's metal3 meeting.

added in today's meeting

dhellmann:
In the meeting on 15 April I believe we said we would hold this open for a few days for further comments. I will approve it on 20 April unless there are objections before then.

PR #61, opened 2019-12-12
Documents how to safely remove a BareMetalHost from a cluster.

mhrivnak:
Let me know if you think this would be better in the cluster-api-provider-baremetal repo. I looked there first, but didn't see other similar docs. I could see it going either there or here, but at least here there seemed to be a critical mass of docs describing how things work.

### DIFF
--- a/design/remove-host.md
+++ b/design/remove-host.md
@@ -1,0 +1,54 @@
+# Remove a Host from a Cluster
+
+At some point you will need to remove a host from a cluster. You may be
+removing failed hardware, downsizing a healthy cluster, or have some other
+reason.
+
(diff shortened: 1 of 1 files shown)


design/remove-host.md
```diff
@@ -0,0 +1,54 @@
...
+guidance on how to do so.
+
+## Steps
+
+These steps are both safe and compatible with automation that scales
+MachineSets to match the number of BareMetalHosts.
+
+### Annotate the Machine
+
+Find the Machine that corresponds to the BareMetalHost that you want to remove.
```
dhellmann: We could mention that the information is available in the host consumerRef.

PR #64, opened 2019-12-23
Add support for NUMA params in BareMetalHost Schema

digambar15:
- Some parameters was missing from existing baremetalhost_crd.yaml schema.
- The missing params mentioned in this spec will be useful for performance and future workloads
- Numa params will be useful for compute intensive applications and NFV applications
- Boot parameters are required because it shows current boot mode along with pxe interface details which will be then used for setting boot mode.


design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Proposal to enhance Schema for Metal3
```
dhellmann: I can imagine that over time we will have a lot of schema enhancement requests. Could you give this proposal a more specific title (and rename the file to match)?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+we recommend following parameters need to be added in schema of `metal3` for `NFV` 
+workloads.
+1. Add enhancement in the existing `metal3` schema for the following parameters:
+
+    CPU Parameters:
+    
+    Parameters for `NUMA` topology:
+     
+    For better performance, selection of CPUs based on the `NUMA` topology 
+    becomes necessary. In case of nodes with `DPDK` aware NICs, the CPUs for
```
dhellmann: This information seems useful for scheduling workloads, but I don't understand how it would be used during provisioning. Can you elaborate on that?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+              type: object
+            type: array
+            ram:
+              properties:
+                numa_node:
+                  type: integer
+                  description: Numa node id
+                size_kb:
+                  type: string
+                  description: memory in kb
```
dhellmann: [Elsewhere](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/apis/metal3/v1alpha1/baremetalhost_types.go#L301) we track RAM using Mebibytes as an integer. Is there any particular reason to switch to a string holding KB here?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+            cpus:
+              properties:
+                cpu:
+                  type: integer
+                  description: cpu id 
+                numa_node:
+                  type: integer
+                  description: Numa node id
+                thread_siblings:
+                  items:
```
dhellmann: I don't know what thread siblings are. Is this a list of integers? What do the integers represent?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+    with the `DPDK` NICs. If hyperthreading is enabled, then selection of the 
+    logical cores requires the knowledge of the siblings.  
+
+    Declarative design idea:
+    ````yaml
+    Cpu
+     …
+    numa_topology:
+      properties:
+        numa_node_0:
```
dhellmann: This implies that nodes have names. Is that right, or are the names made up?

Are the nodes ordered? Should numa_topology be a list of node objects instead of a mapping?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+          properties:
+            nics:
+              properties:
+                numa_node:
```
dhellmann: Is the numa_node value for a NIC ever going to be different from the name as defined on line 85?

We list NICs separately. Does it make sense to add a field to that existing structure to hold the NUMA node ID instead of repeating the NIC name here in this new structure?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+                  type: integer
+                  description: Numa node id
+                name:
+                  type: string
+                  description: network interface name
+              type: object
+            type: array
+            ram:
+              properties:
+                numa_node:
```
dhellmann: Since we only track RAM as a total value elsewhere, it probably makes sense to keep this section to list the breakdown. But the same question applies as above: is the numa_node ID value here ever going to be different from the node ID from line 85?

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+    Where,
+    * `ram` a mapping from memory available to a `NUMA` node
+    * `cpus` a mapping from physical CPU ID to a `NUMA` node and a list of sibling threads
+    * `nics` a mapping from NIC names to `NUMA` node
+
+    Boot Parameters:
+    
+    * current_boot_mode: `uefi`
+        
+        UEFI is installed when device is manufactured. It is the first program which runs when the device is turned on. UEFI replace BIOS.
```
dhellmann: We've had several discussions about how best to represent the boot mode for hosts. I'm not sure we've come to a conclusion, yet. We might make more progress if we separate the proposal for boot mode from the proposal to collect NUMA topology details.

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+    * pxe_interface: `ff:ff:ff:ff:ff:ff`
```
dhellmann: How is this different from [the existing bootMACAddress field](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/apis/metal3/v1alpha1/baremetalhost_types.go#L129)?

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: Sure @dhellmann

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: For provisioning, this is not useful. We are adding this for future workload scenarios where user can look at the CR and plan there workloads accordingly.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: I think it should MB, will correct it.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: list of integers. Integers represents the CPU's

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: Numa node 0 means numa socket 0. It's not a new hardware node.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: This association of NIC defines which nic is assiciated with which NUMA node.
Updating NUMA association with NIC in NIC structure is a good idea. That way, user can identify NUMA to NIC association immidiately by looking at NIC

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: RAM value is same for any NUMA node (socket).

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: Okay, I will create seperate proposal for boot mode

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: Its same but it makes sense if we keep it under boot mode section because that way, user get to know which pxe nic is associated with boot mode.

design/Proposal-to-enhance-schema-for-metal3.md
dhellmann: It seems like this information could be expressed more compactly with a list of lists of integers.

```
cpu_thread_siblings:
   - [1, 17]
   - [10, 26]
   - ...
```

How does a consumer typically use the data? Is there any reason not to favor the more compact representation?

design/Proposal-to-enhance-schema-for-metal3.md
dhellmann: The bootMACAddress *is* the MAC we expect to be used for PXE.

design/Proposal-to-enhance-schema-for-metal3.md
```diff
@@ -0,0 +1,228 @@
...
+![Implementation-flow](enhance-schema-for-metal3-flow.png)
+
+*The changes are illustrated  in Implementation Details section.
+
+### Work Items
+
+1. Introduce new section called `boot` under `hardware` in schema `metal3.io_baremetalhosts_crd.yaml`.
+2. Update the existing sections for `cpu` which will include `NUMA` related parameters in schema.
+3. Define structure for `boot` parameters in `baremetalhosts_types.go`.
+4. Define structure for `NUMA` parameters and call it in `cpu` structure in `baremetalhost_types.go`.
```
dhellmann: Does the gopher cloud library already return the NUMA data?

design/Proposal-to-enhance-schema-for-metal3.md
dhellmann: https://kcsna2019.sched.com/event/Vv0W/present-and-future-of-hardware-topology-awareness-in-kubelet-connor-doyle is relevant for the scheduler using NUMA topology. I wouldn't want to reproduce data in the BareMetalHost that is going to be available somewhere else, like the node.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: No, We will have to add structure for NUMA in gophercloud lib. We will have to talk to gophercloud community for that or create a PR/issue for that. I welcome your help on that.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: We can present NUMA node siblings in list of list for respective nodes. It is possible.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: Yes, I agree

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: It depends on people's need. Good to have all NUMA info can be associated with respective CPU, RAM or NIC so that it will be helpful for user.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: ok.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: Anyway I am creating seperate PR For boot. will address these things there.

design/Proposal-to-enhance-schema-for-metal3.md
digambar15: No, We need to add NUMA structure in gopher library to support this feature. I welcome your help on that.

design/Proposal-to-enhance-schema-for-metal3.md
dhellmann: OK, I think I've talked myself into adding all of the NUMA related information into a new structure, so let's work on defining that.

design/Proposal-to-enhance-schema-for-metal3.md
dhellmann: That gophercloud task should be on the work list here.

digambar15: Yes, will add

zaneb:
I have one crazy idea about this - see what you think.
It might be nice if the data structures for each NUMA node were in some sense equivalent to the one for the entire system, even if that means duplicating a lot of the data. e.g. something like:
```
type NodeDetails struct {
    CPU CPU
    NIC []NIC
    RAMMebibytes int
}

type HardwareDetails struct {
    SystemVendor HardwareSystemVendor
    Firmware     Firmware
    NodeDetails
    NumaNodes    []NodeDetails
    Storage      []Storage
    Hostname     string
}
```
(This is related to my other crazy idea that k8s should avoid going down the OpenStack path of doing complex scheduling that takes into account NUMA and instead just run multiple kubelets on the same machine, one for each NUMA node, separated by cgroups. Note that I haven't shared this idea with anyone yet, so it may be legitimately crazy.)
What's missing from the current CPU/NIC/RAM data that would prevent us from doing this?
* The list of CPU IDs is certainly one. For the whole machine we currently only provide a count, since you can assume the IDs are contiguous.
* The list of hyperthread siblings. This is not specific to NUMA though, so it might be a valuable addition anyway. If formatted as a list of sibling lists as Doug suggested then that could eliminate the need to separately list the CPU IDs (since they'd each occur in the thread sibling list).
* Anything else?

dhellmann:
> I have one crazy idea about this - see what you think.

I like it.

digambar15:
> I have one crazy idea about this - see what you think.
> It might be nice if the data structures for each NUMA node were in some sense equivalent to the one for the entire system, even if that means duplicating a lot of the data. e.g. something like:
> 
> ```
> type NodeDetails struct {
>     CPU CPU
>     NIC []NIC
>     RAMMebibytes int
> }
> 
> type HardwareDetails struct {
>     SystemVendor HardwareSystemVendor
>     Firmware     Firmware
>     NodeDetails
>     NumaNodes    []NodeDetails
>     Storage      []Storage
>     Hostname     string
> }
> ```
> 
> (This is related to my other crazy idea that k8s should avoid going down the OpenStack path of doing complex scheduling that takes into account NUMA and instead just run multiple kubelets on the same machine, one for each NUMA node, separated by cgroups. Note that I haven't shared this idea with anyone yet, so it may be legitimately crazy.)
> What's missing from the current CPU/NIC/RAM data that would prevent us from doing this?
> 
> * The list of CPU IDs is certainly one. For the whole machine we currently only provide a count, since you can assume the IDs are contiguous.
> * The list of hyperthread siblings. This is not specific to NUMA though, so it might be a valuable addition anyway. If formatted as a list of sibling lists as Doug suggested then that could eliminate the need to separately list the CPU IDs (since they'd each occur in the thread sibling list).
> * Anything else?

Looks good to me. It certainly value adds and avoids repetitions.
Let's me update this proposal accordingly.

design/Add-support-for-NUMA-params-in-BareMetalHost-Schema
```diff
@@ -0,0 +1,280 @@
...
+        SystemVendor HardwareSystemVendor  `json:"systemVendor"`
+        Firmware     Firmware              `json:"firmware"`
+        NodeDetails  NodeDetails
+        NumaNodes    []NumaNodes
+        Storage      []Storage             `json:"storage"`
+        Hostname     string                `json:"hostname"`
+    }
+    
+    type NumaNodes struct {
+        NumaNodeId      int
```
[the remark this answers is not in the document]
digambar15: IDs here are NumaNode 0 and NumaNode 1. so we will have only 0 and 1 here.

design/Add-support-for-NUMA-params-in-BareMetalHost-Schema
```diff
@@ -0,0 +1,280 @@
...
+        RAMMebibytes int    `json:"ramMebibytes"`
+    }
+    
+    type CPU struct {
+        Arch           string     `json:"arch"`
+        Model          string     `json:"model"`
+        ClockMegahertz ClockSpeed `json:"clockMegahertz"`
+        Flags          []string   `json:"flags"`
+        Count          int        `json:"count"`
+        threadsiblings [][]int    `json:"thread_siblings"`
```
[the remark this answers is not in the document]
digambar15: Will correct it.

design/Add-support-for-NUMA-params-in-BareMetalHost-Schema
```diff
@@ -0,0 +1,280 @@
...
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+      * [References](#references)
+<!--te-->
+
+## Summary
+
+Enhancing the `BareMetalHost` schema by adding useful parameters that we received
```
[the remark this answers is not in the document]
digambar15: Will update it in next commit.

design/Add-support-for-NUMA-params-in-BareMetalHost-Schema
```diff
@@ -0,0 +1,280 @@
...
+
+## Motivation
+
+To make the existing `BareMetalHost` schema compatible and scalable for heavy computation 
+based use cases.  
+
+### Goals
+
+To compare the introspection data fetched from `Ironic-Inspector` through 
+`Gopher-cloud client` library with `BareMetalHost` schema and add useful parameters
```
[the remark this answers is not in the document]
digambar15: Yes, will correct it.

design/Add-support-for-NUMA-params-in-BareMetalHost-Schema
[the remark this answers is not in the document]
zaneb: In that case it seems like we could just use the index into the list, i.e. do

  NumaNodes    []NodeDetails

above, and just get rid of this NumaNodes struct.

sharajat:
In the above Numa structure we found unecessary repetation of data in case of Nic and cpu, as the structure remains same for the data we fetch from getcpudetails() and cpu data (threadsiblings) from getnumadetails(), So it would be better if we keep Numa topology in a seperate structure inside CPU.
Below is the proposed sample structure of cpu with NUMA -: 

```
cpu
{
	Arch: x86_64 
	Model: Intel(R) Xeon(R) Gold 6126 CPU @ 2.60 GHz 
	ClockMegahertz: 3700 
	Flags: [3 eopt xtopology xtpr] 
	Count: 48 
	NumaTopology: [{
		NumaNodeID: 0 
		Numadetails: {
			Nics: [p2p1 p2p2 p1p2 p1p1 em4 em2 em1 em3] 
			RAM: 99289532 
			ThreadSiblings: [[20 44][8 32][22 46][10 34][0 24][18 42][6 30][12 36][4 28][14 38][2 26][16 40]]
		}
	}, 
	{
		NumaNodeID: 1 
		Numadetails: {
			Nics: [] 
			RAM: 100663296 ThreadSiblings: [[3 27][23 47][9 33][15 39][7 31][21 45][5 29][19 43][11 35][17 41][1 25][13 37]]
		}
	}]
}
```

zaneb:
> In the above Numa structure we found unecessary repetation of data in case of Nic and cpu

I'm not sure if you read my [original comment](https://github.com/metal3-io/metal3-docs/pull/64#issuecomment-572167900) where I proposed this structure, but it explicitly stated this at the beginning (emphasis added):

> It might be nice if the data structures for each NUMA node were in some sense equivalent to the one for the entire system, **even if that means duplicating a lot of the data**.

So if you disagree (which is perfectly fine - I did call it a crazy idea) then perhaps you could state how the downsides of duplicating the data outweigh the upsides of being able to look at a NUMA node as if it were a whole host.

One issue with your proposal is that it only provides information about thread siblings on hosts with NUMA.

arkadykanevsky:
For
ThreadSiblings: [[3 27][23 47][9 33][15 39][7 31][21 45][5 29][19 43][11 35][17 41][1 25][13 37]]
I assume there is built in assumption that every core up to 	Count: 48 
is listed at some NumaNode?
Let's document it.

pramchan:
> For
> ThreadSiblings: [[3 27][23 47][9 33][15 39][7 31][21 45][5 29][19 43][11 35][17 41][1 25][13 37]]
> I assume there is built in assumption that every core up to 	Count: 48
> is listed at some NumaNode?
> Let's document it.

pramchan:
For NUMA
It applies only for Multiprocessor host, as otherwise NUMA the Non Uniform Memory Access for faster access between local-memory and non-local-memory of another processor does not arise. Thus there should be no logical  need for NUMA parameter for single processor host.

design/Proposal-to-enhance-schema-for-metal3.md
sharajat: Numa details in gophercloud merged successfully 
https://github.com/gophercloud/gophercloud/pull/1842

sharajat:
> > In the above Numa structure we found unecessary repetation of data in case of Nic and cpu
> 
> I'm not sure if you read my [original comment](https://github.com/metal3-io/metal3-docs/pull/64#issuecomment-572167900) where I proposed this structure, but it explicitly stated this at the beginning (emphasis added):
> 
> > It might be nice if the data structures for each NUMA node were in some sense equivalent to the one for the entire system, **even if that means duplicating a lot of the data**.
> 
> So if you disagree (which is perfectly fine - I did call it a crazy idea) then perhaps you could state how the downsides of duplicating the data outweigh the upsides of being able to look at a NUMA node as if it were a whole host.
> 
> One issue with your proposal is that it only provides information about thread siblings on hosts with NUMA.

We are testing with your approach on the real hardware, We will get back to you once we get results.

PR #66, opened 2020-01-14
proposal to add missing boot parameters

sharajat:
Proposal to add missing boot parameter


digambar15:
/assign @dhellmann @zaneb 

Hi @dhellmann, As per our discussion on PR #64 we created separate proposal for Boot params. Please review.
Thanks

design/proposal-to-add-boot-parameters-for-metal3-schema.md
```diff
@@ -0,0 +1,135 @@
...
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+      * [References](#references)
+<!--te-->
+
+## Summary
+
+Enhancing the `metal3` schema by adding `BOOT` parameters that we received
```
zaneb: This would be a lot clearer if it said something like "Enhance the `HardwareDetails` section of the BaremetalHost schema". metal3 is a namespace, not a Custom Resource type.

design/proposal-to-add-boot-parameters-for-metal3-schema.md
```diff
@@ -0,0 +1,135 @@
...
+through `gopher-cloud client` library. 
+
+Having fetched BareMetal Node introspection data from Ironic-Inspector and
+compared it with BareMetalHost schema of `metal3` `(metal3.io_baremetalhosts_crd.yaml`), 
+we recommend following parameters need to be added in schema of `metal3`.
+1. Add enhancement in the existing `metal3` schema for the following parameters:
+
+    Boot Parameters:
+    
+    * current_boot_mode: `uefi`
```
zaneb: I don't really see the use of this. In future, the user will be able to select the boot mode they want in the CR (once metal3-io/baremetal-operator#302 or something like it merges). Knowing the boot mode that was used to boot the inspection image is not that interesting.
A list of the *available* boot modes could be interesting. I'm not sure we can get that from inspection though. In fact, it's arguably too late at that point.

design/proposal-to-add-boot-parameters-for-metal3-schema.md
```diff
@@ -0,0 +1,135 @@
...
+        
+        UEFI is installed when device is manufactured. It is the first program which runs when the device is turned on. UEFI replace BIOS.
+    * pxe_interface: `ff:ff:ff:ff:ff:ff`
```
zaneb: The NIC structure already has a flag indicating whether an interface allows PXE. Why is that not sufficient?

design/proposal-to-add-boot-parameters-for-metal3-schema.md
digambar15: Okay.

design/proposal-to-add-boot-parameters-for-metal3-schema.md
digambar15: See, Usually as an Operator/admin, I would like to know existing boot mode before setting another boot mode. This is lot more useful for planning server boot management and it's PXE management accordingly. Because PXE settings are dependent on boot modes, so understanding existing boot mode will be useful for operator. Also it avoids resetting of same boot mode also.

design/proposal-to-add-boot-parameters-for-metal3-schema.md
digambar15: Actually PXE interface is tightly associated with boot, It would be helpful for admin/operator also. Usually boo mode and PXE_interface are placed side by side.

design/proposal-to-add-boot-parameters-for-metal3-schema.md
digambar15: ok

design/proposal-to-add-boot-parameters-for-metal3-schema.md
zaneb: Inspection can't tell you the existing boot mode though. It can only tell you that it was booted once in that mode some time in the past. That's why I'm not sure it's that useful.

digambar15: By looking at existing boot mode,  it is helpful for user to avoid same boot mode again.

sharajat: I agree with Digambar, If user wants to change boot mode later after, It would be really helpful.

zaneb: > That makes it more important to add a field to set the desired boot mode, which this proposal doesn't cover.

Noting for the record that that discussion is already happening simultaneously in #63

design/proposal-to-add-boot-parameters-for-metal3-schema.md
digambar15: OKay, We will just keep current_boot_mode field and remove the pxe_interface.
Thanks @dhellmann

design/proposal-to-add-boot-parameters-for-metal3-schema.md
digambar15: > It's unfortunate that there is a window of time between inspection and provisioning where the current state may be shown incorrectly. That's the case for all of the data we get through inspection, though, and any of the rest of the hardware details can change, too. That makes it more important to add a field to set the desired boot mode, which this proposal doesn't cover. Is there a reason for not including that field?

Hey, We are addressing desired_boot_mode functionality as part of another PR #63

digambar15: > > That makes it more important to add a field to set the desired boot mode, which this proposal doesn't cover.
> 
> Noting for the record that that discussion is already happening simultaneously in #63

Yes

digambar15:
@dhellmann @zaneb We have addresed your comments. Please review.

design/proposal-to-add-boot-parameters-for-metal3-schema.md
zaneb: I'm just confused, because the obvious way to store the "current boot mode" is in the Provisioning data, at the time you boot it, based on the settings the user provided via #63 and that you actually used to boot it. Nobody has yet explained why it would be better to store a boot mode that is anything but current in the HardwareDetails and call it "current boot mode".

PR #78, opened 2020-03-05
design for determining the boot mode from other data

dhellmann:
This is an alternative design to replace the ones that call for adding
a boot mode field to the BareMetalHost.

### DIFF
--- a/design/implicit-boot-mode.md
+++ b/design/implicit-boot-mode.md
@@ -1,0 +1,203 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# implicit-boot-mode
+
+## Status
+
+provisional
+
+## Table of Contents
+
+<!--ts-->
+   * [Title](#title)
+      * [Status](#status)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [User Stories [optional]](#user-stories-optional)
+            * [Story 1](#story-1)
+            * [Story 2](#story-2)
+         * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Work Items](#work-items)
+         * [Dependencies](#dependencies)
+         * [Test Plan](#test-plan)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Drawbacks [optional]](#drawbacks-optional)
+      * [Alternatives [optional]](#alternatives-optional)
+      * [References](#references)
+
+<!-- Added by: stack, at: 2019-02-15T11:41-05:00 -->
+
+<!--te-->
+
+## Summary
+
+The required boot mode for a host depends on several factors,
+including support in the hardware, whether IPv6 is being used, and
+whether virtual media is being used. We do not need to support all
+combinations of those features with all boot modes. Asking the user to
+give us the boot mode explicitly introduces an opportunity for them to
+specify an incorrect value. We have enough information about the host
+already to determine the correct boot mode to use. Therefore we should
+not need to extend the BareMetalHost API in order to support multiple
+boot modes.
+
+## Motivation
+
+When booting using virtual media, Ironic builds a custom ISO for each
+host. Today, it only includes the correct contents for one boot mode
+at a time. Some attempt is made to determine the right value by
+examining the hardware, but that is not always possible. When
+necessary, there is a way to configure a default.
+
+Several proposals have been floated to introduce a way for the user to
+explicitly set a boot mode. This proposal asserts that we can
+successfully manage the x86 hardware supported by metal3 today without
+it. Avoiding a new API field avoids several opportunities for
+misconfiguration.
+
+### Goals
+
+* Describe a method for providing a deterministic boot mode value to
+  Ironic without relying on the user to set the value directly.
+* Ensure that metal3's behavior is always deterministic.
+
+### Non-Goals
+
+* Address boot mode selection for non-x86 hardware.
+
+## Proposal
+
+For each host, Ironic maintains a set of properties telling it how to
+manage the hardware. One of these is the `boot_mode`. Valid values
+are `bios` and `uefi`. When the `boot_mode` property is
+explicitly set, Ironic will try to ensure that the hardware is
+configured accordingly and it may also choose different paths in the
+provisioning process. In particular, there is a branch in the ISO
+building logic to include different content based on the boot mode.
+
+If the `boot_mode` is not specified, Ironic tries to determine the
+right value by looking at the hardware settings. Some BMC types do not
+support this capability, especially when using the IPMI protocol. When
+Ironic cannot determine the right value for itself, and has not been
+told explicitly what value to use, it falls back to a global default
+specified by a configuration option. Metal3 is not typically deployed
+in a way that allows that option to be changed easily.
+
+We want to ensure that metal3's behavior is always deterministic and
+predictable. We also want to keep the API as minimal as possible, and
+to avoid introducing API fields in a way that makes it likely that a
+user would either need to tell us something we can figure out or that
+they could tell us something incorrect.
+
+Given those goals, it is useful to examine the different situations in
+which different boot modes are appropriate, and how to select the
+right mode at the right time.
+
+The first approach would be to try to always require the same boot
+mode for all hosts. Most modern x86 hosts support UEFI. It is
+therefore useful as a default. However, many contributors have
+somewhat older hardware, which works better with the BIOS boot
+mode. While this hardware is not necessarily our target, it isn't
+practical to abandon support entirely by always requiring
+UEFI. Similarly, there are reasons we cannot always require a host to
+use the BIOS boot mode. UEFI supports desirable features, such as
+secure booting, non-x86 hardware, and PXE on IPv6 networks. For these
+reasons, a single boot mode will not work.
+
+The desired features do give us some hints about how we can determine
+the boot mode implicitly, however. For example, we have no plans to
+support virtual media with IPMI. We can also tell whether the
+provisioning network is using IPv4 or IPv6. Using this information, we
+could build a set of rules to determine the boot mode. However, we
+would have to document those rules, and they may not seem obvious to
+the end-user.
+
+At least some of the features are tied to the protocol used to
+communicate with the BMC. We could therefore link the boot mode to the
+BMC driver, and have a deterministic method for choosing the boot mode
+that is also easy to document.
+
+The proposal, therefore, is to always use BIOS mode when using IPMI
+and to prefer UEFI when using Redfish. We can implement this by adding
+a new method to the [AccessDetails
+interface](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/access.go#L27)
+to return the properties for a host, including the `boot_mode`. This
+API is consistent with other similar methods in that class that return
+other sets of data passed to Ironic when a host is registered.
+
+We can be somewhat flexible with the rules by allowing the BMC driver
+in metal3 to look at some of the other settings for the host to decide
+which boot mode to set. For example, when the driver author knows that
+Ironic supports determining the boot mode correctly, it could choose
+to skip setting a mode unless the other settings (use of virtual media
+or IPv6, for example) require a particular value. This would
+complicate the documentation, but may be necessary for some drivers
+where trying to have Ironic set the boot mode frequently causes
+provisioning failures and it is better to use the host's existing
+configuration when possible.
+
+In order to be transparent about what metal3 is doing, we can add a
+status field to show the boot mode used. The field's name could be
+something like `Status.Provisioning.BootMode`, and it would contain a
+string for either `"BIOS"` or `"UEFI"` for now and possibly other
+values later for other hardware platforms.
+
+### User Stories [optional]
+
+#### As a user, I want to provision a host using virtual media instead of PXE
+
+#### As a user, I want to provision a host that uses an IPMI-based BMC
+
+#### As a user, I want to provision a host using an IPv6 provisioning network
+
+### Risks and Mitigations
+
+## Design Details
+
+- Update the [AccessDetails
+  interface](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/access.go#L27)
+  to include the new method, `NodeProperties()`.
+- Update each implementation of the interface to include the new
+  method with at least a static return value.
+- Update the default for Ironic to be UEFI.
+
+### Dependencies
+
+N/A
+
+### Test Plan
+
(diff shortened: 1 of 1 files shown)


design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+configured accordingly and it may also choose different paths in the
+provisioning process. In particular, there is a branch in the ISO
+building logic to include different content based on the boot mode.
+
+If the `boot_mode` is not specified, Ironic tries to determine the
+right value by looking at the hardware settings. Some BMC types do not
+support this capability, especially when using the IPMI protocol. When
+Ironic cannot determine the right value for itself, and has not been
+told explicitly what value to use, it falls back to a global default
+specified by a configuration option. Metal3 is not typically deployed
```
demonCoder95: What is this global default fallback? Is it simple as ```BIOS``` or ```UEFI``` or something else?

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+host. Today, it only includes the correct contents for one boot mode
+at a time. Some attempt is made to determine the right value by
+examining the hardware, but that is not always possible. When
+necessary, there is a way to configure a default.
+
+Several proposals have been floated to introduce a way for the user to
+explicitly set a boot mode. This proposal asserts that we can
+successfully manage the x86 hardware supported by metal3 today without
+it. Avoiding a new API field avoids several opportunities for
+misconfiguration.
```
demonCoder95: I suppose there wouldn't be a need to have explicit documentation, which says 'boot mode is configured automatically'. Since we would want to hide as much of Ironic stuff behind the API as possible. Yes?

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+<!--te-->
+
+## Summary
+
+The required boot mode for a host depends on several factors,
+including support in the hardware, whether IPv6 is being used, and
+whether virtual media is being used. We do not need to support all
+combinations of those features with all boot modes. Asking the user to
+give us the boot mode explicitly introduces an opportunity for them to
+specify an incorrect value. We have enough information about the host
```
demonCoder95: Just out of curiosity, if we _were_ to implement explicit config of ```boot_mode```, just the way it is in Ironic, it would be an addition to what this proposal talks about, yes?

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+support virtual media with IPMI. We can also tell whether the
+provisioning network is using IPv4 or IPv6. Using this information, we
+could build a set of rules to determine the boot mode. However, we
+would have to document those rules, and they may not seem obvious to
+the end-user.
+
+At least some of the features are tied to the protocol used to
+communicate with the BMC. We could therefore link the boot mode to the
+BMC driver, and have a deterministic method for choosing the boot mode
+that is also easy to document.
```
demonCoder95: This would then mean for supporting BMC protocols other than ```IPMI``` or ```Redfish```, say ```iDRAC-wsman```, it will become *_incumbent_* that this ```boot_mode``` implementation is done. And without it, the implementation would be incomplete?

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+and to prefer UEFI when using Redfish. We can implement this by adding
+a new method to the [AccessDetails
+interface](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/access.go#L27)
+to return the properties for a host, including the `boot_mode`. This
+API is consistent with other similar methods in that class that return
+other sets of data passed to Ironic when a host is registered.
+
+We can be somewhat flexible with the rules by allowing the BMC driver
+in metal3 to look at some of the other settings for the host to decide
+which boot mode to set. For example, when the driver author knows that
```
demonCoder95: This sounds good. Agree.

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+At least some of the features are tied to the protocol used to
+communicate with the BMC. We could therefore link the boot mode to the
+BMC driver, and have a deterministic method for choosing the boot mode
+that is also easy to document.
+
+The proposal, therefore, is to always use BIOS mode when using IPMI
+and to prefer UEFI when using Redfish. We can implement this by adding
+a new method to the [AccessDetails
+interface](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/access.go#L27)
+to return the properties for a host, including the `boot_mode`. This
```
demonCoder95: When we say return host ```properties```, are we refering to the Ironic Node ```capabilities```? (I understand we would want to name them differently).

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### As a user, I want to provision a host using an IPv6 provisioning network
+
+### Risks and Mitigations
+
+## Design Details
+
+- Update the [AccessDetails
+  interface](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/access.go#L27)
+  to include the new method, `NodeProperties()`.
```
demonCoder95: This method is supposed to return a structure? Like you mentioned earlier, node ```properties```. Where would be the best place to put such a structure? The same ```AccessDetails``` interface? (Would make sense, since this is where we generally handle vendor-specific stuff.

design/implicit-boot-mode.md
juliakreger: I think they begin to become mutually exclusive. I guess what I'm lacking is a clear definition of why bios/legacy mode is still required. I'd personally love to see the default change, and all the complexity here is really around trying to avoid some of the issues that are present with different BMC implementations. Hiding everything _does_ make it easier for us to define our supported configurations, and reduce the risk of being distracted by someone trying to essentially deploy/operate systems in modes that we wouldn't want them to operate in under normal circumstances.

design/implicit-boot-mode.md
juliakreger: Explicit docs would really still be needed to explain what is done/supported in terms of the way Metal3 operates to set the operator expectation.

design/implicit-boot-mode.md
juliakreger: basically "bios" or "uefi" if memory serves. The default ironic ships is bios, but we've got notes and reminders in our code base to change that to UEFI. The only thing really holding that back (If there were infinite time/humans working on the code) is IPMI and risk of breaking users.

design/implicit-boot-mode.md
juliakreger: It would be idea, but I don't think required. Documentation can always say "for vendor-z, please ensure hardware has been pre-configured to UEFI mode". Realistically, that will still be required for some firmware versions in BMCs today.

design/implicit-boot-mode.md
```diff
@@ -0,0 +1,197 @@
...
+combinations of those features with all boot modes. Asking the user to
+give us the boot mode explicitly introduces an opportunity for them to
+specify an incorrect value. We have enough information about the host
+already to determine the correct boot mode to use. Therefore we should
+not need to extend the BareMetalHost API in order to support multiple
+boot modes.
+
+## Motivation
+
+When booting using virtual media, Ironic builds a custom ISO for each
```
yrobla: Isn't there the possibility of building a hybrid ISO that can work in both modes? This will save us complexity

etingof: This is technically feasible, but needs to be coded in ironic.

May be we could build hybrid ISO only if ironic is not certain what boot mode is in effect.

design/implicit-boot-mode.md
dhellmann: While the point of this proposal is to try not to add a new API field, if we did I see a few ways we could use it:

1. We could require that the user always provide it.
2. We could have a single default and let the API override that. 
3. We could keep the system described here, and let the API override that.

As to why we need to keep IPMI, I tried to explain that briefly down around line 108. The tl;dr version of this proposal is "always use UEFI, except when using IPMI".

design/implicit-boot-mode.md
dhellmann: That's certainly another possible approach. Here I am trying to limit the scope of work. I think we can make this work with 1 patch in the baremetal-operator repository.

design/implicit-boot-mode.md
dhellmann: Right, I think we can expand our driver documentation with a description of the boot mode that would be used, including any pre-configuration that might be needed for hosts that don't support probing or setting the boot mode via the BMC.

design/implicit-boot-mode.md
dhellmann: The BMC driver here in metal3 will always need the new method to be implemented in order to fulfill the go interface.

design/implicit-boot-mode.md
dhellmann: I think we want it to return a map[string]interface{} that can be passed as the Properties field of a CreateOpts when registering the node with ironic. See https://github.com/gophercloud/gophercloud/blob/master/openstack/baremetal/v1/nodes/requests.go#L220

design/implicit-boot-mode.md
dhellmann: See the earlier response about using a map.

design/implicit-boot-mode.md
demonCoder95: I see. And having this default as ```bios```, could it complicate things for us? In this specific context of the operator.

design/implicit-boot-mode.md
demonCoder95: That sounds good. I agree. Where is this driver documentation?

design/implicit-boot-mode.md
demonCoder95: I think 3. sounds good. And because of 3. and given Julia's comment that both of these are mutually exclusive, this work, if not anything else, becomes an augmentation to the future (if any) additions to the API (whenever/however they are needed). 1. and 3. would be the way to go, in my opinion.
I agree with the pain IPMI provides. And I think it sounds more like an Ironic community wide discusison getting everybody on board, before we can move away from the legacy/IPMI setup. Ergo, it's safe to ignore that/keep the way it is in this proposal.

design/implicit-boot-mode.md
demonCoder95: > 
> 
> The BMC driver here in metal3 will always need the new method to be implemented in order to fulfill the go interface.

That's what I thought. I'll have to take this into consideration then. Let me know whenever the work starts on this, I'd like to contribute to the idrac side of things.

design/implicit-boot-mode.md
demonCoder95: Got it. Thanks.

design/implicit-boot-mode.md
demonCoder95: I see. Okay.

design/implicit-boot-mode.md
dhellmann: > That sounds good. I agree. Where is this driver documentation?

There is some very minimal documentation in the API docs at https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#spec-fields but I think we will want to add something new if we start having to explain to users that different drivers will have different behaviors.

design/implicit-boot-mode.md
dhellmann: The default is configurable, so we could change it in the copy of Ironic that we deploy if what is being used now turns into a problem.

dhellmann:
How does everyone feel about this? Does it include enough detail for us to proceed? Should we merge it, or does it need any updates?

sadasu:
A couple of minor comments:
1. Can the proposal say explicitly if the boot mode derived by metal3 would be cluster wide or per host? 
2. Can the Status field of the BareMetal Host CRD be updated to report the actual boot mode (per host to take of the case where the cluster could comprise of dissimilar hosts)?

design/implicit-boot-mode.md
demonCoder95: Sounds good.

design/implicit-boot-mode.md
demonCoder95: Yes, we will need that. Also, the doc above seems to be a description of the API object. Since the boot mode will _not_ be an extension to the API, documenting it in there would be tricky. How about we start another doc with non-API implementation details? Case in point, this boot-mode thing. And I'm sure there will be more as well, as we go along.

demonCoder95:
> How does everyone feel about this? Does it include enough detail for us to proceed? Should we merge it, or does it need any updates?

Given [this PR | https://github.com/metal3-io/baremetal-operator/pull/437], do we still want to pursue this? Or is this work an addendum?

demonCoder95:
> A couple of minor comments:
> 
>     Can the proposal say explicitly if the boot mode derived by metal3 would be cluster wide or per host? My understanding is that it is per host but I am not sure.

It must be per host, since this PR targets the BMO. The operator would be expect to operate on a per BMH basis. CAPM3 is the point to consider things cluster-wide.
 
>     Can the Status field of the BareMetal Host CRD be updated to report the actual boot mode (per host to take of the case where the cluster could comprise of dissimilar hosts)?

We can definitely show it in the status field of the BMH object to reflect current ```bootmode``` of the host. 
The cluster-wide configuration, however, might be a CAPI thing, and CAPM3 would be the point to think about that. Using annotations to pass cluster-wide ```bootmode``` to every provisioned machine might be a way. 

You can also take a look at [this PR](https://github.com/metal3-io/baremetal-operator/pull/437), which implements the boot-mode property as an ironic node ```capability```.

sadasu:
> We can definitely show it in the status field of the BMH object to reflect current `bootmode` of the host.
This the only addition I am requesting.

dhellmann:
I added a note about a status field and also fixed a typo in one of the user stories.

dhellmann:
It seems like we were very close to consensus on this before my most recent minor edits. I would like to approve this using lazy consensus by early next week. Please let me know if there are any objections to that.

maelk:
lgtm

maelk:
That looks good to me!

dhellmann:
@demonCoder95 did you have any more input on this design or do you feel like we're ready to proceed?

demonCoder95:
I think we're good to go! @dhellmann

dhellmann:
> I think we're good to go! @dhellmann

Thanks!

@maelk could you LGTM this, please?

juliakreger:
Greetings @pierrecregut 

Your comment makes me really wonder if we're making a huge mistake by trying to guess and not just providing the ability for users to declare their needed configuration. I know we generally want to abstract and hide as much away as possible, but we're going to continue to run into cases like this I suspect.

Before I mentally continue, can you confirm your hardware? Lenovo has an SR630, where as Dell has an R630. You mentioned both vendors in your message so I want to be sure in case I need to reach out to their engineering groups.

That being said, Your not the first person to hit the continuous flag issue as of recent. I've had some discussions with two different vendors and the feeling seems to be more that it should be honored as the specific UEFI options are newer additions to the schema. In ironic, we're working to try and only reset the value on the field if necessary instead of always trying to send it, and I guess this leads me to wonder what exactly the exchange between the BMC and ironic is at this point. I ask because we've seen a few different variations in touching this settings object, and if this is another variation, we need to be aware of it to try and determine the correct path to navigate. Ideally, if you can provide we/us/I with the ironic conductor container log output, we can try and identify the variation and go from there.

pierrecregut:
@juliakreger  Yes it is a Lenovo SR630. Sorry for the confusion.
I have filled https://storyboard.openstack.org/#!/story/2007527
Here is a more complete log file of the conductor.
[ironic-conductor-logs.txt](https://github.com/metal3-io/metal3-docs/files/4444188/ironic-conductor-logs.txt)

PR #79, opened 2020-03-11
Add a design proposal for metadata processes

maelk:
CAPM3 needs to be handling metadata in speific ways, this proposal outlines them along with a possible solution

### DIFF
--- a/design/metadata-handling.md
+++ b/design/metadata-handling.md
@@ -166,7 +166,7 @@ Some values would be set by default to maintain compatibility:
 - **metal3-namespace**: the name of the BareMetalHost
 - **metal3-name**: The name of the BareMetalHost
 - **local-hostname**: The name of the BareMetalHost
-- **local_hostname**: The name of the BareMetalHost
+- **local_hostname**: The namespace of the BareMetalHost
 
 However, setting any of those values in the metaData secret will override those
 default values.
@@ -373,7 +373,9 @@ spec:
       - dns: "2001:4860:4860::8888"
 status:
   indexes:
-    "machine-1": 0
+    "0": "machine-1"
+  dataNames:
+    "machine-1": nodepool-1-0
   lastUpdated: "2020-04-02T06:36:09Z"
 ```
 
@@ -602,7 +604,7 @@ There will be two cases:
 - An already generated Metal3Data object exists with an ownerReference to this
   Metal3Machine. In that case, the reconciler will verify that the required
   secrets exist. If they do not, they will be created.
-- if no secret exists with an ownerReference to this Metal3Machine, then the
+- if no Metal3Data exists with an ownerReference to this Metal3Machine, then the
   reconciler will create one and fill the respective field with the secret name.
 
 To create a Metal3Data object, the Metal3DataTemplate controller will select an
@@ -612,7 +614,7 @@ available index that is not in the `indexes` field of the status. If the
 object linked to this Metal3DataTemplate and recreate the unavailable indexes.
 It will fill it by extracting the index from the Metal3Data names. The indexes
 always start from 0 and increment by 1. The lowest available index is to be used
-next.
+next. The `dataNames` field contains the map of Metal3Machine to Metal3Data.
 
 Once the next lowest available index is found, it will create the Metal3Data
 object. The name would be a concatenation of the Metal3DataTemplate name and
--- a/design/metadata-handling.md
+++ b/design/metadata-handling.md
@@ -48,8 +48,11 @@ provisional
 Cloud-init templates offer a very powerful feature to render the configuration
 at runtime based on metadata given by Ironic. We want to take advantage of this
 feature to render node specific configuration for nodes in a machine deployment
-or in a control plane kubeadm object. In addition, we want to be able to
-generate node specific network configuration in CAPM3 based on a template.
+or in a control plane kubeadm object dynamically, where this kind of objects
+cannot be created beforehand. In addition, we want to be
+able to generate node specific network configuration in CAPM3 based on a
+template for nodes that are part of a machine deployment or or a control plane
+provider object.
 
 This design proposal introduces a new data template object that contains
 templates for network data and metadata secret generation per node. Once
@@ -76,6 +79,11 @@ some hardware related information, not Kubernetes node information).
 Hence we want a way to generate a node-specific metadata and network data for a
 machine deployment or Kubeadm Control Plane and pass it down to Ironic.
 
+In addition, the network data might require static ip addresses configuration.
+The controller must be able to provide a unique IP address. Hence the controller
+should keep track of the index of the node to be able to render consistently ip
+addresses.
+
 ### Goals
 
 The goals are :
@@ -92,7 +100,8 @@ The goals are :
 
 ### Non-Goals
 
-TBA
+- Implement a new DSL to render the data. All configuration should be done
+  through API.
 
 ## Proposal
 
@@ -570,10 +579,11 @@ extracting the index from the Metal3Data names.
 Once the next available index is found, it will create the Metal3Data object.
 The name would be a concatenation of the Metal3DataTemplate name and index.
 Upon conflict, it will fetch again the list to consider the new list of
-Metal3Data. Upon success, it will render the content values, and create
-the secrets containing the rendered data. The controller will generate the
-content based on the `metaData` or `networkData` field of the Metal3DataTemplate
-Specs.
+Metal3Data and try to create the new object with the new index, this will happen
+until the new object is created successfully. Upon success, it will render the
+content values, and create the secrets containing the rendered data. The
+controller will generate the content based on the `metaData` or `networkData`
+field of the Metal3DataTemplate Specs.
 
 ### The generated secrets
 
--- a/design/metadata-handling.md
+++ b/design/metadata-handling.md
@@ -154,9 +154,22 @@ understandable.
 
 ## Design Details
 
+### BareMetalHost changes
+
 BareMetalHost would have an added field `metaData` in the `Spec` pointing to a
 secret containing the metadata map. Bare Metal Operator would then pass this map
 to Ironic.
+
+Some values would be set by default to maintain compatibility:
+
+- **uuid**: This is the BareMetalHost UID
+- **metal3-namespace**: the name of the BareMetalHost
+- **metal3-name**: The name of the BareMetalHost
+- **local-hostname**: The name of the BareMetalHost
+- **local_hostname**: The name of the BareMetalHost
+
+However, setting any of those values in the metaData secret will override those
+default values.
 
 ### Metal3Machine changes
 
@@ -215,6 +228,19 @@ spec and to the status objects, consisting of a secret
 reference containing the name and the namespace of the secret containing the
 network configuration for this Metal3Machine.
 
+The secret containing the metaData or the network data could be provided by the
+user directly using the `metaData` or `networkData` fields in the spec of the
+Metal3Machine.
+
+The `metaData` and `networkData` field in the `spec` section are for the user
+to give directly a secret to use as metaData or networkData. The `userData`,
+`metaData` and `networkData` fields in the `status` section are for the
+controller to store the reference to the secret that is actually being used,
+whether it is from one of the spec fields, or somehow generated. This is aimed
+at making a clear difference between the desired state from the user (whether
+it is with a DataTemplate reference, or direct `metaData` or `userData` secrets)
+and what the controller is actually using.
+
 A `dataTemplate` field would be added, consisting of an object reference
 to a Metal3DataTemplate object containing the templates for
 the metadata and network data generation for this Metal3Machine.
@@ -223,10 +249,6 @@ the Metal3Data object created for this machine. If the dataTemplate field is set
 but either the `renderedData`, `metaData` or `networkData` fields in the status
 are unset, then the Metal3Machine controller will wait until it can find the
 Metal3Data object and the rendered secrets. It will then populate those fields.
-
-The secret containing the metaData or the network data could be provided by the
-user directly using the `metaData` or `networkData` fields in the spec of the
-object.
 
 When CAPM3 controller will set the different fields in the BareMetalHost,
 it will reference the metadata secret and the network data secret
@@ -274,8 +296,9 @@ spec:
         start: 192.168.0.10
         end: 192.168.0.100
         subnet: 192.168.0.0/24
+        step: 1
     - key: mac
-      bareMetalHostMACByName: "eth0"
+      fromHostInterface: "eth0"
   networkData:
     links:
       - ethernet:
@@ -283,13 +306,13 @@ spec:
           id: "enp1s0"
           mtu: 1500
           macAddress:
-            bareMetalHostMACByName: "eth0"
+            fromHostInterface: "eth0"
       - ethernet:
           type: "phy"
           id: "enp2s0"
           mtu: 1500
           macAddress:
-            bareMetalHostMACByName: "eth1"
+            fromHostInterface: "eth1"
       - bond:
           id: "bond0"
           mtu: 1500
@@ -307,7 +330,7 @@ spec:
           vlanId: 1
           vlanLink: bond0
     networks:
-      - ipv4Dhcp:
+      - ipv4DHCP:
           id: "provisioning"
           link: "bond0"
 
@@ -327,10 +350,10 @@ spec:
               services:
                 - type: "dns"
                   address: "8.8.4.4"
-      - ipv6Dhcp:
+      - ipv6DHCP:
           id: "provisioning6"
           link: "bond0"
-      - ipv6Slaac:
+      - ipv6SLAAC:
           id: "provisioning6slaac"
           link: "bond0"
       - ipv6:
@@ -376,12 +399,20 @@ ways. The following objects are available:
 - **objectName** : renders the name of the object that matches the type given in
   the `object` field.
 - **index**: renders the index of the current object, with the offset from the
-  `offset` field and using the step from the `step` field.
+  `offset` field and using the step from the `step` field. The following
+  conditions must be matched :
+
+    - `offset` >= 0
+    - `step` >= 1
+
+  if the step is unspecified (default value being 0), the controller will
+  automatically change it for 1.
 - **ip**: renders an ip address based on the index, based on the `start` value
   if given or using `subnet` to calculate the start value, and checking that
   the rendered value is not over the `end` value. The increment is the `step`
-  value.
-- **bareMetalHostMACByName**: renders the MAC address of the BareMetalHost that
+  value. If the computed value goes out of bounds, the error status will be set
+  with the error in the error message.
+- **fromHostInterface**: renders the MAC address of the BareMetalHost that
   matches the name given as value.
 
 The **key** is required and one of the object is required. If there are multiple
@@ -425,7 +456,7 @@ The **links/ethernet/type** can be one of :
 The **links/ethernet/macAddress** object can be one of:
 
 - **string**: with the desired Mac given as a string
-- **bareMetalHostMACByName**: with the interface name from BareMetalHost
+- **fromHostInterface**: with the interface name from BareMetalHost
   hardware details.
 
 The **links/bond** object contains the following:
@@ -459,10 +490,10 @@ The **links/vlan** object contains the following:
 The object for the **networks** section list can be one of:
 
 - **ipv4**: an ipv4 static allocation
-- **ipv4Dhcp**: an ipv4 dhcp based allocation
+- **ipv4DHCP**: an ipv4 DHCP based allocation
 - **ipv6**: an ipv6 static allocation
-- **ipv6Dhcp**: an ipv6 dhcp based allocation
-- **ipv6Slaac**: an ipv6 slaac based allocation
+- **ipv6DHCP**: an ipv6 DHCP based allocation
+- **ipv6SLAAC**: an ipv6 SLAAC based allocation
 
 The **networks/ipv4** object contains the following:
 
@@ -550,6 +581,10 @@ spec:
   metal3Machine:
     name: machine-1
     namespace: default
+status:
+  ready: true
+  error: false
+  errorMessage: ""
 ```
 The Metal3Data will contain the index of this node, and links to the secrets
 generated and to the Metal3Machine using this Metal3Data object.
@@ -570,20 +605,24 @@ There will be two cases:
 - if no secret exists with an ownerReference to this Metal3Machine, then the
   reconciler will create one and fill the respective field with the secret name.
 
-To create a Metal3Data object, the controller will select an index for that
-Metal3Machine. The selection happens by selecting the lowest available index.
-The controller will list all existing Metal3Data object linked to this
-Metal3DataTemplate and create a list of unavailable indexes. It will fill it by
-extracting the index from the Metal3Data names.
+To create a Metal3Data object, the Metal3DataTemplate controller will select an
+index for that Metal3Machine. The selection happens by selecting the lowest
+available index. The controller will list all existing Metal3Data object linked
+to this Metal3DataTemplate and create a list of unavailable indexes. It will
+fill it by extracting the index from the Metal3Data names. The indexes always
+start from 0 and increment by 1. The lowest available index is to be used next.
 
-Once the next available index is found, it will create the Metal3Data object.
-The name would be a concatenation of the Metal3DataTemplate name and index.
-Upon conflict, it will fetch again the list to consider the new list of
+Once the next lowest available index is found, it will create the Metal3Data
+object. The name would be a concatenation of the Metal3DataTemplate name and
+index. Upon conflict, it will fetch again the list to consider the new list of
 Metal3Data and try to create the new object with the new index, this will happen
 until the new object is created successfully. Upon success, it will render the
 content values, and create the secrets containing the rendered data. The
 controller will generate the content based on the `metaData` or `networkData`
 field of the Metal3DataTemplate Specs.
+
+Once the generation is successful, the status field `ready` will be set to True.
+If any error happens during the rendering, an error message will be added.
 
 ### The generated secrets
 
@@ -614,6 +653,47 @@ spec:
 The secret will contain the generated data for the host. Once the
 secret reference field on the Metal3Machine is set, the Metal3Machine controller
 will proceed with the provisioning.
+
+### Deployment flow
+
+#### Manual secret creation
+
+In the case where the Metal3Machine is created without a `dataTemplate` value,
+if the `metaData` or `networkData` fields are set (one or both), the
+Metal3Machine reconciler will fetch the secret, set the status field and
+directly start the provisioning of the BareMetalHost using the secrets if given.
+If one of the secrets does not exist, the controller will wait to start the
+provisioning of the BareMetalHost until it exists.
+
+#### Dynamic secret creation
+
+In the case where the Metal3Machine is created with a `dataTemplate` value, the
+Metal3Machine reconciler will fetch that object (or wait until it exists if it
+does not exist yet) and add the Metal3Machine in the ownerReferences of the
+Metal3DataTemplate.
+
+The Metal3DataTemplate would then be reconciled, and its controller will create
+an index for this Metal3Machine if it does not exist yet, and create a
+Metal3Data object with the index and the Metal3Machine in the ownerReference.
+
+The Metal3Data reconciler will then generate the secrets, based on the index,
+the Metal3DataTemplate and the machine. Once created, it will set the status
+field `ready` to True.
+
+Once the metal3Data object is ready, the Metal3Machine controller will fetch
+the secrets that have been created (one or both) and use them to start
+provisioning the BareMetalHost.
+
+#### Hybrid configuration
+
+If the Metal3Machine object is created with a `dataTemplate` field set, but one
+of the `metaData` or `networkData` is also set in the spec, this one will
+override the template generation for this specific secret. i.e. if the user sets
+the three fields, the controller will use the user input secret for both.
+
+This means that some hybrid scenarios are supported, where the user can give
+directly the `metaData` secret and let the controller render the `networkData`
+secret through the Metal3DataTemplate object.
 
 ## Implementation structure
 
--- a/design/metadata-handling.md
+++ b/design/metadata-handling.md
@@ -374,6 +374,10 @@ spec:
     services:
       - dns: "8.8.8.8"
       - dns: "2001:4860:4860::8888"
+status:
+  indexes:
+    "machine-1": 0
+  lastUpdated: "2020-04-02T06:36:09Z"
 ```
 
 This object will be reconciled by its own controller. When reconciled,
@@ -607,10 +611,12 @@ There will be two cases:
 
 To create a Metal3Data object, the Metal3DataTemplate controller will select an
 index for that Metal3Machine. The selection happens by selecting the lowest
-available index. The controller will list all existing Metal3Data object linked
-to this Metal3DataTemplate and create a list of unavailable indexes. It will
-fill it by extracting the index from the Metal3Data names. The indexes always
-start from 0 and increment by 1. The lowest available index is to be used next.
+available index that is not in the `indexes` field of the status. If the
+`indexes` field is empty, the controller will list all existing Metal3Data
+object linked to this Metal3DataTemplate and recreate the unavailable indexes.
+It will fill it by extracting the index from the Metal3Data names. The indexes
+always start from 0 and increment by 1. The lowest available index is to be used
+next.
 
 Once the next lowest available index is found, it will create the Metal3Data
 object. The name would be a concatenation of the Metal3DataTemplate name and
--- a/design/metadata-handling.md
+++ b/design/metadata-handling.md
@@ -149,6 +149,8 @@ BareMetalHost would have an added field `metaData` in the `Spec` pointing to a
 secret containing the metadata map. Bare Metal Operator would then pass this map
 to Ironic.
 
+### Metal3Machine changes
+
 The Metal3Machine will be modified as follow:
 
 ```yaml
@@ -226,6 +228,8 @@ When the Metal3Machine gets deleted, the CAPM3 controller will remove its
 ownerreference from the data template object. This will trigger the deletion of
 the generated Metal3Data object and the secrets generated for this machine.
 
+### The Metal3DataTemplate object
+
 A new object would be created, a Metal3DataTemplate type.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


maelk:
I'll put a lazy consensus on this one. If not specific comments or requests for changes before Wednesday 25th, we'll implement it as is.

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+
+From a user point of view, if this is not well documented or understood, it
+could lead to unexpected errors or behaviours. The errors should be easily
+understandable.
+
+## Design Details
+
+BareMetalHost would have an added field `metaData` in the `Spec` pointing to a
+secret containing the metadata map. Bare Metal Operator would then pass this map
+with added default key/value pairs if not present to Ironic.
```
dhellmann: How would this be different from the existing UserData field? Just in how it is passed to Ironic?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+<!-- Added by: stack, at: 2019-02-15T11:41-05:00 -->
+
+<!--te-->
+
+[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
+
+## Summary
+
+Cloud-init templates offer a very powerful feature to render the configuration
+at runtime based on metadata given by Ironic. We want to take advantage of this
```
dhellmann: Where is Ironic going to provide that data?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+generating the metadata content. It also introduces a new Metadata object that
+stores the list of metadata items for each node.
+
+## Motivation
+
+When deploying a machine deployment or a control plane with CAPI (v1alpha3), the
+KubeadmConfig applied is the same for all the nodes. However, we would want
+control on some fields, such as the node name for example. This can be a
+variable in the cloud-init template, but it needs to be provided through Ironic
+as metadata. And the metadata needs to be different for each of the nodes.
```
dhellmann: What needs to be provided, the variable or the value? What is going to render the templates to real data to be used by the host?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+For example we would want the nodes of a machine deployment to have names like
+`worker-np1-0`, `worker-np1-1`, `worker-np1-2` etc. and those names to be
+unrelated with the BareMetalHost names (that might be `machine-xyz`, reflecting
+some hardware related information, not Kubernetes node information).
```
dhellmann: I expect you'll find most users want those names to be the same, to make mapping from node to host simpler.

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Goals
+
+The goals are :
+
+- to add a metadata field in BareMetalHost to have it configurable
+  for Ironic, and in the Metal3Machine for CAPM3 to be able to pass it down to
+  BareMetalHost.
+- introduce new objects containing information to get the metadata per node,
+  or generate it, and store the values in use to not duplicate them.
```
dhellmann: When you say "new objects" do you mean custom resources or something else? If custom resources, are they new types or existing types?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+for one node only, and be re-used during upgrade.
+
+#### Story 2
+
+As a user, I want to give my own metadata to Ironic as a field on BareMetalHost.
+
+### Implementation Details/Notes/Constraints
+
+- Baremetal Operator should set some default metadata in case they are not
+  provided by the use (Like UUID that is required).
```
dhellmann: I don't think we want the bare metal operator to do any more than pass the data to ironic. Something else should prepare the data, including setting defaults if that's needed.

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+    name: cluster-1
+spec:
+  values:
+    node-0:
+      abc: def
+    node-1:
+      abc: ghi
+  templates:
+    local-hostname: worker-np1-{{ getIndex }}
+```
```
dhellmann: This looks like a ConfigMap. Why do we need a new CRD?

It looks like data for multiple hosts is mixed together in one instance here. Would we really do that? Why not use a separate resource for each host?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+  machines:
+    "default/machine-1":
+      index: "0"
+      value: "node-0"
+```
+
+This object would store the list of values used for each Metal3Metadata object.
+The rationale behind not including it as a status of the previous object is to
+be able to go through the move phase without information loss. This object would
+only be used by the controller and should not be edited by the user.
```
dhellmann: Placing status values in the spec of a different CRD is an anti-pattern. It subverts the idea of "spec" being input and "status" being output in the API. It also leads to race conditions (it's impossible to know whether the "indexes" are correctly up to date, because we can't tell which instance of the CRD to look at and whether it has been reconciled by its controller (that's why random suffixes are used everywhere else in kubernetes).

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+  values:
+    node-0:
+      abc: def
+    node-1:
+      abc: ghi
+  templates:
+    local-hostname: worker-np1-{{ getIndex }}
+```
+
+This object will be reconciled by the Metal3machine controller. When reconciled,
```
dhellmann: Why will the machine controller manage these things instead of a new controller?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+When CAPM3 controller will fill select a BaremetalHost and set the different
+fields there, it will generate the metaData content based on the static metaData
+and then render the metadata from the metaDataTemplate. It will
+get the referenced Metal3Metadata and render metadata from the templates,
+selecting the lowest available index, and from the values, selecting the first
+available id. It will save the index and id in the Metal3MetadataStatus. If the
+Metal3Machine already has an index and an id in the Metal3MetadataStatus, it
+will re-use those.
```
dhellmann: This feels like it's mixing two different sets of responsibilities into one component. Why wouldn't we add a new controller that configures the metadata and have the CAPM3 controller wait until the metadata is resolved before it starts the provisioning?

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+
+When the Metal3Machine gets deleted, the CAPM3 controller will remove its
+references from the Metal3MetadataStatus, making it available for other
+Metal3Machines.
```
dhellmann: We would only have to do this if we try to use a resource as a database. If we take some templates as input and render them into a resource that is associated one-to-one with the host, we can set the owner reference to the Metal3Machine so that when the host is deprovisioned the metadata is removed.

design/metadata-handling.md
```diff
@@ -0,0 +1,286 @@
...
+    kind: Metal3Cluster
+    name: cluster-1
+spec:
+  values:
+    node-0:
+      abc: def
+    node-1:
+      abc: ghi
+  templates:
+    local-hostname: worker-np1-{{ getIndex }}
```
dhellmann: Instead of using a template like this, why not follow the usual pattern of asking for a name "prefix" and adding the suffix to it in the controller?

design/metadata-handling.md
maelk: This is about passing metadata, not userdata. Currently in BMO, they are hard-coded, while they should be configurable and exposed through a metadata field in BMH. BMO would give it to ironic as metadata. here is the PR doing this in Baremetal-operator : https://github.com/metal3-io/baremetal-operator/pull/448

design/metadata-handling.md
maelk: it will be given as metadata : https://cloudinit.readthedocs.io/en/latest/topics/instancedata.html It is accessible from within the userData template as `{{ ds.meta_data.<var_name> }}`

design/metadata-handling.md
maelk: Both actually, the variable name and content need to be both provided to Ironic. The template rendering will be done by the Metal3Machine controller, when creating the metadata secret for the BMH

design/metadata-handling.md
maelk: Probably, this is only an example, it does not have to be about names. We just need a way to render node specific information based on the index of the node or on a list of values. We need it specifically for the stacked storage, so that a node keeps its identity across upgrade.

design/metadata-handling.md
maelk: I meant Custom Resources, and they are new types.

design/metadata-handling.md
maelk: okay, that is possible, but that means that the deployment might fail if the defaults are not set properly by whatever entities generate the metadata. It needs to be very explicitly documented then.

design/metadata-handling.md
maelk: The idea is that this resource would be tied to a machine deployment, or a kubeadm control plane object. That's why there are multiple hosts, since they are automatically created for the machine deployment or KCP.

Yes in the first iteration, this was a configmap, but we benefit from using CR, with the validation opportunity that it brings, not only with openapi, but also with the webhooks.

design/metadata-handling.md
maelk: Very good point, I will update the proposal with this.

design/metadata-handling.md
maelk: the reason was that it is tightly bound to the reconciliation of the metal3machine objects. But your next comment makes this irrelevant, so I will update the proposal

design/metadata-handling.md
maelk: We do not want to assume any behavior. This way, the user has a full control over the final value. Maybe this example is not the best, it could be that the user wants a prefix and a suffix around the index value.

design/metadata-handling.md
maelk: that's a very good point, I will update the proposal

design/metadata-handling.md
maelk: Unfortunately random suffixes do not work for our use case. However, I'll make the changes, so that we do not use this object as a database. I was not very happy with this either, but I did not see a good alternative. You proposed one in your following comments that would solve this problem.

design/metadata-handling.md
maelk: In the new proposal, I have a streamlined the object, using only a template, to render the metadata for all machines referencing this object.

maelk:
@dhellmann What do you think of this new version ? does it fit better wrt your concerns ?

design/metadata-handling.md
```diff
@@ -0,0 +1,417 @@
...
+user directly.
+
+When CAPM3 controller will set the different fields in the BareMetalHost,
+it will reference the metadata secret and the network data secret
+in the BareMetalHost. If `metaData` or `networkData` field is unset, it will
+also remain unset on the BareMetalHost.
+
+When the Metal3Machine gets deleted, the CAPM3 controller will remove its
+references from the data template object. This will trigger the deletion of the
+secrets generated for this machine.
```
dhellmann: This part isn't clear. What "references" are meant here?

design/metadata-handling.md
```diff
@@ -0,0 +1,417 @@
...
+metadata:
+  name: nodepool-1
+  namespace: default
+  ownerReferences:
+  - apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    controller: true
+    kind: Metal3Cluster
+    name: cluster-1
+spec:
+  metaData: |
```
dhellmann: So the content of metaData and networkData are strings that should contain embedded YAML or JSON content? Is metaData always YAML and networkData always JSON, as in this example?

design/metadata-handling.md
```diff
@@ -0,0 +1,417 @@
...
+    abc: def
+    local-hostname: {{ machineName }}
+    index: {{ index }}
+  networkData: |
+    {
+        "links": [
+            {
+                "id": "enp1s0",
+                "type": "phy",
+                "ethernet_mac_address": "{{ '{{ bareMetalHostMACByName "eth0" }}' }}"
```
dhellmann: I'm having trouble parsing this syntax. It looks like nested {{ }} with one set wrapped in single quotes. Is that right? What does the nesting signify?

design/metadata-handling.md
```diff
@@ -0,0 +1,417 @@
...
+- **bareMetalHostName** : returns the BareMetalHost name
+- **index** : returns the Metal3Machine index for the Metal3Metadata object.
+  The index starts from 0.
+- **indexWithOffset** : takes an integer as parameter and returns the sum of
+  the index and the offset parameter
+- **indexWithStep** : takes an integer as parameter and returns the
+  multiplication of the index and the step parameter
+- **indexWithOffsetAndStep** OR **indexWithStepAndOffset**: takes two
+  integers as parameters, order depending on the function name, and returns the
+  sum of the offset and the multiplication of the index and the step.
```
dhellmann: What would this be used for?

design/metadata-handling.md
```diff
@@ -0,0 +1,417 @@
...
+- **index*Hex** : All the `index` functions can be suffixed with `Hex` to
+  get the same value in hexadecimal format.
+- **bareMetalHostMACByName**: takes a string as parameter and returns the MAC
+  address of the nic with the name matching the parameter. This function
+  operates over the list of NICs in the `status.hardwareDetails.NIC` field of
+  the BareMetalHost.
```
dhellmann: How does the thing rendering the template associated with the Metal3Machine know which BareMetalHost to use?

design/metadata-handling.md
```diff
@@ -0,0 +1,417 @@
...
+based on the `metaData` or `networkData` field of the Metal3DataTemplate Specs.
+It will render the content of the secret, selecting the lowest available index
+from the status indexes map. If that map is empty, it will fill it by building
+a map of names of the machines and index used by the machine, extracting the
+index from the secret names in the Metal3Machine specs.
+
+Once the next available index is found, it will update the Metal3DataTemplate
+object. Upon conflict, it will immediately requeue to consider the new state of
+Metal3DataTemplate. Upon success, it will render the content values, and create
+a secret containing the rendered data.
```
dhellmann: I'm still unclear why this is so focused on indexes. You say random values aren't sufficient, but don't explain how. Could you provide more detail about that, please?

design/metadata-handling.md
maelk: yes, the format must be like this

design/metadata-handling.md
maelk: those are helpers function that the user can use to generate an integer based on the index

design/metadata-handling.md
maelk: the data is rendered when the BMH is associated, so the controller can access all the fields from the Metal3Machine or the BMH

design/metadata-handling.md
```diff
@@ -0,0 +1,456 @@
...
+Hence we want a way to generate a node-specific metadata and network data for a
+machine deployment or Kubeadm Control Plane and pass it down to Ironic.
+
+### Goals
+
+The goals are :
+
+- to add a `metaData` field in BareMetalHost to have it configurable
+  for Ironic, and in the Metal3Machine for CAPM3 to be able to pass it down to
+  BareMetalHost.
```
dhellmann: We should describe how we're going to merge the values that come from the template with the set of values the BMO provides on its own. The template can override values, but if there is no template or the template doesn't include one of the 4-5 things we always send then those values will still be there.

design/metadata-handling.md
dhellmann: That's going to be very difficult to validate, and probably use.

Do you know any of the values that need to be templated in advance? Could we define a CRD with more of an API for setting those values, and have it generate the YAML or JSON to put into the secret and pass to the host?

For example, a structure like this would let the kubernetes API service do some of the validation for us.

```
metaData:
  - fieldName: abc
    literalValue: def
  - fieldName: local-hostname
    dataSource: "{{ machineName }}"
  - fieldName: index
    dataSource: "{{ index }}"
    dataType: int
networkData:
  links:
    - - fieldName: id
        literalValue: enp1s0
      - fieldName: type
        literalValue: phy
      - fieldName: ethernet_mac_address
        dataSource: "{{ bareMetalHostMACByName \"eth0\" }}"
    - - fieldName: id
        literalValue: enp2s0
      - fieldName: type
        literalValue: phy
      - fieldName: ethernet_mac_address
        dataSource: "{{ bareMetalHostMACByName \"eth1\" }}"
```

The network data section would be even better if it just used the fields directly. Something like

```
networkData:
  links:
    - id: enp1s0
      type: phy
      ethernetMacAddress: "{{ bareMetalHostMACByName \"eth0\" }}"
    - id: enp2s0
      type: phy
      ethernetMacAddress: "{{ bareMetalHostMACByName \"eth1\" }}"
```

The implementation of that form would be a little more complex, because everything might have to be treated as a possible source of template data. But maybe not all fields need to be templates? Or maybe it looks more like

```
networkData:
  links:
    - id: enp1s0
      type: phy
      ethernetMacAddressSource: eth0
    - id: enp2s0
      type: phy
      ethernetMacAddressSource: eth1
```

design/metadata-handling.md
```diff
@@ -0,0 +1,456 @@
...
+                "type": "dns",
+                "address": "8.8.8.8"
+            }
+        ]
+    }
+```
+
+This object will be reconciled by its own controller. When reconciled,
+the controller will add a label pointing to the Metal3Cluster that has nodes
+linking to this object. The spec contains a `metaData` and a `networkData` field
```
dhellmann: What is the label for?

design/metadata-handling.md
dhellmann: The advice I've had from our k8s API experts internally is to favor fully defined APIs over embedded domain-specific languages. I think that makes a lot of sense, for several reasons, so I've tried to describe how we might be able to build such an API for some of these values in my comment above. I think if we work out more detailed example uses of all of these things, we can find ways to express them in an API without relying on the user to learn a new DSL.

design/metadata-handling.md
```diff
@@ -0,0 +1,456 @@
...
+  Metal3Machine. In that case, the reconciler will verify that the required
+  secrets exist. If they do not, they will be created.
+- if no secret exists with an ownerReference to this Metal3Machine, then the
+  reconciler will create one and fill the respective field with the secret name.
+
+To create a Metal3Data object, the controller will select an index for that
+Metal3Machine. The selection happens by selecting the lowest available index.
+The controller will list all existing Metal3Data object linked to this
+Metal3DataTemplate and create a list of unavailable indexes. It will fill it by
+extracting the index from the Metal3Data names.
```
dhellmann: What happens if a hole is created in the list? Say I create 3 hosts, then delete the second. Is index `1` available after that? Or is the next lowest index `3`?

design/metadata-handling.md
```diff
@@ -0,0 +1,456 @@
...
+
+Once the next available index is found, it will create the Metal3Data object.
+The name would be a concatenation of the Metal3DataTemplate name and index.
+Upon conflict, it will fetch again the list to consider the new list of 
+Metal3Data. Upon success, it will render the content values, and create
+the secrets containing the rendered data. The controller will generate the
+content based on the `metaData` or `networkData` field of the Metal3DataTemplate
+Specs.
```
dhellmann: If the templates are only rendered at the point when the host is associated with the machine, then the secrets could also be named based on the machine name to ensure no conflict.

If we need an "index" for a machine, then could the thing creating the machine add an index to its metadata somehow? Normally I would think that is the MachineSet controller, but I don't know exactly how that works in the latest version of the APIs so I'm not sure if we have a place to plug in when the Machine is created. Doing it from the MachineSet avoid any conflicts, because only 1 goroutine should be reconciling the MachineSet and creating Machines for it.

design/metadata-handling.md
maelk: okay, I will add the description. We will always set by default the UUID, we could maybe get rid of the rest of the default values, as they were a temporary workaround for until we have this feature

design/metadata-handling.md
maelk: This is a very good idea for the validation. We could also take an approach of letting the user define a map directly for the metadata, so it will be validated as a map[string]string. For the network, I also like the approach, though I would not bake in the functions, since the user might want to directly give the mac address for example. 

I am a bit unsure if we would take away some features from the network data. Considering the definition https://docs.openstack.org/nova/latest/_downloads/9119ca7ac90aa2990e762c08baea3a36/network_data.json there are different types of links, networks etc. Either we need to implement it so that we validate again all the fields, so re-implementing the nova specs, or we leave more room for the user, with the opportunity for mistakes

The label does not have a particular use, except for clarity.

design/metadata-handling.md
maelk: yes, it will select 1 as this is the lowest index.

About the additional DSL, I don't really see yet how we could do that with the same level of flexibility. I can work on defining more use-cases so that we can discuss it further

design/metadata-handling.md
maelk: This is a very good idea for the validation. We could also take an approach of letting the user define a map directly for the metadata, so it will be validated as a map[string]string. For the network, I also like the approach, though I would not bake in the functions, since the user might want to directly give the mac address for example. 

I am a bit unsure if we would take away some features from the network data. Considering the definition https://docs.openstack.org/nova/latest/_downloads/9119ca7ac90aa2990e762c08baea3a36/network_data.json there are different types of links, networks etc. Either we need to implement it so that we validate again all the fields, so re-implementing the nova specs, or we leave more room for the user, with the opportunity for mistakes

design/metadata-handling.md
maelk: About the additional DSL, I don't really see yet how we could do that with the same level of flexibility. I can work on defining more use-cases so that we can discuss it further

design/metadata-handling.md
maelk: yes, the secrets name will contain the metal3machine name.

About the index, I am not sure that doing it in the machineset controller would be optimal, considering that it is bringing a requirement that is specific to Metal3 into CAPI impacting all providers. Considering this proposal, there also should be only one controller creating the indexes, the controller that reconciles the Metal3Data. No other controller should be doing this.

design/metadata-handling.md
dukov: network data is yaml formatted (see https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L992).

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+  the metadata map.
+- **string**: renders the given string as value in the metadata
+- **objectName** : renders the name of the object that matches the type given in
+  the `object` field.
+- **index**: renders the index of the current object, with the offset from the
+  `offset` field and using the step from the `step` field.
+- **ip**: renders an ip address based on the index, based on the `start` value
+  if given or using `subnet` to calculate the start value, and checking that
+  the rendered value is not over the `end` value. The increment is the `step`
+  value.
```
dhellmann, line 383: What happens if the computed value goes out of bounds?

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+The `metaData` field contains a list of items that will render data in different
+ways. The following objects are available:
+
+- **key**: This is compulsory and is the key used for the rendered value in
+  the metadata map.
+- **string**: renders the given string as value in the metadata
+- **objectName** : renders the name of the object that matches the type given in
+  the `object` field.
+- **index**: renders the index of the current object, with the offset from the
+  `offset` field and using the step from the `step` field.
```
dhellmann, line 379: Can offset or step be negative? 0?

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+
+##### The networks specifications
+
+The object for the **networks** section list can be one of:
+
+- **ipv4**: an ipv4 static allocation
+- **ipv4Dhcp**: an ipv4 dhcp based allocation
+- **ipv6**: an ipv6 static allocation
+- **ipv6Dhcp**: an ipv6 dhcp based allocation
+- **ipv6Slaac**: an ipv6 slaac based allocation
```
dhellmann, line 465: The convention is usually to use the same case for all letters in an abbreviation, so I think we would want `ipv6DHCP` and `ipv6SLAAC` but otherwise this makes sense.

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+
+- **id**: the network name
+- **link**: The name of the link to configure this network for
+- **routes**: the list of route objects
+
+The **networks/ipv6Slaac** object contains the following:
+
+- **id**: the network name
+- **link**: The name of the link to configure this network for
+- **routes**: the list of route objects
```
dhellmann, line 516: The first time I read through this section I thought we didn't need 5 different types (ipv4, ipv6, ipv4Dhcp, ipv6Dhcp, ipv6Slaac) because they're all just variations and we could have a "type" field. But after considering it, I like using the separate types because it makes API validation a bit easier and it's easier to describe which fields are needed for each type.

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+    name: nodepool-1
+spec:
+  index: 0
+  metaData:
+    name: machine-1-metadata
+    namespace: default
+  networkData:
+    name: machine-1-metadata
+    namespace: default
+  metal3Machine:
```
dhellmann, line 550: I expected to see a metal3Data reference on the machine instead of having the reference point from the data to machine. What benefit do we get from doing it this way?

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+the `dataTemplate` references a Metal3DataTemplate, that object will be reconciled.
+There will be two cases:
+
+- An already generated Metal3Data object exists with an ownerReference to this
+  Metal3Machine. In that case, the reconciler will verify that the required
+  secrets exist. If they do not, they will be created.
+- if no secret exists with an ownerReference to this Metal3Machine, then the
+  reconciler will create one and fill the respective field with the secret name.
+
+To create a Metal3Data object, the controller will select an index for that
```
dhellmann, line 573: Let's be specific when we say "the controller" -- which controller? The one for the template CRD?

design/metadata-handling.md
dhellmann: OK, it would be good to be explicit about that case. The algorithm is to pick the lowest unallocated number starting with the offset and incrementing by the step, right?

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+
+Once the next available index is found, it will create the Metal3Data object.
+The name would be a concatenation of the Metal3DataTemplate name and index.
+Upon conflict, it will fetch again the list to consider the new list of
+Metal3Data and try to create the new object with the new index, this will happen
+until the new object is created successfully. Upon success, it will render the
+content values, and create the secrets containing the rendered data. The
+controller will generate the content based on the `metaData` or `networkData`
+field of the Metal3DataTemplate Specs.
+
```
dhellmann, line 587: Let me see if I have the sequence right based on the description above.

For the fully automated flow, I think we have:

1. The user creates a Metal3DataTemplate (M3DT). It may contain any combination of metadata and network data.
2. The controller for M3DT creates a Metal3Data (M3D) with a reference to the template from 1, also giving it a unique index value and linking it to the Metal3Machine that will use it (how does it know that value?).
3. The M3D controller creates 2 Secrets and updates the spec.metaData and spec.networkData fields of the M3D to point to the Secrets. Either of the secrets may be empty, based on the data given. (Can you have an empty secret?)
4. The machine controller uses the secrets in the Metal3Machine to update the host as part of provisioning.

For the fully manual flow, I think we have:

1. The user creates two Secrets containing metadata and network data respectively.
2. The user creates a M3D with references to the secrets.
3. The user creates a Metal3Machine linked to the M3D from 2.
4. The machine controller uses the secrets in the Metal3Machine to update the host as part of provisioning.

For the partially automated flow, say where the user creates a metadata secret by hand:

1. The user creates one Secret containing metadata text.
2. The user creates a Metal3DataTemplate containing network data instructions.
3. The controller for M3DT creates an M3D with a reference to the secret from 1 and the template from 2, also giving it a unique index value and linking it to the Metal3Machine that will use it (how does it know that value?).
4. The M3D controller creates 1 Secret and updates the spec.networkData field of the M3D to point to the new Secret. It does not create a new Secret with metadata, regardless of the contents of the M3D, because the M3D already has a metadata secret set in its spec field.
5. The machine controller uses the secrets in the Metal3Machine to update the host as part of provisioning.

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+- **objectName** : renders the name of the object that matches the type given in
+  the `object` field.
+- **index**: renders the index of the current object, with the offset from the
+  `offset` field and using the step from the `step` field.
+- **ip**: renders an ip address based on the index, based on the `start` value
+  if given or using `subnet` to calculate the start value, and checking that
+  the rendered value is not over the `end` value. The increment is the `step`
+  value.
+- **bareMetalHostMACByName**: renders the MAC address of the BareMetalHost that
+  matches the name given as value.
```
dhellmann, line 385: We're looking for a macAddress so the `MAC` is implicit. The `ByName` doesn't clearly say the name of what.

Maybe we can spell it as `fromHostInterface` instead?

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+Metal3Data object and the rendered secrets. It will then populate those fields.
+
+The secret containing the metaData or the network data could be provided by the
+user directly using the `metaData` or `networkData` fields in the spec of the
+object.
+
+When CAPM3 controller will set the different fields in the BareMetalHost,
+it will reference the metadata secret and the network data secret
+in the BareMetalHost. If any of the `metaData` or `networkData` status fields
+are unset, that field will also remain unset on the BareMetalHost.
```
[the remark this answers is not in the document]
maelk: I added a `ready` field in the Metal3Data object.

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+
+When the Metal3Machine gets deleted, the CAPM3 controller will remove its
+ownerreference from the data template object. This will trigger the deletion of
+the generated Metal3Data object and the secrets generated for this machine.
```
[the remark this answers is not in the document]
maelk: definitely

design/metadata-handling.md
```diff
@@ -0,0 +1,668 @@
...
+          link: "bond0"
+
+      - ipv4:
+          id: "Baremetal"
+          link: "vlan1"
+          ipAddress:
+            start: "192.168.0.10"
+            end: "192.168.0.100"
+            subnet: "192.168.0.0/24"
+            step: 1
```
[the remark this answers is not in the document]
maelk: it was just forgotten in the metadata section

design/metadata-handling.md
maelk: the conditions are offset >= 0 and step >=1. The actual index value will always start from 0 and have increments by 1, but the rendering here can take an offset and a step to change the rendered value based on the index.

maelk: I added the following:
```
The following
  conditions must be matched :

    - `offset` >= 0
    - `step` >= 1

  if the step is unspecified (default value being 0), the controller will
  automatically change it for 1. 
```

design/metadata-handling.md
maelk: If the computed value goes out of bounds, the error status will be set
  with the error in the error message.

design/metadata-handling.md
maelk: done, thanks

design/metadata-handling.md
maelk: done, thank you

design/metadata-handling.md
maelk: I followed the exact same train of thoughts when writing it :)

design/metadata-handling.md
maelk: It's for the template rendering to get it quickly, rather than having to parse the ownerReferences.

design/metadata-handling.md
maelk: the Metal3DataTemplate controller, thanks

design/metadata-handling.md
maelk: the index itself, does not have an offset or a step. it will only increment by 1. The rendering of the index is what takes an offset and a step. I'll make it clearer

design/metadata-handling.md
maelk: I have added a whole section about the flow. About your questions : 
- the M3DT reconciler gets the M3Machine from the ownerreferences on the M3DT
- The secret would not be created if the `metaData` or `networkData` field is not set on the M3DT. in that case, the secret pointer is nil.
- for the partially automated flow, the user would give the secret directly in the Metal3Machine spec. In that case, the M3D controller will simply not create a `metaData` secret. If the `metaData` field is set on the M3DT, then it will also be created in the M3D, but not used in the M3Machine, because the `metaData` field in the spec is set.

design/metadata-handling.md
```diff
@@ -0,0 +1,750 @@
...
+### BareMetalHost changes
+
+BareMetalHost would have an added field `metaData` in the `Spec` pointing to a
+secret containing the metadata map. Bare Metal Operator would then pass this map
+to Ironic.
+
+Some values would be set by default to maintain compatibility:
+
+- **uuid**: This is the BareMetalHost UID
+- **metal3-namespace**: the name of the BareMetalHost
```
dhellmann, line 166:
```suggestion
- **metal3-namespace**: the namespace of the BareMetalHost
```

design/metadata-handling.md
dhellmann: It sounds like the M3M controller will wait for the M3D to have any missing secret references, then copy those to the status fields of the M3M. And if the M3M spec includes references created by the user, those will be copied to the status instead. That way by looking at the status fields we can tell which secrets are actually being used by the M3M. Is that right?

maelk: yes, this is correct.

design/metadata-handling.md
```diff
@@ -0,0 +1,750 @@
...
+
+A `dataTemplate` field would be added, consisting of an object reference
+to a Metal3DataTemplate object containing the templates for
+the metadata and network data generation for this Metal3Machine.
+A `renderedData` field will be added in the status and will be a reference to
+the Metal3Data object created for this machine. If the dataTemplate field is set
+but either the `renderedData`, `metaData` or `networkData` fields in the status
+are unset, then the Metal3Machine controller will wait until it can find the
+Metal3Data object and the rendered secrets. It will then populate those fields.
+
```
dukov, line 252: It's not clear what happened if dataTemplate, metaData and networkData are set together. Which one has higher priority?

design/metadata-handling.md
```diff
@@ -0,0 +1,750 @@
...
+  lastUpdated: "2020-04-02T06:36:09Z"
+```
+
+This object will be reconciled by its own controller. When reconciled,
+the controller will add a label pointing to the Metal3Cluster that has nodes
+linking to this object. The spec contains a `metaData` and a `networkData` field
+that contain a template of the values that will be rendered for all nodes.
+
+The `metaData` field will be rendered into a map of strings in yaml format,
+while `networkData` will be rendered into a json string that fulfills the
```
dukov, line 386: gophercloud requires map[string]interface{} for network data

maelk: fixed, thank you

maelk:
@dhellmann I added some status to make it more usable (found during the early stage implementation), otherwise this is now good to go for me. Should we put a lazy consensus on this ? Maybe by Wednesday ?

maelk:
Let's put a lazy consensus on this one by Friday, end of work day, eastern time. If there no additional comments, we'll proceed with the merge after that.

dhellmann:
This version looks good to me.

maelk:
hum, that merged a bit earlier than expected, I had not put a hold, sorry for that. If there are any comments, we can address them in a second PR

PR #80, opened 2020-03-18
Update design of Machine Remediation

n1r1:
Previous design was using new CRD, this one doesn't rely on CRD and utilizes BMO reboot api

### DIFF
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -95,7 +95,7 @@ itself, allowing it to host workloads.
 
 ## Proposal
 
-This proposal calls for a new machine remediation controller (MRC) in CAPBM
+This proposal calls for a new machine remediation controller (MRC) in CAPM3
 which watches for the presence of a `host.metal3.io/external-remediation` Machine
 annotation. If present, the controller will locate the Machine
 and BareMetalHost host objects via their annotations, and will annotate the BareMetalHost
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -99,10 +99,10 @@ This proposal calls for a new machine remediation controller (MRC) in CAPM3
 which watches for the presence of a `host.metal3.io/external-remediation` Machine
 annotation. If present, the controller will locate the Machine
 and BareMetalHost host objects via their annotations, and will annotate the BareMetalHost
-CR with a `reboot.metal3.io/machine-remediation`. Upon host power off, the controller
-will delete the node and then remove the annotation, to power on the host by
-utilizing [Baremetal-Operator reboot API](https://github.com/metal3-io/metal3-docs/blob/master/design/reboot-interface.md).
-
+CR with a `reboot.metal3.io/machine-remediation` to power cycle the host by
+utilizing the [Baremetal-Operator reboot API](https://github.com/metal3-io/metal3-docs/blob/master/design/reboot-interface.md).
+Upon host power off, the controller will delete the node and then remove the annotation,
+allowing it to be powered back on.
 After restart/crash of the new controller it will check for
 `reboot.metal3.io/machine-remediation` annotation on all hosts,
 and continue remediation process, to make sure that such host is
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -111,17 +111,21 @@ not left in powered off state.
 ## Assumptions
 
 MHC checks that each machine has a node, and if it doesn't after some timeout,
-it will mark the machine as unhealthy. There could be situations where this timeout
-expires few seconds after MRC powered on the machine, resulting in one or more
-redundant reboot(s).
-Current MHC implementation checks that `now() - Machine.LastUpdated < timeout`.
-MRC deletes the unhealthy node, which updates `Machine.LastUpdated`.
-We assume that this implementation won't change.
+it will mark the machine as unhealthy. If this timeout expires soon after the MRC powered
+on the host, it could result in one or more redundant reboot(s).
+However, the current MHC implementation checks that `now() - Machine.LastUpdated < timeout`
+and the MRC deletes the unhealthy node, which updates `Machine.LastUpdated`.
+
+We assume that this implementation won't change, that `timeout` is set long enough
+for a node to be both provisioned *and* rebooted, which implies that any `Node` 
+hitting this timeout is truly malfunctioning.
+
+Further, we require that equivalent functionality will exist in parallel implementations.
 
 In addition, MRC has some operations to perform between node deletion (which triggers
 update of `Machine.LastUpdated`) to powering on the host, and these operations
 are also counted against that timeout. These operations are removing
-MAO's annotation and removing the reboot annotation. We assume that this
+the unhealthy annotation from the Machine and removing the reboot annotation. We assume that this
 time is negligible, especially compared to a newly provisioned host, and in case of
 temporary failure to remove one of the annotations we will risk in additional reboot.
 
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -159,6 +159,26 @@ hosts referenced by node-less machines
 `host.metal3.io/external-remediation` annotation (by removing
 `reboot.metal3.io/machine-remediation` annotation from host)
 
+See truth table below:
+
+|Node Ref exists?|Needs remediation annotation|Powered on?|Reboot annotation|MRC action|
+|:-:|:-:|:-:|:-:|---|
+|0|1|1|0|Add reboot annotation|
+|1|1|1|0|Add reboot annotation|
+|1|1|0|1|Delete node|
+|0|0|1|0|nothing|
+|0|0|1|1|nothing|
+|0|1|0|0|nothing|
+|0|1|1|1|nothing|
+|1|0|0|0|nothing|
+|1|0|1|0|nothing|
+|1|0|1|1|nothing|
+|1|1|0|0|nothing|
+|1|1|1|1|nothing|
+|0|1|0|1|Remove needs-remediation annoation|
+|0|0|0|1|Remove reboot annotation|
+|1|0|0|1|Remove reboot annotation|
+
 ### Work Items
 
 - Create a new machine remediation controller in openshift/CAPBM repo
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -185,13 +185,16 @@ No changes are required to preserve the existing behaviour.
 
 ### Version Skew Strategy
 
-If CAPBM will be downgraded and BMO version will include the reboot functionality,
-we might end with a powered off host. This can happen if remediation process started,
-MRC requested for a power off by adding suffixed reboot annotation, and then CAPBM
-being downgraded to a version without MRC. There will be no one to remove that annotation
-and BMO will keep the host in powered off state until annotation removal.
-This will be solved as soon as the version skew is gone and they both use either
-reboot-supported versions or not at all.
+CAPBM and BMO are delivered via the same image (both are installed by MAO, but running
+in a different pod), therefore an upgrade or downgrade affecting one component will
+eventually be applied to the other and any version skew is temporary.
+During the transient case that CAPBM does not include the MRC, Host can be left in powered
+off state, since MRC could add reboot annotation, and then it get downgraded, and there will
+be no controller to remove the reboot annotation.
+During the transient case that BMO does not support the reboot API, MRC could request for a reboot,
+which won't commence until BMO upgrades to a version which supports reboot annotation.
+In both cases, the impact is just delayed remediation and will be resolved once there's
+no version skew. 
 
 ## Drawbacks [optional]
 
--- a/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
+++ b/design/annotation-for-power-cycling-and-deleting-failed-nodes.md
@@ -167,7 +167,8 @@ hosts referenced by node-less machines
 
 ### Work Items
 
-- Create a new machine remediation controller in CAPBM repo
+- Create a new machine remediation controller in openshift/CAPBM repo
+- Backport to upstream
 
 ### Dependencies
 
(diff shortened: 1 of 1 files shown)


n1r1:
An implementation of this enhancement is available [here](https://github.com/n1r1/cluster-api-provider-baremetal/blob/machine-remediation/pkg/controller/machine_remediation/machine_remediation_controller.go) for reference

fmuyassarov:
Don’t you think that we also need re-provision and re-introspection apart from powering off?

**Why re-introspection:**
In some cases where some physical characteristics of the host are changed due to some hardware related failure, for example disk replacement which was required because of the disk failure.
So, re-introspection would also be needed to regather facts about the same host again since it might have a new hardware characteristics due to the replacement ( e.g. disk replacement)

**Why re-provisioning**
In some cases only rebooting of the host might be enough to get back the host to the healthy condition.

/cc @dhellmann

n1r1:
@fmuyassarov I think it's a good idea and can be done in a separate PR. i.e. have an escalated remediation - first start with power-cycle and if problem persists we might want to try re-provisioning, etc.

beekhof:
I agree with keeping the scope focused on establishing an e2e power based flow for now.
Creating a remediation escalation path that includes re-provisioning and introspection is an excellent idea that is on our todo list, but we want to establish the basics first.

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -95,35 +95,51 @@ itself, allowing it to host workloads.
...
+utilizing [Baremetal-Operator reboot API](https://github.com/metal3-io/metal3-docs/blob/master/design/reboot-interface.md).
+
+After restart/crash of the new controller it will check for
+`reboot.metal3.io/machine-remediation` annotation on all hosts,
+and continue remediation process, to make sure that such host is
+not left in powered off state.
+
+## Assumptions
+
+MHC checks that each machine has a node, and if it doesn't after some timeout,
```
fmuyassarov: Does the MHC will be modifying CAPI level Machine object? Or should it be BareMetalMachine object?
Because, what I understood first is that, MHC will watch for Machine object, if Machine meets unhealthiness condition, then MHC will set the `host.metal3.io/external-remediation` annotation on corresponding BareMetalMachine object, meaning that MRC will be also watching for BareMetalMachine and if `host.metal3.io/external-remediation` annotation is present, then annotate corresponding BareMetalHost object with `reboot.metal3.io/machine-remediation`.
Could you please confirm, if I'm right that - MHC will watch for CAPI level Machine object but will annotate  the CAPM3 level BareMetalMachine object not the Machine object with `host.metal3.io/external-remediation`?
/cc @jan-est

fmuyassarov:
Ok, I agree @n1r1 , @beekhof 

Another question I have is that, 

> I agree with keeping the scope focused on establishing an e2e power based flow for now.
> Creating a remediation escalation path that includes re-provisioning and introspection is an excellent idea that is on our todo list, but we want to establish the basics first.

Sure, thanks @n1r1 , @beekhof

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
n1r1: MHC doesn't modify BareMetalHost. If a Machine is unhealthy, it will annotate the **Machine** with `host.metal3.io/external-remediation`. Now MRC kicks in (watches both Machine and BareMetalHost)

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -95,35 +95,51 @@ itself, allowing it to host workloads.
...
+expires few seconds after MRC powered on the machine, resulting in one or more
+redundant reboot(s).
+Current MHC implementation checks that `now() - Machine.LastUpdated < timeout`.
+MRC deletes the unhealthy node, which updates `Machine.LastUpdated`.
+We assume that this implementation won't change.
+
+In addition, MRC has some operations to perform between node deletion (which triggers
+update of `Machine.LastUpdated`) to powering on the host, and these operations
+are also counted against that timeout. These operations are removing
+MAO's annotation and removing the reboot annotation. We assume that this
```
dhellmann: There is no machine-api-operator upstream here in metal3. Is that going to change the timing at all?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -166,8 +205,16 @@ No changes are required to preserve the existing behaviour.
 
 ### Version Skew Strategy
 
-By shipping the new controller with the baremetal-operator that it consumes, we
-can prevent any version mismatches.
+CAPBM and BMO are delivered via the same image (both are installed by MAO, but running
```
dhellmann: This information is all specific to OpenShift. The issue still exists, but you should rewrite this to not refer to the OpenShift-specific configuration or CAPBM. For example, "If CAPM3 and the BMO are in different pods, an upgrade or downgrade ..."

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
fmuyassarov: @dhellmann  
We are actually quite interested in having MRC & MHC on upstream. During the last community meeting I was asking about how problematic would it be to backport MRC code from openshift/capbm (once it is there) to Metal3. But, as you also mentioned, the API versions are different, and there is no CRD. So, I'm actually starting to write MRC in parallel and wanted to make sure that writing it in parallel is better rather than waiting for the code first to be landed on openshift/capbm and then port to upstream?

dhellmann: Oh, it seems a shame to duplicate the effort of building it.

Is the only difference the type of resource the controller has to watch or update?

beekhof: Wait a second... there are `BareMetalMachine`s now too?
Can someone point me to the design doc where they were introduced?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
n1r1: Since there's no machine-api-operator in metal3, no one will ever declare the machine is unhealthy (machine-healthcheck-controller [MHC] is part of MAO).
Somehow, MHC functionality should be added to metal3, and that metal3-MHC will have to consider these assumptions.

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
fmuyassarov: > Wait a second... there are `BareMetalMachine`s now too?
> Can someone point me to the design doc where they were introduced?

Maybe these pictures will help you understand the current design:
[v1alpha3](https://github.com/Nordix/metal3-docs/blob/v1a3/crs-feruz/images/v1alpha3_crs.svg)
[v1alpha2](https://github.com/metal3-io/metal3-docs/blob/master/crs_reference.md)

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
n1r1: done

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
```diff
@@ -95,35 +95,51 @@ itself, allowing it to host workloads.
...
-This proposal calls for a new controller which watches for the presence of a
-specific Node annotation.  If present, the controller will locate the Machine
-and BareMetalHost host objects via their annotations, and (in serial) use the
-BareMetalHost API to power off the machine, delete the Node, and then power the
-machine back on again.
-
-In order to track the state of the recovery workflow, a [new
-CRD](https://github.com/kubevirt/machine-remediation/blob/master/pkg/apis/machineremediation/v1alpha1/machineremediation_types.go)
-is proposed.
+This proposal calls for a new machine remediation controller (MRC) in CAPBM
```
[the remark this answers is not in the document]
n1r1: done

jan-est:
Here is the overall design of CAPM3 
https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/architecture.md 

Idea in CAPM3 is that it watches cluster api level Machine object and if any changes appear it will update CAPM3 Metal3Machine.

Cluster API is already provide MachineHealthCheck for Machines: 
https://cluster-api.sigs.k8s.io/tasks/healthcheck.html

If Cluster API MachineHealthCheck is enough for us we might not need new healthcheck controller at all. We could just set annatations to Metal3Machines and/or BareMetalHost objects based on changes in Machine object. 

On the other hand that might not be enough for us and we need to implement our own Metal3MachineHealthCheck controller that also watches Cluster API Machine object and on top of that do some baremetal specific health checks before adding those annatations into Metal3Machine and/or BareMetalHost objects. 

I am not quite sure how we should do health check in upsteam CAPM3 level? But for sure before we start implementing anything, we need to agree how to do it?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
fmuyassarov: > @fmuyassarov Can you clarify whether
> 
> > I'm actually starting to write MRC in parallel
> 
> means you're working on the backport already, or starting a completely new implementation?

I was just trying out backporting option while experimenting with how current implementation of MRC would work in the context of CAPM3, but didn’t start full implementation.

beekhof: > I was just trying out backporting option while experimenting with how current implementation of MRC would work in the context of CAPM3, but didn’t start full implementation.

Ah, very cool.

Regarding the diagrams, they certainly help explain what but why. Is this pattern of every object having a provider specific shadow mandated by the cluster API?

fmuyassarov: Are you referring to the relationship/mapping between CAPI API version and Provider API version?

beekhof: I mean why Machines and Clusters have BareMetalMachines and BareMetalClusters

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
beekhof: > If Cluster API MachineHealthCheck is enough for us we might not need new healthcheck controller at all. We could just set annotations to Metal3Machines and/or BareMetalHost objects based on changes in Machine object.

The Cluster API MachineHealthCheck came from the same folks that wrote the one in machine-api-operator.  So my expectation is that they would (or could) be compatible.

beekhof:
Urgh, posted this in the wrong place previously...
 
> If Cluster API MachineHealthCheck is enough for us we might not need new healthcheck controller at all. We could just set annotations to Metal3Machines and/or BareMetalHost objects based on changes in Machine object.

The Cluster API MachineHealthCheck came from the same folks that wrote the one in machine-api-operator. So my expectation is that they would (or could) be compatible

beekhof:
Just checked, its essentially the same without an equivalent of https://github.com/openshift/machine-api-operator/commit/584f36b09a349d8fa9132735f5e8b36b8d04846c

BUT

Perhaps this is a good opportunity to imagine what the solution would look like if the decision to reboot vs. [delete the machine](https://github.com/kubernetes-sigs/cluster-api/blob/e65fb5c6cd271ab9dd80f28d1fc7f0ac79ddae4a/controllers/machinehealthcheck_targets.go#L306)  was made in the MachineHealthCheck code.

jan-est:
> 
> Perhaps this is a good opportunity to imagine what the solution would look like if the decision to reboot vs. [delete the machine](https://github.com/kubernetes-sigs/cluster-api/blob/e65fb5c6cd271ab9dd80f28d1fc7f0ac79ddae4a/controllers/machinehealthcheck_targets.go#L306) was made in the MachineHealthCheck code.

Yes I agree. Reboot is something we need to figure out. So should we implement our own metal3MachineHealthCheck in to CAPM3? What do you think @beekhof @fmuyassarov

beekhof:
> > Perhaps this is a good opportunity to imagine what the solution would look like if the decision to reboot vs. [delete the machine](https://github.com/kubernetes-sigs/cluster-api/blob/e65fb5c6cd271ab9dd80f28d1fc7f0ac79ddae4a/controllers/machinehealthcheck_targets.go#L306) was made in the MachineHealthCheck code.
> 
> Yes I agree. Reboot is something we need to figure out. So should we implement our own metal3MachineHealthCheck in to CAPM3? What do you think @beekhof @fmuyassarov

IMHO reboot should be a first-class remediation option that is understood by MHC.
That would put it in the best position to allow:
- delete/reprovision only
- reboot only
- delete after N failed reboot attempts 

I wouldn't be a fan of duplicating https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/machinehealthcheck_targets.go#L278 in m3MHC

jan-est:
> > > Perhaps this is a good opportunity to imagine what the solution would look like if the decision to reboot vs. [delete the machine](https://github.com/kubernetes-sigs/cluster-api/blob/e65fb5c6cd271ab9dd80f28d1fc7f0ac79ddae4a/controllers/machinehealthcheck_targets.go#L306) was made in the MachineHealthCheck code.
> > 
> > 
> > Yes I agree. Reboot is something we need to figure out. So should we implement our own metal3MachineHealthCheck in to CAPM3? What do you think @beekhof @fmuyassarov
> 
> IMHO reboot should be a first-class remediation option that is understood by MHC.
> That would put it in the best position to allow:
> 
>     * delete/reprovision only
> 
>     * reboot only
> 
>     * delete after N failed reboot attempts
> 
> 
> I wouldn't be a fan of duplicating https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/machinehealthcheck_targets.go#L278 in m3MHC

@beekhof yes we need to add our own remediation logic and not to duplicating one from cluster api. I also give +1 for your remediation proposal with reboot as a first-class option. 

So I think we need to write our own metal3machinehealtcheck controller to reconcile metal3machines and add annotations on BMH? And implement MHC the way @beekhof  suggested. But should we do this as a CR based implementation?

design/annotation-for-power-cycling-and-deleting-failed-nodes.md
fmuyassarov: Because this is the way of representing those Cluster and Machine objects in infrastructure provider level. Basically, each infra provider has their own objects (INFRAXCluster, INFRAXMachine...so on) which are eventually linked to Cluster API objects (Machines & Clusters...son on). See [AWS example](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/982d2200ded3123e3803cd0d3c9197f4af5380d0/templates/cluster-template.yaml#L12).

zaneb: My 2c: it makes sense to me to keep the annotation on Machine (not BareMetalMachine) because the remediation system shouldn't need to know details about what type of platform it is running on. (Ultimately, we would hope to make this part of the Cluster API and supported by all providers.)

n1r1: Yep, we're working on having some of this upstream

n1r1:
Closing this issue as we're currently focusing on upstream (CAPI) proposals, and this is out of date.

/close

PR #87, opened 2020-05-08
extend the design principle document

dhellmann:
Add a few more design principles based on recent design reviews in
this repository.

### DIFF
--- a/design/api-design-principles.md
+++ b/design/api-design-principles.md
@@ -73,3 +73,41 @@ assumptions through the metal3 API.
 Don't make assumptions about what the BaremetalHost will be used
 for. The Machine API is not the only consumer and running a Kubernetes
 node is not the only thing users may want to do with the Host.
+
+### Not Every Feature Needs to Go into the baremetal-operator
+
+Metal3 is designed to take advantage of the microservice nature of
+kubernetes. New features may require changing the BareMetalHost
+schema, but just as often it will be possible to add a new feature
+using a new API.
+
+Provisioning hosts can be complicated. The BareMetalHost API is
+designed so the `baremetal-operator` can eliminate all of the
+complexity beyond basic host management and provisioning
+operations. Other APIs and controllers drive the decisions about how
+to configure a host, which image to use, etc. For example, the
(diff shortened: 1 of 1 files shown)


design/api-design-principles.md
```diff
@@ -92,3 +96,41 @@ assumptions through the metal3 API.
...
+`baremetal-operator` so that we can plug Metal3 into the cluster API
+ecosystem, but also so different hosts can be configured in different
+ways using the Machine API and so the BareMetalHost API can be used by
+tools other than the cluster API provider.
+
+### Make Components Optional
+
+We want to avoid making the Metal3 components so tightly coupled that
+in order to use any of them a user has to deploy all of them. Where
+possible, features should be optional or pluggable so that users can
```
zaneb, line 122: This was confusing to read because 'components' sounds like we're saying e.g. cluster-api-provider-metal3 should be optional and you can use the baremetal-operator in other ways, but saying that features should be optional sounds like it might be trying to say something different.

dhellmann: Yeah, I was trying not to be too specific with examples, but maybe I should.

I want to be able to use the BareMetalHost without a Metal3Template or its controller. I want to be able to use the new IPAM stuff @maelk is working on without that assuming that a template owns the IP claim. I also want to make sure we can use the BareMetalHost and baremetal-operator without the cluster-api-provider-metal3, in case someone (*cough* Red Hat *cough*) wants a different machine API implementation.

So, I do want features to be optional and implementations replaceable, and the way to achieve that is to not have the implementations tightly coupled.

Suggestions on better wording?

dhellmann:
Rebased to get the markdownlint job output.

design/api-design-principles.md
russellb: just s/features/components/ ?

russellb:
lgtm, but maybe you want to tweak the wording on the sentence that zane flagged first?

design/api-design-principles.md
dhellmann: Done and rebased to drop the ToC since that global change landed in master.

PR #89, opened 2020-05-08
stop using generated table of contents

dhellmann:
As part of adding linter validation to this repo we added a customized
version of the script for generating table of contents at the top of
each file. It's a bit hacky, and feedback on slack indicated that
droping the ToC entirely instead of fixing the script might be OK.


russellb:
this is fine with me.

PR #90, opened 2020-05-13
Add pivoting/move documentation

kashifest:
This PR adds the pivoting/move documentation.

### DIFF
--- a/docs/move.md
+++ b/docs/move.md
@@ -1,0 +1,143 @@
+# Move
+
+In the context of Metal3 Provider, _Move_ or _Pivoting_ is the process of moving
+CAPI and CAPM3 objects from one management k8s cluster to another. In Metal3,
+this process is performed using the
+[clusterctl](https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html) tool
+provided by CAPI. For the rest of the document we will refer this process as
+_Move_.
+
+The following objects are moved during this process:
+
+**Main Objects:** Cluster, Machine, MachineSet, MachineDeployment,
+KubeadmControlplane, Metal3Cluster, Metal3Machine, and BareMetalHost
+
+**Related Objects:** Cluster Secrets, BareMetalHost Secrets, Configmaps,
+KubeadmConfig, Metal3MachineTemplate
+
+## Prerequisite
+
+* **`clusterctl` is used to initialize both the bootstrap and target cluster.**
+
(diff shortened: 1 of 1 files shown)


docs/pivoting.md
```diff
@@ -0,0 +1,142 @@
+# Pivoting/Move
```
dhellmann, line 1: I'm not sure it's necessary to use both terms every time we refer to the process. Which term is preferred? Let's use that in the title, and mention the alternative once in the introduction, then use the preferred term everywhere else.

docs/pivoting.md
```diff
@@ -0,0 +1,142 @@
...
+* **BMH objects have correct status annotation.**
+
+  Since BMH _status_ holds important information regarding the BMH itself, we
+  have to move BMH with status and reconstruct the BMH with correct status in
+  target  cluster before it is being reconciled. This is now done through BMH
+  status annotation in BMO.
+
+* **Maintain connectivity towards provisioning network.**
+
+  Baremetal machines boot over a network with a dhcp server. This requires
```
dhellmann, line 45:
```suggestion
  Baremetal machines boot over a network with a DHCP server. This requires
```

docs/pivoting.md
kashifest: Agree. I have changed it accordingly.

docs/pivoting.md
kashifest: Done

docs/pivoting.md
```diff
@@ -0,0 +1,143 @@
+# Move
```
dhellmann, line 1: The file is called "pivoting". Can we make the file and title match?

docs/pivoting.md
```diff
@@ -0,0 +1,143 @@
...
+## Prerequisite
+
+* **`clusterctl` is used to initialize both the bootstrap and target cluster.**
+
+  If the provider components are not installed using `clusterctl`, it will not
+  be able to identify the objects to move.  Initializing the cluster using
+  `clusterctl` essentially adds the following labels in the CRDs of each object
+  related.
+
+  ```bash
```
dhellmann, line 27: this is YAML text, right?

docs/pivoting.md
kashifest: Done

docs/pivoting.md
kashifest: Indeed. I have changed it.

PR #93, opened 2020-05-18
add process doc for migrating reviewer permissions to owners files

dhellmann:

PR #94, opened 2020-05-20
add policy for managing reviewers

dhellmann:
Reviewers have fewer permissions than maintainers, so we don't need such
a detailed policy for managing them.

### DIFF
--- a/processes/managing-reviewers.md
+++ b/processes/managing-reviewers.md
@@ -1,0 +1,50 @@
+# Managing Reviewers
+
+## Status
+
+implementable
+
+## Summary
+
+After the migration to use OWNERS files to manage reviewers, it is
+easier for us to add reviewers to separate repositories. This document
+describes the process for adding reviewers to a metal3 project repo.
+
+### Goals
+
+1. Describe a process for adding reviewers.
+2. Keep the process light-weight.
+
+### Non-Goals
+
+1. Change the process for adding [maintainers](../maintainers/).
+
+## Proposal
+
+Anyone can propose a patch to update an OWNERS file in a repository to
+add a reviewer. The patch should be submitted as a standalone PR,
+rather than being linked to any other contribution.
+
+The reviewer list for each repository will be pruned over time to
+remove contributors who are inactive.
+
+Reviewers may also be removed for behaving in a manner that other
+maintainers view as detrimental to the project, following the process
+described for maintainers in [Revoking Approval
+Access](../maintainers#revoking-approval-access).
+
+Pull requests to add or remove reviewers from OWNERS files should be
+approved using the same policy as other changes: One person with
+approval permission and another with at least reviewer permission must
+accept the PR.
+
+### Risks and Mitigations
+
+Ideally new reviewers will have already contributed to the project,
+either through code, documentation, reviews, or design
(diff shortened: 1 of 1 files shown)


dhellmann:
This is tied to #93, so we should give time for discussion.

/hold

processes/managing-reviewers.md
```diff
@@ -0,0 +1,49 @@
...
+remove contributors who are inactive.
+
+Reviewers may also be removed for behaving in a manner that other
+maintainers view as detrimental to the project, following the process
+described for maintainers in [Revoking Approval
+Access](../maintainers#revoking-approval-access).
+
+Two contributors with approval permission for the repository are
+required to approve any PR to add or remove contributors from an
+OWNERS file.
```
russellb, line 38: I wonder if it should be 1 approver + 1 reviewer or approver.  That would match the workflow enforced for all PRs by prow right now.

processes/managing-reviewers.md
```diff
@@ -0,0 +1,49 @@
...
+1. Change the process for adding [maintainers](../maintainers/).
+
+## Proposal
+
+Anyone can propose a patch to update an OWNERS file in a repository to
+add a reviewer. The patch should be submitted as a standalone PR,
+rather than being linked to any other contribution.
+
+The reviewer list for each repository will be pruned over time to
+remove contributors who are inactive.
```
maelk, line 29: should we put a clear criteria (like no commits in the last x months for example) to make it objective ?

processes/managing-reviewers.md
dhellmann: I could support that, sure.

processes/managing-reviewers.md
dhellmann: I always hesitate to make hard rules like that. What happens in a stable repo for a feature-complete component, where there is no real need to change code for several months in a row? We wouldn't want to remove all of the reviewers.

maelk: ok. Then I guess anybody can propose to prune the reviewer list as long as it is approved by one approver and one reviewer as mentioned by Russell ?

dhellmann: That seems OK to me. I would expect anyone reviewing a patch to the OWNERS file to be careful with approving it.

processes/managing-reviewers.md
dhellmann: Updated

dhellmann:
I propose that we use lazy consensus rules to approve this by 12 June. That's a couple of days after our next community meeting, which gives us plenty of time to discuss it here as well as synchronously.

processes/managing-reviewers.md
```diff
@@ -0,0 +1,50 @@
...
+Reviewers may also be removed for behaving in a manner that other
+maintainers view as detrimental to the project, following the process
+described for maintainers in [Revoking Approval
+Access](../maintainers#revoking-approval-access).
+
+Pull requests to add or remove reviewers from OWNERS files should be
+approved using the same policy as other changes: One person with
+approval permission and another with at least reviewer permission must
+accept the PR.
+
```
jan-est, line 40: I think this is a good policy. +1 for this one

PR #95, opened 2020-05-22
Re-inspection API proposal

fmuyassarov:
This proposal adds a new API to request inspection.

### DIFF
--- a/design/re-introspection-interface.md
+++ b/design/re-introspection-interface.md
@@ -1,0 +1,88 @@
+# re-inspection API
+
+## Status
+
+implementable
+
+## Summary
+
+A declarative API is proposed to request the baremetal operator to
+inspect a `Ready` BareMetalHost.
+
+## Motivation
+
+We would like to have an interface to allow a user to re-gather hardware
+inventory of a `Ready` BareMetalHost when a hardware replacement is made.
+When a user of the underlying infrastructure makes some changes to the actual
+server (e.g. replace or add NIC, disk, etc.), the latest hardware inventory
+including those changes need to be re-collected and updated on the spec of the
+corresponding BareMetalHost object without having to delete it.
+
+Implementation of this proposal is based on using annotation (similar to
+[Reboot API](https://github.com/metal3-io/metal3-docs/blob/master/design/reboot-interface.md))
+to request inspection of a `Ready` BareMetalHost.
+Once the annotation is set on BareMetalHost, the baremetal operator will
+request hardware inspection of the host from Ironic.
+
+### Goals
+
+- A declarative API to perform inspection
+- Use this API for future Metal³ remediation controller
+
+### Non-Goals
+
+- Automated mechanism to trigger inspection
+
+## Proposal
+
+We follow Reboot API implementation and expect to implement the similar
+annotation based interface for a user to trigger inspection.
+
+`inspect.metal3.io` annotation form on BareMetalHost object
+triggers the controller to query Ironic for inspection of a host. This form
+has set-and-forget semantics and controller removes the annotation once
+inspection is completed.
+While the host is being inspected, the BareMetalHost will stay in
+`Inspecting` state until the process is completed.
+
+Re-inspection API can be requested only when BareMetalHost is in `Ready`
+state. If a re-inspection request is made while BareMetalHost is any other
+state than `Ready`, the request will be ignored. This is important in order to
+not reboot a BareMetalHost (e.g. when `spec.provisioning.state == provisioned`)
+or avoid having unintended inconsistent states.
+
+|BMH state|Externally provisioned|API action|Annotation|
+|---|---|---|---|
+|Ready|No|move to Inspecting state|delete|
+|Inspecting|No|nothing|delete|
+|Provisioning|No|nothing|keep it until BMH is in Ready state|
+|Provisioned|No|nothing|keep it until BMH is in Ready state|
+|Provisioned|Yes|nothing|keep it until BMH is in Inspecting state|
+
+After completing inspection, previous inspection data should be updated
+both in Ironic and on the spec of the BareMetalHost object. Both
+`status.operationHistory.inspect.start` and
+`status.operationHistory.inspect.end` timestamps should be updated accordingly.
+
+### User stories
+
+#### Story1
+
+- As a cluster admin, I would like to have a simple way of triggering
+  inspection for my server after I replace NIC, disk, etc.
+
+## Future Enhancements
+
+Re-inspection API and reboot API interfaces can be modified to have a
+more formal interface for future Metal³ remediation controller.
+
+## Alternatives
+
+One alternative approach to keep hardware details updated is to run Ironic
+Python Agent (IPA) as a pod on the node which would be contstantly updating the
+hardware details of the host.
+
+## References
+
+- Reboot API [proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/reboot-interface.md)
+- Reboot API [implementation](https://github.com/metal3-io/baremetal-operator/pull/424)


design/re-introspection-interface.md
```diff
@@ -0,0 +1,76 @@
...
+triggers the controller to query Ironic for introspection of a host. This form
+has set-and-forget semantics and controller removes the annotation once
+introspection is completed.
+While the host is being re-introspected, the BareMetalHost will stay in
+`Inspecting` state until the process is completed.
+
+Re-introspection API can be requested only when BareMetalHost is in `Ready`
+state. If a re-introspection request is made while BareMetalHost is any other
+state than `Ready`, the request will be ignored, i.e., annotation will
+not be set on BareMetalHost object. This is important in order to not reboot a
```
dhellmann, line 52: If the user sets the annotation and we're going to ignore it, is that an error? Do we remove the annotation, or leave it there? What happens to a provisioned host with the annotation that is then deprovisioned? Would that trigger introspection again?

design/re-introspection-interface.md
```diff
@@ -0,0 +1,76 @@
...
+both in Ironic and on the spec of the BareMetalHost object. Both
+`status.operationHistory.inspect.start` and
+`status.operationHistory.inspect.end` timestamps should be updated accordingly.
+
+### User stories
+
+#### Story1
+
+- As a cluster admin, I would like to have a simple way of triggering
+  introspection for my server after I replace NIC, disk, etc.
```
dhellmann, line 66: Another approach we have discussed for keeping the hardware details up to date is to run introspection from a pod running on the node. You could include that in the alternatives section of this document.

zaneb:
Why not just have the user delete the HardwareDetails if they want to trigger this, rather than have an annotation?

design/re-introspection-interface.md
maelk: That is an interesting approach. However it would work only if the hardware change is not breaking. In case the user would change the ethernet card, the mac address would change. Considering that the network configuration of the host is based on the old mac address, the node might not come back up, and the daemon set would then not run and be unable to report the new hardware. We probably still need to introspect with IPA in some corner cases. The rest of the cases would probably be covered by the daemon set approach indeed.

design/re-introspection-interface.md
maelk: Could it be that we leave the annotation there until the host is in a state where we can perform the operation ? How would you otherwise prevent the annotation to be set ?

fmuyassarov: I think we should not introspect a provisioned BMH, so that we don’t end up with deprovisioning it.

I can think of three different options about what to do with the annotation set on a provisioned BMH.

1. If the user sets the annotation on a provisioned BMH nothing will happen from the user’s perspective, while from the operator's side we still remove the annotation without calling introspection. 
2. Keep the annotation until `actionDeprovisioning` is called. After deprovisioning is completed, call the introspection and remove the annotation.
3. Keep the annotation until ` actionDeprovisioning` is called. After deprovisioning is completed, remove the annotation but don’t call the introspection.  

Third option is somehow similar to Reboot API, in a sense that you can’t ask for a power off while the BMH is in Ready state, but if you still set the annotation (suffixed) on a Ready BMH then nothing will happen but the annotation will be kept and removed during the provisioning If I’m not wrong.

However, the second option makes more sense to me and seems safe because introspection will be called only after deprovisioning and the user will not have to delete the BMH in order to keep the hardware details up to date.

fmuyassarov:
> Why not just have the user delete the HardwareDetails if they want to trigger this, rather than have an annotation?

I think annotation would be easier for end users that don't have internal knowledge of BMO.
If we trigger introspection by deleting HardwareDetails then I’m guessing we would do it from the code which is I think something that isn’t trivial for end users.

jan-est:
> > Why not just have the user delete the HardwareDetails if they want to trigger this, rather than have an annotation?
> 
> I think annotation would be easier for end users that don't have internal knowledge of BMO.
> If we trigger introspection by deleting HardwareDetails then I’m guessing we would do it from the code which is I think something that isn’t trivial for end users.

Using annotation is better aligned with other apis. I think it could be confusing from the user’s point of view if introspection would work differently than, for example, reboot or pause.

design/re-introspection-interface.md
fmuyassarov: @dhellmann Does second option make sense to you?

n1r1: Another option could be an admission webhook which validates that the annotation being set only on `Ready` state.
Otherwise, a clear error is being returned to the user.

zaneb:
> > Why not just have the user delete the HardwareDetails if they want to trigger this, rather than have an annotation?
> 
> I think annotation would be easier for end users that don't have internal knowledge of BMO.
> If we trigger introspection by deleting HardwareDetails then I’m guessing we would do it from the code which is I think something that isn’t trivial for end users.

`kubectl patch bmh myhost --type=merge --patch='{"status": {"hardware": null}}'`

But answering my own question, we can't do this because the HardwareDetails are in the Status not the Spec, which most users should not have RBAC access to write to.

design/re-introspection-interface.md
zaneb: I would suggest to ignore the annotation unless we are in the Ready or Inspecting states.
In the Ready state, if the annotation exists (and ExternallyProvisioned is not set) then move to the Inspecting state.
In the Inspecting state, delete the annotation if it exists.
(This is another way of describing option 2.)

The reason option 3 makes sense for the reboot one is that when you provision the server, it *has* been rebooted so that counts as success (and we go ahead and remove the annotation). The same does not apply to inspection, so that wouldn't be appropriate here.

dhellmann: The admission web hook @n1r1 mentions would certainly produce the clearest result. Although we (or at least I) have been trying to avoid adding web hooks.

Maybe this is really something else to solve as part of the API decomposition? If the inspection data is its own resource, deleting it could trigger reprovisioning.

design/re-introspection-interface.md
dhellmann: Ironic also uses the MAC addresses to identify a host, so if it changes then it may be necessary to re-register the host or at least manually update the MAC in the host CRD and have the operator update ironic.

design/re-introspection-interface.md
zaneb: > Maybe this is really something else to solve as part of the API decomposition? If the inspection data is its own resource, deleting it could trigger reprovisioning.

Yeah, I think that could be the way to go long-term (which is why I asked about just deleting the hardware details from the CR). But given that the data is in the Status, I think the annotation is a reasonable way of approximating the same semantics in the short term.

fmuyassarov: I agree with @n1r1 's point regarding webhooks, but as @zaneb mentioned that could be something as a future enhancement (I can include it in the proposal). Maybe for now we can live with annotations, but soon we will anyway need to somehow formalize those annotations we use for introspection & reboot & power off/on.

fmuyassarov: > I would suggest to ignore the annotation unless we are in the Ready or Inspecting states.
> In the Ready state, if the annotation exists (and ExternallyProvisioned is not set) then move to the Inspecting state.
> In the Inspecting state, delete the annotation if it exists.
> (This is another way of describing option 2.)

Thanks, will add it.

> The reason option 3 makes sense for the reboot one is that when you provision the server, it _has_ been rebooted so that counts as success (and we go ahead and remove the annotation). The same does not apply to inspection, so that wouldn't be appropriate here.

Yeah, make sense.

hardys: Could we also consider the case where a host previously failed to inspect, and we want to retry?  Currently there's no way to recover from that situation (temporary network outage etc) without re-creating the BMH, so using this annotation could be more convenient in that case?

zaneb: @hardys tbh I think we should probably just change that behaviour so it always tries again.

jan-est:
This looks good to me

design/re-introspection-interface.md
```diff
@@ -0,0 +1,90 @@
...
+`reintrospection.metal3.io` annotation form on BareMetalHost object
+triggers the controller to query Ironic for introspection of a host. This form
+has set-and-forget semantics and controller removes the annotation once
+introspection is completed.
+While the host is being re-introspected, the BareMetalHost will stay in
+`Inspecting` state until the process is completed.
+
+Re-introspection API can be requested only when BareMetalHost is in `Ready`
+state. If a re-introspection request is made while BareMetalHost is any other
+state than `Ready`, the request will be ignored, i.e., annotation will
```
zaneb, line 51: "i.e. annotation will not be set" is confusing. The user can set the annotation whenever they like; it's what we do about it that's at issue.

design/re-introspection-interface.md
```diff
@@ -0,0 +1,90 @@
...
+BareMetalHost (e.g. when `spec.provisioning.state == provisioned`) or avoid
+having unintended inconsistent states.
+
+|BMH state|Externally provisioned|API action|Annotation|
+|---|---|---|---|
+|Ready|False|call introspection|delete|
+|Inspecting|False|nothing|delete|
+|Provisioning|False|nothing|keep it until BMH is in Ready state|
+|Provisioned|False|nothing|keep it until BMH is in Ready state|
+|Provisioned|True|nothing|delete|
```
zaneb, line 62: This state should be ExternallyProvisioned I think? I don't see why we would bother to delete the annotation in this state.

design/re-introspection-interface.md
```diff
@@ -0,0 +1,90 @@
...
+Re-introspection API can be requested only when BareMetalHost is in `Ready`
+state. If a re-introspection request is made while BareMetalHost is any other
+state than `Ready`, the request will be ignored, i.e., annotation will
+not be set on BareMetalHost object. This is important in order to not reboot a
+BareMetalHost (e.g. when `spec.provisioning.state == provisioned`) or avoid
+having unintended inconsistent states.
+
+|BMH state|Externally provisioned|API action|Annotation|
+|---|---|---|---|
+|Ready|False|call introspection|delete|
```
zaneb, line 58: Rather than "call introspection" I would say "move to Inspecting state".

It may not be possible to delete the annotation as part of that state change; I'm not sure it's possible to atomically update both status and annotations at the same time. In any event, there's no need for this. Just delete the annotation in the Inspecting state.

design/re-introspection-interface.md
```diff
@@ -0,0 +1,90 @@
...
+### Non-Goals
+
+- Automated mechanism to trigger re-introspection
+
+## Proposal
+
+We follow Reboot API implementation and expect to implement the similar
+annotation based interface for a user to trigger re-introspection.
+
+`reintrospection.metal3.io` annotation form on BareMetalHost object
```
zaneb, line 42: Ironic terminology freely mixes 'inspection' and 'introspection', but the BMH API is pretty consistent at using 'inspection'. I'd suggest we use `inspect.metal3.io` for consistency with the existing `reboot.metal3.io` annotation.

design/re-introspection-interface.md
fmuyassarov: Fixed

design/re-introspection-interface.md
fmuyassarov: Yes, I was't really sure what to do on state. Fixed, thanks.

design/re-introspection-interface.md
fmuyassarov: Updated.

design/re-introspection-interface.md
fmuyassarov: Good point, thanks. Updated.

design/re-introspection-interface.md
```diff
@@ -0,0 +1,88 @@
...
+### Non-Goals
+
+- Automated mechanism to trigger inspection
+
+## Proposal
+
+We follow Reboot API implementation and expect to implement the similar
+annotation based interface for a user to trigger inspection.
+
+`inspection.metal3.io` annotation form on BareMetalHost object
```
zaneb, line 41: I slightly prefer `inspect` (imperative verb) to `inspection` (noun) for consistency with reboot, but I'll let others comment.

design/re-introspection-interface.md
```diff
@@ -0,0 +1,88 @@
...
+not reboot a BareMetalHost (e.g. when `spec.provisioning.state == provisioned`)
+or avoid having unintended inconsistent states.
+
+|BMH state|Externally provisioned|API action|Annotation|
+|---|---|---|---|
+|Ready|No|move to Inspecting state|delete|
+|Inspecting|No|nothing|delete|
+|Provisioning|No|nothing|keep it until BMH is in Ready state|
+|Provisioned|No|nothing|keep it until BMH is in Ready state|
+|Provisioned|Yes|nothing|keep it until BMH is in Ready state|
```
zaneb, line 60: Note that Externally provisioned is actually a [separate state](https://github.com/metal3-io/baremetal-operator/blob/master/docs/baremetalhost-states.md#externally-provisioned).

design/re-introspection-interface.md
```diff
@@ -0,0 +1,88 @@
...
+
+Re-inspection API can be requested only when BareMetalHost is in `Ready`
+state. If a re-inspection request is made while BareMetalHost is any other
+state than `Ready`, the request will be ignored. This is important in order to
+not reboot a BareMetalHost (e.g. when `spec.provisioning.state == provisioned`)
+or avoid having unintended inconsistent states.
+
+|BMH state|Externally provisioned|API action|Annotation|
+|---|---|---|---|
+|Ready|No|move to Inspecting state|delete|
```
zaneb, line 56: Last column should be "keep it until BMH is in Inspecting state".

fmuyassarov: Updated.

design/re-introspection-interface.md
fmuyassarov: I agree with you. It was my bad with copy-paste.Updated.

fmuyassarov:
@dhellmann , @zaneb is the current state of the proposal OK to be merged?

russellb:
/lgtm

based on feedback from doug and zane

PR #96, opened 2020-05-27
template: move design details section before implementation details

dhellmann:
It has been bothering me that we talk about implementation details
before design details. I think that was a mistake in importing the
original version of the template, so let's switch the order of those
sections.


maelk:
/hold
/lgtm
Feel free to remove the hold when ready

russellb:
lgtm

/hold cancel

PR #98, opened 2020-06-03
WIP: host discovery design

dhellmann:
Addresses https://github.com/metal3-io/baremetal-operator/issues/41

### DIFF
--- a/design/host-discovery-controller/host-discovery.md
+++ b/design/host-discovery-controller/host-discovery.md
@@ -1,0 +1,358 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# host-discovery
+
+## Status
+
+implementable
+
+## Summary
+
+Ironic Inspector is capable of discovering new hosts. This should be
+an optional capability of metal3. When we discover a new host, it
+should result in automatically creating a corresponding
+`BareMetalHost` resource.
+
+## Motivation
+
+There are several workflows for adding nodes to existing clusters that
+would benefit from not forcing the user to register the hosts in
+advance. Especially in large data centers, it may be desirable to
+build a small cluster and then attach additional hosts to the
+provisioning network and deploy them by simply powering them on and
+using PXE to boot an agent that knows how to reach back to the
+cluster.
+
+### Goals
+
+1. Describe how we can take advantage of Ironic’s existing discovery
+   features to automatically add hosts to clusters.
+2. Describe how the cluster can, and cannot, manage a host without BMC
+   credentials.
+
+### Non-Goals
+
+1. Describe a way to set BMC credentials automatically for discovered
+   hosts.
+2. Attempt to make it possible to fully manage a host without BMC
+   credentials.
+
+## Proposal
+
+### User Stories
+
+#### Adding a host and automatically provisioning
+
+*As a cluster admin, I want to add a node to a cluster without
+pre-configuring the BMC and have the cluster automatically provision
+the host as a node.*
+
+Provisioning the hosts to make them nodes can take quite a while. It
+would be nice to be able to start that process for a bunch of hosts,
+then configure their BMC credentials later to enable full management
+of the hosts. There are also situations where the BMC credentials
+might never be added (proof-of-concept or test clusters that are never
+going to be “production ready”).
+
+By combining Ironic’s discovery feature, Ironic’s fast-track
+provisioning feature, MachineSet scaling, and some new work in metal3
+we should be able to make it possible for a user to:
+
+1. Attach a host to the provisioning network
+2. Power the host on
+3. Have a `BareMetalHost` created with hardware inventory details
+4. Optionally classify the host (using the
+   `hardware-classification-controller`)
+5. Automatically scale the appropriate MachineSet based on labels on
+   the host
+6. Automatically provision the host as a node in the cluster
+
+#### Adding a host and not automatically provisioning
+
+*As a cluster admin, I want to add a node to a cluster without
+pre-configuring the BMC and NOT have the cluster automatically
+provision the host as a node.*
+
+This use case is different from the previous case because the
+`MachineSet` would not scale automatically and so the host would not
+be provisioned. This gives cluster admins an opportunity to review the
+host before provisioning, to ensure that it should be a member of the
+cluster and is not misconfigured (PXE booting from the wrong
+environment) or a malicious actor.
+
+#### Replacing a faulty node
+
+*As a data center staff member, I want to replace a faulty node in a
+cluster quickly.*
+
+The idea here is to be able to take a “broken” host out of service and
+replace it with a standby host without having access to the kubernetes
+API or UI at all. If we assume the broken host is broken enough that
+it is completely offline, it should be possible to remove it from the
+rack and replace it with another physical host, which is then
+provisioned automatically.
+
+## Design Details
+
+When Ironic is configured to support its existing discovery feature,
+unknown hosts that boot the Ironic agent will have their hardware
+details registered with `ironic-inspector` and have a node created in
+the Ironic database. Typically this happens via the PXE boot
+configuration, but in the future it could happen in other ways such as
+having a bootable ISO image with the agent on it. How the agent ends
+up on the host should not matter in the rest of the workflow.
+
+The data collected for an unknown host includes the same inspection
+data available for any other host, but does not include the BMC
+credentials. We can use the data to partially create a `BareMetalHost`
+resource for the host by polling the Ironic API to compare the list of
+nodes it knows about with the list of host resources. A new
+`host-discovery-controller` will manage this polling, using a new
+`BareMetalHostDiscovery` API to control it.
+
+The `BareMetalHostDiscovery` API will include a `resourceNameTemplate`
+field to hold instructions for building a resource name for a new
+node, including a prefix, suffix, and enum for selecting known
+hardware details.
+
+For a host with name `"the-host-name"`, this specification
+
+```yaml
+resourceNameTemplate:
+  prefix: "string-literal1-"
+  suffix: "-string-literal2"
+  hardwareDetails: hostname
+```
+
+would produce a resource name
+`"string-literal1-the-host-name-string-literal2"`.
+
+The `hardwareDetails` field is required. The `prefix` and `suffix`
+fields are ignored if not provided.
+
+The `hardwareDetails` enum can be one of:
+
+- `hostname` -- Using the discovered hostname for the host.
+- `ip` -- The IP address of the first NIC for the host with `.`
+  replaced by `-`.
+- `serial-number` -- Using the hardware serial number for the host.
+- `boot-mac` -- Using the boot MAC address of the host with `:`
+  replaced by `-`.
+- `provisioning-id` -- The identifier assigned by the underlying
+  provisioning tool when the host was registered automatically.
+
+Changing the a `BareMetalHostDiscovery` resource will have no effect
+on previously discovered host resources.
+
+For each host known to Ironic, the `host-discovery-controller` will
+look for a `BareMetalHost` resource with the same `bootMACAddress`. If
+any host resource is found matching the MAC, no further action will be
+taken.
+
+For any host known to Ironic but not represented by a `BareMetalHost`
+resource, the `host-discovery-controller` will create a
+`BareMetalHost` resource with the `bootMACAddress` filled in based on
+the MAC of the first discovered NIC and using a name built from the
+template values.
+
+The Ironic provisioner in the `baremetal-operator` will be updated to
+search for nodes in Ironic matching a host in kubernetes based on the
+`bootMACAddress` value given in the host. This ensures that the
+`baremetal-operator` can match the host resource to the data already
+in Ironic and fill in status values like the hardware details. This
+allows the existing logic in the `baremetal-operator` to manage the
+hardware details and avoids having two controllers splitting that
+responsibility.
+
+The [state
+machine](https://github.com/metal3-io/baremetal-operator/blob/master/docs/baremetalhost-states.md)
+for the `baremetal-operator` will expand to support importing hardware
+details for a discovered host.  When a new host appears without
+credentials, it will move from the pseudo-state `Created` to a new
+transient state `Importing`, during which the details will be fetched
+from `ironic-inspector`.
+
+After the hardware details are found and the host status is updated so
+that `!NeedsHardwareInspection()` is true, the host will move from
+`Importing` to the `Discovered` state.
+
+When a user adds BMC credentials to a host in the `Discovered` state
+today it moves to `Registering` and the `Provisioner` is responsible
+for configuring the host in Ironic. That logic may need to be modified
+to support partial updates of existing nodes, to set the values that
+could not be discovered.
+
+Because discovered hosts will already have hardware details, there is
+no need to re-inspect them if they move out of the `Discovered`
+state. The state machine for the `baremetal-operator` will therefore
+need a new transition between `Discovered` and `MatchProfile` when the
+condition `!externallyProvisioned && !NeedsHardwareInspection()` is
+true. That transition will bypass the `Inspecting` state and allow the
+host to eventually move to the `Ready` state.
+
+Today there is a race between a server booting and a host resource for
+it being created. If the server boots before the host resource exists
+and the node is registered in Ironic, the host resource is effectively
+broken and the user must clean up both the resource and Ironic
+manually. Supporting discovery, and especially matching host resources
+to Ironic nodes using MAC address, means we can fix the race. The
+`IronicProvisioner` needs to be updated to recognize when a node with
+the same MAC already exists. If the node in Ironic has no name,
+indicating that it was discovered, the provisioner should set the name
+and update any other values not set through discovery. If the node
+does have a name and the name does not match the name of the
+`BareMetalHost` then the host resource being reconciled should enter
+the `Error` state and an error message should be attached explaining
+that it is a duplicate host.
+
+### Implementation Details/Notes/Constraints
+
+The state management changes described above are likely to require
+changes to the `Provisioner` API. The details of those changes will be
+determined during implementation.
+
+The changes to the state machine would not change the fact that many
+of the metal3 features (fencing and power management) require BMC
+credentials. Discovered hosts would not enter the states where those
+features are allowed until BMC credentials are provided. The host
+controller should skip trying to maintain the power state and should
+not honor hard reboot requests for hosts without BMC credentials.
+
+### Risks and Mitigations
+
+If multiple `BareMetalHostDiscovery` resources exist, there may be
+multiple attempts to create a host from discovered information by
+different goroutines in the `host-discovery-controller`. If the name
+template information is the same, the conflict will be caught. If the
+name template information is different, then multiple hosts will be
+created. Resolving this will require adding a webhook in the
+`baremetal-operator` to be invoked when a resource is created and to
+block duplicate hosts. See [the v1alpha2
+migration](https://github.com/metal3-io/metal3-docs/pull/101) plans
+for details related to allowing webhooks. In the future, we may add
+rules to the `BareMetalHostDiscovery` API to control the naming
+convention applied to hosts with different characteristics, but there
+are no concrete plans to do so now.
+
+Data on discovered hosts will be lost if automatic provisioning is
+enabled. The feature is disabled by default, so the risk is mitigated
+because users will be aware when it is turned on.
+
+Ironic inspector creates discovered nodes with an `auto_discovered`
+flag set in the introspection data. We could limit the polling in the
+`baremetal-operator` to only look for hosts with the flag set. It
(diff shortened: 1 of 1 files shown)


design/host-discovery.md
```diff
@@ -0,0 +1,290 @@
...
+6. Update `IronicProvisioner` so it will recognize a manually
+   registered host that matches a discovered node in the Ironic
+   database.
+7. Document how to fully enable and disable the feature.
+8. Extend the `DemoProvisioner` to support the new states.
+9. Add a method to `BareMetalHost` to report whether power control
+   operations are supported. For now, this will only look at whether
+   there are BMC credentials.
+10. Update the bare metal controller to skip trying to apply power
+   state management operations to hosts that do not support them.
```
maelk, line 235: Would there be a need to change some parameters for Ironic too ? For example, Ironic tries to set the boot mode to local after provisioning a node, in order to reboot from the local disk. So if a user wants to automatically provision a discovered node, the node must be configured to boot always from hard drive with a temporary override for one PXE boot to start IPA, and Ironic should not try to change the boot order. Is this something that should be of a concern ? At least the requirements on the node configuration should probably be documented

hardys: We may want to change https://github.com/metal3-io/ironic-inspector-image/blob/master/inspector.conf#L12 from the default of ipmi to the noop driver.

As you say that implies no automation of e.g boot order, so we'd have to document that the node should be configured to boot from disk, with a one-time PXE boot to enable deployment.

There could also be non-pxe use-cases, e.g it may be possible to create a bootable iso which an admin could boot manually, e.g for small Poc deployments.

dhellmann: Updating the driver is item 1 in the list. Docs are item 7. I extended the description of the doc updates needed to include more details.

dhellmann: It used to be that if you configured a host to boot from disk then PXE, if there was no image on the disk the host would skip that step and just PXE. I assume that's still the case, but maybe things changed?

maelk: Are we sure that there won't be an image on the disk ? the node could be booted with its old drive (not cleaned yet).

dhellmann: There are definitely going to be some limitations in what this approach will support. Since it's primarily for new hosts, it seems reasonable to assume no OS. If that's not the case, then the host would need to be the boot order set as @hardys suggested with the one-time PXE option.

ohadlevy:
Back in the days I defaulted to always boot over pxe, and in case the host
wasn't up to provisioning it would have a pxe configuration to boot from
disk... This way it's not required to change boot order at all but change
tftp / ipxe output

dhellmann:
> Back in the days I defaulted to always boot over pxe, and in case the host wasn't up to provisioning it would have a pxe configuration to boot from disk... This way it's not required to change boot order at all but change tftp / ipxe output

That would also be a good option. We could investigate adding support for managing the dnsmasq settings for individual hosts to Ironic or the BMO, if it proves necessary.

We're also trying to support non-PXE cases using manually attached ISOs, so if we rely entirely on configuring the PXE response we'll have trouble with those cases and have to provide instructions for the user anyway.

maelk:
There was an interesting discussion yesterday in the airship PTG related to this IPAM configuration. The idea was to look into whether Ironic could decouple from Neutron and Dnsmasq to provide this exact feature. so maybe that could be taken care of by Ironic instead of BMO in that case

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+
+*As a data center staff member, I want to replace a faulty node in a
+cluster quickly.*
+
+The idea here is to be able to take a “broken” host out of service and
+replace it with a standby host without having access to the kubernetes
+API or UI at all. If we assume the broken host is broken enough that
+it is completely offline, it should be possible to remove it from the
+rack and replace it with another physical host, which is then
+provisioned automatically.
```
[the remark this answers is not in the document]
dhellmann: No, not necessarily. They would have to create a Machine to trigger provisioning, but with the automatic MachineSet scaling piece in place that would happen as soon as the new host appeared. They would then just need to clean up the old host and its machine.

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+that it is a duplicate host.
+
+### Implementation Details/Notes/Constraints
+
+The state management changes described above are likely to require
+changes to the `Provisioner` API. The details of those changes will be
+determined during implementation.
+
+The changes to the state machine would not change the fact that many
+of the metal3 features (provisioning and power management) require BMC
```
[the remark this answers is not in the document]
dhellmann: I think I meant "fencing" not "provisioning" here. I'll fix that in an update. The work you link to is definitely a prerequisite for the goal of this work, though.

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+credentials. Discovered hosts would not enter the states where those
+features are allowed until BMC credentials are provided. The host
+controller should skip trying to maintain the power state and should
+not honor reboot requests for hosts without BMC credentials.
```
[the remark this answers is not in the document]
dhellmann: Yes, I should be clearer here that it is hard reboot that isn't supported.

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+enabled. The feature is disabled by default, so the risk is mitigated
+because users will be aware when it is turned on.
+
+Ironic inspector creates discovered nodes with an `auto_discovered`
+flag set in the introspection data. We could limit the polling in the
+`baremetal-operator` to only look for hosts with the flag set. It
+isn’t really clear how useful that would be, so the description above
+does not assume such a limit.
+
+There does not appear to be a way to remove inspection data from
```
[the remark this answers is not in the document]
dhellmann: Yes, here I had been considering the BMO looking at the ironic-inspector API rather than the ironic API, and I was trying to work out how we could tell the difference between a host discovered by ironic but deleted from the cluster and one that we actually want to bring into the cluster.

If we end up with a host in k8s that the user doesn't want, its data will be deleted from ironic when the host CR is removed. If the same host posts discovery info again, the same records will be re-created. I don't think there's a way to avoid that, other than not booting the discovery image on the host.

Maybe this bit is a digression that we don't need in this design doc.

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+serial numbers will also be available. Users can also place data in
+the `spec.description` field, if necessary.
+
+The [state
+machine](https://github.com/metal3-io/baremetal-operator/blob/master/docs/baremetalhost-states.md)
+for the `baremetal-operator` will expand so that hosts in the
+`Discovered` state that appear in the Ironic database can have their
+hardware details fetched from the `ironic-inspector` API, even if the
+host has no BMC credentials. When a new host appears without
+credentials, it will move from `Created` to a new state `Importing`
```
zaneb, line 138: Note that there is no state named "created". When the resource is initially created, the state is an empty string. Currently the host will remain in this (non-)state until creds are available to try, at which point it will move to "registering". This pre-registering state is designated as "Created" on the state machine diagram, but it isn't actually a state.
In the near future we should implement the "discovered" state so that the empty state immediately moves to either "discovered" or "registering" (depending on whether creds are available).

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+to `0` then polling for discovered hosts will be disabled.
+
+The discovery loop will create a `BareMetalHost` with its name set to
+the Ironic UUID for the discovered node. This will ensure a unique
+name, and will allow the `IronicProvisioner` in the
+`baremetal-operator` to match the host resource to the Ironic node
+automatically. Resource names are immutable, so this may make it
+difficult for a user to keep track of hosts, but other discovered
+identifying information such as the MAC address, IP address(es), and
+serial numbers will also be available. Users can also place data in
```
zaneb, line 129: A thing I've learned from TripleO is that people care really really deeply about controlling names. The hacks we had to do to make that possible were onerous, manual, and made everything more brittle than it needed to be. But while people might be mildly irritated if everything occasionally falls over in a heap, they will go somewhere else in a heartbeat if you make them accept automatically generated names.

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+data available for any other host, but does not include the BMC
+credentials. We could use the data to partially create a
+`BareMetalHost` resource for the host by polling the Ironic API to
+compare the list of nodes it knows about with the list of host
+resources. This polling is independent of the normal controller
+reconciliation loop, and should run in a single separate goroutine in
+the `baremetal-operator` process, started after the process wins the
+leader election. A command line flag will control the frequency of
+polling with a default of every `60` seconds. If the frequency is set
+to `0` then polling for discovered hosts will be disabled.
```
zaneb, line 120: Is there any advantage to creating a CRD to control the discovery (its reconcile loop could double as the thing that actually polls Ironic)?
One that I can think of is that a CRD could provide users some way of templating the resource name. At the time of creating the resource we can theoretically get all of the HardwareDetails from ironic-inspector, right? It would also give you an interface for turning discovery on and off, maybe limiting it to particular networks, &c.

design/host-discovery.md
```diff
@@ -0,0 +1,292 @@
...
+
+The [state
+machine](https://github.com/metal3-io/baremetal-operator/blob/master/docs/baremetalhost-states.md)
+for the `baremetal-operator` will expand so that hosts in the
+`Discovered` state that appear in the Ironic database can have their
+hardware details fetched from the `ironic-inspector` API, even if the
+host has no BMC credentials. When a new host appears without
+credentials, it will move from `Created` to a new state `Importing`
+during which the details will be fetched from `ironic-inspector`. The
+host will continue to reconcile in the `Importing` state until details
```
zaneb, line 140: Aren't the details available immediately? The thing that creates the host in ironic is IPA reporting the inspection results, isn't it? It seems like we could just handle this in the Discovered state without needing to create a separate state.
As written, this would appear to create problems when a user manually creates a Host without adding the credentials initially, which is a thing we will likely also want to support.

design/host-discovery.md
dhellmann: Do you have any suggestions for better ways to pick identifying information?

dhellmann: Nevermind, the templating idea you propose is pretty interesting, let me explore that.

design/host-discovery.md
dhellmann: I clarified that the created state is a 'pseudo-state'. I also moved the new state before discovered. I want a new state here because it's transient, but the details might not be present if the user creates the host without credentials and then powers on the host.

I'm going to go ahead and resolve this conversation because as part of the update I've rebased and moved this file into a new directory so the thread is going to be detached from the source. Let's start a new thread on the new file.

design/host-discovery.md
dhellmann: The latest draft explores using a CRD and separate controller.

design/host-discovery.md
dhellmann: I've tried to address this in the new draft. If a host is created without credentials (by something other than the new controller, say the user or an installer tool) and then the host is powered on, we will have to wait for the details to show up in the ironic database.

dhellmann:
Based on similar discussions about keeping the annotation up to date for pivoting, I updated this design to simplify the workflow a bit. Rather than having the baremetal-operator use the MAC to find an ironic node and update the host, we can have the new discovery controller create the host with the status annotation set so the baremetal-operator only needs to copy the data from one field to the other in the resource, as it already does. This will result in fewer changes in the baremetal-operator, and we won't need a new state in the state machine.

kirankt:
> Based on similar discussions about keeping the annotation up to date for pivoting, I updated this design to simplify the workflow a bit. Rather than having the baremetal-operator use the MAC to find an ironic node and update the host, we can have the new discovery controller create the host with the status annotation set so the baremetal-operator only needs to copy the data from one field to the other in the resource, as it already does. This will result in fewer changes in the baremetal-operator, and we won't need a new state in the state machine.

I think this doc is very well written and many things are clear. Few questions: 
- Is BMO still responsible for deploying *all* ironic services, including `ironic-inspector`? 
- Which entity will be responsible to launch the host discovery controller? SLO?

maelk:
Hi @kirankt ! We have reworked a bit the deployment process. BMO is now separated from ironic, we have kustomizations to deploy ironic and all related containers. BMO is deployed standalone or as part of CAPM3, with clusterctl. I would expect new controllers related to BMO to also be deployed as part of CAPM3 when using clusterctl, or if not using CAPI, be deployed as part of BMO, included in the BMO kustomization. What do you think about this @dhellmann ?

dhellmann:
> @kirankt wrote:
> * Is BMO still responsible for deploying _all_ ironic services, including `ironic-inspector`?

The baremetal-operator doesn't actually manage the deployment at all. Upstream, kustomize is used to generate the deployment instructions. As @maelk mentioned adding a new controller will mean adding that controller to the deployment. Downstream in OpenShift, when we adopt the feature, we will also need to update the operator we have that generates our deployment settings.

> * Which entity will be responsible to launch the host discovery controller? SLO?

Downstream in OpenShift, yes, I don't see us adopting this before the new operator for managing metal3 is done.

> @maelk wrote:
> I would expect new controllers related to BMO to also be deployed as part of CAPM3 when using clusterctl, or if not using CAPI, be deployed as part of BMO, included in the BMO kustomization. What do you think about this @dhellmann ?

I intend for this to be a completely new stand-alone controller, to let us use kubebuilder instead of operator-sdk. That may complicate deployment a bit, but the feature is optional so maybe that's not a problem. 

I see some stuff in the CAPM3 repo under `config` that looks like it brings in the BMO when CAPM3 is deployed, so it would make sense to do something similar with the new controller. Does that seem reasonable?

maelk:
> I intend for this to be a completely new stand-alone controller, to let us use kubebuilder instead of operator-sdk. That may complicate deployment a bit, but the feature is optional so maybe that's not a problem.
> 
> I see some stuff in the CAPM3 repo under `config` that looks like it brings in the BMO when CAPM3 is deployed, so it would make sense to do something similar with the new controller. Does that seem reasonable?

For CAPM3 definitely, we'll pull it in as part of CAPM3 deployment. Then it should have its own deployment to be independent of BMO, and be included in CAPM3, the same way we do it for ip-address-manager (with releases) or BMO (without releases).

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+taken.
+
+For any host known to Ironic but not represented by a `BareMetalHost`
+resource, the `host-discovery-controller` will create a
+`BareMetalHost` resource with the `bootMACAddress` filled in based on
+the MAC of the first discovered NIC and using a name built from the
+template values and with an annotation containing the hardware details
+and provisioner ID that should be added to the status fields.
+
+When a user adds BMC credentials to a host in the `Discovered` state
```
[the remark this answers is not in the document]
dhellmann: Yeah, the original draft of this predates the unmanaged work, I think.

dhellmann:
This definitely needs a lot of rework, given other work that has happened since it was originally written.

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+today it moves to `Registering` and the `Provisioner` is responsible
+for configuring the host in Ironic. That logic may need to be modified
+to support partial updates of existing nodes, to set the values that
+were not not set when the node was discovered.
```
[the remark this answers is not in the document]
dhellmann: We talked today about possibly adding an `UnmanagedProvisioned` state for hosts that we have provisioned using fast-track but that we can't do power control of.

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+`BareMetalHost` resource.
+
+## Motivation
+
+There are several workflows for adding nodes to existing clusters that
+would benefit from not forcing the user to register the hosts in
+advance. Especially in large data centers, it may be desirable to
+build a small cluster and then attach additional hosts to the
+provisioning network and deploy them by simply powering them on and
+using PXE to boot an agent that knows how to reach back to the
```
dtantsur, line 28: You don't *have* to use PXE, it could be an ISO image with IPA just as well.

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+provisioned automatically.
+
+## Design Details
+
+When Ironic is configured to support its existing discovery feature,
+unknown hosts that boot the Ironic agent will have their hardware
+details registered with `ironic-inspector` and have a node created in
+the Ironic database. Typically this happens via the PXE boot
+configuration, but in the future it could happen in other ways such as
+having a bootable ISO image with the agent on it. How the agent ends
```
dtantsur, line 107: It can right now, we just need to prepare an ISO with the right parameters (ironic URL, inspector URL, inspection parameters).

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+fields are ignored if not provided.
+
+The `hardwareDetails` enum can be one of:
+
+- `hostname` -- Using the discovered hostname for the host.
+- `ip` -- The IP address of the first NIC for the host with `.`
+  replaced by `-`.
+- `serial-number` -- Using the hardware serial number for the host.
+- `boot-mac` -- Using the boot MAC address of the host with `:`
+  replaced by `-`.
```
dtantsur, line 145: Just as a note: the boot MAC can only be detected during PXE (not ISO boot).

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+
+For any host known to Ironic but not represented by a `BareMetalHost`
+resource, the `host-discovery-controller` will create a
+`BareMetalHost` resource with the `bootMACAddress` filled in based on
+the MAC of the first discovered NIC and using a name built from the
+template values and with an annotation containing the hardware details
+and provisioner ID that should be added to the status fields.
+
+When a user adds BMC credentials to a host in the `Discovered` state
+today it moves to `Registering` and the `Provisioner` is responsible
```
dtantsur, line 165: Ironic can now be configured to provision a node without having credentials, see my demo https://youtu.be/A9Q8qrEhyh0. I don't suggest we do it by default, but it can be a handy option.

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+the Ironic agent image. This could cause confusion, if the users
+intent was to remove the host completely. It could also be seen as a
+feature, if it is considered a way to reprovision a host.
+
+### Work Items
+
+1. Create the new git repository for the controller and API and set up
+   CI jobs.
+2. Write the new controller
+3. Update ironic-inspector-image to add unknown hosts using the `fake`
```
dtantsur, line 230: Should be `manual-management` and we'll need to make a few updates to ironic-image to enable this mode (add `agent` power interface, add `noop` management interface, enable fast track and netboot fallback - see my demo linked above).

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+Instead of polling the Ironic API looking for new hosts, we could
+watch the `dnsmasq` logs. However, in some configurations PXE may be
+disabled or hosts may be booted in other ways even with PXE enabled,
+so we would not see all discovered hosts.
```
dtantsur, line 287: Strictly speaking DHCP does not imply PXE, but somebody could be using their own DHCP + ISO boot.

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+discovered. This would be less compute intensive, but offers an
+opportunity to "miss" the notification of a new host, and does not
+provide an opportunity to configure the names for the new host
+resources.
+
+Instead of polling the Ironic API, we could use a notification plugin
+to invoke a callback when a new host is discovered. This would be less
+compute intensive, but offers an opportunity to "miss" the
+notification of a new host, and does not provide an opportunity to
+configure the names for the new host resources.
```
dtantsur, line 300: Not sure I get the part about configuring names, but we could probably do both polling (infrequent) and notifications.

design/host-discovery-controller/host-discovery.md
```diff
@@ -0,0 +1,325 @@
...
+- [early draft of this
+  proposal](https://docs.google.com/document/d/1hlMSdim53CQtqlYOnx1ptq3qKHKhP9LhDrRxzCutGoM/edit)
+- [baremetal-operator issue
+  41](https://github.com/metal3-io/baremetal-operator/issues/41)
+- [Ironic node discovery
+  spec](https://specs.openstack.org/openstack/ironic-inspector-specs/specs/ironic-node-auto-discovery.html)
+- [Ironic node discovery
+  docs](https://specs.openstack.org/openstack/ironic-inspector-specs/specs/ironic-node-auto-discovery.html)
+- [WIP changes for
+  baremetal-operator](https://github.com/metal3-io/baremetal-operator/pull/545)
```
dtantsur, line 325: Docs on credential-less deployment: https://docs.openstack.org/ironic/latest/admin/agent-power.html

dtantsur (commented):
I'd be curious to see this revived.

PR #101, opened 2020-06-16
WIP: BareMetalHost v1alpha2 API Migration

dhellmann:
Addresses https://github.com/metal3-io/baremetal-operator/issues/434

### DIFF
--- a/design/baremetalhost-v1alpha2.md
+++ b/design/baremetalhost-v1alpha2.md
@@ -1,0 +1,317 @@
+# BareMetalHost v1alpha2 API Changes
+
+## Status
+
+provisional
+
+## Summary
+
+We need to continue the API version migration for the `BareMetalHost`
+resource type to complete the work described in
+<https://github.com/metal3-io/baremetal-operator/issues/434>. The fix
+would be a relatively small API change, but the effect of that
+backwards-incompatible change is larger than anticipated. Therefore, I
+propose that we wait and include the change along with other, more
+significant, reworking of the API so that we don't take on the extra
+work multiple times.
+
+## Motivation
+
+The `baremetal-operator` repository still relies on a vendored copy of
+the `operator-sdk` project because the `BareMetalHost` CRD includes a
+field with type `float`, which is not formally supported by either
+`operator-sdk` or `kubebuilder`. Changing the field type will require
+a backwards-incompatible API change, so we need to change our API
+version and support upgrading existing users.
+
+The mechanism for handling API upgrades in kubernetes is a "conversion
+webhook". The kubebuilder project supports generating the scaffolding
+for the REST API for the conversion for `go` types that support the
+`Conversion` interface. The `baremetal-operator` uses the
+`operator-sdk` project, which does not yet have this
+feature. Therefore, part of the process of upgrading the API will also
+require us to convert to kubebuilder.
+
+We have also discussed [future API
+evolution](https://groups.google.com/d/msg/metal3-dev/QfMJpyG0Zss/5gdmRbm6CAAJ)
+that would be more significant than changing the type of one field. Those
+changes will also be backwards-incompatible and require effort to
+upgrade. By combining the incremental change and the more significant
+change we can avoid doing the upgrade work multiple times.
+
+### Goals
+
+- Implement an API version update to `v1alpha2` for the
+  `BareMetalHost` CRD.
+- Clean up a few issues with the `BareMetalHost` API for which the
+  fixes are not backwards-compatible.
+- Anticipate other more significant changes to the API.
+- Support in-place upgrades of existing `BareMetalHost` resources.
+
+### Non-Goals
+
+- Describe a stable `v1` API.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a cluster administrator, I want to update the version of metal3 in
+my existing cluster without losing any data or features.
+
+## Design Details
+
+### Choosing a New Type for the Clock Speed Field
+
+Our problematic field holds the clock speed of a CPU in fractional
+megahertz. We assume the fractional portion is noise, so we could
+truncate the value without losing any information.
+
+We have three options for the new type. The advice from the
+`controller-tools` maintainers is to use `resource.Quantity` or
+`int32` for floating point values. We receive the value from
+gophercloud as a string, so we could also simply pass the string
+through unmodified.
+
+The `Quantity` type adds extra characters representing the unit
+conversion, so that a `1 GHz` value would be expressed as `1Gi`, which
+would be a bit ugly for someone trying to consume the value without
+access to the `go` library for parsing the string.
+
+The CPU speed is simply being reported, and not used by the
+`baremetal-operator` code. We could, therefore, use a `string` field
+without losing any functionality. The value we get from `gophercloud`
+is already a string.
+
+Since the fractional portion of the speed can be dropped, we can use
+an `int32` to hold the value. An integer value will make it easier for
+the `hardware-classification-controller` to continue to use the value
+when applying rules.
+
+The best approach is to use an `int32` field.
+
+### Converting Existing Data
+
+It is possible to [use a web
+hook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#configure-customresourcedefinition-to-use-conversion-webhooks)
+to convert resources from their storage format to a specific API
+format. We do not want users to continue to consume the `v1alpha1`
+format, however, because we want to be able to drop support
+quickly. Therefore, we should not write a conversion web hook.
+
+Instead, we should add startup code to the baremetal-operator to read
+all host resources in their current format and write them (with type
+conversion but no other changes) to the new storage format. See option
+2
+[here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version)
+for more discussion.
+
+### Dropping Hardware Profile Field
+
+Now that we have support for more verbose root device hints and we
+have the `hardware-classification-controller`, we no longer need the
+hardware profile fields. We can remove them as part of the
+backwards-incompatible changes as we move to the new API version.
+
+### Converting to Kubebuilder
+
+The `baremetal-operator` was built with the `operator-sdk`, which is
+actively developed and maintained but has fewer features than the
+alternative, `kubebuilder`, including especially the ability to
+generate the scaffolding for the conversion webhooks for API changes.
+
+`kubebuilder` is also built on the same code generator project as
+`operator-sdk`, and enforces the same limitation on the use of `float`
+type for fields. In order to support the same `v1alpha1` API, we will
+need to apply the same work-around in a forked copy of the tool.
+
+It is possible, but not trivial, to convert between the two tools. The
+expectations are different for code layout, CRD naming, build tools,
+and other aspects. Moving the `baremetal-operator` to `kubebuilder`
+will introduce inherent compatibility issues in the results.
+
+The `BareMetalHost` CRD as defined today has the group `"metal3.io"`,
+so that its full name is `baremetalhost.metal3.io`. Kubebuilder
+expects the domain and group portion of a CRD to be different, so it
+wants an additional value as a prefix to `"metal3.io"` (for example,
+`"hardware.metal3.io"`). We could work around this by modifying the
+output generated by the tool, but working against the intent of the
+tools is how we ended up in a situation where we need to fix the API,
+so we should avoid doing that again.
+
+Given that we're going to end up with the resource renamed anyway, it
+makes sense to consider the more significant API changes proposed to
+the API on the mailing list. The specific changes to make will be
+addressed by a separate proposal.
+
+### Risks and Mitigations
+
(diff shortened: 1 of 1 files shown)


design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,223 @@
...
+## Design Details
+
+### Choosing a New Type for the Clock Speed Field
+
+The advice from the `controller-tools` maintainers is to use
+`resource.Quantity` or `int32` for floating point values.
+
+Our problematic field holds the clock speed of a CPU in fractional
+megahertz, so in order to use an integer we would need to represent
+the speed in terms of kilohertz.
```
zaneb, line 53: I think if we represent it as an integral number of MHz, that's fine.
Originally it was an integral number of GHz, which was pretty low-resolution. I first changed it to a float, but then later changed it to MHz. In retrospect I should have changed it back to an int at that point.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,223 @@
...
+
+The `Quantity` type adds extra characters representing the unit
+conversion, so that a `1 GHz` value would be expressed as `1Gi`, which
+would be a bit ugly for someone trying to consume the value without
+access to the `go` library for parsing the string.
```
zaneb, line 58: Speed isn't really a quantity, and this sounds a bit icky. +1 for avoiding.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,223 @@
...
+### Risks and Mitigations
+
+We do not know all of the ways that consumers of the API are using the
+CPU speed values currently being reported. Given that most kubernetes
+tools appear to assume that floating point types are not to be used in
+APIs, it is likely that most consumers are either ignoring the value
+or using it in simple ways (displaying it, logging it, etc.) where
+loss of precision from the type conversion to a float is not
+significant.
+
```
zaneb, line 101: One possible source of risk is that the CAPM3/CAPBM needs to be updated in lockstep because we're dropping support for the old version immediately.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,223 @@
...
+### Change Type in Place, Retain v1alpha1
+
+The API version for `BareMetalHost` is `v1alpha1`. Upstream in metal3,
+we could reasonably say that the way to fix the problem with the CPU
+speed field is to change the type of the field and that no backwards
+compatibility or upgrade support is provided. Fixing the type in place
+allows the `operator-sdk` tools to work properly again.
+
+Any existing users would be broken by the type change, because
+existing data could no longer be deserialized from storage properly.
```
zaneb, line 154: Have we verified that this is actually the case? All numbers in JSON are stored as floats anyway.

design/baremetalhost-v1alpha2.md
dhellmann: That's a good point. There are "release" steps in the work items list. I should include "since to providers" as part of that.

design/baremetalhost-v1alpha2.md
dhellmann: If we change from float to string, I think they're going to be broken. If we change to int, maybe not all of them will? I would expect any golang consumers to break either way. Other languages may be more flexible.

design/baremetalhost-v1alpha2.md
dhellmann: My host reports fractional MHz even for GHz. Perhaps those are representation errors (the numbers aren't identical, so that's fun).

```
$ grep 'cpu MHz' /proc/cpuinfo
cpu MHz		: 2585.951
cpu MHz		: 2399.982
cpu MHz		: 2469.934
cpu MHz		: 2400.288
cpu MHz		: 2468.399
cpu MHz		: 2417.853
cpu MHz		: 2496.642
cpu MHz		: 2412.698
cpu MHz		: 2449.773
cpu MHz		: 2399.325
cpu MHz		: 2542.550
cpu MHz		: 2399.882
cpu MHz		: 2578.015
cpu MHz		: 2399.956
cpu MHz		: 2606.565
cpu MHz		: 2404.802
cpu MHz		: 2552.326
cpu MHz		: 2412.515
cpu MHz		: 2556.837
cpu MHz		: 2400.665
cpu MHz		: 2598.182
cpu MHz		: 2400.156
cpu MHz		: 2630.709
cpu MHz		: 2412.368
cpu MHz		: 2535.367
cpu MHz		: 2400.012
cpu MHz		: 2446.512
cpu MHz		: 2402.527
cpu MHz		: 2401.146
cpu MHz		: 2412.242
cpu MHz		: 2542.022
cpu MHz		: 2412.191
cpu MHz		: 2415.755
cpu MHz		: 2399.906
cpu MHz		: 2566.748
cpu MHz		: 2399.979
cpu MHz		: 2583.524
cpu MHz		: 2401.964
cpu MHz		: 2543.863
cpu MHz		: 2409.357
```

dhellmann: I suppose we could just truncate the values and turn `2409.357` to `2409`? That's still in MHz though.

zaneb: The number that IPA reports is actually the `CPU max MHz` from `lscpu`. It does report it to 4 decimal places, but I would hazard a guess that this is ~always an integer on real systems.

Annoyingly, thanks to turbo boost even this is not necessarily the actual nominal clock speed:

```
Model name:          Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
Stepping:            1
CPU MHz:             1265.260
CPU max MHz:         3100.0000
CPU min MHz:         1200.0000
```
```
Model name:                      Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
Stepping:                        10
CPU MHz:                         800.032
CPU max MHz:                     4200.0000
CPU min MHz:                     400.0000
```

The output in `/proc/cpuinfo` is just measuring noise in the dynamic frequency scaling to 3 decimal places.

dhellmann: OK, so it looks like we could use an int32 with units of MHz and just truncate the float we get, turning `4200.0000` into `4200`. We still receive a string from gophercloud, so we would need to do the conversion but we're already doing the conversion to float so that's not a huge change and it will save consumers from having to do their own type conversion.

dhellmann:
Updated to propose using an integer and to include updating the go modules after releases.

dhellmann:
Updated to propose some version numbers in the release steps.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,236 @@
...
+- Remove hardware profile from spec and status.
+- Remove the state machine states for setting the hardware profile.
+- Set `v1alpha2` as the storage version
+- Add conversion code to the `baremetal-operator`
+- Update the inspection code to no longer convert the value given by
+  Ironic Inspector before storing it.
+- Release `baremetal-operator` version `0.2.0` with code that can
+  upgrade existing resources.
+- Update the module references in `cluster-api-provider-metal3` and
+  `hardware-classification-controller`.
```
maelk, line 124: Those will require a release. Considering that BMO is now deployed as part of CAPM3 for CAPI, we must release a new version of CAPM3 (that might only include the new BMO). We will need to make sure that the consumers do not skip that version (if the next release without v1alpha1 is already out) as it would break the deployment.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,236 @@
...
+- Drop support for `v1alpha1`
+  - Remove the API code
+  - Remove the conversion code
+  - Remove the `operator-sdk` code from the `tools` directory
+- Release `baremetal-operator` version `0.3.0`.
+- Update the module references in `cluster-api-provider-metal3` and
+  `hardware-classification-controller`.
```
maelk, line 131: This will also require a new release of those.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Dependencies
+
+The prerequisite work was done as part of
+[#434](https://github.com/metal3-io/baremetal-operator/issues/434).
+
+### Test Plan
+
+- Test updating the `baremetal-operator` on an existing cluster with
```
maelk, line 140: we should probably include an e2e test that drives an upgrade with all pieces, including CAPI / CAPM3 and HWCC. What I am after is to test the integration around the new feature during the upgrade, to make sure that CAPM3 will be able to use the BMH v1a2 after upgrade from v1a1, not only with new v1a2 BMHs.
This might be more complex to put in place, as we'll need in that case to do the setup through clusterctl and then proceed with the upgrade.

design/baremetalhost-v1alpha2.md
dhellmann: Good points. I've added a comment about releasing the consumers. I'm not sure what to do about ensuring that users don't skip releases on an upgrade. Is a release note somewhere sufficient?

design/baremetalhost-v1alpha2.md
dhellmann: Given the nature of the changes planned, it seems unlikely that we'll have any issue. Putting an upgrade test in place does make sense, though. What sort of test scenarios should we include? Just upgrading and ensuring the new controllers don't put resources into an error state isn't enough. We probably need to reprovision a host so we can see that everything actually works. Maybe we can build that test before starting the rest of this work, based on existing CAPBM releases. What do you think?

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,236 @@
...
+`int32` for floating point values. We receive the value from
+gophercloud as a string, so we could also simply pass the string
+through unmodified.
+
+The `Quantity` type adds extra characters representing the unit
+conversion, so that a `1 GHz` value would be expressed as `1Gi`, which
+would be a bit ugly for someone trying to consume the value without
+access to the `go` library for parsing the string.
+
+The CPU speed is simply being reported, and not used by the
```
mhrivnak, line 65: There has been interest in the past in using automation to parse specs such as this, compare each to some standard or minimum, and apply labels to hosts based on those comparisons. To support that goal, it will help to keep parsability and comparability in mind in choosing the type. A string would probably not be ideal for that purpose.

design/baremetalhost-v1alpha2.md
maelk: IMO, the tests, if added, should probably interact with BMO through consumers, for example updating a machine deployment from CAPI or modifying a HWCC. There's some similar work on-going in metal3-dev-env for upgrade verifications. THis could be synced, because the delta does not seem that big.

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,236 @@
...
+- Add conversion code to the `baremetal-operator`
+- Update the inspection code to no longer convert the value given by
+  Ironic Inspector before storing it.
+- Release `baremetal-operator` version `0.2.0` with code that can
+  upgrade existing resources.
+- Update the module references in `cluster-api-provider-metal3` and
+  `hardware-classification-controller` and release them.
+- Drop support for `v1alpha1`
+  - Remove the API code
+  - Remove the conversion code
```
mhrivnak, line 127: Does this assume that anyone on 0.1.0 will upgrade to 0.2.0 on their way to 0.3.0? What would happen if someone upgrades from 0.1.0 -> 0.3.0 directly? If that needs to be prevented, how would that be done?

design/baremetalhost-v1alpha2.md
```diff
@@ -0,0 +1,236 @@
...
+`baremetal-operator` code. We could, therefore, use a `string` field
+without losing any functionality. The value we get from `gophercloud`
+is already a string.
+
+Since the fractional portion of the speed can be dropped, we can use
+an `int32` to hold the value. An integer value will make it easier for
+the `hardware-classification-controller` to continue to use the value
+when applying rules.
+
+The best approach is to use an `int32` field.
```
kirankt, line 75: +1 for int32

dhellmann:
I reorganized the work list a bit to reflect the phases better and to move some steps around because we're going to have to support storing the data using the old format for at least 1 more release so we can update the clients in CAPM3 and HCC to use the new format.

design/baremetalhost-v1alpha2.md
maelk: From CAPI point of view, I don't think there is any way to force this. We can add it to the readme and release notes, but using clusterctl, it would be very easy to overlook this, as the command by default pulls the latest version

mhrivnak: I also think it'll be difficult to guarantee that people upgrade to 0.2.0.

Once the controller itself is using v1alpha2, then all we need the v1alpha1 structs for is the conversion webhook, right? Maybe we can "freeze" that code and have operator-sdk stop interacting with it during `operator-sdk generate` or similar. Moving the v1alpha1 code to a different package might be sufficient for it to be ignored.

dhellmann: The tool only includes information in the generated CRD for versions that it reads.

dhellmann:
After more investigation into the implementation details, I have updated the plan to describe converting to kubebuilder and putting off all of the work until the future when we have decided what to do about the more significant API changes we have discussed on the mailing list.

PR #110, opened 2020-06-30
Update pivoting documentation

kashifest:
Update the pivoting documentation namely:

-  Update the list of objects to be pivoted
-  Add important information regarding configurations.

### DIFF
--- a/docs/move.md
+++ b/docs/move.md
@@ -13,7 +13,8 @@ The following objects are moved during this process:
 KubeadmControlplane, Metal3Cluster, Metal3Machine, and BareMetalHost
 
 **Related Objects:** Cluster Secrets, BareMetalHost Secrets, Configmaps,
-KubeadmConfig, Metal3MachineTemplate
+KubeadmConfig, Metal3MachineTemplate, Metal3IPPool, Metal3IPClaim, IPAddress,
+Metal3data, Metal3DataClaim, Metal3DataTemplate
 
 ## Prerequisite
 
@@ -32,7 +33,9 @@ KubeadmConfig, Metal3MachineTemplate
 
   So if the clusters are not initialized using `clusterctl`, all the CRDS of the
   objects to be moved to target cluster needs to have these labels both in
-  bootstrap cluster and target cluster.
+  bootstrap cluster and target cluster. **Note**: This is not recommended, since
+  the way `clusterctl` identifies objects to manage might change in future, so
+  it's always safe to install CRDs and controllers through `clusterctl` init.
 
 * **BMH objects have correct status annotation.**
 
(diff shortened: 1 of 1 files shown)


docs/move.md
```diff
@@ -130,11 +133,35 @@ This can now be achieved with the following procedure:
...
+
+The following requirements are essential for the move process to run
+successfully:
+
+1.
+The move process should be done when the BMHs are in a steady state. BMHs
+should not be moved while any operation is on-going i.e. BMH is in provisioning
+state. This will result in failure since the IP of the BMH might change after
+the move and the DHCP-leases from the management cluster are not moved to target
+cluster.
```
maelk, line 146: the bigger risk is that the interaction between IPA and ironic get broken since ironic's database might not be repopulated. What you said is valid and should be there, but we should also add that doing this would compromise the state of your cluster.

docs/move.md
```diff
@@ -130,11 +133,35 @@ This can now be achieved with the following procedure:
...
+
+2.
+Before the move process is initialized, it is important to delete the ironic
+pod/ironic containers. If ironic is deployed in cluster the deployment is named
+`metal3-ironic`, if it is deployed locally outside the cluster, the containers
+to be deleted are `ironic`, `ironic-inspector`, `ironic-endpoint-keepalived`,
+`ironic-dnsmasq`, `ironic-httpd` and `mariadb`. Deleting the ironic before move
+process makes sure that the old ironic does not interfere with the operations of
+the new ironic deployed in target cluster and also it makes sure that there is
+only one dnsmasq existent at any time.
```
maelk, line 156: I would invert those two sentences to emphasize on why you should not have two ironic at once. Also, we should emphasize that the database of the first ironic instance is not cleaned when we move the BMHs because we move them without reconciling

docs/move.md
```diff
@@ -130,11 +133,35 @@ This can now be achieved with the following procedure:
...
+
+3.
+The provisioning bridge where the `ironic-endpoint-IP` is supposed to be
+attached should have a static IP assignment on it before the ironic
+pod/containers start to operate in the target cluster. This is important since
+`ironic-endpoint-keepalived` container will only assign the `ironic-endpoint-IP`
+on the provisioning bridge in target cluster when it has an IP on it. Otherwise
+it will fail to attach the IP and ironic will be unreachable.
```
maelk, line 164: maybe specify that this interface is used to host the DHCP server so it cannot be configured to use DHCP ?

docs/move.md
```diff
@@ -32,7 +33,9 @@ KubeadmConfig, Metal3MachineTemplate
 
   So if the clusters are not initialized using `clusterctl`, all the CRDS of the
   objects to be moved to target cluster needs to have these labels both in
-  bootstrap cluster and target cluster.
+  bootstrap cluster and target cluster. **Note**: This is not recommended, since
+  the way `clusterctl` identifies objects to manage might change in future, so
+  its always safe to install CRDs and controllers through `clusterctl` init.
```
fmuyassarov, line 38:
```suggestion
  it's 
```

docs/move.md
```diff
@@ -130,11 +133,35 @@ This can now be achieved with the following procedure:
 
 9. Delete the bootstrap cluster
 
-**Important Note** The move process should be done when the BMHs are
-in a steady state. BMHs should not be moved while any operation is on-going i.e.
-BMH is in provisioning state. This will result in failure since the IP of the
-BMH might change after the move and the DHCP-leases from the management cluster
-are not moved to target cluster.
+## Impotant Notes
```
fmuyassarov, line 136:
```suggestion
## Important 
```

docs/move.md
```diff
@@ -130,11 +133,35 @@ This can now be achieved with the following procedure:
...
+
+The following requirements are essential for the move process to run
+successfully:
+
+1.
+The move process should be done when the BMHs are in a steady state. BMHs
```
fmuyassarov, line 142:
```suggestion
ready?
```

kashifest: it should be steady actually

docs/move.md
kashifest: Done. Thanks

docs/move.md
kashifest: Done. Thanks

docs/move.md
kashifest: Done. Thanks

docs/move.md
kashifest: Done. Thanks

docs/move.md
kashifest: Done. Thanks

kashifest:
Thanks @maelk & @fmuyassarov. I have addressed your comments.

maelk:
lgtm from @fmuyassarov too so
/approve

PR #120, opened 2020-07-01
describe unmanaged state for hosts

dhellmann:
Related to https://github.com/metal3-io/baremetal-operator/pull/569/

Co-Authored-By: Zane Bitter <zbitter@redhat.com>

### DIFF
--- a/design/baremetal-operator/unmanaged-state.md
+++ b/design/baremetal-operator/unmanaged-state.md
@@ -1,0 +1,141 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# unmanaged-state
+
+## Status
+
+implementable
+
+## Summary
+
+This design proposes adding a new state to the state machine to be
+applied to hosts that are known but cannot be managed because the BMC
+credentials are not.
+
+## Motivation
+
+As a precursor to being able to provision hosts without their BMC
+credentials, we want to be able to define hosts without BMC
+credentials so that the credentials can be added at a later
+time. Consumers of the hosts, including the
+`cluster-api-provider-metal3`, need to be able to recognize such hosts
+and treat them differently.
+
+### Goals
+
+1. Define a new `Unmanaged` state and the changes in each component to
+   support it.
+
+### Non-Goals
+
+1. Discuss the full host discovery design.
+
+## Proposal
+
+We had previously discussed a `Discovered` state but never implemented
+it, choosing to use the `OperationalStatus` value of `Discovered`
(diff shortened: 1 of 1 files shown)


dhellmann:
Based on @n1r1's reviews of https://github.com/metal3-io/baremetal-operator/pull/569 I thought I had better put together a document explaining the API change there.

/cc @n1r1 @zaneb @maelk @kirankt @honza @stbenjam

design/baremetal-operator/unmanaged-state.md
```diff
@@ -0,0 +1,139 @@
...
+applied to hosts that are known but cannot be managed because the BMC
+credentials are not.
+
+## Motivation
+
+As a precursor to being able to provision hosts without their BMC
+credentials, we want to be able to define hosts without BMC
+credentials so that the credentials can be added at a later
+time. Consumers of the hosts, including the
+`cluster-api-provider-metal3`, need to be able to recognize such hosts
```
kirankt, line 26: Should we also include `cluster-api-provider-baremetal` here now that the projects differ significantly.

dhellmann: I left that out because it's a downstream concern, but the changes would essentially be the same.

n1r1:
Looks good to me. Some minor comments:

1. Even if it's obvious for some users, I think we should mention explicitly that BMO won't be able to apply the desired state for unmanaged hosts. For example, any change to `Online` won't have any effect until BMC credentials are supplied. 
2. The change in consumers is not only where they look at the host state. It's also where they are not. For example, downstream in OCP, there's remediation code which assumes that any host has power management. So I think that any consumer who's trying to change `spec.Online` is potentially affected by this change

dhellmann:
> Looks good to me. Some minor comments:
> 
> 1. Even if it's obvious for some users, I think we should mention explicitly that BMO won't be able to apply the desired state for unmanaged hosts. For example, any change to `Online` won't have any effect until BMC credentials are supplied.

I think that's covered by https://github.com/metal3-io/baremetal-operator/pull/569/files#diff-0f426376b964e609c5c4618eb8bed3b2R372

> 2. The change in consumers is not only where they look at the host state. It's also where they are not. For example, downstream in OCP, there's remediation code which assumes that any host has power management. So I think that any consumer who's trying to change `spec.Online` is potentially affected by this change

I don't think the remediation code should change to behave differently when credentials are not present. That expects the caller to understand too much about how the baremetal operator works with hosts.

Even with credentials present, they might be wrong, and therefore the remediation would fail. The BMC might be frozen, and so remediation would fail. The BMC might be disconnected, and so remediation would fail. The calling code needs to deal with all of those cases. It should be able to ask for a reboot and then if that reboot does not happen for any reason it should be able to make a new decision about what to do.

PR #122, opened 2020-07-11
Update Metal3-dev-env images

fmuyassarov:
Improve existing dev-env related images.


fmuyassarov:
> s the



> nit: metal3-dev-env includes the switches, not only the vms

Sorry I didn't you. Do you want me to add switches on the picture ?

maelk:
in your new image, you added an overall square for metal3-dev-env, but it does not contain the switches, that are actually part of it

fmuyassarov:
> in your new image, you added an overall square for metal3-dev-env, but it does not contain the switches, that are actually part of it

Actually, four boxes on the left side of the overall square represent switches/bridges.

maelk:
yes, but those boxes are not in the metal3-dev-env box, they are out of it. that was my point

fmuyassarov:
> yes, but those boxes are not in the metal3-dev-env box, they are out of it. that was my point

Ah, ok. Will fix that.

maelk:
/remove-lifecycle stale
I think we can still get this in, the change remains valid now
/approve

PR #123, opened 2020-07-15
describe baremetal-operator migration to kubebuilder

dhellmann:
We have recently encountered several situations where we wanted to
take advantage of kubernetes API features that are not supported by
the operator-sdk.  Migrating from operator-sdk to kubebuilder will
unblock several efforts, and it will be enough work on its own that
documenting the change separately seems prudent. This proposal
describes the work we will need to do to complete that migration, with
references to a proof-of-concept implementation.

### DIFF
--- a/design/baremetal-operator/kubebuilder-migration.md
+++ b/design/baremetal-operator/kubebuilder-migration.md
@@ -1,0 +1,217 @@
+# kubebuilder-migration
+
+## Status
+
+implementable
+
+## Summary
+
+We have recently encountered several situations where we wanted to
+take advantage of kubernetes API features that are not supported by
+the old version of operator-sdk currently used with the
+baremetal-operator (especially web hooks for validation and converting
+between different API versions, and support for v1 of the
+CustomResourceDefinition API).  The newer version of operator-sdk (1.0
+and later) is more closely aligned with kubebuilder, so updating to
+the new version would be about the same amount of work as switching to
+kubebuilder. Updating the tool will unblock several efforts, and that
+will be enough work own that documenting the change separately seems
+prudent.
+
+Since the results are effectively the same, and the other projects in
+metal3 use kubebuilder, this proposal describes the work we will need
+to do to migrate from operator-sdk to kubebuilder, with references to
+a proof-of-concept implementation.
+
+## Motivation
+
+We chose to build the baremetal-operator using the operator-sdk
+because it was familiar to the original designers and known to be
+under active development. The version of operator-sdk we are using
+does not support all of the features we need, and in the time since
+that initial choice, the priorities of the operator-sdk team have
+decided to integrate the work with kubebuilder. The newer version of
+kubebuilder has features that will be useful to us (web hooks, API
+version migration, etc.).
+
+### Goals
+
+- Use a tool more widely used for complex controller implementation.
+- Update our tools to gain support for features such as web hooks and
+  scaffolding for multiple API versions for the same resource type.
+
+### Non-Goals
+
+- Make any API changes to the BareMetalHost type.
+
+## Proposal
+
+As discussed in [the original v1alpha2 API
+proposal](https://github.com/metal3-io/metal3-docs/pull/101/), we will
+need to carry a modified version of kubebuilder to support float field
+types.  We can carry the fork either within the baremetal-operator
+repository or in a separate repository under the metal3-io org on
+github, but it will involve fewer large patches to the
+baremetal-operator repository if we fork what we need in a separate
+repository and have the Makefile in the baremetal-operator repository
+build the tools from our fork.
+
+Kubebuilder initializes a project repository using a different layout
+than the old version of operator-sdk. It also uses a different
+approach to registering controllers, API types, etc. So, migrating
+will involve moving some existing code to new places and changing some
+of it. In particular:
+
+- kubebuilder places all packages at the top of the repository, not
+  under `pkg`
+- kubebuilder places all controllers in the same directory
+- main.go instantiates reconciler objects directly instead of calling
+  a factory, so some previously private parameters such as the
+  provisioner factory may need to be made public
+
+The [kubebuilder3
+branch](https://github.com/dhellmann/baremetal-operator/tree/kubebuilder3)
+of dhellmann's fork of the baremetal-operator repository is a
+proof-of-concept implementation of the migration. The host API is out
+of date, and the kubebuilder code probably is as well, but the same
+basic steps can be repeated when we are ready to do the work.
+
+As much effort as possible should be made to preserve the git history
+of files such as the API definition and the controller
+implementation. Unfortunately, these significant files will also have
+to be changed beyond their location within the repository, so
+preserving all of the history may not be possible.
+
+### User Stories
+
+#### Story 1
+
+As a contributor to the baremetal-operator, I want to use familiar
+tools and workflows so I can take advantage of my experience
+contributing elsewhere in the kubernetes community to work more
+efficiently.
+
+#### Story 2
+
+As a contributor to the baremetal-operator, I want to be able to make
+API changes to the BareMetalHost and support conversion web hooks to
+change between different storage versions.
+
+#### Story 3
+
+As a contributor to the baremetal-operator, I want to be able to use
+admission web hooks to validate BareMetalHost resources.
+
+## Design Details
+
+The process for creating the proof-of-concept migration was basically:
+
+1. Move everything in the git repository out of the way.
+2. Replace the copy of operator-sdk with a copy of kubebuilder, with a
+   similar patch to support float fields.
+3. Use kubebuilder init to create a new project in the repository,
+   setting the domain to an empty string.
+4. Use kubebuilder to create an API and controller for BareMetalHost,
+   then replace the implementations with the existing ones.
+5. Restore the rest of the existing code from the pkg directory,
+   moving it to the root of the repository.
+6. Fix everything so it compiles and the tests run.
+7. Add missing targets to the Makefile (lint, sec, openapi generation,
+   etc.).
+
+### Implementation Details/Notes/Constraints
+
+Using `git mv` as often as possible will help link the new files to
+their history in any old locations.
+
+Using an empty string for the domain argument to `kubebuilder init`
+means that we can specify a group of `"metal3.io"` when creating the
+host API, and avoid ending up renaming the resource type. Future APIs
+should include a more well-scoped group ("networking.metal3.io" or
+"hardware.metal3.io") but changing the group name for the existing
+resource will make it exceptionally difficult to upgrade.
+
+### Risks and Mitigations
+
+Although operator-sdk and kubebuilder both use the same underlying
+code generation library, and are trying to achieve parity, they are
+likely to be using different versions and options from time to
+time. The new implementation will therefore introduce updated
+dependencies, and the longer we put off the work the more likely those
+dependencies are to require code changes. We will mitigate some of
+that risk by forking the tool(s), which lets us lock in a specific
+version to use.
+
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,192 @@
...
+5. Restore the rest of the existing code from the pkg directory,
+   moving it to the root of the repository.
+6. Fix everything so it compiles and the tests run.
+7. Add missing targets to the Makefile (lint, sec, openapi generation,
+   etc.).
+
+### Implementation Details/Notes/Constraints
+
+Using `git mv` as often as possible will help link the new files to
+their history in any old locations.
```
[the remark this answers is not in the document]
dhellmann: That's true. The controller needed enough changes that I thought a git mv, commit, then edit would preserve more history.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,192 @@
...
+describes the work we will need to do to complete that migration, with
+references to a proof-of-concept implementation.
+
+## Motivation
+
+We chose to build the baremetal-operator using the operator-sdk
+because it was familiar to the original designers and known to be
+under active development. In the time since that initial choice, the
+priorities of the operator-sdk team have taken the project in a
+somewhat different direction than we need. Meanwhile, kubebuilder has
```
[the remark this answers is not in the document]
dhellmann: There is an issue from more than a year ago for [webhooks](https://github.com/operator-framework/operator-sdk/issues/1217) which would be needed for API changes. The design doc was being written about a month ago.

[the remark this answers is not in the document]
dhellmann: [This comment](https://github.com/operator-framework/operator-sdk/issues/1217#issuecomment-631263001) doesn't give me a lot of confidence, though:

> operator-sdk create webhook is now available, but only for Go projects created with operator-sdk init (or those that have migrated to that layout). Since init is not the default and we do not have migration documentation written yet, I will not close this yet.

I still consider migrating to be the best choice. The PoC took me about a day of work.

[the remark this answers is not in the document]
zaneb: I know that operator-sdk is supposed to be converging somewhat with kubebuilder... does that comment imply that `operator-sdk init` will create projects with the kubebuilder layout anyway (as opposed to `operator-sdk new`, which is what we have)?
If so that's a good reason to switch to the new layout regardless of whether we drop operator-sdk.

dhellmann:
I have updated this to reflect what we've learned about the plans to integration operator-sdk and kubebuilder.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,202 @@
...
+## Design Details
+
+The process for creating the proof-of-concept migration was basically:
+
+1. Move everything in the git repository out of the way.
+2. Replace the copy of operator-sdk with a copy of kubebuilder, with a
+   similar patch to support float fields.
+3. Use kubebuilder init to create a new project in the repository,
+   setting the domain to an empty string.
+4. Use kubebuilder to create an API and controller for BareMetalHost,
```
zaneb, line 109: There's a migration guide for legacy Operator SDK projects: https://sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide/

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,202 @@
...
+## Status
+
+implementable
+
+## Summary
+
+We have recently encountered several situations where we wanted to
+take advantage of kubernetes API features that are not supported by
+the old version of operator-sdk currently used with the
+baremetal-operator.  The newer version of operator-sdk is more closely
```
zaneb, line 12: Specifically, versions after 1.0 I think.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Status
+
+implementable
+
+## Summary
+
+We have recently encountered several situations where we wanted to
+take advantage of kubernetes API features that are not supported by
+the old version of operator-sdk currently used with the
```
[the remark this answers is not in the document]
dhellmann: Yes, that's the main thing. We need to update to get the conversion web hooks so we can deprecate that float field and move to a non-alpha version of the API.

design/baremetal-operator/kubebuilder-migration.md
dhellmann: I added a note to that effect.

design/baremetal-operator/kubebuilder-migration.md
dhellmann: Added to the references list at the end.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,202 @@
...
+
+Bug fixes will be more challenging to backport after the migration,
+because the code layout will be different. This risk is mitigated by
+the fact that we do not maintain old versions of the
+baremetal-operator today, and maintain backwards compatibility in all
+changes. The new implementation should also be API-compatible with the
+older implementation. We can create a maintenance branch for anyone
+using the baremetal-operator who wishes to support the older
+implementation for any reason.
+
```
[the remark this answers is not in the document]
dhellmann: Good point, I made a note of that here and in the work items list.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,202 @@
...
+## Drawbacks
+
+None
+
+## Alternatives
+
+We could use a new version of the operator-sdk instead of
+kubebuilder. The amount of work involved would be effectively the
+same, though, and we should benefit from standardizing on tools for
+all metal3 repositories.
```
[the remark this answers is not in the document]
dhellmann: Yes, that's right. If we ever do decide to provide an operator compatible with OLM then we should be able to use the operator-sdk to do that without having to rebuild the whole thing again.

dhellmann:
> The massive amount of churn required to basically stay in place makes me very uneasy, but I guess the webhook generation stuff would be nice.

Me, too. I wish there was a better migration strategy, but this feels like a one time thing since the 2 teams have agreed to maintain compatibility after this point. And the web hooks are required to migrate from an alpha version to a more stable version.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Design Details
+
+The process for creating the proof-of-concept migration was basically:
+
+1. Move everything in the git repository out of the way.
+2. Replace the copy of operator-sdk with a copy of kubebuilder, with a
+   similar patch to support float fields.
+3. Use kubebuilder init to create a new project in the repository,
+   setting the domain to an empty string.
```
zaneb, line 110: Implementation thought: I wonder if it would pay to do this init in a fresh branch, then merge the master branch into it (either in one big merge commit, or maybe do a series of incremental commits followed by `git merge -s ours master` if that still does the Right Thing for git blame &c.).

zaneb (commented):
Let's bite this bullet then.
/approve

design/baremetal-operator/kubebuilder-migration.md
dhellmann: I expect to have a series of patches that don't work individually but do work when the PR is submitted. Using moving files and committing before editing in the same branch might appear cleaner in the history than creating new files and merging? I have no idea, but I could experiment if you think the commits from https://github.com/dhellmann/baremetal-operator/commits/kubebuilder3 aren't clean enough to review.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,213 @@
...
+### Risks and Mitigations
+
+Although operator-sdk and kubebuilder both use the same underlying
+code generation library, and are trying to achieve parity, they are
+likely to be using different versions and options from time to
+time. The new implementation will therefore introduce updated
+dependencies, and the longer we put off the work the more likely those
+dependencies are to require code changes. We will mitigate some of
+that risk by vendoring a copy of the tool, which lets us lock in a
+specific version to use.
```
maelk, line 140: I don't think there is a need to vendor a copy of kubebuilder to lock the version. In CAPM3 we do that with the go.mod in the hack/tools folder.

maelk: Of course modifying it to support float fields require either vendoring or maintaining our own fork of the tool and using a replace command (that would avoid huge commits in BMO). I would rather be inclined to use the second option.

design/baremetal-operator/kubebuilder-migration.md
```diff
@@ -0,0 +1,213 @@
...
+### Work Items
+
+1. Tag a release and create a maintenance branch for the
+   baremetal-operator.
+2. Follow a process as described in the Design Details section above
+   to migrate to kubebuilder.
+3. Tag a release to use for including in the cluster API providers.
+4. Rebase all open PRs after the new implementation is approved.
+5. Update the imports in CAPM3.
+6. Update the imports and vendoring in CAPBM.
```
maelk, line 171: Is this really needed considering we deprecated the repo ?

maelk (commented):
lgtm, I just have some comments regarding the vendoring of kubebuilder. We probably don't need it, we can probably get by with a fork

design/baremetal-operator/kubebuilder-migration.md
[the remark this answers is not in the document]
dhellmann: Newer kubebuilder does also use newer versions of the CRD type, which give us features like ignoring extra fields in the API.

design/baremetal-operator/kubebuilder-migration.md
dhellmann: Vendoring it made it obvious how to run a specific version, so it seemed simpler. I could look at forking instead, and having the Makefile build from the fork.

design/baremetal-operator/kubebuilder-migration.md
dhellmann: I need to do that step downstream where the repo is still used. I can clarify that in the text or remove the task entirely.

maelk:
I have been working with the Tilt setup for CAPM3 and BMO, and moving to a kubebuilder structure (with a config folder, and a main.go at the root of the repo) will put us in line with the expectations for providers in CAPI. That's also an argument for moving to Kubebuilder :)

design/baremetal-operator/kubebuilder-migration.md
maelk: Actually I was a bit confusing things. We do not do it right now in CAPM3, we have instead an installation script that installs a specific version in a local folder : https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/hack/tools/install_kubebuilder.sh Sorry for the confusion

dhellmann: OK, we could either publish built releases on GitHub and use an install script like that or we could have the scripts use an image. I'll spend a little time investigating which would be simpler.

dhellmann: I've updated the text to describe the approach we discussed in the community meeting today.

dhellmann:
I believe the latest draft addresses all of the review comments as well as the input from today's community meeting discussion.

maelk:
We agreed on a lazy consensus on 23rd EOD in the meeting today, right ?

dhellmann:
> We agreed on a lazy consensus on 23rd EOD in the meeting today, right ?

That was my understanding, yes.

maelk:
lgtm

PR #125, opened 2020-07-17
Add dnsServers field to IPAM

maelk:
This PR adds the dnsServers field as part of the IPPool resource and modifies the Metal3DataTemplate to make use of this new field. This was requested in https://github.com/metal3-io/ip-address-manager/issues/8

### DIFF
--- a/design/ip-address-manager/ip-address-management-for-networkdata.md
+++ b/design/ip-address-manager/ip-address-management-for-networkdata.md
@@ -169,6 +169,8 @@ spec:
       subnet: 192.168.0.0/24
       gateway: 192.168.0.1
       prefix: 24
+      dnsServers:
+        - 8.8.8.8
     - start: 192.168.1.10
       end: 192.168.1.15
       subnet: 192.168.1.0/24
@@ -176,6 +178,8 @@ spec:
       prefix: 24
   gateway: 192.168.1.1
   prefix: 24
+  dnsServers:
+    - 8.8.4.4
   preAllocations:
     "RenderedData-10": 192.168.0.9
     "RenderedData-9": 192.168.0.8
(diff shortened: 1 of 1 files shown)


maelk:
Considering this is a minor API change, I'll start the implementation on the side. I'll adapt wrt the comments if needed.

dhellmann:
This looks fine as a standalone change. I wonder how it might integrate with the network management design, though. [That design](https://docs.google.com/document/d/1l2OIM6pxo3bX99ccPsLtjKlSsiXZFAU4Dvx6OIa_M3k/edit) also includes a "Network" type that will be involved in specifying the VLAN(s) to use for hosts. Perhaps we should spend a little time trying to unify the approaches, so we do not end up with a bunch of separate networking concepts in different parts of the API.

maelk:
I would indeed think that the addressing and the network configuration are different things at that level. Maybe we want to link a pool and a vlan, but in that case, we know the network and addressing configuration objects to reference when creating the Metal3Machine. The link between the two could be maintained by the higher level entity using Metal3. We would not want to assume anything on behalf of the user.

dhellmann:
What drives the DNS server settings more, the layer 2 details like VLAN or the layer 3 details like the IP pool? Routing seems clearly tied to the layer 3 settings. Is DNS really related to either directly, or is it a separate entity?

maelk:
I would say it is tied to the layer 3. Depending on your IP configuration you would need to hit a specific dns server or another.Considering for example that dns servers are advertised in DHCP offers, I would definitely keep it within the IPAM context.

design/ip-address-manager/ip-address-management-for-networkdata.md
```diff
@@ -169,13 +169,17 @@ spec:
       subnet: 192.168.0.0/24
       gateway: 192.168.0.1
       prefix: 24
+      dnsServers:
+        - 8.8.8.8
```
hardys, line 173: Not related to this PR but I notice the prefix is duplicated since the subnet is specified as a CIDR - is that intentional?

hardys:
> I would say it is tied to the layer 3. Depending on your IP configuration you would need to hit a specific dns server or another.Considering for example that dns servers are advertised in DHCP offers, I would definitely keep it within the IPAM context.

I agree - I added an unrelated question as I missed the original PR but this lgtm

design/ip-address-manager/ip-address-management-for-networkdata.md
maelk: yes, it is for two different uses. the CIDR is used to give the boundaries of the allocation addresses, while prefix is used for the configuration. Maybe an example could make this clearer : Let's say we have a /24 on that network. However, we want the IP addresses to be allocated from the first /25 only, then the CIDR will be /25, but the prefix given will be /24 so that the host can reach the whole subnet. Does this make sense @hardys  ?

dhellmann:
This looks good to me. @hardys had a question, so I'll leave it open until he has had a chance to review the answer.

/approve

maelk:
@hardys and @dhellmann This has now been implemented since there was no further feedback. Is it ok to merge ?

PR #131, opened 2020-08-12
Add Metal3 architectural diagrams

fmuyassarov:
Co-authored-by: Maël Kimmerlin <mael.kimmerlin@est.tech>


maelk:
Would you mind giving more meaningful names to the files please ? Also, maybe svg or png would be a better format if possible

hroyrh:
A small comment - in  Slide2.jpg image shouldn't we use "Control Plane Nodes" instead of "Masters".

fmuyassarov:
> Would you mind giving more meaningful names to the files please ? Also, maybe svg or png would be a better format if possible

Oh forgot that. Thanks.

fmuyassarov:
> A small comment - in Slide2.jpg image shouldn't we use "Control Plane Nodes" instead of "Masters".

Thanks. fixed.

dhellmann:
These look good to me. Are we ready to merge them?

/approve

fmuyassarov:
> For the images/architectural diagrams/components.svg I think it would help to update the diagram to reflect 2 controllers - KCP and CABPK

CAPBK is already there on picture, on the right most block. For KCP we didn't include it because here we wanted to show the references for mainly cluster and machines objects. We could add later another diagram with all the CAPI objects like KCP, MachineHealthCheck, etc.

fmuyassarov:
> These look good to me. Are we ready to merge them?
> 
> /approve

Yes I think.

PR #132, opened 2020-08-18
design for automatic configuration of the network

zhouhao3:
This proposal is to expand the scope of Metal³ to include an API to manage physical network devices.

Google Docs address: https://docs.google.com/document/d/1QT3ojXlfy0J0TjbpmOR86mgeN-mZuEUa81v_R755Urs
Issue address: https://github.com/metal3-io/baremetal-operator/issues/570
Demo address: https://github.com/Hellcatlk/network-operator

Co-Authored-By: Andrea Fasano(andfasano)
Co-Authored-By: Maël Kimmerlin(maelk)
Co-Authored-By: Song Shukun <song.shukun@fujitsu.com>
Co-Authored-By: Zou Yu <zouy.fnst@cn.fujitsu.com>
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>


zhouhao3:
PTAL

zhouhao3:
@maelk @dhellmann @zaneb Can you give us some suggestions? If this method is feasible, we will proceed to the coding phase.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+      SwitchPort CR to match the network configuration.
+    - The Switch controller will configure the switch based on the SwitchPort
+      CR.
+    - After the switch configuration is completed, the BMH is connected to the
+      provisioning network. Ironic will perform introspection and update the
+      status of the BMH. The `.status.hardware.nics` in the BMH CR will contain
+      the LLDP information collected on each NICs.
+    - CAPM3 monitors BMH CRs and obtains the information of each interface
+      stored in `.status.hardware.nics`. CAPM3 will also create the
+      corresponding Interface CR if it is not created.
```
maelk, line 137: I would assume the interface CRs must be created by the admin and linked to the BMH. otherwise there is no way to know what interface CR is linked to which OS interface, as far as I know (except mac addresses, but do we want to force the admin to give all the mac addresses of all the interfaces ? )

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+- User:
+  - Creates production Network CRs.
+  - Creates a Cluster and specify its `.spec.networkRef`.
```
maelk, line 140: that cannot be done at the cluster level. a cluster must have multiple networks, and some set of hosts might have some connectivity to vlan A B and C while some other in the same cluster could be connected to A D and E. In my opinion, the network references must be defined per interface for the Metal3Machine(template).

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+    - CAPM3 finds a ready BMH.
+    - CAPM3 selects an Interface from the Interface CRs owned by this BMH to
+      connect to the network specified by the user.
```
maelk, line 143: which one ? that matters because it needs to be linked to the cloud-init config provided to that host.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+    - CAPM3 writes the chosen Interface CR information into the Network CR.
```
maelk, line 144: I would propose to do this the reverse way, as described in a previous comment

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+    - Network Controller and Switch Controller will complete the automatic
+      configuration of the switch.
+    - After CAPM3 observes that the SwitchPort configuration is complete, it
+      triggers provisioning via the provisioning network.
```
maelk, line 148: this means delaying the provisioning, we can do that in two ways, either using the pause annotation (but that might be risky if there are changes on the ironic side, they will not be reflected), or using a new annotation created for that purpose, or not providing the image when selecting the BMH (but that is relying on some BMO behaviour that is not really stated as part of the API)

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+    - When the provisioning is completed, CAPM3 removes the bmh's Interface for
+      provisioning from the provisioning Network CR, thus disconnecting the BMH
+      from the provisioning network.
```
maelk, line 151: Is this desired ? In some cases probably, but it should be up to the user. For example if we pivot, to have a self hosted cluster, the nodes that run ironic must have access to the provisioning network. so that should be up to the user.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+### Changes to Current API
+
+#### Metal3Cluster CRD
+Add a `networkRef` field to `.spec` to point to the corresponding Network
+resource.
+
+```yaml
+Spec:
+ ...
+ networkRef: sample-network
```
maelk, line 162: I don't think we should modify the Metal3Cluster. the network configuration should be at the node level, not cluster level.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+ ...
+```
+
+#### CAPM3
+- Before BareMetalHost provisioning, CAPM3 obtains all the NICs information. If
+  the corresponding Interface CR has not been created, it will create one.
```
maelk, line 168: I don't think CAPM3 can do that, it does not have the information linking an OS interface to the Interface CR it is creating, and the switchport behind. LLDP could allow you to do that to a certain extent, but we should consider cases where we do not have LLDP too.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+- If a BMH has multiple interfaces, CAPM3 selects one to use.
```
maelk, line 169: again, it should not be random decision. the user should be able to select a specific OS interface, i.e. a specific CR interface linked to that OS interface.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+- Writes the ref information of the selected Interface into `.spec.interfaces`
+  in the corresponding Network CR.
```
maelk, line 171: how can CAPM3 link a specific interface CR to a specific Network CR ? a Metal3Machine might need to reference multiple network CR, and it needs to be linked to a specific OS interface, then through BMH, to a specific Interface CR

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+ hardware:
+ …
+ nics:
+   // The following data is obtained by Baremetal Operator and
+   // filled with each lldp information.
+   - name: eth0
+     ...
+     lldp:
+   - name: eth1
+     ...
```
maelk, line 198: The admin should be able to link the BBMH interfaces in order with the interfaces CRs, for example :
```suggestion
 spec:
   interfaces:
     0 : 
       name: interface-0-cr
       namespace: metal3
     1:
       name: interface-1-cr
       namespace: metal3
 status:
 …
 hardware:
 …
 nics:
   // The following data is obtained by Baremetal Operator and
   // filled with each lldp information.
   - name: eth0
     ...
     lldp:
   - name: eth1
     ...
```

maelk: That would mean that we would change the Metal3machine definition like this : 

```yaml
spec:
  interfaces:
    0:
      name: network-cr-for-iface-0
      namespace: metal3
    1:
      name: network-cr-for-iface-1
      namespace: metal3
```
In that case the order matters, hence using a map with integer keys to match the order in the BMH

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+### Add new CRD and controller
+
+#### Interface CRD
+```yaml
+apiVersion: network.metal3.io/v1alpha1
+kind: Interface
+metaData:
+  name: bm0-eth0
+spec:
+  portID: 10
```
maelk, line 210: what is this ID ? and why do we need to switch port mac address here if we can get it from the switchport CR maybe ?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+#### Interface Controller
+- Watch
+  - Interface CR
+- Configuration Operations
+  - Whenever a new Interface CR is created, Interface Controller will find the
+    corresponding SwitchPort CR according to `.spec.portID` and
+    `.spec.switchMacaddress`, and write the reference of the SwitchPort CR into
+    `spec.neighborRef`.
```
maelk, line 226: How would the controller know which switch this is the port of ? I think we should rather give the switch reference in the spec. We could get the switch name through lldp and match it with a switch CR, but we must consider cases where we do not have LLDP, so the linking must be done manually by the admin.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+  networkType: physicalSwitch
+  // The VLANs available
+  vlans:
+    - id:
+    - id:
+  // The untagged VLAN ID
+  vlanId:
+  interfaces:
+    - name: bm0-eth0
+    - name: bm1-eth1
```
maelk, line 245: that should rather be in the status, otherwise there are higher risks of misconfiguration, if it can be done by the user. I proposed to reverse that workflow.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+metadata:
+  name: example-switch
+spec:
+  accessDetail:
+    ip: 192.0.0.1
+    OS: fos
+    port: 22
+    // Store the username and password of the switch
+    // in the form of secret
+    accessSecret: example-switch-secret
```
maelk, line 275: we should probably have some info that can be matched through LLDP like the actual switch name (not the CR name). Maybe we could add here also a map of the switch ports, referencing the switchport CRs maybe, and listing things like allowed configurations for that port.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+  // vlans indicates the vlan information to be changed to.
+  // Changed by network controller.
+  // The VLANs available
+  vlans:
+    - id:
+    - id:
+  // The untagged VLAN ID
+  vlanId:
+  ownerRef:
+    name: sample-switch
```
maelk, line 297: that's in metadata.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+### Risks and Mitigations
+
+TBD
+
+### Work Items
+- Define the Switch, SwitchPort, Network and Interface CRD
+- Implementation of Switch, Network and Interface Controller
+- Change the API of Metal3Cluster and BareMetalHost
+- Unit tests
+- E2e tests in metal3-dev-env
```
maelk, line 334: This proposal does not say anything about a test setup. Metal3-dev-env switches are for now not configurable. it means we would need to switch to OVS probably (the setup with libvirt must be checked), and the maybe run an SDN controller on top, to not directly interact with OVS ? However, metal3-dev-env should probably still support deployments without the network APIs part, i.e. as it is now, being backwards compatible. This kind of changes should be addressed in the proposal too.

maelk (commented):
First let me apologize for the delay in the review. I was on holidays, and then had to work through my backlog. 
That is definitely a step in the good direction, thank you for the submission. There are some areas that are very unclear to me, specifically around the persona definitions and user stories, the permissions and isolation side of it was completely left out. Then around the workflows. I think the current approach should in some part be reversed. Also, there will not be only one network CR per cluster. each node can have one network CR per interface. and the configuration might be different for all nodes. Also you should consider physical hosts connected to multiple switches, where the user must be able to select specifically an interface for a configuration to be able to match it with the configuration done in cloud-init.

Overall I think this would probably need a call with all interested parties to discuss the design and make decisions on the approach we want to take. I am afraid that doing it over a PR would not be manageable considering the extent of the discussions that are expected.

And finally, please create new commits for now when you do modifications so that we can track it easily.

Thank you for this submission, let's keep the discussion rolling.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+- Action: allow, deny
+- Port status: up, down (status of the switch port)
+
+## Design Details
+
+Resources and controllers scheduled to be added and modified in the current
+design:
+- Switch CRD:
+  - Switch CR represents a physical switch. It contains the system information
+    (vendor, OS, ...) and the access information of the switch.
```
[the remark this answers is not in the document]
dhellmann: If we move ports to a separate namespace, we have to worry about duplication and conflicts.

Perhaps tenants should not be consuming ports directly at all, but using something like a claim to ask for configuration changes on the switch? I'm not sure that really resolves the issue with conflicts, though.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+    metal machine. It contains device information, LLDP information including
+    the connected switch’s mac address and the switch portID of the link, and
+    the reference referring to the device which this interface is connected to.
+- Interface controller:
+  - Monitors Interface CRs. When there is a new Interface CR, find the
+    corresponding SwitchPort CR according to the provided information and store
+    the reference of that SwitchPort CR in the Interface CR.
+- CAPM3:
+  - Gets the LLDP information of all the interfaces in the host. If there is no
+    corresponding Interface CR, create one.
```
dhellmann, line 96: Why would CAPM3 create these? Why not a new controller?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+  - If a host has multiple interfaces, selects one for network configuration.
```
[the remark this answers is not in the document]
dhellmann: If a host has multiple interfaces, I would assume they all need some configuration. So it's not a matter of picking one, it's a matter of determining the correct configuration for each interface. We still need more detail about how that would work, but perhaps lines 102-103 describe the manual process for establishing the right links?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+    configuration.
+- Interface CRD:
+  - Interface CR represents a network interface, this can be a NIC in a bare
+    metal machine. It contains device information, LLDP information including
+    the connected switch’s mac address and the switch portID of the link, and
+    the reference referring to the device which this interface is connected to.
+- Interface controller:
+  - Monitors Interface CRs. When there is a new Interface CR, find the
+    corresponding SwitchPort CR according to the provided information and store
+    the reference of that SwitchPort CR in the Interface CR.
```
dhellmann, line 93: Why does the switch port also need to be updated?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+
+The main implementation process is described below:
+- After creating a BMH, the administrator puts an interface into the
+  provisioning network.
+
+  ![automatically configure 1](images/automatically_configure_1.png)
+
+- After an Interface is put into the provisioning network, the LLDP information
+  of other interfaces can be obtained and network configuration can be
+  performed.
```
dhellmann, line 109: This section needs to be rewritten to cover what is doing each action (obtaining the LLDP information and performing the network configuration).

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+
+  ![automatically configure 2](images/automatically_configure_2.png)
+
+The workflow of a basic user story:
+- Administrator:
+  - Creates a Network resource used as provisioning network.
```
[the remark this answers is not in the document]
dhellmann: I think we're going to want 2 levels to the API, to manage the RBAC. The low level will include the actual physical connections, and only admins should be able to use those resources. The higher level will describe a desired configuration from the tenant's perspective in a more general way that can be mapped to the hardware configuration (2 separate networks with specific VLAN IDs, for example).

We should also consider the SDN case, and whether that changes the way we need to define networks to go beyond just a VLAN.

It doesn't feel like we should be inventing this from nothing. We should be able to learn from the APIs of other tools like AWS, OpenStack, etc. What do those look like? How do they separate the RBAC concerns?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+      fills it in `.spec.neighborRef` in the Interface CR.
+  - Adds the Interface CR information to `.spec.interfaces` in the provisioning
+    Network CR.
+    - The Network controller watches Network CRs and for each interfaces in its
+      `.spec.interfaces`, the controller will modify the corresponding
+      SwitchPort CR to match the network configuration.
+    - The Switch controller will configure the switch based on the SwitchPort
+      CR.
+    - After the switch configuration is completed, the BMH is connected to the
+      provisioning network. Ironic will perform introspection and update the
```
dhellmann, line 132: The baremetal-operator is going to try to perform inspection as soon as it has BMC credentials for the host, which happens up on line 116. If the host has to be attached to the provisioning network before that inspection process will work, then the admin is going to have to do that very early in the process, and the tenant won't get to pick the network.

design/automatic-network-configuration.md
dhellmann: Do people really run clusters with the hosts on different networks? Do you mean for the secondary networks or the main one where the ingress traffic is?

design/automatic-network-configuration.md
dhellmann: Yes, I think we can set the ownerRef of the host to avoid having something else claim it and the when we're ready to provision set the image details.

design/automatic-network-configuration.md
dhellmann: I think the API would be easier to understand if the interfaces were linked to NICs via a name or other value instead of just the order.

design/automatic-network-configuration.md
maelk: I meant secondary networks. But there could be cases where the primary network would be different between hosts, as long as routing is in place (for example different availability zones ? )

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+- Define a Kubernetes API for configuring network switches.
+- Automatically configure the network infrastructure for a host when adding it
+  to a cluster.
+
+### Non-Goals
+- Add new Switch API to BMO.
+
+## Proposal
+This document proposes to add a new mechanic to automatically perform physical
+network device configuration before provisioning a BareMetalHost.
```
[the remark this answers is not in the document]
zhouhao3: After deprovisioning, we plan to remove the interface from the previous configuration. (It will be explained in detail in a new document)

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+It contains 4 new objects (Switch, SwitchPort, Interface and Network object), 3
+new controllers (Switch controller, Interface controller and Network
+controller) and related changes to current code including CAPM3, BareMetal
+Operator, Metal3Cluster and BareMetalHost.
+
+### User Stories
+
+#### Story 1
+As a consumer of Metal³, when adding a machine resource to the cluster, it can
+automatically put the adapted BMH into the correct Switch's VLAN.
```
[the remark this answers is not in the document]
zhouhao3: In the new document, we will take these situations into consideration.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+LAG/mLAG.
+
+#### Story 4
+As an admin, when creating a cluster, it is possible to configure ACL settings
+in a switch for each bare metal server according to its functionality.
+The ACL setting:
+- Direction: ingress or egress
+- Protocol: SSH, HTTP, …all
+- Action: allow, deny
+- Port status: up, down (status of the switch port)
```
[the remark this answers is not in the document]
zhouhao3: Because we are not familiar with this, many situations are not considered. If you have a good idea about this, you can help us improve this story, thank you very much.

design/automatic-network-configuration.md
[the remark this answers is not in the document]
zhouhao3: Regarding the issue of RBAC, we will explain in the new design document.

zhouhao3 (commented):
Thank you very much for your suggestions. We will redesign this document based on your suggestions. After the document is updated, we will tell you the main update points, and hope to get your suggestions at that time.

Hellcatlk:
@dhellmann @maelk We have updated the design, Contains new three CRDs:
1. NetWorkDevice CRD
    -  contains all network devices under the same network.
2. ClusterNetwork CRD
    - describes some networks needed by cluster.
3. NIC CRD
    - contains the information of the network card of the node and the corresponding relationship of the network devices.

And a new controller:
1. ClusterNetwork controller
    - Managing the ClusterNetwork CR

PTAL

maelk:
Thanks for the update, i'll try to have a look by the end of the week

maelk:
I took a look at the new proposal. Could we maybe have a meeting to discuss it ? We could bring more people on this proposal. We can schedule something in the cluster-api-baremetal channel on slack if ok for you

sshukun:
Agree about the meeting. We also think it is difficult to design the API only by us and we need the knowledge and experience about network from the community. But please schedule the meeting after October 8th as we will soon enter a national holiday.

maelk:
Which timezone are you in ? (to see which slots to propose)

sshukun:
Most of us are in UTC+8, the others are in UTC+9

maelk:
I have created a doodle for next week, please select the times that would work for you : https://doodle.com/poll/ai96y8ev7iu22hcn

maelk:
@zhouhao3 @Hellcatlk @sshukun You have not answered the doodle. Can you please answer it asap ?

sshukun:
Sorry for the late response, I have answered the doodle.

maelk:
Let's make it Friday 12:00 UTC. I will send an invite to the metal3-dev group. do you want to receive it on a specific email address ?

dtantsur:
Similarly to how we use ironic instead of re-implementing all various quirks of hardware management, have you considered using some existing backend? I know some people use ironic together with https://opendev.org/x/networking-ansible.

Also, not sure if you know it or not, but ironic has support for automating VLAN changes to use a separate provisioning network (completely isolated from other flows). Currently it requires OpenStack Neutron, but we're open to revisiting standalone networking operations if there is enough interest.

sshukun:
@maelk  No, the metal3-dev group will be enough. Thanks for arranging the meeting.

maelk:
Notes from the meeting : https://docs.google.com/document/d/1ES_I9P1-EI2xS6hWqOGXMBIcwk4EPaae9Zv755VHVj4/edit (recording link at the top)

sshukun:
Thanks for the note and recording. The meeting is very helpful.

maelk:
Once you'll have time to rework the base of the proposal, we could setup a second one, to discuss in more details the permissions and also the different restrictions we could put in place, such as allowing only specific vlans for a specific tenant.

sshukun:
@maelk  Sure, let us have the second one. But currently we need to do some research and study about network, we will notify you once we are prepared for the second meeting.

sshukun:
@maelk Sorry for taking such a long time. We have reconsidered the design and created a new version. In this version, we try to abstract network environment and its configuration, and make it easier to be utilized by other components like CAPM3. But permissions and restrictions are still not mentioned in the new version.
So I think it is time for us to have the second meeting to discuss about permissions, restrictions and also the new version itself.

zaneb (commented):
There appear to be a lot of CRDs here that have no obvious purpose other than slicing up the data into smaller chunks - there's no controller needed to reconcile them and no obvious API-related reasons (like atomic uniqueness enforcement via naming) for them to exist. This adds many layers of indirection that will be expensive to resolve and cause unnecessarily slow convergence to eventual consistency, and appears to require very complex manual configuration to set up.
It seems to me that the BMH could link directly to a switch, with the port being just a text field. If you're lucky you even can fill it all in automatically using LLDP data from introspection.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,509 @@
...
+      desPortRange: # 22, 22-30
+status:
+  networkBinding:
+    - name:
+      namespace:
+    - name:
+      namespace:
+```
+
+#### PortLink CRD
```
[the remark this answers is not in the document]
zhouhao3: We modified the design document according to your suggestion, PTAL.

zhouhao3:
@maelk @zaneb @dhellmann We have updated the design document, I hope you can take the time to review and give us some suggestions.
In addition, we hope to have a second meeting with the community to discuss permissions, and restrictions.

zhouhao3:
@maelk @zaneb @dhellmann PTAL

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,485 @@
...
+    of the switch, and initializes all ports in `.status.ports`.
+
+* Administrator deletes `Switch` CR
+  * The Switch Controller detects whether the number of configured ports is 0.
+    If it is not 0, it is not allowed to delete the CR, otherwise delete the
+    `finalizers` and delete the CR. The controller determines whether a port is
+    configured by checking the port's `.networkConfiguration`. If this field is not
+    nil, the port is configured.
+
+* The NetworkBinding Controller modify `.status.ports`
```
maelk, line 426: I would say that the status of a resource should only be modified by its own controller

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,485 @@
...
+### Changes to Current API
+
+#### Metal3Machine CRD
+
+```yaml
+spec:
+  # Specify the network configuration that metal3Machine needs to use
+  networkConfigurationRef:
+  - name: nc1
+    namespace: default
```
maelk, line 218: I am still unclear on how, as a user, you specify which port of the BMH should have which configuration. How do you pick the switch port on which to apply this configuration ? In your proposal, the metal3Machine seems to be the only point where you can bring together the switch configuration (from linking it to the BMH port) and the cloud-init data that you will pass to the BMH. (It could also be done at the BMH level if there was no CAPI in the picture).

The user will need to specify the network configuration to be applied via cloud-init on the host. How can this be mapped with the configuration applied on the switch. The goal is to ensure that if you configure interface 1 of the OS, the proper configuration is applied on the switch in a deterministic manner to ensure both configurations match.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,485 @@
...
+    `.status.networkBindingRefs` is 0, if it is not 0, it is not allowed to
+    delete the CR, otherwise the Controller deletes the `finalizers` and deletes
+    the CR.
+
+* User modify `NetworkConfiguration` CR
+  * The NetworkConfiguration Controller deletes all the old `NetworkBinding` CRs
+    and creates a new one for each deleted `NetworkBinding` CR according to the
+    contents in `.status.networkBindingRefs`.
+
+#### NetworkBinding CRD
```
maelk, line 324: From what you put in the NetworkConfiguration and NetworkBinding, it looks like the two could be merged. What would be the value of keeping them split ?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,485 @@
...
+    namespace: def
+status:
+  ports:
+  - portID:
+    # Current network configuration applied to this port
+    networkConfiguration:
+    # Reference refer to desired networkConfiguration CR
+    networkConfigurationRef:
+      name:
+      namespace:
```
maelk, line 402: I don't think we need both. We can put the Ref to the current one, and update it when the port is updated, no need to keep both current and desired. At the end what matters is the desired one, and it will be overwritten if it changes. until that point it is valid.

maelk (commented):
Sorry for the delay in the review! I'll try to be more active with this proposal in the future

design/automatic-network-configuration.md
sshukun: Our thought is that if the configuration of one port is changed, we need to first clear the previous configuration applied on that port and then apply new configuration. If only have the desired one, the cleaning will be difficult as we don't know the previous configuration. So we need to keep both current and desired here.

design/automatic-network-configuration.md
zhouhao3: We do this to better divide permissions. Some users may only be able to create or modify NetworkBinding, but not modify NetworkConfiguration.

design/automatic-network-configuration.md
zhouhao3: updated.

zhouhao3 (commented):
@maelk  Thank you for your review, we hope to have a second meeting with the community to discuss permissions, and restrictions. Please let us know when you are free.

design/automatic-network-configuration.md
zhouhao3: We do this to better divide permissions. Some users may only be able to create or modify NetworkBinding, but not modify NetworkConfiguration.

design/automatic-network-configuration.md
zhouhao3: The user cannot specify which interafce or which port to configure. In our opinion, the user does not know which BMH will be selected and which switchport is connected to this BMH.
The user just declares which network configuration is required for the created Metal3Machine, and then the configuration of the specific port is performed by the operator.

maelk: That's a problem. How do you configure properly the network configuration for the OS via cloud-init if you don't know which interface is configured with that configuration ?

design/automatic-network-configuration.md
maelk: I don't think the user itself should create the network binding in most of the workflows. We should consider rather being able to reference the desired configuration maybe on the device for which it will be applied. that would remove the need for that intermediary object without changing the permissions requirements since they would not need modifications rights on the network configuration. CAPM3 could then set that reference when the workflow is automated.

design/automatic-network-configuration.md
maelk: I would argue that you can get the current configuration from the switch, and you anyways would want to reset everything to a default state if de-provisioning or configure everything according to the new configuration, so the old one should not matter. For example you don't care which vlan were configured when you know that you want only X and Y. you'll remove all others than X and Y.

maelk:
I think we could start with a discussion on this current proposal first, trying to involve more people there. We can discuss the permissions etc. afterwards once we have the base in place and agreed by everybody. We can make a doodle to ask for disponibilities

maelk:
I think we could start with a discussion on this current proposal first, trying to involve more people there. We can discuss the permissions etc. afterwards once we have the base in place and agreed by everybody. We can make a doodle to ask for disponibilities

design/automatic-network-configuration.md
zhouhao3: After the user declares the networkConfiguration used by Metal3Machine, CAPM3 selects the appropriate BMH according to the configuration requirements, then selects the network card that meets the network configuration requirements, and then creates NetworkBounding according to the port connected to the network card to trigger the network configuration.

maelk: That is for configuring the switch. But how do you map that with the network configuration that needs to be done on the host by cloud-init for example ? How do you know what interface is connected to which vlan once you are on the host running cloud-init ?

design/automatic-network-configuration.md
zhouhao3: Our idea is to decouple this new operator from CAPM3 as much as possible, so that CAPM3 does not need to know the APIs of other types of devices.

sshukun:
@maelk Agree with the idea of involving more people to the discussion. We can also explain our thoughts more detailly there.

design/automatic-network-configuration.md
maelk: I agree with that, but CAPM3 will have to be aware of those types anyways, considering that it will be the one doing the mapping between the desired config at the cluster level, and the baremetal host (via the metal3machine(Template))

zhouhao3: If `CAPM3` is asked to modify the `configuration` reference used by each port, how to notify each device when the `configuration CR` is modified?

In our original design, we refer to the implementation principle of `deployment`. When the `configuration` is modified, all related `networkBinding` will be deleted and then created to trigger the device configuration update.

zhouhao3: Like I said above, I think the `NetworkBinding CRD`  should still be retained, so that when the `Configuration CR`  is modified, it can be very convenient to trigger the corresponding switch to modify the configuration.(By deleting and then creating  `NetworkBinding CR`). @maelk @zaneb @dhellmann WDYT?

zhouhao3:
@maelk @dhellmann @zaneb  We want to organize a meeting to discuss this proposal together. We created a doodle: https://doodle.com/poll/tc6wcv4gkxzpsnvr 
Please select the times that would work for you.

zhouhao3:
In the v6 version, we made the following updates:
     1. Put the functions of link aggregation and lldp into Non-Goals as functions to be implemented in the future
     2. Added the handling of smartNIC
     3. The port structure has been changed, and nextRef is added to indicate the port it is connected to.
In addition, we also wrote about the api design of this proposal.
https://github.com/Hellcatlk/networkconfiguration-operator

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+adding it to a cluster.
+- Do network configuration when deprovisioning the hosts.
+- Design a network abstraction that can represent any of the target networking
+configuration, independently of the controller used.
+
+## Non-Goals
+
+- Implement a network controller which is aimed at being an integration with
+an existing controller only.
+- Implement a solution for a specific underlying infrastructure.
```
dtantsur, line 37: We need to have a vendor-agnostic reference implementation. On the meeting last week I've suggested using ansible-networking e.g. via https://pypi.org/project/network-runner/

I also don't think metal3 should be in a business of writing vendor-specific backends.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+additional methods, and then trigger the NetworkConfiguration operator to
+perform network configuration.
+
+### User Stories
+
+#### Story 1
+
+As a consumer of Metal³, when adding a machine resource to the cluster, according
+to the network configuration of the cluster, the physical host is connected to the
+corresponding logical layer-2 network segment.
```
dtantsur, line 57: User stories are usually formulated in terms of users' desires ("I want my traffic isolated from ..."), not technical implementation.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+Abstract all devices in the cluster (such as Switch, SmartNic).
+Each type of network device must have its own CRD.
+
+At the same time, each device has its own specific interface, and
+needs to implement corresponding interfaces for different vendors
+and device types. The following is an example of a switch interface:
+```go
+// Switch is a interface for different protocol
+type Switch interface {
+    // PowerOn enable switch
```
dtantsur, line 105: I don't think we should manage power of switches. Also "enable" and "power on" are different things in bare metal world. Could you clarify what these 2 calls do and also consider if they're needed at all, especially for the first version.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+    kind: SmartNicConfiguration
+    namespace: default
+    nicHint:
+      name: eth1
+      smartNIC: true
+```
+
+#### BareMetalHost
+
+Add the `ports` field to `.spec` to indicate the connection information and
```
dtantsur, line 284: MAC addresses may not be known until inspection, how will it be handled here?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+  secret:
+    name: switch-example-secret
+    namespace: default
+```
+
+###### NETConfSwitch CRD
+
+```yaml
+apiVersion: v1alpha1
+kind: CLISwitch
```
dtantsur, line 372: Wrong kind here and below?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+an existing controller only.
+- Implement a solution for a specific underlying infrastructure.
+- Implement a solution for link aggregation in the future.
+- Implement a solution for lldp in the future.
+
+## Proposal
+
+This document proposes to add a new mechanic to automatically perform physical
+network device configuration before provisioning a BareMetalHost. First we
+added the NetworkConfiguration operator to implement the network configuration
```
dtantsur, line 45: Where will the new operator live? I'd prefer it to be a separate repository to avoid overcomplicating BMO.

dtantsur (changes requested):

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+#### Deletion workflow
+
+1. User sets deletion timestamp on all resources (that automatically triggers
+the deletion of the Provider Switch because of OwnerRefs).
+2. Switch deletion is blocked until all SwitchPorts are deleted.
+3. Wait for switch port to be idle.
+4. If switchPort is idle, then remove the finalizer.
+5. Once no ports are left for the Switch, then remove the finalizer on
+the ProviderSwitch and on the Switch.
+
```
rhjanders, line 254: First of all thank you for submitting this for review, I think it is a great effort and a very valuable potential feature.

I would like to suggest that we consider handling state divergence as well. It is not an obvious issue at this stage in the design process but it is a known and often very disruptive issue in existing baremetal SDN solutions. It can be caused either by issues in the underlying SDN mechanism, network disruptions during configuration or errors on the physical switch level. I think it would be beneficial to have a workflow reconciling any possible state divergence on this level so that all the underlying implementations are expected to implement it also.

Some ideas for discussion:
* query configuration for a specific host/port, compare with what is expected, apply expected configuration if different
* walk the entire fabric looking for state divergence
* give the operator the ability to configure the "fabric walk" as a periodic task as well as an optional automatic reconciliation.

Those may be quite specific ideas but I do think it would be beneficial to consider them early. These comments are based on hard-learned real-life lessons from my operator days. 

I look forward to your feedback w/r/t this idea and I am happy to make more specific comments in the relevant parts of the document if there is interest.

zhouhao3: We have discussed this issue before, but we could not get a conclusion, so we leave the issue remained [here](https://github.com/metal3-io/metal3-docs/pull/132/files#diff-4d8a601ed0393d3ce33f572829bb22a35659f3b85d7a3c959275865ddf6d3e99R683-R702).
In [POC](https://github.com/Hellcatlk/network-operator), we designed a function to obtain the configuration of the switch periodically, and then compare it with the configuration specified by the user. If it is not the same, reconfigure the switch.

design/automatic-network-configuration.md
sshukun: We also intend to create a separate repository for the new operator and make sure to make the least impact to CAPM3 and BMO.

design/automatic-network-configuration.md
zhouhao3: Tnanks, We are trying to use network-runner, we will update this proposal later.

zhouhao3: We tried to use network-runner as the backend to configure the switch, but due to the limited functions of network-runner, the actual configuration of the switch port could not be obtained. In our design, after the port is configured with the user-specified network configuration, the actual configuration of the port must be continuously obtained. When the actual configuration is different from the user-specified configuration (for example, someone logs in to the switch and changes the port configuration) , The port will be reconfigured.
Since network-runner cannot obtain port configuration, it cannot meet our needs.

dtantsur: Please let them know and try to work with them on improving that. Any alternative backend is welcome, of course. But we cannot land API that has no generic implementation at all.

zhouhao3: We have posted a PR in network-runner to add get_port_conf feature, and we are currently advancing this PR. In the demonstration, we used the part of the code we wrote.
btw, We think that the POC and this proposal have been completed, can you review it as a whole? We hope that this proposal can be merged as soon as possible so that we can continue with the rest of the work.

design/automatic-network-configuration.md
zhouhao3: updated.

zhouhao3:
@dtantsur @zaneb @rhjanders @hardys @andfasano Hi, We think that the [POC](https://github.com/Hellcatlk/network-operator) and this proposal have been completed. Because this PR has not been reviewed for a long time, can you please review it? Only after this proposal is merged, we can continue to advance to realize this proposal. Thanks!

jan-est:
I think we can merge this proposal. It has been open for a long time. So let's not block this any longer.

fmuyassarov:
/cc @zaneb @dtantsur @maelk @rhjanders @sshukun @hardys @andfasano @jan-est 
Do you think we can agree on lazy consensus to merge the proposal? @sshukun gave a [demo](https://www.youtube.com/watch?v=LmVUftg5c1s) as well on today's community meeting .

maelk:
this lgtm. We can put the lazy consensus, IMO

andfasano:
I'm also fine with the lazy consesus

fmuyassarov:
If there are no objections, I would suggest to have a week of lazy consensus from today. As such, if there are no comments to be addresses by 15.10, we should be good to merge IMO.

design/automatic-network-configuration.md
zhouhao3: In our design, the person who created the BMH needs to know the MAC address

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,355 @@
...
+#### Story 4
+As an admin, when creating a cluster, it is possible to configure ACL settings
+in a switch for each bare metal server according to its functionality.
+The ACL setting:
+- Direction: ingress or egress
+- Protocol: SSH, HTTP, …all
+- Action: allow, deny
+- Port status: up, down (status of the switch port)
+
```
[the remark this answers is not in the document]
zhouhao3: Added the [Permissions](https://github.com/metal3-io/metal3-docs/pull/132/files#diff-4d8a601ed0393d3ce33f572829bb22a35659f3b85d7a3c959275865ddf6d3e99R526) section to explain the permissions in detail.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+#### Story 1
+
+As a consumer of Metal³, I hope that after adding the machine to the cluster,
+the cluster can automatically put the machine into the corresponding network.
+
+#### Story 2
+
+As a consumer of Metal³, when a machine is deleted from the cluster, the
+corresponding network configuration is automatically deleted.
```
zaneb, line 61: An obvious (to me) potential use case not covered here is when you need to connect the Host to the provisioning network for the purposes of provisioning, but you then want to remove it from that network once it has booted for security reasons.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+
+At the same time, each device has its own specific interface, and
+needs to implement corresponding interfaces for different vendors
+and device types. The following is an example of a switch interface:
+
+```go
+// Switch is a interface for switch backend
+type Switch interface {
+  // IsAvaliable check switch is avaliable or not
+  IsAvaliable() error
```
zaneb, line 106: typo: Available

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+
+1. CAPI / User set the deletion timestamp on the `Metal3Machine` CR.
+2. CAPM3 calls the `deconfigureNetwork()` to clears the contents of
+  `BMH.ports.configuration`.
+3. BMO deconfigures the ports.
+    1. `BMH.ports.provisioningConfiguration` is empty
+        1. BMO clears the configuration field in the switchPort CR
+          corresponding to the network card.
+        2. SwitchPort controller starts to deconfigure the network by calling
+          `device.DeConfigurePort()`. Return the port to a `cleaning` state.
```
zaneb, line 236: What does the "Idle" state look like after cleaning? Is there some default config for ports that are not explicitly configured?
How do you differentiate between ports that are 'deconfigured' and ports that are simply not managed by the controller (e.g. uplink ports)?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+  # The type of OS this switch runs
+  os: fos
+  # The ip address of the switch
+  ip: 192.168.0.1
+  # The port to use for SSH connection
+  port: 22
+  # Only for ovs switch
+  bridge:
+  # The login credentials of switch
+  secret:
```
zaneb, line 351: Something like credentialsSecret would make this more explicit.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+    name: switch-example-secret
+    namespace: default
```
zaneb, line 353: There's no reason to allow the Secret to be in a different namespace.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: Switch
+metadata:
+  name: switch-example
+spec:
+  provider:
+    kind: AnsibleSwitch
+    name: ansibleswitch-sample
```
zaneb, line 366: Would it be reasonable to assume that the provider CR has to have the same name as the Switch CR?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+
+- No multi-tenancy : one set of controllers deployed, all objects are
+  reconciled by the same controllers.
+- "Limited" multi-tenancy : one set of CRs (the most up-to-date), one set of
+  webhooks and a single network controller with clusterRole and in its own
+  namespace. All other controllers (CAPI + CAPM3) per namespace. Permission
+  separation between namespaces.
+- Full multi-tenancy : one set of CRs (the most up-to-date), one set of
+  webhooks. All other controllers per namespace. No permission separation
+  within the same namespace.
```
zaneb, line 524: I would not worry so much about running multiple controllers for different namespaces (this is never going to work well for k8s due to the CRD and webhooks being global, as you've identified).

What is more interesting is making sure that roles can be separate. So e.g. an administrator can assign a given Host to one of several tenants (each with their own namespace) and allow the tenants to control how it is provisioned (e.g. by using CAPM3, or something else) while making sure that e.g. only the administrator can define which networks the Host can connect to.

This will only actually be possible once the BaremetalHost itself is decomposed in some way along the lines of my draft proposal in https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit?usp=sharing (you must be a member of the Metal³ Google Group to view this doc), but it is worth thinking about how the network configuration would work in such a scenario. In particular, I think this scenario suggests that the network config should in future be associated with a BaremetalAllocation, not a BaremetalImage (i.e. deployment), and it follows from that perhaps the network configuration should precede the Host being selected by the CAPM3 rather than being part of the provisioning done by CAPM3.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+|:-|:-|:-|
+|Operator|None||
+
+#### Scenario: Infra team and consumers
+
+Description: There is an owner of the infra and some consumers. The infra
+team is the owner of the management cluster and manages the physical resources.
+A consumer is the manager of the target cluster, responsible for its lifecycle.
+The consumer gets resources from the infra team (tenant). No consumer action
+should impact another consumer. Metal3 setup on a shared management cluster,
```
zaneb, line 583: I agree that no consumer action should impact another consumer, but the rules below don't enforce that because they assume that the only resource that is exclusive to a consumer is the Port.
In reality, the common multitenancy use case would be to assign different VLANs to different consumers. Allowing each consumer to configure their own ports in any way they want would permit them to connect their hosts to any VLAN.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,738 @@
...
+- Configurability might be missing some coverage, operator manually changes it.
+  Some exceptions might be reasonable (e.g. credentials change)
+- Might be heavy process (can be fixed by reducing the frequency,
+  jittering etc.)
+- Other actors might increase complexity.
+- No full ownership guaranteed of the hardware
+- Could break during an upgrade
+
+Possibility to add a boolean in the CR to control whether we reconcile and
+re-apply the configuration if there is a delta, or not.
```
zaneb, line 636: IMHO it's a mistake to make it configurable. I'd suggest you always re-apply, but start with a very long polling interval if you're worried about resource usage.
Don't let people get used to the idea that they can manually change the config behind the controller's back and not have it reverted.

zaneb:
I finally finished the review I started 3 months ago. Apologies for the delay.
I'm not looking to block this - many of the comments are just clarifications that could easily be addressed in a follow-up. However, you may want to spend some time thinking about the multitenancy part.

design/automatic-network-configuration.md
zhouhao3: got it, thanks!

design/automatic-network-configuration.md
zhouhao3: This is indeed a problem that we did not consider, and we intend to solve it in the following ways:
Each user can specify any vlan. If this vlan is used by other users in the switch, it will randomly find an unused vlan and put the port in this new vlan. But the user does not know, only the operator knows and records this information. This allows different users to specify the same vlan, and the actual configuration is not in the same vlan. WDYT?

design/automatic-network-configuration.md
zhouhao3: Is it necessary to do this? I think it's strange that different CRDs have the same name.

design/automatic-network-configuration.md
zhouhao3: >What does the "Idle" state look like after cleaning? 

When performing the clean operation, we will call the `delete_port` function of `network_runner`, which will be processed according to the definition of each manufacturer, mainly to delete the vlan-related configuration.

>  Is there some default config for ports that are not explicitly configured?

No, ports without a specified configuration file will not be configured

> How do you differentiate between ports that are 'deconfigured' and ports that are simply not managed by the controller

We have a `Disabled` field in `Switch.ports` to indicate whether this port is available or controllable.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+In the network operator, we abstract the following roles:
+
+#### Device
+
+Abstract all devices in the cluster (such as Switch, SmartNic).
+Each type of network device must have its own CRD.
+
+At the same time, each device has its own specific interface, and
+needs to implement corresponding interfaces for different vendors
+and device types. The following is an example of a switch interface:
```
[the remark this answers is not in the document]
zhouhao3: We may not describe it clearly here. What we want to express is that each specific network device has its own CRD, and has its own interface, and some corresponding methods are written in the interface. After implementing this method, it can be called by the Switch instance.

design/automatic-network-configuration.md
zhouhao3: yes, I will improve this story later.

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,801 @@
...
+- Do network configuration when deprovisioning the hosts.
+- Design a network abstraction that can represent any of the target networking
+configuration, independently of the controller used.
+
+## Non-Goals
+
+- Implement a network controller which is aimed at being an integration with
+an existing controller only.
+- Implement a solution for a specific underlying infrastructure.
+- Implement a solution for link aggregation in the future.
```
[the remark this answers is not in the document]
zhouhao3: This is a temporary non-goal, We plan to implement the basic functions first, and then add other functions in the next phase.

zhouhao3 (commented):
Thank you very much for your review, we need more consideration and discussion to solve the multi-tenant issue.

design/automatic-network-configuration.md
zaneb: I would very strongly advise against that.

My intention here was to point out one of the ways in which the current model fails to account for multitenancy _of the network_ as discussed above. (I suspect this is also what @maelk was getting at when he [originally suggested](https://github.com/metal3-io/metal3-docs/pull/132#discussion_r485369047) documenting the approach to RBAC.) We should look at the problem in that context.

design/automatic-network-configuration.md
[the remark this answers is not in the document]
zaneb: What I think you're saying is that the connection/authentication data is stored in a separate CRD but the actual implementation is not done by a separate controller for the provider CRD, instead the Switch controller calls the method on some object which must be hard-coded into the Switch controller?

This eliminates the benefit of using the Cluster API model of separate provider CRDs (e.g. Metal3Machine), which is that you can add new providers without modifying the code of the main controller (Machine in the case of Cluster API, Switch in this case). There's very little difference between what you appear to have and just including the connection/authentication data in a free-form field in the main Switch API (which is how the Cluster API originally did it in v1alpha1).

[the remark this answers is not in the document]
zhouhao3: There is a problem with what I said yesterday. 
In our proposal, the configuration of each device may have multiple back-end implementations. 
For example, in the case of Switch, we use ansible as its back-end implementation, but there may be other back-end implementations to configure the switch in the future. This requires that no matter which method is added, this method must implement the `Switch interface`, so that you do not need to modify the controller code and directly use this new method to configure the switch.
I will add a detailed description in the proposal later.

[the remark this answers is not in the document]
zaneb: You _do_ need to modify the controller code to support another backend, because something needs to import the new backend and call [`Register()`](https://github.com/Hellcatlk/network-operator/blob/master/pkg/backends/switches/switches.go#L21-L23) with it.

The (newer) Cluster API approach is to have a separate controller for each platform (e.g. AnsibleSwitch) so that new controllers can be added without modifying the core (Switch) controller at all. They communicate only through the CRD interfaces, not through shared golang code.

design/automatic-network-configuration.md
zaneb: I could imagine that being useful, although the config needs to be in a CRD, not in a file.

design/automatic-network-configuration.md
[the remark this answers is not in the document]
zhouhao3: > What I think you're saying is that the connection/authentication data is stored in a separate CRD but the actual implementation is not done by a separate controller for the provider CRD, instead the Switch controller calls the method on some object which must be hard-coded into the Switch controller?

Yes, your understanding is correct. I will revise this document to make it clearer.

zhouhao3:
@zaneb We have updated the proposal based on your suggestions, including updated user stories, added a new CR to restrict users' use of vlan, etc. Please review again, thank you!
In addition, we are still considering the separation of roles that [you mentioned](https://github.com/metal3-io/metal3-docs/pull/132#discussion_r709514153), and we will continue to discuss with you when we have a conclusion

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,641 @@
...
+Indicates the content of the specific configuration for one port.
+Like `DevicePort`, `Configuration` also must have different CRDs for
+different device types (eg. `SwitchPortConfiguration` CRD).
+
+#### ResourceLimit
+
+Indicates the limit added to a user. It is created by the administrator
+for each user to limit the use of resources (eg. the allowed VLAN range).
+Like `DevicePort`, `ResourceLimit` also must have different CRDs for
+different device types (eg. `SwitchResourceLimit` CRD).
```
zhouhao3, line 116: Hi @zaneb, we use the following methods to limit the vlan (or other resources) that tenants can use in a multi-tenant scenario:
1. The administrator creates ResourceLimit for each user based on the negotiation with the user
2. The controller will verify whether the VLAN to be configured is within the allowed range before configuring the switch port.
3. When users need more VLANs, they need to negotiate with the administrator, and then the administrator changes the corresponding ResourceLimit.

However, this method relies on the communication between the user and the administrator, and may be more suitable for On-Premise scenarios. In the case where the user cannot communicate with the administrator in time (such as the public cloud), it may take a long time when the user wants to use more VLANs. What do you think of this situation?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,641 @@
...
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: SwitchResourceLimit
+metadata:
+  name: user-limit
+  namespace: user1
+spec:
+  # Indicates the range of VLANs allowed by the user
+  vlanRange: 5-10
```
zhouhao3, line 404: @zaneb  We clarify the vlan range that users can use through the vlanRange field. But when the user does not use so many VLANs, other users cannot use it either, which will cause a waste of resources. 
So we think it is better to limit the number of VLANs that users can use, and then combine the mapping [method](https://github.com/metal3-io/metal3-docs/pull/132#discussion_r710014679) we said before to ensure that users will not use the same VLAN. WDYT?

zhouhao3:
There are some problems with the previous design:
1. When the administrator assigns some usable vlans to a user, the code will not detect that the vlan will not be repeated
2. When the administrator reduces the vlan that the user can use, the code will not detect whether the user is using this vlan.

So we updated the proposal and added [SwitchResourceLimit Controller](https://github.com/metal3-io/metal3-docs/pull/132/files#diff-4d8a601ed0393d3ce33f572829bb22a35659f3b85d7a3c959275865ddf6d3e99R413-R433) for monitoring.

design/automatic-network-configuration.md
zaneb: That doesn't sound uncommon for the case where you want to increase your quota for something in most public clouds.

zhouhao3:
Hi @zaneb, as I said above, we have modified the implementation structure of this proposal. Could you please review again? Thanks!

design/automatic-network-configuration.md
dtantsur: @zaneb does the current version address your concerns? I assume the configuration-port split does it?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,616 @@
...
+- SwitchPortConfiguration object immutable.
+- Changes would be done by new configuration and linking the new configuration
+  in the ports.
+- Reconciliation triggered for each port.
+- Required to keep the CAPI model support. New configuration should not impact
+  old machine running.
+
+#### Clean port definition
+
+Defined by infra team, which ports are for PXE/iPXE boot, which are not.
```
dtantsur, line 538: Where is this defined?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,616 @@
...
+
+Two cleaning configurations:
+
+1. PXE/iPXE boot (allows IPA booting).
```
dtantsur, line 542: What about non-PXE implementation? Do they fall here as well?

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,616 @@
...
+    1. Indicates that the port is being configured.
+    2. The state handler configures the port's network and checks configuration
+    progress. If finished set the state of CR to `Active` state.
+- Active - Steady State
+    1. Indicates that the port configuration is complete.
+    2. The state handler check whether the target configuration is consistent
+    with the actual configuration, return to `Configuring` state and clean
+    `status.configurationRef` when inconsistent.
+    3. If deletionTimestamp is set -> go to `Cleaning`
+- Cleaning
```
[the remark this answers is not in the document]
zhouhao3: thanks, all done.

design/automatic-network-configuration.md
zhouhao3: @zaneb The main remaining problem is this. Do you think our current design can solve this problem?
Can you and @dtantsur review again? If there are no other questions, can you give an approval? This is very important to us, thank you!

zhouhao3:
@zaneb @dtantsur Hi, Because this PR has not been reviewed for a long time, I have no choice but to ask you to review it again.
For the convenience of review, you can refer to the [information](https://github.com/metal3-io/metal3-docs/pull/132#discussion_r741644236) I mentioned above.
btw, I know that reviewing this PR may take some time, so please let me know if you are reviewing. Thanks!

design/automatic-network-configuration.md
```diff
@@ -0,0 +1,617 @@
...
+different device types (eg. `SwitchResource` CRD).
+
+#### Relative controllers
+
+For each type of device, there will be a series of CRDs, but only one
+indispensable controller to handle the configuration job.
+That controller need to monitor the corresponding `DevicePort`,
+and whenever the `Configuration` of a `DevicePort` is changed,
+the controller should be able to interact with the `Device` through
+the `ProviderDevice` to do the configuration job.
```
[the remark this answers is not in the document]
dtantsur: So, what's the resolution here? I see ProviderDevice still mentioned below, do we keep it or remove it?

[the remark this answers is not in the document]
zhouhao3: We think it would be better to keep the ProviderDevice, and we also explained the reasons above, but also need to see @zaneb 's views on this.

zhouhao3:
@zaneb Hi, This PR requires you to see if the current version address your concerns. We cannot move forward until you confirm it, so can you review it?

zhouhao3:
Hi, @zaneb PTAL.

kashifest:
/hold

PR #134, opened 2020-08-31
describe RAID API for BareMetalHost

dhellmann:
This design describes changes needed to allow RAID storage
configuration as part of managing hosts.

See https://github.com/metal3-io/baremetal-operator/pull/292 for the
implementation.

### DIFF
--- a/design/baremetal-operator/raid-api.md
+++ b/design/baremetal-operator/raid-api.md
@@ -1,0 +1,202 @@
+# raid-api
+
+## Status
+
+implementable
+
+## Summary
+
+This design describes changes needed to allow RAID storage
+configuration as part of provisioning hosts.
+
+## Motivation
+
+While it is reasonable to assume that cloud instances and VMs only
+need one drive with minimal configuration, many classes of server
+hardware need to have RAID storage configured to present fast and
+reliable storage to the OS. While it is possible today to
+pre-configure the hosts, it would be more convenient for users if
+metal3 could do that work for them.
+
+### Goals
+
+1. Provide an API to specify the RAID storage configuration to use
+   when provisioning a host.
+2. Outline the controller/provisioner logic changes needed to pass the
+   instructions to Ironic to have the storage configured.
+3. Support basic RAID levels `0`, `1`, and `1+0`.
+
+### Non-Goals
+
+1. Provide default RAID configuration.
+2. Retain existing RAID configuration on a host if it matches the
+   desired configuration at the time of provisioning (we make no
+   assumptions about preserving the state of the storage system or its
+   contents).
+3. Re-provision a host if the RAID settings change. (Users are
+   expected to explicitly deprovision the host and then reprovision
+   it.)
+4. Support every possible RAID configuration.
+
+## Proposal
+
+### User Stories
+
+1. As a user, I can specify the RAID configuration for a host before
+   provisioning it.
+2. As a user, I can choose between hardware and software RAID for each
+   host.
+
+## Design Details
+
+### API Changes
+
+The `BareMetalHost` API will change to accept RAID configuration
+inputs and to reflect the settings that were used when a host is
+provisioned (in case they are changed after provisioning).
+
+New `spec` fields will be added in a section called `raid`, with
+separate fields describing the hardware and software RAID settings. If
+the hardware settings are provided, the software settings are ignored.
+
+For hardware RAID, the inputs describe how to combine physical disks
+into a RAID volume via the RAID controller hardware in the host.
+
+```yaml
+spec:
+  raid:
+    hardwareVolumes:
+      - sizeGibibytes: 1000
+        level: 1
+        name: volume-name
+        rotational: true
+        numberOfPhysicalDisks: 2
+```
+
+- The RAID level for a volume is required and should be expressed as
+  one of `"1"`, `"0"`, or `"1+0"`.
+- The size for a volume is optional and should be expressed in
+  gibibytes.
+- The number of disks to use for the volume defaults to a minimum
+  value based on the RAID level.
+
+For software RAID, the inputs describe how to configure the operating
+system to see physical disks in a RAID configuration through software
+drivers, without using any special hardware.
+
+```yaml
+spec:
+  raid:
+    softwareVolumes:
+      - sizeGibibytes: 1000
+        level: 1
+        physicalDisks:
+          - deviceName: /dev/sda
+          - deviceName: /dev/sdb
+```
+
+- The RAID level for a volume is required and should be expressed as
+  one of `"1"`, `"0"`, or `"1+0"`.
+- The size for a volume is optional and should be expressed in
+  gibibytes.
+- The disks to use are specified using the same syntax as the
+  `rootDeviceHints`.
+
+Software RAID is constrained in other ways:
+
+- There must be 0, 1, or 2 software RAID devices.
+- The first software RAID volume must use RAID level 1 because
+  otherwise GRUB will need to understand how to assemble the `mdraid`
+  device to be able to boot anything. RAID-1 is safe in this regard
+  because any mirror looks just like a normal disk.
+- The second software RAID volume may use RAID level `0`, `1`, or
+  `1+0`.
+
+Regardless of the type of RAID used, the first volume defined is used
+as the root volume by default. If the host also has `rootDeviceHints`
+configured, those hints override this behavior, and the RAID devices
+are not flagged as the root volume.
+
+The status fields of the `BareMetalHost` expand to include the RAID
+configuration used for the most recent provisioning operation.
+
+```yaml
+status:
+  provisioning:
+    raid:
+      hardwareVolumes: ...
+      softwareVolumes: ...
+```
+
+### Controller Changes
+
+Ironic applies RAID configuration during a cleaning phase, separate
+from the provisioning phase. In order for the metal3 controller to
+track this new phase, a new "preparing" state is introduced to the
+metal3 state machine, between "ready" and "provisioning". The
+controller and provisioner need to handle the new state.
+
+### Implementation Details/Notes/Constraints
+
+If the RAID instructions do not meet the validation criteria given
+above, the host is not provisioned and Ironic is not given any
+cleaning or provisioning instructions. The error is reported through
+the host status, and reconciliation stops until the settings are
+modified again.
+
+Valid RAID settings are stored in the status fields when provisioning
+starts, and only those values are used by the provisioner to ensure
+that if the spec fields are updated after provisioning has started
+Ironic is not given conflicting instructions. This is similar to how
+the root device hints are managed.
+
+### Risks and Mitigations
+
+Reconfiguring the RAID controller can lead to unexpected data loss, so
+during provisioning the existing configuration is only erased when
+explicit settings are given via the API. To manually configure a host,
+the user can pre-configure it and not provide RAID instructions to
+metal3.
+
+Leaving the storage of a host configured after it is deprovisioned can
+result in data exposure, so the host's RAID configuration is always
+erased during deprovisioning.
+
+Not all hosts support hardware RAID configuration. Software RAID is
+available as an option, and if hardware RAID setup fails the host will
+report an error message and fail to provision.
+
+### Work Items
+
+See [the implementation](https://github.com/metal3-io/baremetal-operator/pull/292).
+
+### Dependencies
+
+None
+
+### Test Plan
+
+- Unit tests for determining the instructions to pass to Ironic.
+
+### Upgrade / Downgrade Strategy
+
+Older versions of the baremetal-operator will ignore the RAID fields.
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+We could rely on the user to pre-configure the host's RAID setup, but
+that would move a lot of burden to the user that we could eliminate by
+taking advantage of the existing Ironic feature.
+
+## References
+
+- [original issue](https://github.com/metal3-io/baremetal-operator/issues/206)
+- [implementation](https://github.com/metal3-io/baremetal-operator/pull/292)


dhellmann:
/cc @zaneb @longkb @dtantsur @Hellcatlk @jtwill98

Here is the design writeup based in part on https://github.com/metal3-io/baremetal-operator/pull/292. There are a few differences that need some input, such as the status field handling and error handling. Please take a look.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,200 @@
...
+  `rootDeviceHints`.
+
+Software RAID is constrained in other ways:
+
+- The first volume is always used as the root volume.
+- There must be 0, 1, or 2 software RAID devices.
+- The first software RAID volume must use RAID level 1 to ensure the
+  host boots in the event of a disk failure.
+- The second software RAID volume may use RAID level `0`, `1`, or
+  `1+0`.
```
dhellmann, line 111: I don't see anywhere in the implementation that enforces the rest of these rules, either.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,202 @@
...
+
+*NOTE: Specifying the RAID level for the root volume seems like a
+recommendation we could make but not something we should really be
+requiring. Does something actually enforce that rule?*
+
+Regardless of the type of RAID used, the first volume defined is used
+as the root volume.
```
dhellmann, line 114: How does this rule interact with the root device hints? What if a user specifies both?

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+- The size for a volume is optional and should be expressed in
+  gibibytes.
+- The disks to use are specified using the same syntax as the
+  `rootDeviceHints`.
+
+Software RAID is constrained in other ways:
+
+- There must be 0, 1, or 2 software RAID devices.
+- The first software RAID volume must use RAID level 1 to ensure the
+  host boots in the event of a disk failure.
```
dhellmann, line 109: Specifying the RAID level for the root volume seems like a recommendation we could make (changing "must" to "should") but not something we should really be requiring. Does something actually enforce that rule? I don't see it in the current PR.

Hellcatlk: Sorry, i forgot to check the root volume. Now, I added it.

design/baremetal-operator/raid-api.md
Hellcatlk:
```	// +kubebuilder:validation:Enum="0";"1";"1+0"
	Level string `json:"level" required:"true"`
```
Can it not ensure the rule?

design/baremetal-operator/raid-api.md
Hellcatlk:
```
is_root_volume - Set to true if this is the root volume. At most one logical disk can have this set to true; the other logical disks must have this set to false. The root device hint will be saved, if the RAID interface is capable of retrieving it. This is false by default.
```

design/baremetal-operator/raid-api.md
dhellmann: Yes, the API validation will enforce the enum values.

design/baremetal-operator/raid-api.md
dhellmann: Maybe @dtantsur, @iurygregory, or @elfosardo can help us understand the way root device hints and the root volume flag on the RAID setup interact.

design/baremetal-operator/raid-api.md
dhellmann: I copied this from the comments in the implementation pull request, but I don't think metal3 should be enforcing that rule at all. Is there is a technical reason for it? Would deployment fail if we did not enforce this rule?

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+          - deviceName: /dev/sda
+          - deviceName: /dev/sdb
+```
+
+- The RAID level for a volume is required and should be expressed as
+  one of `"1"`, `"0"`, or `"1+0"`.
+- The size for a volume is optional and should be expressed in
+  gibibytes.
+- The disks to use are specified using the same syntax as the
+  `rootDeviceHints`.
```
zaneb, line 103: This is true, but for now deviceName is the only hint supported.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+
+Reconfiguring the RAID controller can lead to unexpected data loss, so
+during provisioning the existing configuration is only erased when
+explicit settings are given via the API. To manually configure a host,
+the user can pre-configure it and not provide RAID instructions to
+metal3.
+
+Leaving the storage of a host configured after it is deprovisioned can
+result in data exposure, so the host's RAID configuration is always
+erased during deprovisioning.
```
zaneb, line 159: This seems to conflict with the last sentence of the previous paragraph.

I'm also not sure this is true. Does the automated cleaning clean the physical disks, or just the logical ones?

dhellmann: It's meant to be aspirational, so if the code does not do that now it should be changed to do that.

Should we say that the RAID settings are only erased if it looks like the BMO configured RAID on the host in the first place (i.e., there are RAID settings in the status fields)?

zaneb: The only time I am definitely sure that RAID settings are erased are next time you provision.

design/baremetal-operator/raid-api.md
demonCoder95: I'm not sure if this rule exists in Ironic as well, from the docs, it "has to be" RAID-1 in interest of not having a failed boot. and I agree, this could be documented as a "recommendation" rather than a hardcore requirement.

design/baremetal-operator/raid-api.md
demonCoder95: Yes, and ironic enforces this rule in it's implementation of software RAID, so we're good to go.

design/baremetal-operator/raid-api.md
demonCoder95: I'm sure since the ``is_root_volume`` is an optional flag (its default value being false), the following would make the most sense:
1. The first logical disk is root device of the configuration, if there is no ``is_root_device`` flag on any disk.
2. If the ``is_root_device`` is flag set on any logical disk, it would take precedence over statement 1.
3. If the ``is_root_device`` is flag is set on more than 1 logical disks, Ironic would complain, meaning the conductor will fail deployment with an error, moving Node to ``deploy_failed/clean_failed`` state.
4. The drivers can pull out this flag, to implement some wrapping functionality, if desired. Since the flag is saved in the Ironic database.

design/baremetal-operator/raid-api.md
demonCoder95: > Should we say that the RAID settings are only erased if it looks like the BMO configured RAID on the host in the first place (i.e., there are RAID settings in the status fields)?

It would appear to be kind of artificial. The recommended approach would be to rely on Ironic for core functionality like this and only implement wrappers in BMO if it makes sense in a K8s environment. 

From what I know so far, Ironic can remove the ``logical_disks`` only if the ``delete_configuration`` clean/deploy step is used. In other words, it only removes ``logical_disks`` as of now. And I'm unsure if this would go so far as to zap the physical disks as well, I'm guessing it wouldn't, that looks like an added security measure that would take way too long. Imagine zapping 10TBs of spinners, and that's a highly optimistic number.

If we really desire this functionality of complete physical disk erase, Ironic would be a right place to put it, leaving BMO implementation unchanged, which would simply depend on Ironic to clean the physical disks in either the ``delete_configuration`` step or some other mechanism.

design/baremetal-operator/raid-api.md
demonCoder95: This looks like a great starting point. Unless we have a dire need for supporting many other device hints? Do we have some use cases in mind for those?

demonCoder95 (commented):
I think this looks really good. A few ideas to wrap up and we can go ahead with merging this!

Thanks a lot @dhellmann :)

design/baremetal-operator/raid-api.md
Hellcatlk: I did not find any enforcement measures in the source code of ironic:

[raid.py](https://github.com/openstack/ironic/blob/master/ironic/common/raid.py)
```
def _check_and_return_root_volumes(raid_config):
    """Returns root logical disks after validating RAID config.
    This method checks if multiple logical disks had 'is_root_volume'
    set to True and raises an exception if it is True. Otherwise,
    returns the root logical disk mentioned in the RAID config.
    :param raid_config: target RAID configuration or current RAID
        configuration.
    :returns: the dictionary for the root logical disk if it is
        present, otherwise None.
    :raises: InvalidParameterValue, if there were more than one
        root volume specified in the RAID configuration.
    """
    logical_disks = raid_config['logical_disks']
    root_logical_disks = [x for x in logical_disks if x.get('is_root_volume')]
    if len(root_logical_disks) > 1:
        msg = _("Raid config cannot have more than one root volume. "
                "%d root volumes were specified") % len(root_logical_disks)
        raise exception.InvalidParameterValue(msg)

    if root_logical_disks:
        return root_logical_disks[0]

def validate_configuration(raid_config, raid_config_schema):
    """Validates the RAID configuration passed using JSON schema.
    This method validates a RAID configuration against a RAID configuration
    schema.
    :param raid_config: A dictionary containing RAID configuration information
    :param raid_config_schema: A dictionary which is the schema to be used for
        validation.
    :raises: InvalidParameterValue, if validation of the RAID configuration
        fails.
    """
    try:
        jsonschema.validate(raid_config, raid_config_schema)
    except json_schema_exc.ValidationError as e:
        # NOTE: Even though e.message is deprecated in general, it is said
        # in jsonschema documentation to use this still.
        msg = _("RAID config validation error: %s") % e.message
        raise exception.InvalidParameterValue(msg)

    # Check if there are multiple root volumes specified.
    _check_and_return_root_volumes(raid_config)
```
[raid_config_schema.json](https://github.com/openstack/ironic/blob/master/ironic/drivers/raid_config_schema.json)
```

{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "raid configuration json schema",
    "type": "object",
    "properties": {
        "logical_disks": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "raid_level": {
                        "type": "string",
                        "enum": [ "JBOD", "0", "1", "2", "5", "6", "1+0", "5+0", "6+0" ],
                        "description": "RAID level for the logical disk. Valid values are 'JBOD', '0', '1', '2', '5', '6', '1+0', '5+0' and '6+0'. Required."
                    },
                    ......
                    ......
                    ......
                },
                "required": ["raid_level", "size_gb"],
                "additionalProperties": false,
                "dependencies": {
                    "physical_disks": ["controller"]
                }
            },
            "minItems": 1
        }
    },
    "required": ["logical_disks"],
    "additionalProperties": false
}
```

But the description from [docs.openstack.org](https://docs.openstack.org/ironic/latest/admin/raid.html#software-raid) clearly states this:
>Building Linux software RAID in-band (via the Ironic Python Agent ramdisk) is supported starting with the Train release. It is requested by using the agent RAID interface and RAID configuration with all controllers set to software. You can find a software RAID configuration example in Examples for target_raid_config.
>
>There are certain limitations to be aware of:
>- Only the mandatory properties (plus the required controller property) from Target RAID configuration are currently supported.
>- The number of created Software RAID devices must be 1 or 2. If there is only one Software RAID device, it has to be a RAID-1. If there are two, the first one has to be a RAID-1, while the RAID level for the second one can 0, 1, or 1+0. As the first RAID device will be the deployment device, enforcing a RAID-1 reduces the risk of ending up with a non-booting node in case of a disk failure.
>- Building RAID will fail if the target disks are already partitioned. Wipe the disks using e.g. the erase_devices_metadata clean step before building RAID:

demonCoder95: There are use cases where it wouldn't be suggestive to enforce RAID-1, for instance if the node only has 1 disk on it. Additionally, root disks, in most environments these days only carry the operating system, and no persistent data, in the effort to ensure overall system stability in case there is node failure. That will be another reason to not require a RAID-1 enforcement, maybe Ironic does that for a reason.

design/baremetal-operator/raid-api.md
dhellmann: Is the use of names a limitation of Ironic or of the implementation in the baremetal-operator? Perhaps we should just change `physicalDisks` to be a list of device names?

design/baremetal-operator/raid-api.md
dhellmann: @dtantsur do you have any insight into this?

design/baremetal-operator/raid-api.md
dhellmann: If we wait for Ironic to enforce all of the rules, then the user may wait 30 minutes to receive an error message. I think we're going to want to anticipate those sorts of failures and report problems early, before we even attempt to provision.

design/baremetal-operator/raid-api.md
Hellcatlk: I implemented this in `baremetal-operator`, and [modify gophercloud](https://github.com/gophercloud/gophercloud/pull/1982) to support `physicalDisks` in software raid.

design/baremetal-operator/raid-api.md
demonCoder95: And I presume this kube builder verification will be able to point those inconsistencies out? Or what are we missing here?

design/baremetal-operator/raid-api.md
demonCoder95: > Is the use of names a limitation of Ironic or of the implementation in the baremetal-operator? Perhaps we should just change `physicalDisks` to be a list of device names?

I'm fine with replacement with simply disk names, but this discussion is about specific mention of the _root device hints_, which for now, is simply the name of the device, other parameters perhaps could be very particular details like models, vendor names, rotational (boolean) and so on. The whole list is here
https://docs.openstack.org/ironic/pike/install/include/root-device-hints.html

I was just wondering if it's okay to only have a ``deviceName`` as the root device hint. Which I think is sufficient for most use cases. And unless we can figure out another use case where people would like to choose their root devices based on vendor names, or SCSI bus addresses etc?

design/baremetal-operator/raid-api.md
rpioso: The ironic `idrac` hardware type/driver's **hardware** RAID does not enforce `RAID-1` for `logical_disks` which are root volumes.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+        rotational: true
+        numberOfPhysicalDisks: 2
+```
+
+- The RAID level for a volume is required and should be expressed as
+  one of `"1"`, `"0"`, or `"1+0"`.
+- The size for a volume is optional and should be expressed in
+  gibibytes.
+- The number of disks to use for the volume defaults to a minimum
+  value based on the RAID level.
```
rpioso, line 81: Hardware RAID offered by ironic hardware types/drivers may have additional requirements. For example, [here](https://github.com/openstack/ironic/blob/72baa1da581e42042e488abbdb293e5f9f26b9a2/ironic/drivers/modules/drac/raid.py#L1233-L1241) the `idrac` hardware type requires that the physical disks be specified when size is set to `MAX`.

design/baremetal-operator/raid-api.md
rpioso: The `idrac` hardware type/driver is not capable of retrieving the **root device hint**, so it does not save it.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+status:
+  provisioning:
+    raid:
+      hardwareVolumes: ...
+      softwareVolumes: ...
+```
+
+### Controller Changes
+
+Ironic applies RAID configuration during a cleaning phase, separate
```
rpioso, line 129: Has consideration been given to applying the RAID configuration during deployment? Ironic deploy steps facilitate that.

design/baremetal-operator/raid-api.md
rpioso: A few comments, ...

1. Deletion of the existing RAID configuration can be requested during creation. That is expressed by `create_configuration`'s `delete_existing` parameter.
2. The `idrac` hardware type/driver does not zap the physical disks when it deletes the hardware RAID configuration. Other hardware types may behave the same.
3. Secure erase could be a means to quickly erase a large amount of storage.

design/baremetal-operator/raid-api.md
zaneb: No, my comment was not about root device hints, it's about the selection of disks for RAID.
I missed that @Hellcatlk has implemented that other selection criteria in gophercloud, so IIUC this description is now accurate and my comment is obsolete.

design/baremetal-operator/raid-api.md
Hellcatlk: Yes, but not all BMCs require this, and it is difficult to make BMO consistent with "ironic" in a timely manner. So can we let `ironic` check this?

Hellcatlk: Yes, but not all BMCs require this, and it is difficult to make BMO consistent with "ironic" in a timely manner. So can we let ironic check this?

design/baremetal-operator/raid-api.md
Hellcatlk: @dhellmann  If BMC supports RAID, `delete_configuration` step will done every time.

design/baremetal-operator/raid-api.md
demonCoder95: Oh I see. So we can scratch number 4 in case of iDRAC. It appears to be a stretch goal anyway, thanks for pointing it out @rpioso

design/baremetal-operator/raid-api.md
demonCoder95: That sounds good!

design/baremetal-operator/raid-api.md
demonCoder95: I see. That's alright then, no need to discuss the hints.

design/baremetal-operator/raid-api.md
demonCoder95: > 1. Deletion of the existing RAID configuration can be requested during creation. That is expressed by `create_configuration`'s `delete_existing` parameter.

That would eliminate the need to explicitly invoke the ``delete_configuration`` step? am I assuming corrently?

> 2. The `idrac` hardware type/driver does not zap the physical disks when it deletes the hardware RAID configuration. Other hardware types may behave the same.

Just as I thought, makes sense, since it will take wayyyy too long and would not be desired in majority of use cases.

> 3. Secure erase could be a means to quickly erase a large amount of storage.


That could be looked into, as an alternative of disk zap, is we want to have this functionality, it's a good option.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+
+### API Changes
+
+The `BareMetalHost` API will change to accept RAID configuration
+inputs and to reflect the settings that were used when a host is
+provisioned (in case they are changed after provisioning).
+
+New `spec` fields will be added in a section called `raid`, with
+separate fields describing the hardware and software RAID settings. If
+the hardware settings are provided, the software settings are ignored.
```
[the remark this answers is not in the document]
demonCoder95: Then this begs the question, do we want this in BMO or not, and ensure this is implemented accordingly, without _presuming_ Ironic to do anything?

design/baremetal-operator/raid-api.md
demonCoder95: We could add this to our ``idrac`` implementation as a small check in the code. Yes?

design/baremetal-operator/raid-api.md
demonCoder95: Not in this PR, we could extend this implementation for deployment step based configuration as well, if we see sufficient interest in it.

design/baremetal-operator/raid-api.md
rpioso: > > 1. Deletion of the existing RAID configuration can be requested during creation. That is expressed by `create_configuration`'s `delete_existing` parameter.
> 
> That would eliminate the need to explicitly invoke the `delete_configuration` step? am I assuming corrently?
> 
Digging deeper, the `delete_existing` parameter is an optional argument which was added relatively recently. While the `idrac` driver supports it, not all drivers do.

design/baremetal-operator/raid-api.md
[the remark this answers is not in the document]
dhellmann: Yes, good question. This came from the proposed implementation, so I assume it was backed by a limitation of Ironic.

What does it mean to ask for both hardware and software RAID?

design/baremetal-operator/raid-api.md
dhellmann: If we can anticipate these sorts of issues in metal3, that would be good. Otherwise it may take quite a while between the user providing the RAID settings and Ironic reporting an error.

It's also OK, IMHO, for us to limit the configuration options metal3 supports to be a subset of what Ironic supports, if we're focusing on making the hosts suitable to be nodes in a cluster.

So, we could do something like default to listing all of the physical disks when the size is MAX, even if that is not the default behavior of Ironic.

design/baremetal-operator/raid-api.md
dhellmann: Is it safe to remove this requirement, then?

design/baremetal-operator/raid-api.md
dhellmann: Perhaps I misunderstood the implementation. I thought the cleaning steps did happen before provisioning. If not, then I think we need to change the implementation so that's where the setup occurs.

design/baremetal-operator/raid-api.md
dhellmann: > The only time I am definitely sure that RAID settings are erased are next time you provision.

The Ironic state machine can move from deleting to cleaning before going back to available or manageable. So, I think we can make Ironic wipe the RAID settings when it deprovisions the host.

dhellmann: > A few comments, ...
> 
> 1. Deletion of the existing RAID configuration can be requested during creation. That is expressed by `create_configuration`'s `delete_existing` parameter.

Yes, the implementation will always start by deleting what is there. I think we should also delete those settings when deprovisioning a host.

> 2. The `idrac` hardware type/driver does not zap the physical disks when it deletes the hardware RAID configuration. Other hardware types may behave the same.
> 3. Secure erase could be a means to quickly erase a large amount of storage.

Sure, we can look at adding a secure wipe option as a separate enhancement.

design/baremetal-operator/raid-api.md
Hellcatlk: In the current implementation, the `cleaning steps` did before `provisioning`.

design/baremetal-operator/raid-api.md
Hellcatlk: > > A few comments, ...
> > 
> > 1. Deletion of the existing RAID configuration can be requested during creation. That is expressed by `create_configuration`'s `delete_existing` parameter.
> 
> Yes, the implementation will always start by deleting what is there. I think we should also delete those settings when deprovisioning a host.

Do we still need to execute `delete_configuration` in `preparing`, if we execute `delete_configuration` in `deprovisioning`?

design/baremetal-operator/raid-api.md
demonCoder95: I believe so, since it isn't a hard requirement for every driver. It could either change into a ``note`` perhaps? Saying something like "Check if your driver enforces RAID-1 for the first volume" or something akin to it.

design/baremetal-operator/raid-api.md
demonCoder95: I believe so. Since the node being provisioned for the first time will still need the ``delete_configuration`` in the ``preparing`` stage.

design/baremetal-operator/raid-api.md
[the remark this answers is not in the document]
demonCoder95: I believe this question is easier answered if we think of it as ``is there a _desired_ situation/use-case where we'd like to have both enabled at the same time``? If this isn't the case. It's safe to enforce this, to avoid ambiguous behavior. 

Whether a limitation from Ironic enforces this will be secondary if we don't need this in BMO.

design/baremetal-operator/raid-api.md
demonCoder95: That's actually a very reasonable and good approach, since after all we're breaking fresh ground here, implementing RAID support. We can always extend this to encase all of what Ironic can offer us.

design/baremetal-operator/raid-api.md
demonCoder95: > Perhaps I misunderstood the implementation. I thought the cleaning steps did happen before provisioning. If not, then I think we need to change the implementation so that's where the setup occurs.

Like @Hellcatlk has mentioned, It _is_ happening before provisioning. 

What @rpioso meant is, there are two mechanisms for delivering configurations to a node. One is as a cleaning step (performed earlier) and the other is a deployment step (performed later) in the "ironic node life-cycle". 

The latter can be added as an addendum to this later, if there is desire for it in the BMO. AFAIK, there is no objective preference between the two, it's a matter of choice.

design/baremetal-operator/raid-api.md
dhellmann: One of the goals of metal3 is that users shouldn't need to go read the Ironic documentation in order to understand how to use it. So, if there is some driver that does require RAID-1 for root volumes we should document that in the metal3 docs as well. Do we think there is a driver with that requirement? Or are we just not sure that there isn't a driver like that?

Does the difference in driver behavior within Ironic mean that we need some driver-specific validation in metal3?

design/baremetal-operator/raid-api.md
dhellmann: The kubebuilder validations should be good enough. I just wanted to point out that we need to do our own validation as much as possible.

design/baremetal-operator/raid-api.md
dhellmann: Instead of using a flag, could we say that the first item in the array will be set as the root volume and if RAID is used then any root device hints are ignored?

design/baremetal-operator/raid-api.md
dhellmann: Yes, that's right. We have to clean the host before we provision because we don't know what state it is in and we need to reset it. We have to clean the host when we deprovision it because we want to leave it in a blank state for the next use.

design/baremetal-operator/raid-api.md
dhellmann: It sounds like the current implementation is OK, then.

design/baremetal-operator/raid-api.md
demonCoder95: I see. I'm sure we're doing some validation in the implementation as well.

design/baremetal-operator/raid-api.md
demonCoder95: Well, the documentation in Ironic says that it _should_ be RAID-1 (which looks like a recommendation to me). And @rpioso just pointed out that it is not _necessary_ in case of ``idrac``. This appears to me more like a recommendation/usage intricacy than a hardline that needs to be validated in code. 

I agree with the part about documenting this specifically in metal3, one way or the other.

design/baremetal-operator/raid-api.md
demonCoder95: AFAIK, the implementation always does a ``delete_configuration`` before ``apply_configuration``. And I'm sure it would also be doing a ``delete_configuration`` in deprovisioning. So we're covered here.

design/baremetal-operator/raid-api.md
demonCoder95: Yes, it is.

design/baremetal-operator/raid-api.md
demonCoder95: I agree with the second part of the statement, it will make sense since RAID will be abstracting the hardware and presenting logical volumes for consumption as root volumes. 

About the first part of the statement, what's the significance of asserting it? Will it somehow simplify our explanation? Make the implementation more effective? I'm unclear. 
I assume we want to abstract away Ironic-land things, and in that regard, getting rid of an Ironic-land flag and using the BMO-land array references would help us in that endeavor? Is that what we're looking for?

dhellmann: Yes, I'm looking for ways to make the metal3 API simple to understand. I don't want to over-simplify, though. I'm not sure why someone would, for example, define 3 RAID volumes and say the 2nd one should be the root volume for the OS. Is that a real use case? Is it a use case we care about for provisioning nodes?

demonCoder95: I agree. It's stretching it too far to presume it to be a real use case. I'm inclined to believe that to ignore the ``root_device_hints`` in case of RAID is good to go for us. I'd say what's in the implementation at this point should be good enough, unless the implementation processes the ``root_device_hints`` in case of RAID (which I doubt it does).

design/baremetal-operator/raid-api.md
zaneb: > > The only time I am definitely sure that RAID settings are erased are next time you provision.
> 
> The Ironic state machine can move from deleting to cleaning before going back to available or manageable. So, I think we can make Ironic wipe the RAID settings when it deprovisions the host.

Yes, we could, by doing manual cleaning (although this would be a giant pain because we'd probably have to add a 'Cleaning' provisioning state after Deprovisioning). But currently we only do automatic cleaning, and its not clear to me that that will clean the physical disks (and thus wipe the RAID config off them) and not just the contents of the RAID volumes.

Hellcatlk: The node will only be actually used in the "active" state, so we only need to execute `delete_configuration` clean step by `automated cleaning` in `deleting` state to avoid data exposure,  right?

design/baremetal-operator/raid-api.md
Hellcatlk: I think `software raid` is supported by `ironic_python_agent`, and found the requirements for `software raid`.

[hardware.py](https://github.com/openstack/ironic-python-agent/blob/ddbba070216233e695f90095334ab7ddaef563f2/ironic_python_agent/hardware.py#L1815-L1901)

```python
    def validate_configuration(self, raid_config, node):
        """Validate a (software) RAID configuration
        Validate a given raid_config, in particular with respect to
        the limitations of the current implementation of software
        RAID support.
        :param raid_config: The current RAID configuration in the usual format.
        """
        LOG.debug("Validating Software RAID config: {}".format(raid_config))

        if not raid_config:
            LOG.error("No RAID config passed")
            return False

        logical_disks = raid_config.get('logical_disks')
        if not logical_disks:
            msg = "RAID config contains no logical disks"
            raise errors.SoftwareRAIDError(msg)

        raid_errors = []

        # Only one or two RAID devices are supported for now.
        if len(logical_disks) not in [1, 2]:
            msg = ("Software RAID configuration requires one or "
                   "two logical disks")
            raid_errors.append(msg)

        # All disks need to be flagged for Software RAID
        for logical_disk in logical_disks:
            if logical_disk.get('controller') != 'software':
                msg = ("Software RAID configuration requires all logical "
                       "disks to have 'controller'='software'")
                raid_errors.append(msg)

            physical_disks = logical_disk.get('physical_disks')
            if physical_disks is not None:
                if (not isinstance(physical_disks, list)
                        or len(physical_disks) < 2):
                    msg = ("The physical_disks parameter for software RAID "
                           "must be a list with at least 2 items, each "
                           "specifying a disk in the device hints format")
                    raid_errors.append(msg)
                if any(not isinstance(item, dict) for item in physical_disks):
                    msg = ("The physical_disks parameter for software RAID "
                           "must be a list of device hints (dictionaries)")
                    raid_errors.append(msg)

        # The first RAID device needs to be RAID-1.
        if logical_disks[0]['raid_level'] != '1':
            msg = ("Software RAID Configuration requires RAID-1 for the "
                   "first logical disk")
            raid_errors.append(msg)

        # Additional checks when we have two RAID devices.
        if len(logical_disks) == 2:
            size1 = logical_disks[0]['size_gb']
            size2 = logical_disks[1]['size_gb']

            # Only one logical disk is allowed to span the whole device.
            if size1 == 'MAX' and size2 == 'MAX':
                msg = ("Software RAID can have only one RAID device with "
                       "size 'MAX'")
                raid_errors.append(msg)

            # Check the accepted RAID levels.
            current_level = logical_disks[1]['raid_level']
            if current_level not in SUPPORTED_SOFTWARE_RAID_LEVELS:
                msg = ("Software RAID configuration does not support "
                       "RAID level %s" % current_level)
                raid_errors.append(msg)
            physical_device_count = len(self.list_block_devices())
            if current_level == '5' and physical_device_count < 3:
                msg = ("Software RAID configuration is not possible for "
                       "RAID level 5 with only %s block devices found."
                       % physical_device_count)
                raid_errors.append(msg)
            if current_level == '6' and physical_device_count < 4:
                msg = ("Software RAID configuration is not possible for "
                       "RAID level 6 with only %s block devices found."
                       % physical_device_count)
                raid_errors.append(msg)
        if raid_errors:
            error = ('Could not validate Software RAID config for %(node)s: '
                     '%(errors)s') % {'node': node['uuid'],
                                      'errors': '; '.join(raid_errors)}
            raise errors.SoftwareRAIDError(error)
```

demonCoder95: It does say that "every logical disk needs 'software_raid' controller", but does that mean _all_ the disks? or _all the software RAID_ disks? We're trying to determine what happens if both RAIDs are given.

design/baremetal-operator/raid-api.md
demonCoder95: We also need to leave the node "cleaned" after deleting it. So we will need to remove all RAID volumes/disks (hardware/software) when node is deleted.

demonCoder95: And AFAIK, ``automated_cleaning`` doesn't do that for us.

design/baremetal-operator/raid-api.md
Hellcatlk: I think that mean all the disks, beacause the different of `software raid` and `hardware raid` is `controller`,and if only include software RAID disks the check is meaningless.

So i think if we're trying to determine `software raid` and `hardware raid` at the same time, an error will be reported.

demonCoder95: someone from Ironic community told me they can both be configured if "done carefully". That looks to me like a complication, and we should not have this in our implementation. I think it is sufficient to document it as a recommendation for the user to _not_ apply them both at the same time, since it can become complicated.

However, if we decide that we need to restrict users from using one at a time, we can do that as well by reporting an error to user, without passing it onto Ironic. As @dhellmann said, BMO should be considered as something independent of Ironic.

Hellcatlk: I am not sure but if we configure `software raid` in a `manual clean` and configure `hardware raid` in another `manual clean`, is it possible to configure software RAID and hardware RAID on a same node?

design/baremetal-operator/raid-api.md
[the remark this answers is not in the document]
dtantsur: I think wanting both hardware and software RAID is a pretty exotic case that we don't want to support (just imagine recovering failed RAID in this case).

design/baremetal-operator/raid-api.md
dtantsur: The root disk must be RAID-1 because otherwise Grub will need to understand how to assemble mdraid to be able to boot anything. RAID-1 is safe in this regard because any mirror looks just like a normal disk.

design/baremetal-operator/raid-api.md
dtantsur: `is_root_volume` makes ironic automatically overwrite root device hints with the created volume. If we don't use this flag, we need to set root device hints to point at the desired volume, otherwise ironic will apply its usual logic. In any case, I think "root volume is the first volume and must use RAID-1" is a sane thing to assert on our level. Will save us a lot of headache.

design/baremetal-operator/raid-api.md
dtantsur: The benefit of using deploy steps is that we don't need new states on the BMO side, everything happens during "Provisioning"

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+    raid:
+      hardwareVolumes: ...
+      softwareVolumes: ...
+```
+
+### Controller Changes
+
+Ironic applies RAID configuration during a cleaning phase, separate
+from the provisioning phase. In order for the metal3 controller to
+track this new phase, a new "preparing" state is introduced to the
```
dtantsur, line 131: Yep, exactly this thing is not needed if we use deploy steps.

design/baremetal-operator/raid-api.md
dtantsur: I'd be open to an ironic change to allow `delete_configuration` during automated cleaning.

design/baremetal-operator/raid-api.md
demonCoder95: I agree with @dtantsur. Let's stick to this.

design/baremetal-operator/raid-api.md
demonCoder95: @Hellcatlk Let's say for now, we aren't going to allow this, since it will lead to unnecessary complications. Let's "discourage" this in code for now, and document this. If a need comes up later to have both, a patch can be developed accordingly.

@dtantsur I see. In that case, we can stick to RAID-1 on the root disk.

design/baremetal-operator/raid-api.md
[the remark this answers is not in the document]
demonCoder95: I agree. I suggest we "discourage" this and leave it for the situation that (hopefully never) arises.

design/baremetal-operator/raid-api.md
demonCoder95: I see. We can have this enhancement later then. Let's leave the implementation on ``clean_step`` for now.

demonCoder95:
Thanks for the review @dtantsur . @dhellmann what do you think is left for us to do now? I believe all the discussions have been concluded?

I believe we can "revise" the spec to add the discussions into it, and move onto conformance of this with BMO#292?

design/baremetal-operator/raid-api.md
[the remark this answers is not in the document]
dhellmann: OK, I think we agree on the limitation as stated in the doc so I am going to resolve this thread.

design/baremetal-operator/raid-api.md
dhellmann: @dtantsur That's an exceptionally clear explanation for the requirement. I hope that's in some documentation somewhere. :-)

I will add that detail to this design doc before we merge it.

design/baremetal-operator/raid-api.md
dhellmann: I think we're saying that metal3 should set the `is_root_volume` flag on the first volume and not give the user an option to set it themselves on other volumes. I will update this part of the doc with enough detail to make that clear.

design/baremetal-operator/raid-api.md
dhellmann: If we have a way to implement this without adding more states to the metal3 state machine, that seems like a better approach. Are there any down sides?

dtantsur: I seem to vaguely recall that iDRAC RAID configuration had problems when used via a deploy step. I'll let @rpioso comment on that.

Also using deploy steps is a bit inconvenient right now (see https://opendev.org/openstack/ironic-tempest-plugin/src/branch/master/ironic_tempest_plugin/tests/scenario/baremetal_standalone_manager.py#L577-L595 for an example), but we're planning on fixing it in ironic (tracked in https://storyboard.openstack.org/#!/story/2008043).

Hellcatlk: Currently, does `gophercloud` support `deploy step`? I did not find it in `gophercloud`.

design/baremetal-operator/raid-api.md
demonCoder95: Makes sense. Let's make things easier for the user.

design/baremetal-operator/raid-api.md
demonCoder95: I doubt it would. But as @dtantsur says, the deploy steps could cause issues, so it's safe to stick to clean steps for now.

Hellcatlk:
@dhellmann Is there anything else that blocked this PR?

design/baremetal-operator/raid-api.md
namnx228: The idea that the first RAID volume is always the root device, and it only uses RAID 1 provides a lot of convenience to the user. I just have one concern that what if we want the root device is a disk that is not a RAID disk? What if we only want to have RAID disks to store workload? How do you think @dhellmann @demonCoder95 @dtantsur ?

demonCoder95: For that we have what we call the [``root device hints``](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#rootdevicehints) which are a separate part of the BMH specification. 
If we only want RAID volumes for workload, I'm assuming we'll use ``root device hints`` to choose a physical disk as a root disk, and leave the rest up to RAID to consume to create RAID volumes for the workloads.

namnx228: @demonCoder95 but as discussed above, I can see that we have a plan to ignore `rootDeviceHint` if RAID volume is used. 
>Instead of using a flag, could we say that the first item in the array will be set as the root volume and if RAID is used then any root device hints are ignored?

For the use cased that I mentioned, I believe we should choose the first RAID volume as the root device when `rootDeviceHint` is not set. Otherwise, we should apply the hint to choose the root device.

jtwill98:
With this looking complete (no new comments for 12 days), what's it going to take to close this issue and the corresponding 292 issue.  Are there some concerns that have not been addressed?

Hellcatlk:
1. Software RAID
    - The first volume is root volume and must use RAID-1(as same as the implementation).
    - Add more validation.

2. How to configure RAID
    - `Deploy Step` is better than `Clean Step(manual clean)`, but there are some problems, so we don’t need to use `Deploy Step` immediately.

3. `is_root_device` flag and `root_device_hints`

    I see the two options:
    - If RAID is used, the` root device` is the first first volume(as same as the implementation).
    - If `root_device_hints` is nil and RAID is used, the` root device` is the first first volume.

4. After deprovisioning
    - Also need to delete RAID configure(modify `ironic` to allow delete_configuration during automated cleaning).

The above is where I see that the implementation needs to be modified. Is there anything missing or wrong? @dhellmann

namnx228:
> `is_root_device` flag and `root_device_hints`
> 
> I see the two options:
> 
>     * If RAID is used, the` root device` is the first first volume(as same as the implementation).
> 
>     * If `root_device_hints` is nil and RAID is used, the` root device` is the first first volume(this will affect the first point).

I suppose it can be a good idea to follow the option 2 because there can be the case that we want to have the root volume in a normal disk, and apply RAID only to disks storing workload. WDYT @dhellmann ?

design/baremetal-operator/raid-api.md
demonCoder95: I think the discussion was not about ``rootDeviceHints`` as @zaneb mentioned earlier. It will make complete sense to me, if ``rootDeviceHints`` were used to override the default behavior of "always using first RAID volume as root disk".

digambar15 (approved):
Looks good!

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,198 @@
...
+This design describes changes needed to allow RAID storage
+configuration as part of provisioning hosts.
+
+## Motivation
+
+While it is reasonable to assume that cloud instances and VMs only
+need one drive with minimal configuration, many classes of server
+hardware need to have RAID storage configured to present fast and
+reliable storage to the OS. While it is possible today to
+pre-configure the hosts, it would be more convenient for users is
```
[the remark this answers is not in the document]
dhellmann: Fixed in the next draft.

design/baremetal-operator/raid-api.md
dhellmann: The next draft will say that the default is to use the first RAID volume as the root device unless rootDeviceHints are set.

design/baremetal-operator/raid-api.md
dhellmann: @dtantsur @zaneb Could we use manual cleaning during deprovisioning? Or should we just say that the RAID configuration is left intact after deprovisioning, for now?

dhellmann:
I have updated the design based on earlier review comments. There is one remaining open question.

design/baremetal-operator/raid-api.md
```diff
@@ -0,0 +1,202 @@
...
+- The size for a volume is optional and should be expressed in
+  gibibytes.
+- The disks to use are specified using the same syntax as the
+  `rootDeviceHints`.
+
+Software RAID is constrained in other ways:
+
+- There must be 0, 1, or 2 software RAID devices.
+- The first software RAID volume must use RAID level 1 because
+  otherwise Grub will need to understand how to assemble the `mdraid`
```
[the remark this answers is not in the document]
dhellmann: Thanks, fixed.

design/baremetal-operator/raid-api.md
Hellcatlk: @dhellmann If we use `manual cleaning` during deprovisioning, we need to add  a new state. Will this be allowed?

maelk:
looks good to me

namnx228:
That looks really great to me. Thanks for the update and for this design.

namnx228 (approved):

design/baremetal-operator/raid-api.md
dhellmann: @Hellcatlk if we need a new state to keep track of what is happening, then yes, I would have no problem accepting it. We should look at how much work it would be to do what @dtantsur suggested, and allow `delete_configuration` during automated cleaning. That seems like it may be a better solution overall.

Hellcatlk: I see, thank you.

demonCoder95: @dhellmann  I suggest, we can leave it intact for now and work on an incremental change later, in interest of time. Let's finish this and quickly merge the implementation PR as well, we have a couple more PRs coming up afterwards that will need review, and it is going to take a substantial amount of time. Really appreciate the expedition.

Hellcatlk: @dtantsur PTAL

maelk:
Should we put a lazy consensus on this one ?

dhellmann:
> Should we put a lazy consensus on this one ?

Yes, I think so.

I haven't looked in a while, but I think the current implementation differs from this design in a few important ways. I don't expect the design to change substantially, so we could also start updating that code.

demonCoder95:
> > Should we put a lazy consensus on this one ?
> 
> Yes, I think so.
> 
> I haven't looked in a while, but I think the current implementation differs from this design in a few important ways. I don't expect the design to change substantially, so we could also start updating that code.

@Hellcatlk Can you please look where the design differs from the implementation and harmonize the two? We have some patches pending on this work and we would really like to speed up things. Thank you!

Hellcatlk:
> > > Should we put a lazy consensus on this one ?
> > 
> > 
> > Yes, I think so.
> > I haven't looked in a while, but I think the current implementation differs from this design in a few important ways. I don't expect the design to change substantially, so we could also start updating that code.
> 
> @Hellcatlk Can you please look where the design differs from the implementation and harmonize the two? We have some patches pending on this work and we would really like to speed up things. Thank you!

@dhellmann @demonCoder95 The implementation has been modified according to this document, except for the part after deprovisioning.

Hellcatlk:
> > Should we put a lazy consensus on this one ?
> 
> Yes, I think so.
> 
> I haven't looked in a while, but I think the current implementation differs from this design in a few important ways. I don't expect the design to change substantially, so we could also start updating that code.

Do we have a dead line about the ` lazy consensus `?

demonCoder95:
> > > > Should we put a lazy consensus on this one ?
> > > 
> > > 
> > > Yes, I think so.
> > > I haven't looked in a while, but I think the current implementation differs from this design in a few important ways. I don't expect the design to change substantially, so we could also start updating that code.
> > 
> > 
> > @Hellcatlk Can you please look where the design differs from the implementation and harmonize the two? We have some patches pending on this work and we would really like to speed up things. Thank you!
> 
> @dhellmann @demonCoder95 The implementation has been modified according to this document, except for the part after deprovisioning.

That seems to me like we're done. Since the part _after deprovisioning_ is what we need to discuss as a future improvement (adding ``delete_configuration`` as part of automated cleaning). @dhellmann Let's review these quickly and get them merged.

dhellmann:
Today in the community meeting we agreed to use lazy consensus to approve this 1 week from today. So, unless there are significant changes requested by 4 Nov we can merge this then and go ahead with updating the implementation. Thanks for all of the reviews and assistance with getting this proposal into good shape!

demonCoder95:
That sounds great. Thank you everyone! And thanks @dhellmann :)

design/baremetal-operator/raid-api.md
zaneb: I believe @dtantsur is referring to his proposal #138 to add sub-states within Provisioning. This would allow us to track the progress of cleaning without adding a new provisioning state, and would give more accurate information to the user about what's actually happening. https://github.com/metal3-io/metal3-docs/pull/138#discussion_r504718910 (e.g. we don't actually wait for cleaning to finish before moving from Preparing->Provisioning, so it's actually quite misleading if you care about low-level details for debugging purposes.)
There's nothing to implement in gophercloud, we just need to decide if this is the path forward and add the new sub-state to the CRD.
If we add the Preparing state first then subsequently switching to sub-states is a lot more painful than doing it from the outset.

design/baremetal-operator/raid-api.md
dtantsur: I think https://review.opendev.org/744117 will actually be enough to be able to use raid.delete_configuration during cleaning.

dtantsur (approved):

Hellcatlk:
@dhellmann Could this PR be merged?

PR #139, opened 2020-09-23
design: limit number of hosts simultaneously provisioned

andfasano:
Design proposal to introduce a limit to the number of hosts that the BMO can provision/register simultaneously.

Co-authored-by: Dmitry Tantsur <dtantsur@redhat.com>

### DIFF
--- a/design/baremetal-operator/limit-hosts-provisioning.md
+++ b/design/baremetal-operator/limit-hosts-provisioning.md
@@ -1,0 +1,136 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# limit-hosts-provisioning
+
+## Status
+
+provisional
+
+## Summary
+
+This proposal aims to introduce a limit to the number of hosts that the
+BareMetal Operator can provision or register simultaneously as a safety
+measure to avoid overloading excessively the system
+
+## Motivation
+
+In the current implementation BMO takes care of handling the complete
+provisioning cycle for a bare metal host without considering the pressure
+level applied to the underlying Ironic services (DHCP included).
+A user that desires to scale up several hundreds of nodes in a single step
+could suffer issues produced from the sudden resources depletion
+
+### Goals
+
+1. Avoid any system instability due the excessive number of hosts being
+   provisioned simultaneously
+
+### Non-Goals
+
+1. Optimize the deployment time of large batches of nodes
+2. Allow the user to configure such behavior
+3. Modify the current Ironic service layout
+
+## Proposal
+
+BMO reconciliation loop already checks for Ironic availability before
+processing an update and the request gets rescheduled if Ironic is not ready.
+After that, a new check will verify if Ironic is overloaded or not so that
+the request could be rescheduled or processed as usual.
+
+### Implementation Details/Notes/Constraints
+
+The new check will verify the number of currently hosts being provisioned/
+registered and, based on a configurable threshold (default set to 20), will
+determine if the current update can be processed or needs to be rescheduled.
+The difficult part consists in determining the number of hosts currently being
+provisioned, since Ironic does not offer any way for querying about its
+workload.
+The basic idea is to identify those nodes that already passed DHCP/PXE -
+considering it the most limiting factor for concurrent provisioning - by
+selecting the nodes that are in a waiting state and have
+_driver\_internal\_info_ field set by the agent.
+Following a simplified pseudo-code:
+
+```bash
+provisioning_slots = GetEnv("MAX_CONCURRENT_PROVISIONING_HOSTS", 20)
+
+for each Node
+  if Node.provision_state in
+    ("cleaning", "clean wait", "inspecting",
+    "inspect wait", "deploying", "wait call-back", "rescuing",
+    "rescue wait", "unrescuing", "deleting")
+    &&
+    Node.driver_internal_info.agent_url is present and not empty
+  then
+    provisioning_slots--
+
+if provisioning_slots < 0 then
+  reschedule current update
+
+```
+
+### Risks and Mitigations
+
+- In a worst case scenario we could have a high number of “bad” hosts under
+  provisioning but into an error state not recoverable. If such number is equal
+  to the configured threshold, then the “bad” hosts could prevent other hosts
+  to be correctly provisioned. A fair scheduling approach will be required
+  to ensure that the available provisioning slots will be distributed evenly
+  (the recently added _BMH ErrorCount_ field could be used for that)
+- The operator concurrency level can be configured via the environment
+  variable `BMO_CONCURRENCY` (default value set to 3). If the threads
+  activity is not properly orchestrated, when this variable is set
+  to a value greater than `MAX_CONCURRENT_PROVISIONING_HOSTS` then the operator
+  could perform a number of requests to Ironic greater than the configured
+  throttle limit - when a sufficient high number of hosts are provisioned
+  simultaneously. A minimal approach could consist in at least tracing
+  a log warning message when `BMO_CONCURRENCY` > `MAX_CONCURRENT_PROVISIONING_HOSTS`.
+
+### Work Items
+
+- Define a new configuration variable for the Ironic provisioner based on the
+  environment variable `MAX_CONCURRENT_PROVISIONING_HOSTS`. Default value set
+  to 20.
+- Extend the Ironic provisioner `IsReady` method to include the check described in
+  the Implementation Details section
+- Add a warning log trace if `BMO_CONCURRENCY` > `MAX_CONCURRENT_PROVISIONING_HOSTS`,
+  as described in the Risks and Mitigations section.
+- Unit tests for the above work
+
+### Dependencies
+
+None
+
+### Test Plan
+
+- Unit test
+- metal3-dev-env integration test
+- Stress test with a number of simultaneous provisioning nodes greater than
+  the configured threshold
+
+### Upgrade / Downgrade Strategy
+
+None
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+None
+
+## References
+
+- Ironic state diagram. <https://docs.openstack.org/ironic/latest/_images/states.svg>
+- Ironic states. <https://docs.openstack.org/ironic/latest/contributor/api/ironic.common.states.html>


design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Method 1
+
+Identify those nodes that already passed DHCP/PXE - considering it the most
+limiting factor for concurrent provisioning. The Ironic node’s field
+_driver\_internal\_info_ field can be analyzed for the presence of agent
+heartbeat (ie, for the presence of fields like _agent\_url_). A preliminary
+analysis revealed that this approach can have some drawbacks in case of pod
+restart: as soon as BMO comes back (along with the Ironic services) it will
+find already existing BMH resources as provisioned, and thus they will get
```
dtantsur, line 64: Not sure it's a problem, we don't have to provide any throttling for or anyhow consider hosts that are already provisioned. By the way, normally provisioned hosts also may not have an `agent_url`. You need to limit your algorithm to Ironic nodes in certain states (`*wait*`, `*ing`).

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Method 2
+
+A simpler and more crude way is counting the current number of Ironic
+nodes not in the _active_ state (as per Ironic state
+[diagram](https://docs.openstack.org/ironic/latest/_images/states.svg):
+if the node is not in the _active_ state, then the assumption is that Ironic
+will still have some work to do on it, so the node consumes one provisioning
+slot. If this approach is considered too coarse-grained then a list of precise
+Ironic states could be then identified
```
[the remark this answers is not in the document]
dtantsur: Wait states also happen on asynchronous actions (e.g. while ironic is waiting for the agent to download and write image, which takes minutes). I think the final solution should combine methods 1 and 2.

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,120 @@
...
+
+BMO reconciliation loop already checks for Ironic availability before
+processing an update and the request gets rescheduled if Ironic is not ready.
+After that, a new check will verify if Ironic is overloaded or not so that
+the request could be rescheduled or processed as usual.
+
+### Implementation Details/Notes/Constraints
+
+The new check will verify the number of currently hosts being provisioned/
+registered and, based on a fixed threshold (initial propose is 20), will
```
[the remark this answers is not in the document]
andfasano: How do you suggest to make it configurable? Ie through an environment variable like `DEPLOY_KERNEL_URL`? The value of `20` could be used as a default value.

design/baremetal-operator/limit-hosts-provisioning.md
andfasano: It sounds as a good approach. I'll update the proposal trying to enumerate the correct states.

design/baremetal-operator/limit-hosts-provisioning.md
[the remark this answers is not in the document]
andfasano: Could you please clarify this point about the combination? Do you maybe mean to check first the `driver_internal_info` field and subsequently verify if the node is in one of the selected states (`*wait*, *ing`)?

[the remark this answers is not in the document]
dtantsur: The other way around: check the provision state first (ruling out anything that cannot be PXE booting at this point), then do the heuristics based on driver_internal_info.

design/baremetal-operator/limit-hosts-provisioning.md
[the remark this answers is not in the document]
andfasano: Added env var

design/baremetal-operator/limit-hosts-provisioning.md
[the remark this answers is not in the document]
andfasano: Thanks, I've reviewed the details implementation section accordingly by adding a detailed states list (to be reviewed) and some pseudo-code

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,120 @@
...
+could suffer issues produced from the sudden resources depletion
+
+### Goals
+
+1. Avoid any system instability due the excessive number of hosts being
+   provisioned simultaneously
+
+### Non-Goals
+
+1. Optmize the deployment time of large batches of nodes
```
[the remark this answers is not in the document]
andfasano: Done

design/baremetal-operator/limit-hosts-provisioning.md
andfasano: See previous comment

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,147 @@
...
+selecting the nodes that are in a waiting state and have
+_driver\_internal\_info_ field set by the agent.
+Following a simplified pseudo-code:
+
+```bash
+provisioning_slots = GetEnv("MAX_CONCURRENT_PROVISIONING_HOSTS", 20)
+
+for each Node
+  if Node.provision_state in
+    ("verifying", "adopting", "cleaning", "clean wait", "inspecting",
```
dtantsur, line 65: Let's remove `verifying` and `adopting`, despite their name they're not relevant here.

dtantsur (commented):
One suggestion inline, otherwise looks great.

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,121 @@
...
+1. Optimize the deployment time of large batches of nodes
+2. Allow the user to configure such behavior
+3. Modify the current Ironic service layout
+
+## Proposal
+
+BMO reconciliation loop already checks for Ironic availability before
+processing an update and the request gets rescheduled if Ironic is not ready.
+After that, a new check will verify if Ironic is overloaded or not so that
+the request could be rescheduled or processed as usual.
```
dhellmann, line 44: Does that mean a change in the Provisioner interface? What will that look like?

andfasano: My initial idea was not to change the Provisioner interface, and to embed the implementation of this new check directly within the `IsReady` method: from the point of view of the controller, it could make sense to have the provisioner "not ready" to process the current update (since its workload is too high) and to reschedule the request. In this way, only the `ironic` package will be affected, and the current reconcile loop will remain unchanged.

dhellmann: OK, that makes sense.

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Risks and Mitigations
+
+In a worst case scenario we could have a high number of “bad” hosts under
+provisioning but into an error state not recoverable. If such number is equal
+to the configured threshold, then the “bad” hosts could prevent other hosts
+to be correctly provisioned. A fair scheduling approach will be required
+to ensure that the available provisioning slots will be distributed evenly
+(the recently added _BMH ErrorCount_ field could be used for that)
```
dhellmann, line 85: https://github.com/metal3-io/ironic-image/issues/203 made me think a little more about this proposal to use ironic's state to control the number of simultaneous operations we allow.

There's a hidden assumption in this design that the throttle limit is always going to be higher than the concurrency level of the baremetal-operator. If that is not true, then when we query Ironic and find slots available, we might start too many operations at once because we have too many concurrent threads thinking they have a slot available. We should document this assumption here as another risk. I don't think we should try to protect against it in the source code in any way.

andfasano: That's a good point and I agree it's better to document such assumption, I'll update the proposal. In the code, we could anyhow add a warning log trace when `BMO_CONCURRENCY` is greater than the throttle limit

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,129 @@
...
+  activity is not properly orchestrated, when this variable is set
+  to a value greater than `MAX_CONCURRENT_PROVISIONING_HOSTS` then the operator
+  could perform a number of requests to Ironic greater than the configured
+  throttle limit - when a sufficient high number of hosts are provisioned
+  simultaneously. A minimal approach could consist in at least tracing
+  a log warning message when `BMO_CONCURRENCY` > `MAX_CONCURRENT_PROVISIONING_HOSTS`.
+
+### Work Items
+
+None
```
dhellmann, line 97: I think we have enough detail worked out to expand this section now.

dhellmann:
/approve

I'm happy with this version.

@zaneb @maelk do you have any feedback?

maelk (commented):
lgtm

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,136 @@
...
+  to ensure that the available provisioning slots will be distributed evenly
+  (the recently added _BMH ErrorCount_ field could be used for that)
+- The operator concurrency level can be configured via the environment
+  variable `BMO_CONCURRENCY` (default value set to 3). If the threads
+  activity is not properly orchestrated, when this variable is set
+  to a value greater than `MAX_CONCURRENT_PROVISIONING_HOSTS` then the operator
+  could perform a number of requests to Ironic greater than the configured
+  throttle limit - when a sufficient high number of hosts are provisioned
+  simultaneously. A minimal approach could consist in at least tracing
+  a log warning message when `BMO_CONCURRENCY` > `MAX_CONCURRENT_PROVISIONING_HOSTS`.
```
zaneb, line 93: I don't really think it matters if we go a little over (the number 20 is completely arbitrary anyway). The key is to prevent a thundering herd; if we do that then concurrency issues will be minimal.

zaneb (commented):
I think ultimately we should be asking Ironic to tell us what its capacity is rather than guessing, but I'm not going to let perfect be the enemy of good here.
/approve

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,136 @@
...
+the request could be rescheduled or processed as usual.
+
+### Implementation Details/Notes/Constraints
+
+The new check will verify the number of currently hosts being provisioned/
+registered and, based on a configurable threshold (default set to 20), will
+determine if the current update can be processed or needs to be rescheduled.
+The difficult part consists in determining the number of hosts currently being
+provisioned, since Ironic does not offer any way for querying about its
+workload.
```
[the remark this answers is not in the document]
andfasano: The proposal was meant to address the problem locally, very likely not in the perfect way (it's an heuristic) but with an approach that could be quickly implemented directly by us within a (hopefully) short period cycle. If/when the problem will be addressed by Ironic then this approach could be easily removed if needed. In a long-term strategy, and especially in presence of a multi-conductors layout topology, I'd like Ironic to take care directly of the problem.

maelk:
Let's put a lazy consensus here ?

andfasano:
As discussed during the community meeting we agreed to put a lazy consensus of one week (up to 4 Nov). For that date the proposal could be merged in case no relevant changes will be applied. Thanks everybody for all the useful feedbacks and comments

design/baremetal-operator/limit-hosts-provisioning.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+### Risks and Mitigations
+
+- In a worst case scenario we could have a high number of “bad” hosts under
+  provisioning but into an error state not recoverable. If such number is equal
+  to the configured threshold, then the “bad” hosts could prevent other hosts
+  to be correctly provisioned. A fair scheduling approach will be required
+  to ensure that the available provisioning slots will be distributed evenly
+  (the recently added _BMH ErrorCount_ field could be used for that)
```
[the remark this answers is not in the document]
dhellmann: Should the text say that we'll use the random backoff approach, or is it enough to say "fair"?

design/baremetal-operator/limit-hosts-provisioning.md
dhellmann: Agreed.

dhellmann (approved):

design/baremetal-operator/limit-hosts-provisioning.md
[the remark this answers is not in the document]
andfasano: Th reference to `ErrorCount` was meant for that, but didn't want to restrict too much the possible solutions

maelk:
The lazy consensus is expired, so merging this : 
/lgtm

PR #147, opened 2020-10-30
Support for new parameters in HWCC.

Ashughorla:
Add support for new parameters under CPU ,DISK and NIC. Introducing new parameter Firmware and SystemVendor.
This addresses the [Issue-53](https://github.com/metal3-io/hardware-classification-controller/issues/53)

Kindly review.

@dhellmann @digambar15 @russellb @maelk @jtwill98


design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
+# Add Support for extra hardware parameters in Disk, RAM, CPU and NIC
+ 
+## Table of Contents
+ 
+<!--ts-->
```
dhellmann, line 5: We've mostly stopped using the table of contents in the other design documents. Did we fail to remove it from the template?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+classifies the hosts based on user hardware requirements.
+We would like to add more parameters which helps user to 
+identify the ideal host. This feature helps user to avoid runtime failures
+and also increases the performance for workload deployments.
+
+### Goals
+
+Automatically classify bare metal hosts based on user-provided rules
+in a way that other controllers can use, without having to evaluate
+all of those rules themselves.
```
dhellmann, line 38: This is really the goal of the whole hardware classification controller project, right? Is there a more narrowly defined goal that caused you to choose the specific fields being proposed here?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+
+We will carry testing of HWCC only on Dell hardware. Testing of HWCC on 
+hardware from other vendors will not be goal of this proposal.
+
+## Proposal
+ 
+We are submitting proposal based on the issue raised in HWCC: 
+[Issue 53](https://github.com/metal3-io/hardware-classification-controller/issues/53)
+
+### Investigation Details
```
dhellmann, line 48: It seems like the examples in this section are parts of host resources. It would be helpful to have examples of the new API here as well. The larger example farther down is good, but having a small example for each of the new fields will make it clear how we expect someone to use them.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+   kind: HardwareClassification
+   metadata:
+     name: profile1
+   spec:
+     hardwareCharacteristics:
+       CPU:
+          arch: x86_64
+          minimumCount: 4
+          minimumSpeed: 4.3
+       Disk:
```
dhellmann, line 169: Changing `disk` from a single field to a list like this is not backwards compatible, and that will increase the complexity of the implementation significantly because we will have to provide conversion a webhook and publish 2 versions of the CRD. I suggest instead that we add a new field to describe the combinations of disks. Maybe with a name like `diskPools`?

Is this example showing "and" or "or" rules? Does the host have to have 2 SSD and 1 RAID volume, or will a host with either of those match?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+1. CPU
+
+    `arch` will come under CPU parameter.
+
+    User can provide `arch` type based on requirement.
+
+    Investigation Details: Below is sample introspection data for CPU.
+     `arch` parameter will be added under `cpu` section.
+    Types of architecture are 32-bit (x86) and 64-bit 
+    (x86-64, IA64, and AMD64).
```
dhellmann, line 61: Is this going to be an enum match or a string match? If a string match, will it support substrings?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+
+    User can provide multiple disk type (RAID/HDD/SSD/NVMe), count and
+     individual size based on requirement.
+
+    Investigation Details: Below is sample introspection data for Disk.
+
+     1. HDD - Identified using `rotational` parameter's value as true
+      and `hctl` value as "0:0:N:0".  
+     2. SSD - Identified using `rotational` parameter's value as false
+      and `hctl` value as "0:0:N:0" or "N:0:0:0".
```
dhellmann, line 81: Is the `rotational` parameter not sufficient to determine if it is a spinning disk or not? Why do we care where the device is on the SCSI bus?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+     3. NVMe - Identified with the combination of `model` name and
+      `rotational` parameter's value as false.
+     4. RAID - Identified with the combination of
+
+        a) `rotational` parameter's value as true.
+
+        b) If `hctl` have value as "N:0:0:0" or "0:N:0:0".
```
dhellmann, line 88: Is that `hctl` rule going to be right for all RAID on all platforms?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+        cpu:
+           arch: "x86_64"
+    ```
+
+2. Disk
+
+    `type` will come under disk parameter.
+
+    User can provide multiple disk type (RAID/HDD/SSD/NVMe), count and
+     individual size based on requirement.
```
dhellmann, line 74: Some of the rules described here seem like they may be vendor or even host-specific. It might be better to expose the rotational and hctl fields, and let users define classification rules that interpret the values they find, than for us to hard-code some of those rules into a "type" value ourselves.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+            hctl: "0:0:N:0"
+     ```
+3. NIC
+
+     Adding new parameter `vendorName` under NIC. User can provide NIC
+      requirement in the form of `vendor`.
+
+     Investigation details: Below is sample introspection data for NICs.
+      `model` field under `nics` contains combination of Vendor ID and
+       Product ID.
```
dhellmann, line 105: Both of these are strings. Do we need to support sub-string matching or regular expressions?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+         mac: 24:6e:96:aa:bb:cc
+         model: 0x8086 0x1234
+         name: eth3
+         pxe: false
+     ```
+4. Firmware
+
+    Firmware is new parameter to be introduced in HardwareClassification
+     which contains BIOS information. User can specify BIOS requirements
+     by providing `vendor` and `version` details.
```
dhellmann, line 120: Both of these are strings. Do we need to support sub-string matching or regular expressions?

Do we need a version-type-specific rule? For example, should a version rule of `1.5` match a value of `1.5.6`?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,243 @@
...
+             bios:
+               date: "10/17/2018"
+               vendor: "Dell Inc."
+               version: "1.5.6"
+    ```
+5. SystemVendor 
+
+    SystemVendor is new parameter to be introduced which helps user to 
+    classify hosts of certain manufacturer and product. User needs to 
+    provide `manufacturer` and `productName` in yaml.
```
dhellmann, line 137: Both of these are strings. Do we need to support sub-string matching or regular expressions?

dhellmann (commented):
This looks like it is going to make the classification controller much more useful.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Removed ##Table of Contents and added ##Status as per new template.
Thanks for correction.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Examples for each of new fields are provided in comment section of Issue.
https://github.com/metal3-io/hardware-classification-controller/issues/53#issuecomment-720363210

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Yes, it looks like that. We will update it and make it more specific to this Proposal.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Yes, we will declare that enum in CRD validation.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Yes, we are using sub-string matching for Vendor name.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: We are going to use exact match for string. 
For example, `1.5.6` match with `1.5.6` only.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Yes, we are going to use sub-string match for these parameters.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: good to keep it as a enum then string.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: You means to say, we have to define rotational and hctl field as a type instead of RAID, HDD, SSD, NVMe. or you want to keep both ?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: rorational defines whether it's a HDD or SSD, we need SCSI bus details for finding RAID disks.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: Not sure, we need to run tests for other vendors. Currently we have tested it against only Dell servers.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: It's anyway hardcoded values, Its fine to keep it as a string.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: Fine with bringing version-type-specific rule.

digambar15 (changes requested):
Needs improvement

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: OK, could you copy that information into this document so that someone reviewing it has all of the context in one place?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: I mean to have the fundamental values we can discover as fields, and let the user combine them to create their own rules. So we would only have `rotational` and `hctl` fields (although maybe it would be better to use the same data structure that we use for `rootDeviceHints` on the BareMetalHost).

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: Is there some technical reason why RAID always appears in `0:0:N:0` or `N:0:0:0` positions?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: This is why I think the rules should let the user give an HCTL value in some form and we should not hard-code that logic into the controller.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: Either is fine with me. We should document our decision here.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: We should document those rules here.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: We should document that here.

digambar15 (changes requested):

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
digambar15: yeah. I gave a thought on this. When it comes to multiple disks, current proposed solution may not work fully. 
When we add diskPools for all disk types, problem is how can we satisfy the minimum and maximum requirement for all the expected characteristics. Need to think more on this.

digambar15 (changes requested):

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: Perhaps I misunderstood the intent of listing multiple items here under `disk`, and we just need a better name for the new field?

Ashughorla:
Updated the proposal addressing the review comments.

@dhellmann @maelk @digambar15

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: This comment still needs to be addressed, I think? We cannot change `disk` to a list, so if we need a list we need a new field name and we need to describe which field takes precedence when both are present.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+  user is looking for particular vendor and their hardware for his
+  application to support or deploy on.
+  Firmware will be useful to add as a new parameter if user is looking
+  for specific feature/s in particular releases or versions.
+
+### Investigation Details
+
+Add support for new parameters:
+
+1. CPU
```
dhellmann, line 54:
```suggestion
1. `hardwareCharacteristics.cpu.arch`
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+
+   `arch` will come under CPU parameter.
+
+   User can provide `arch` type based on requirement.
```
dhellmann, line 58: I think you mean here that the field is optional?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+  Firmware will be useful to add as a new parameter if user is looking
+  for specific feature/s in particular releases or versions.
+
+### Investigation Details
+
+Add support for new parameters:
+
+1. CPU
+
+   `arch` will come under CPU parameter.
```
dhellmann, line 56:
```suggestion
The new `arch` field will be compared against the host `status.hardwareDetails.cpu.arch` field for an exact string match.
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+
+   User can provide `arch` type based on requirement.
+
+   Investigation Details: Below is sample introspection data for CPU.
+   `arch` parameter will be added under `cpu` section.
```
dhellmann, line 61: The arch parameter is already part of the data collected during inspection.

```suggestion
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+     hardware:
+       nics:
+       - ip: ""
+         mac: 24:6e:96:aa:bb:cc
+         model: 0x8086 0x1234
+         name: eth3
+         pxe: false
+     ```
+
+1. Firmware
```
dhellmann, line 137:
```suggestion
1. `hardwareCharacteristics.firmware.bios`
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+
+   Firmware is new parameter to be introduced in HardwareClassification
+   which contains BIOS information. User can specify BIOS requirements
+   by providing `vendor` and `version` details.
```
dhellmann, line 141:
```suggestion
   by providing `hardwareCharacteristics.firmware.bios.vendor` and `hardwareCharacteristics.firmware.bios.version`.
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+
+   The classification of `vendor` and `version` will be done using
+   exact match of the values provided by user in profile.
```
dhellmann, line 144:
```suggestion
   exact string match of the values provided by user in profile.
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+    ```yaml
+    hardware:
+        firmware:
+             bios:
+               date: "10/17/2018"
+               vendor: "Dell Inc."
+               version: "1.5.6"
+    ```
+
+1. SystemVendor
```
dhellmann, line 158:
```suggestion
1. `hardwareCharacteristics.systemVendor`
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+
+   SystemVendor is new parameter to be introduced which helps user to
+   classify hosts of certain manufacturer and product. User needs to
+   provide `manufacturer` and `productName` in yaml.
```
dhellmann, line 162:
```suggestion
   provide `hardwareCharacteristics.systemVendor.manufacturer` and `hardwareCharacteristics.systemVendor.productName`.
```

Does the user *need* to provide both values? Or can they provide one or the other?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,268 @@
...
+
+   The classification of `manufacturer` and `productName` will be done
+   using exact match of the values provided by user in profile.
```
dhellmann, line 165:
```suggestion
   using exact string match of the values provided by user in profile.
```

dhellmann (changes requested):
I still think the API change details in the Proposal section are presented in a somewhat confusing way. I have suggested some changes to the CPU architecture, BIOS, and system vendor sections as examples of how I think it could be clarified. The other sections are going to require more significant work, since there are more significant API changes.

In order to help you make some progress, I suggest that we remove the disk and NIC sections from this proposal, complete the work on the CPU, BIOS, and system vendor sections so we can approve this PR, and then use separate PRs to work out the bigger API changes needed for the other areas.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done as suggested.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done as suggested.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done as suggested.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done as suggested.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done as suggested.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Yes, like other field "arch"  field is also optional.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done.

Ashughorla:
As suggested, removed Disk's type and NIC's vendor parameters from this proposal.
@dhellmann

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: Since the disk and NIC sections are removed above, you could remove them from this example, too.

dhellmann (commented):
There's just a little bit of cleanup to make the example match the text, but otherwise I am happy with this version.

Thank you!

Ashughorla:
@dhellmann 
Removed the Disk, NIC and RAM from the example YAML as suggested.
Thank you.

jan-est:
Looks good to me

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,192 @@
...
+
+We will carry out testing of HWCC on Dell hardware.
+
+## Proposal
+
+We are submitting proposal based on the issue raised in HWCC:
+[Issue 56](https://github.com/metal3-io/hardware-classification-controller/issues/56)
+
+* CPU architecture will be useful if the user wants to have 32-bit
+  or 64-bit processor as per his requirements.
```
dtantsur, line 34:
```suggestion
  or 64-bit processor as per their requirements.
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,192 @@
...
+               vendor: "Dell Inc."
+               version: "1.5.6"
+    ```
+
+1. `hardwareCharacteristics.systemVendor`
+
+   SystemVendor is new parameter to be introduced which helps user to
+   classify hosts of certain manufacturer and product. User needs to
+   provide `hardwareCharacteristics.systemVendor.manufacturer` and
+   `hardwareCharacteristics.systemVendor.productName` in yaml.
```
dtantsur, line 97: Is it really useful to require `productName`? Also this duplicates `vendor` in `firmware`, why do we need both?

dtantsur (changes requested):
I'm a bit confusion by duplication of vendors

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
```diff
@@ -0,0 +1,192 @@
...
+   apiVersion: metal3.io.sigs.k8s.io/v1alpha1
+   kind: HardwareClassification
+   metadata:
+     name: profile1
+   spec:
+     hardwareCharacteristics:
+       CPU:
+          arch: x86_64
+          minimumCount: 4
+          minimumSpeed: 4.3
```
dtantsur, line 136: This needs units

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: All the fields are optional.
For example :  if user decide to fetch host only on SystemVendor information. Then there is no need to provide vendor in Firmware..

There is a grammatical mistake. Updating it as "User **can** provide `hardwareCharacteristics.systemVendor.manufacturer` and
   `hardwareCharacteristics.systemVendor.productName` in yaml".

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Done.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Units will be updated in User Guide under [User guide](https://github.com/metal3-io/hardware-classification-controller/blob/master/docs/user-guide.md)

digambar15 (approved):
This approach looks good to me.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: This phrasing may be a little more clear:

```suggestion
   User can optionally provide `arch` type.
```

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: The host API has both of these fields as separate, so it makes sense to do the same here. I thought that the host schema was designed to match the inspection data. Does the inspection data not have those as separate fields?

Is the firmware vendor always the same as the system vendor?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: It would be better if the field names included the units, so that users don't have to guess or remember or look it up every time they want to use it. We also don't want `float` fields because some values don't pass cleanly from text to `float64`, so it'll need to be `minimumSpeedMHz` defined as an `int` or `int32` and the value in the example would become `4300`.

Ashughorla: Yes, we will append units as per the field names.

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
Ashughorla: Removing Vendor from Bios because value under `systemVendor.Manufacturer` and `Bios.Vendor` are same `Dell Inc.` for dell hardware.

fmuyassarov:
@dtantsur @digambar15 @dhellmann ,
If there are no further comments to be addressed from you, can we maybe set the lazy consensus on the proposal?

design/hardware-classification-controller/support-for-new-parameters-hwcc.md
dhellmann: > Removing Vendor from Bios because value under `systemVendor.Manufacturer` and `Bios.Vendor` are same `Dell Inc.` for dell hardware.

We support more than Dell hardware, though, right? Are those fields always the same for all vendors?

We originally used separate fields in the host API because Ironic reports the values separately. They are semantically different, so even if the values are the same I think it is useful to keep both fields.

Ashughorla: Yes, it is possible those fields may differ for other vendors.
Reverting the change.

dhellmann:
I'm happy with this draft.

/approve

digambar15:
> @dtantsur @digambar15 @dhellmann ,
> If there are no further comments to be addressed from you, can we maybe set the lazy consensus on the proposal?

YEs. I am fine with the current design.

dhellmann:
This has been open for a week since the last approvals, so I think we can go ahead and merge it.

/lgtm

Issue #146, opened 2020-10-27
Sync a set of labels placed on a BMH to the corresponding K Node object


Arvinderpal:
There are use cases where certain information is inherently tied to the BMH but at the same time, is valuable to the users/operators/schedulers in the K workload cluster. For example:

1. As a user, I would like to place my workloads across hosts that are in different failure zones. For example, I may want replicas of a specific workload spread out across different server racks or geographical locations. 

2. As a user, I would like to place my security sensitive workloads on machines that meet specific security requirements. For example, certain hosts may be in a fortified zone within a data center or certain hosts may have strong hardware based security mechanisms in place (e.g. hardware attestation via TPM). 

In Kubernetes, labels on Node objects are the primary means by which to solve this problem. The ask here is for CAPM3 to synchronize a specific set of labels placed on a BMH object with labels on the corresponding K Node running on that BMH. CAPM3 is already capable of mapping BMH<->Node, so a controller that keeps the labels in sync may be a straightforward addition. The synchronization would be limited to only a specific set of labels matching a certain prefix. For example, the user may specify `my-prefix.metal3.io/` as their prefix (e.g. via a command-line flag). Labels placed on the BMH that match this prefix would be synchronized with the labels on the K Node object. For example, 
```
kind: BareMetalHost
name: node-0
metadata:
  labels:
    my-prefix.metal3.io/rack: xyz-123
    my-prefix.metal3.io/zone: security-level-0
...
---
kind: Node
name: worker-node-0 
metadata:
  labels:
    my-prefix.metal3.io/rack: xyz-123
    my-prefix.metal3.io/zone: security-level-0
...
```

Proposal doc: https://docs.google.com/document/d/1qMCkggaLGQLHNPnEVGYjpabSrO-DTvSsr8uIW06W0ck/edit?usp=sharing 

Related:
There is a related issue in the CAPI community linked below. The proposal there is for CAPI to synchronize labels placed on MachineDeployment objects with the K Nodes created from that deployment. While similar, they are addressing different things. However, the proposed CAPI approach also uses prefixes to limit the scope of the synchronization.

CAPI: [Support syncing a set of labels from MachineDeployment/MachineSet/Machine to Nodes](https://github.com/kubernetes-sigs/cluster-api/issues/493)

Arvinderpal:
@maelk @dhellmann PTAL.

dhellmann:
[...]

PR #149, opened 2020-11-09
Add design proposal for label sync mechanism between BMHs and K Nodes

Arvinderpal:
Issue: https://github.com/metal3-io/metal3-docs/issues/146

### DIFF
--- a/design/sync-labels-bmh-to-node.md
+++ b/design/sync-labels-bmh-to-node.md
@@ -1,0 +1,373 @@
+# Synchronize Labels between BaremetalHosts and Kubernetes Nodes
+
+There are use cases where certain information is inherently tied to the BMH but
+at the same time, is valuable to the users/operators/schedulers in the K
+workload cluster. For example, a user may wish to place her workloads across
+hosts that are in different failure zones. In Kubernetes, labels on Node objects
+are the primary means by which to solve this problem. The proposal is for CAPM3
+to synchronize a specific set of labels placed on a BMH with labels on the
+corresponding Kubernetes Node running on that BMH.
+
+## User Stories
+
+As a user I want replicas of a specific workload spread out across different
+server racks or geographical locations.
+
+As a user, I would like to place my security sensitive workloads on machines
+that meet specific security requirements. For example, certain hosts may be
+in a fortified zone within a data center or certain hosts may have strong
+hardware based security mechanisms in place (e.g. hardware attestation via TPM).
+
+## Goals
+
+* Implement a label synchronization mechanism between BMH and Node objects
+* Limit the scope of synchronization to only labels matching a configured set
+  of prefixes
+* Support a wide range of label prefixes including those in *.kubernetes.io/*
+
+## Non-Goals
+
+* Synchronize taints or annotation on a BMH with those on the Node
+
+## Proposal
+
+CAPM3 can already map a BMH object to the corresponding Node object. The
+proposal seeks to extend CAPM3 to also perform label synchronization
+between the two.
+
+Synchronization is not just a one time copy from BMH to Node on creation.
+More specifically, synchronization accomplishes the following:
+
+1. Addition/Removal of a label, within the set of predefined prefixes, on the
+  BMH will result in the addition/removal of that label on the corresponding
+  Node. Labels outside the prefix set will be ignored.
+2. Addition/Removal of a label, within the set of predefined prefixes,
+  _directly_ on the Node will result in the removal/re-adding of that label
+  on the Node.
+
+The synchronization should be limited to only labels matching a certain set of
+prefixes. The primary reason for this limitation is to avoid stepping on labels
+managed by other entities. Here CAPM3 would own the set of labels that match the
+prefixes. For example, the user may specify _my-prefix.foobar.io/_ as their
+prefix. Labels placed on the BMH that match this prefix would be synchronized
+with the labels on the Node object. For example,
+
+```yaml
+    kind: BareMetalHost
+    name: node-0
+    metadata:
+      labels:
+        my-prefix.foobar.io/rack: xyz-123
+        my-prefix.foobar.io/zone: security-level-0
+        some-other-prefix.blah.io/cow: moo
+    ...
+
+    ---
+    kind: Node
+    name: worker-node-0
+    metadata:
+      labels:
+        my-prefix.foobar.io/rack: xyz-123
+        my-prefix.foobar.io/zone: security-level-0
+    …
+
+```
+
+It's assumed that labels beginning with the specified prefix(es) are owned by
+CAPM3 and CAPM3 is solely responsible for keeping the set of labels in sync
+between BMH and Node objects. In other words, for the given set of label
+prefixes, the BMH is the source of truth and CAPM3 will enforce their
+presence on the Node.
+
+Additionally, there are certain well-known prefixes (e.g.
+kubelet.kubernetes.io/*, beta.kubernetes.io/*, kubernetes.io/*, k8s.io/*) that
+should be disallowed. Note the distinction between *.kubernetes.io/* (allowed)
+versus kubernetes.io/* (disallowed).
+
+### Synchronization Approach
+
+Synchronization must handle two scenarios: (i) a label is added/removed on a
+BMH and must be added/removed on the corresponding Node and (ii) a label is
+added/removed on the Node and must be removed/re-added on the Node to bring it
+in sync with the labels on BMH.
+
+The actual implementation of the synchronization logic may be encapsulated
+into an additional controller running within CAPM3; however, in the discussion
+below we simply reference CAPM3 as the entity doing the work.
+
+Scenario (i) can be handled as follows:
+
+1. CAPM3 will set up a watch on the BMH resource. It only requires read-only
+  access similar to what is done today:
+
+    baremetalhosts,verbs=get;list;watch
+
+2. On each BMH reconcile event, the controller will fetch the corresponding
+  Node. The _mapping_ can be accomplished by:
+
+    BMH.ConsumerRef --> Metal3Machine.OwnerRef --> Machine.Status.NodeRef
+
+3. Synchronize the labels and update the Node in the workload cluster. CAPM3
+  already fetches the Node objects from the workload clusters today using a
+  remote client. We would leverage that same capability here.
+
+Scenario (ii) can be handled with a full-sync of all BMH and Node objects at a
+preconfigured synchronization interval.
+
+An alternative would be to set up a watch on the Node resource types in every
+workload cluster. The benefit being that we can quickly synchronize any label
+changes on the Node object itself. However, this introduces complications when
+the management cluster != workload cluster and additionally, when there are
+many workload clusters being managed by a single management cluster. How such
+a watch could be implemented within the controller-runtime model would require
+further investigation.
+
+Controller-runtime supports requeuing of a resource after a specified time has
+elapsed via the `RequeueAfter` field in its Result object. We can use this
+to ensure that synchronization always happens on every BMH at least once
+within a desired interval.
+
+### Configuration
+
+We need to expose two additional configuration variables to the user: the set
+of label prefixes and the full-sync interval.
+
+Several alternatives were also considered (see Alternatives section).
+
+#### Metal3Cluster
+
+We can have the prefixes specified at the Metal3Cluster level. For example,
+
+```yaml
+kind: Metal3Cluster
+name: test1
+spec:
+  metal3-label-sync-prefixes: "my-prefix.foobar.io, my-prefix.kubernetes.io"
+  metal3-label-sync-interval: "30s"
+...
+```
+
+Alternatively, we can use an annotation on the Meal3Cluster:
+
+```yaml
+kind: Metal3Cluster
+name: test1
+  metadata:
+    annotations:
+      metal3.io/metal3-label-sync-prefixes: "my-prefix.foobar.io, my.thing"
+      metal3.io/metal3-label-sync-interval: "30s"
+```
+
+We need to map the BMH to the Metal3Cluster object. One path to doing this is
+as follows:
+
+```diff
+    BMH.ConsumerRef --> Metal3Machine.OwnerRef -->
+    CAPI.Machine.Labels{cluster.x-k8s.io/cluster-name} -->
+    CAPI.Cluster.InfrastructureRef.Namespace/Name --> Metal3Cluster
+```
+
+Similarly, when Metal3Cluster is updated, we need to fetch all associated
+BMHs. One path to doing this is as follows:
+
+```diff
+    Metal3Cluster.OwnerRef --> CAPI.Cluster -->
+    Label-Selector{CAPI.Cluster.Name} --> CAPI.MachineList -->
+    Machine.Spec.InfrastructureRef.Namespace/Name -->
+    Metal3Machine.Annotation{metal3.io/BareMetalHost} --> BMH
+```
+
+The last step is to map the Metal3Machine to the BMH. Currently, an annotation
+is added to the Metal3Machine of the format  `metal3.io/BareMetalHost:
+Namespace/Name`. We can use this annotation to fetch the corresponding BMH.
+
+    Kind:         Metal3Machine
+    Annotations:  metal3.io/BareMetalHost: metal3/bmh-node-0
+
+### Other Issues
+
+#### Prefix Deletion
+
+Removal of a prefix, p1, from the configuration, has two consequences:
+
+1. Any additional labels with prefix p1 placed on the BMH will not be
+  synchronized to the corresponding Node.
+
+2. Any existing labels with prefix p1 present on the BMH or Node objects
+  will not be removed. Removal of these labels is the users responsibility.
+
+#### Delay between Node create and label sync
+
+The Node object is first created when kubeadm joins a node to the workload
+cluster (i.e. kubelet is up and running). There may be a delay (potentially
+several seconds) before the CAPM3 node synchronization logic kicks in to apply
+the labels on the Node.
+
+Kubernetes supports both eqaulity and inequality requirements in label selection.
+
+In an equality based selection, the user wants to place a workload on node(s)
+matching a specific label (e.g. Node.Labels contains `my.prefix/foo=bar`). The
+delay in CAPM3 applying the label on the node, may cause a subsequent delay in
+the placement of the workload, but this is likely acceptable.
+
+In an inequality based selection, the user wants to place a workload on node(
+s) that do not contain a specific label (e.g. Node.Labels not contain
+`my.prefix/foo=bar`). The case is potentially problematic because it relies on
+the absense of a label and this can occur during the delay.
+
+One way to address this is to use kubelet's `--node-labels` flag. In CAPI, we
+can potentially utilize `kubeletExtraArgs` within the KubeadmConfig spec for
+this purpose. However, this would require further investigation.  In
+particular, since the KubeadmConfig objects are generally created from
+KubeadmConfigTemplate, can we dynamically patch the objects with the labels
+from the BMH?
+
+Another possibility is to utilize the Metal3DataTemplate for labeling the
+nodes. This would need to further investigation as well.
+
+Yet another possiblity is to use a temporary label (e.g. `label-sync-pending`)
+through kubelet's `--node-labels` at node creation. The label can subseqently
+be removed by CAPM3 when it has applied the desired set of labels. The user
+must construct their label-selectors with this label in mind. For example:
+
+    Node.Labels not contain `label-sync-pending` AND Node.Labels not contain `my.prefix/foo=bar`...
+
+This approach can also be implemented entirely by the user. For example, the
+user may specifiy through the KubeadmConfigTemplate a label similar to
+`label-sync-pending`. Subsequently, when the node is ready to accept workloads
+(e.g. the desired labels have been applied by CAPM3), they can manually remove
+this label.
+
+There is also a security vulnerability associated with kubelet's ability to
+self-label the Node object. This has been
+[addressed](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md).
+However, as a consequence, BMH labels that fall within the *.kubernetes.io/*
+space cannot be specified via kubelet's `--node-labels` flag.
+
+Any solution to this issue is beyond the scope of this proposal. We should
+however document this limitation.
+
+## Alternatives
+
+* A provider-agnostic mechanism could be implemented in CAPI itself. For
+  example, labels placed on a provider specific Machine object (e.g.
+  Meta3Machine) could be read by CAPI and applied to corresponding Kubernetes
+  Node. [Here](https://github.com/kubernetes-sigs/cluster-api/issues/3504) is
+  one such use case for the AWS provider. However, there is reluctance in the
+  CAPI community to support such a mechanism. Additionally, for metal3, such a
+  scenario would still require that CAPM3 synchronize labels from BMH to
+  Metal3Machine.
+* Specifing labels via KubeadmConfigTemplate's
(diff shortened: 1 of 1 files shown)


design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,172 @@
...
+1. At a configurable interval (e.g. 15 seconds), fetch the complete list of BMHs.
+2. For each BMH, fetch the corresponding Node using the mapping discussed above. 
+3. Synchronize the labels and update the Node in the workload cluster.
+
+An alternative would be to set up a watch on the Node resource types in every workload cluster. The benefit being that we can quickly synchronize any label changes on the Node object itself. However, this introduces complications when the management cluster != workload cluster and additionally, when there are many workload clusters being managed by a single management cluster. How such a watch could be implemented within the controller-runtime model would require further investigation. 
+
+
+## Configuration
+
+We need to expose two additional configuration variables to the user: the set of label prefixes and the full-sync interval. We can introduce two new command-line flags for CAPM3, and also utilize default values if the flags are not specified:
```
dhellmann, line 111: User-configurable values like this should really be managed through an API, not command line options. Maybe in this case we could use an annotation on the host?

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,172 @@
+# Synchronize a Set of Labels on the BaremetalHost with the corresponding Kubernetes Node
+
+There are use cases where certain information is inherently tied to the BMH but at the same time, is valuable to the users/operators/schedulers in the K workload cluster. For example, a user may wish to place her workloads across hosts that are in different failure zones. In Kubernetes, labels on Node objects are the primary means by which to solve this problem. The proposal is for CAPM3 to synchronize a specific set of labels placed on a BMH with labels on the corresponding Kubernetes Node running on that BMH.
+
+
+# User Stories
+
+As a user I want replicas of a specific workload spread out across different server racks or geographical locations.
+
+As a user, I would like to place my security sensitive workloads on machines that meet specific security requirements. For example, certain hosts may be in a fortified zone within a data center or certain hosts may have strong hardware based security mechanisms in place (e.g. hardware attestation via TPM).
```
dhellmann, line 10: It seems like both of these use cases could be achieved by using different MachineDeployments, based on physical location or security capabilities. Have you explored that as an option?

In our use case, we wanted to add labels based on data exposed through inspection on the host. That same information may also be available through node-feature-discovery, so we are looking at using that tool instead.

Arvinderpal: I think both approaches you mentioned are complementary to what is being proposed here. The differences I see are:

1. MD labels are only applicable at creation time. Adding/Removing a new label requires a complete MD rollout. This is highly undesirable in a baremetal environment. The proposed approach would offer more flexibility. 
2. MD grouping may not always provide the needed granularity; effectively MD == workload type in what you suggested. Depending on the how granular the user wants their labeling, this may lead to a large number of MDs. 
3. There are cases where a logical grouping like MD is not appropriate when trying to capture physical grouping. For example, you may want a specific label on all hosts in the same rack but logically hosts in the rack may belong to different MDs. 
4. Certain information may not be discoverable by host inspection alone. For example, user intent cases such as I want to label certain hosts as temporary so only workloads that can tolerate interruptions are scheduled there. Also, the label sync may be a convenient alternative for users who don't want to implement/deploy controllers for feature discovery. 
5. There are security concerns with kubelet's `--node-labels` flag (as mentioned in this proposal). This is not a deal breaker but definitely undesirable. 

The CAPI proposal mentioned in related section would allow for labels to be added/removed without a rollout, but that is still in the proposal phase.

I can add some (or all) of the points above to the proposal if they provide more clarity.

design/sync-labels-bmh-to-node.md
Arvinderpal: My concern would be consistency since the same annotation would have to be duplicated across BMHs. The prefix, while user configurable, is a static value that should not require re-configuring. I would imagine that a cluster admin would define a standard set of prefixes to be used internally.

Also, annotations can be (accidentally) added/removed/changed. We could address this by having an immutable field in the BMH just for specifying prefixes.

dhellmann: OpenShift cluster administrators do not deploy metal3 themselves and do not have access to the command line arguments for the cluster-api provider. So, command line arguments won't work at all in an OpenShift cluster.

I suggested an annotation on the host because it wasn't obvious that we would always want the same set of labels copied. If we do think it will always be the same set, then maybe a new API resource to describe which labels to copy would make sense. That would let the implementation for copying them be based on a new controller within CAPM3 that could reconcile on the new API type. If we had an API that offered a list of label prefixes and then a labelSelector to pick which hosts it applies to, that would let users have a way to copy all or some of the labels from all or some of the hosts.

design/sync-labels-bmh-to-node.md
dhellmann: It does sound like MachineDeployments will not achieve the goals. Listing that in the alternatives section would be good.

Arvinderpal: Done.

design/sync-labels-bmh-to-node.md
Arvinderpal: I have updated the proposal. 
Also, see the section on using Metal3Cluster. Could this be a good place to keep this info?

Arvinderpal (commented):
Thank again for the feedback.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,191 @@
...
+```
+kind: Metal3Cluster
+name: test1
+spec:
+  bmh-label-sync-prefixes: "my-prefix.metal3.io, my-prefix.kubernetes.io"
+  bmh-label-sync-interval: "30s"   
+...
+```
+
+We need to map the BMH to the Metal3Cluster object. Similarly, when Metal3Cluster is updated, we need to fetch all associated BMHs. We can investigate this further if this is a better approach.
```
maelk, line 159: This is already partially done, when we reconcile the metal3machine we fetch all objects including Cluster and Metal3cluster : https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/controllers/metal3machine_controller.go#L110 and https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/controllers/metal3machine_controller.go#L137. The reverse is done here : https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/controllers/metal3machine_controller.go#L368 . It is quite straightforward to adapt to get the link with the BMH. IMO this approach is the best, considering that it does not require additional objects and controls the behaviour per cluster.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,191 @@
...
+
+We need to map the BMH to the Metal3Cluster object. Similarly, when Metal3Cluster is updated, we need to fetch all associated BMHs. We can investigate this further if this is a better approach. 
+
+## Other Issues
+
+### Delay between Node create and label sync
+
+The Node object is first created when kubeadm joins a node to the workload cluster (i.e. kubelet is up and running). There may be a delay (potentially several seconds) before the CAPM3 node synchronization logic kicks in to apply the labels on the Node. This may not be of concern to most users. However, it's worth asking if we can reduce or eliminate this delay?
+
+One way to address this is to use kubelet's `--node-labels` flag. In CAPI, we can potentially utilize `kubeletExtraArgs` within the KubeadmConfig spec for this purpose. However, this would require further investigation.  In particular, since the KubeadmConfig objects are generally created from KubeadmConfigTemplate, can we dynamically patch the objects with the labels from the BMH?
```
maelk, line 167: There is already a similar mechanism in place in Metal3DataTemplate where we refer to a single label, we could extend it to cover any labels that match the prefix

maelk (commented):
This looks good to me. Some small comments inline. Regarding the delay int he label applications, we could consider extending the label copy functionality of Metal3DataTemplate to cover the delay (that would not work for *.kubernetes.io though, as explained in the proposal) by copying all the labels matching a specific prefix.

design/sync-labels-bmh-to-node.md
Arvinderpal: Thanks. I have noted that.
It also seems to me that there may be user-level mechanisms that can get around this. I have added that to the section. 
In general, I think this is something we have to see if users actually have an issue with. My preference would be to just document this for the time being. If the need arises we can take it up later.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,217 @@
...
+    metal3.io/baremetalhost-label-sync-profile: some-ns/label-sync-profile-0
+prefixes:
+  - my-prefix.metal3.io
+  - my-prefix.kubernetes.io
+label-sync-interval: "30s"  
+```
+
+It's the admin's/user's responsibility to create the linkage by labeling BMHs (e.g `metal3.io/baremetalhost-label-sync-profile: some-ns/label-sync-profile-0`). 
+
+### Metal3Cluster
```
Arvinderpal, line 146: @maelk I'm also inclined to use Metal3Cluster as the top level object where prefixes get specified. Based on your feedback, I have updated the proposal with how the BMH <--> Metal3Cluster mapping may work. PTAL.

@dhellmann What to do you think about using Metal3Cluster instead of a new API type?

Arvinderpal (commented):
@maelk Thanks for the feedback. PTAL

maelk:
this looks good to me. Please fix the markdown lint issues.

Arvinderpal:
> this looks good to me. Please fix the markdown lint issues.

Markdowns have been fixed. Unfortunately, in the process the comments also became "outdated". They are still visible in the Comments tab though. 

@dhellmann I believe the main question is on the use of Metal3Cluster vs new a API object.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,353 @@
...
+and then scan through them to find the Node matching the ProviderID. This does
+not seem scalable for large clusters.
+Field-selectors could potentially be employed and would require further
+investigation. However,there is no obvious advantage over the earlier approach
+for doing this mapping.
+
+Scenario (ii) can be handled with a full-sync of all BMH and Node objects at a
+preconfigured synchronization interval:
+
+1. At a configurable interval (e.g. 15 seconds), fetch complete list of BMHs.
```
dhellmann, line 128: Instead of looping and fetching a list, it could watch hosts and then always return a request to re-enqueue with a delay. That would take advantage of the controller runtime logic for managing the work queue and keep the Reconcile() function a bit simpler.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,353 @@
...
+#### Annotations on BMH
+
+In this approach, an annotation is added to a BMH, in addition to the labels
+to be synchronized, which reflects the prefixes to be utilized. For example,
+
+```yaml
+  kind: BareMetalHost
+  metadata:
+    annotations:
+      cluster.x-k8s.io/label-prefixes: "my-prefix.metal3.io, my-prefix.kubernetes.io"
```
dhellmann, line 163: Annotations can be non-string types, so we could use a list as the value.

Should the name for the annotation be scoped to metal3.io, or is this a cluster API feature?

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,353 @@
...
+      cluster.x-k8s.io/label-sync-interval: "30s"
```
dhellmann, line 164: Do we need to make the sync interval different for each host?

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,353 @@
...
+corresponding `BareMetalHostLabelSyncProfile` (if any) that matches the BMH.
+To avoid fetching all `BareMetalHostLabelSyncProfile`, we can make use of a
+label on the BMH, in namespace/name format, that references the
+`BareMetalHostLabelSyncProfile` (see example below). We can drop the namespace
+requirement if it's in the same namespace as the BMH.
+
+The less common case is where the `BareMetalHostLabelSyncProfile` is updated (
+e.g. a prefix is removed). In this case, we will need to fetch all BMHs
+associated and perform a sync operations across the impacted Nodes. Label
+selectors could be used for this purpose.
```
dhellmann, line 192: This sounds a lot like what the hardware-classification-controller had to do.

I think it's going to be a lot easier to get the syncing logic right if we can figure out a way to watch resources in a "normal" way instead of trying to batch the changes. Maybe you could elaborate more on why that's not possible? I understand why we can't watch Nodes, since they may be in a separate cluster. Why can't we watch hosts? Or why can't we leave it up to each cluster to manage the labels for its own hosts & nodes?

Arvinderpal: > I think it's going to be a lot easier to get the syncing logic right if we can figure out a way to watch resources in a "normal" way instead of trying to batch the changes. Maybe you could elaborate more on why that's not possible? 

The controller watches the BareMetalHostLabelSyncProfile for changes (e.g. prefix updates). However, the profile is associated with 1 or more BMHs. In order to retrieve these BMHs, we make use of labels. I don't see where we are batching the changes. 
We do watch BMHs as discussed in the paragraph above this one, but that is for changes to the BMHs itself. 
If there is another way, please let me know. 

> I understand why we can't watch Nodes, since they may be in a separate cluster. Why can't we watch hosts? Or why can't we leave it up to each cluster to manage the labels for its own hosts & nodes?

I'm not sure I fully understand your comment. Are you suggesting another controller running in the workload cluster that watches BMHs in the mgmt cluster? In general, the pattern of a workload cluster initiating a call to the mgmt cluster introduces more complexities. For example, (i) placing appropriate credentials in the workload cluster, (ii) what happens when the mgmt cluster is something like KinD which is not externally visible, .... The CAPI community generally frowns on this pattern as well.

Arvinderpal (commented):
@dhellmann Thank you for the feedback. PTAL

design/sync-labels-bmh-to-node.md
dhellmann: Where is the controller that watches `BareMetalHostLabelSyncProfile` going to run? How is it going to find/modify Nodes? Where are the hosts that it will read to know which labels to apply to those nodes?

Arvinderpal: The controller runs on the management cluster and is part of CAPM3. If you recall, this new api type is an alternative to using command-line flags to pass the prefixes (which would not work for openshift). Nothing else has changed.

Also, please see the below section on using Metal3Cluster as an alternative. There was some discussion on that with @maelk. The discussion can still be found under the Comments tab.

dhellmann: OK, I suggested watching hosts and nodes as well as the new sync profile resource. You said that wasn't possible because those resources might not be in the same cluster as the profile, so I was confused. If that controller isn't in the same place as the Node it is going to change, how does it change the Node's labels? This is probably obvious to you, but my distro uses these components in a different way (there is no cluster pivot, everything is always just in the same cluster) so I apologize if this is a really basic question.

Arvinderpal: Sorry about the confusion. 

If the mgmt cluster == workload cluster, the process likely simplified. Unfortunately, we can' make that assumption. The process for updating the Nodes in the workload cluster is described [here in this proposal](https://github.com/metal3-io/metal3-docs/pull/149/files#diff-413f1e6764de1ed31caea7e636db03eaea13a1bbd33d2b4a2b52ef507237718cR110). CAPI maintains the credentials for accessing each workload cluster inside a Secret. There is precedence for fetching/updating resources in the workload cluster in both capm3 and capi, so we would take the same approach here.

Arvinderpal:
@dhellmann Can I ask for another review of this proposal? 
I can move to building a PoC once we agree on the decision between using Metal3Cluster or a new resource type. I believe that is the only open question left. 
Thank you

maelk:
What is the status on this proposal ? Should we set a lazy consensus during this week's community meeting ?

Arvinderpal:
A lazy consensus would be great. 
I working on the PoC and should have it ready soon. 
The only open item was how to specify the set of label prefixes. I have taken the approach of adding it to Metal3Cluster. Currently, it's an annotation on the Metal3Cluster which I think is simpler.  But, if desired, I can make that an explicit field in the spec. The doc has been updated accordingly.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,374 @@
...
+
+An alternative would be to set up a watch on the Node resource types in every
+workload cluster. The benefit being that we can quickly synchronize any label
+changes on the Node object itself. However, this introduces complications when
+the management cluster != workload cluster and additionally, when there are
+many workload clusters being managed by a single management cluster. How such
+a watch could be implemented within the controller-runtime model would require
+further investigation.
+
+### Configuration
```
dhellmann, line 140: There are several alternative approaches listed in this section, but it isn't clear which approach is really being proposed for use. I think it can't be the command-line flag approach. The annotation approach also seems to have been ruled out, is that right? Could you move the rejected approaches to the alternatives sections at the bottom of the document so it is clear?

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,374 @@
...
+The common case is when a label on the BMH is updated, we must first find the
+corresponding `BareMetalHostLabelSyncProfile` (if any) that matches the BMH.
+To avoid fetching all `BareMetalHostLabelSyncProfile`, we can make use of a
+label on the BMH, in namespace/name format, that references the
+`BareMetalHostLabelSyncProfile` (see example below). We can drop the namespace
+requirement if it's in the same namespace as the BMH.
+
+The less common case is where the `BareMetalHostLabelSyncProfile` is updated (
+e.g. a prefix is removed). In this case, we will need to fetch all BMHs
+associated and perform a sync operations across the impacted Nodes. Label
```
dhellmann, line 191: What would the sync do? Remove the previously synced labels from the nodes?

Arvinderpal: AFIK, when a prefix is removed, it's not entirely clear which prefix was removed (i.e. we can't do a diff against earlier state of the object). 
This make removal of labels associated with the removed prefix difficult if not impossible. Any existing labels will be left as is (the user can remove them). I documented this in the "Prefix Deletion" section below. 

When a prefix is added, any existing labels matching the prefix will be synced immediately.

Arvinderpal (commented):
@dhellmann Thank you for the feedback. PTAL again.

design/sync-labels-bmh-to-node.md
dhellmann: That doesn't seem ideal. Using a separate CRD to configure the controller to sync 1 prefix would let us trap the delete and handle it.

Arvinderpal: If I understand correctly, you are suggesting creating a new API type where each object of that type will be a specific prefix. 

Couple of comments:
1. While this would allow us to catch the above delete case, this would require more work to put together. My thought was to simply document the procedure. I would image this being a rare situation -- I can't image a scenario where users need to frequently update prefix values. 
2. More importantly, how do you envision mapping a "prefix object" to the set of BMHs it should apply to? With Metal3Cluster we get this logical mapping for free: Metal3Cluster -> Metal3Machine/Machine -> BMH. I suppose we could use labels, much like we do with the other API type I described in the doc; however, the Metal3Cluster grouping seems more appropriate to me as it does not require the user/admin to explicitly add labels to resources to keep them linked.

dhellmann: The new resource could either link to a Metal3Cluster or use a labelSelector to choose hosts. It may be more work, initially, but would provide a more robust solution. We can start with the annotation approach, and add an API later, I suppose.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,374 @@
...
+
+    BMH.ConsumerRef --> Metal3Machine.OwnerRef --> Machine.Status.NodeRef
+
+3. Synchronize the labels and update the Node in the workload cluster. CAPM3
+  already fetches the Node objects from the workload clusters today using a
+  remote client. We would leverage that same capability here.
+
+An alternative means to accomplish the mapping is to use the ProviderID:
+
+    BMH.ConsumerRef --> Metal3Machine.Spec.ProviderID --> Node.Spec.ProviderID
```
zaneb, line 116: Use the NodeRef.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,374 @@
...
+
+### Configuration
+
+#### Command-Line Flag
+
+We need to expose two additional configuration variables to the user: the set
+of label prefixes and the full-sync interval. We can introduce two new
+command-line flags for CAPM3, and also utilize default values if the flags are
+not specified: (i) `--bmh-label-prefixes` with a nil default value and (ii)
+`--label-sync-interval` with a 15-second default value.
```
zaneb, line 148: 15s is really short. That seems like a lot of activity.

design/sync-labels-bmh-to-node.md
zaneb: +1, annotation on one of the CAPM3 objects that applies to at least a whole MachinePool seems like the right way to configure this to me.

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,374 @@
...
+CAPM3 can already map a BMH object to the corresponding Node object. The
+proposal seeks to extend CAPM3 to also perform label synchronization
+between the two.
+
+Synchronization is not just a one time copy from BMH to Node on creation.
+More specifically, synchronization accomplishes the following:
+
+1. Addition/Removal of a label, within the set of predefined prefixes, on the
+  BMH will result in the addition/removal of that label on the corresponding
+  Node. Labels outside the prefix set will be ignored.
```
zaneb, line 43: Does prefix here mean "the entire part before the /" or `strings.HasPrefix()`? (I hope the former.)

Arvinderpal: It's the entire prefix part before the /

design/sync-labels-bmh-to-node.md
```diff
@@ -0,0 +1,374 @@
...
+2. Addition/Removal of a label, within the set of predefined prefixes,
+  _directly_ on the Node will result in the removal/re-adding of that label
+  on the Node.
+
+The synchronization should be limited to only labels matching a certain set of
+prefixes. The primary reason for this limitation is to avoid stepping on labels
+managed by other entities. Here CAPM3 would own the set of labels that match the
+prefixes. For example, the user may specify _my-prefix.metal3.io/_ as their
```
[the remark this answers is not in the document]
Arvinderpal: Good point. I'll use something else in the examples.

design/sync-labels-bmh-to-node.md
Arvinderpal: Yup! I'm using NodeRef in the POC. 
I'll actually remove this alternative from the doc.

design/sync-labels-bmh-to-node.md
Arvinderpal: I'll bump it up to 60 seconds.

Arvinderpal (commented):
Thank you for the feedback.

design/sync-labels-bmh-to-node.md
Arvinderpal: Sounds good. The initial approach utilizes the annotation but I'm happy to follow up with an API approach if there is desire for it.

fmuyassarov:
Thanks @Arvinderpal for the rebase!

fmuyassarov:
@dhellmann can we get merged this if looks okay?

dhellmann (approved):
This draft looks good to me. I know @zaneb and others had comments, so I'll leave it up to them to decide when it's ready to be merged.

Arvinderpal:
@zaneb Can we merge this and move on to the review of the [feature PR](https://github.com/metal3-io/cluster-api-provider-metal3/pull/152)?

dhellmann:
I think this has sat open long enough for feedback, and we can move ahead now.

/lgtm

PR #150, opened 2020-11-10
Add BMH live-iso proposal

hardys:
Adds a proposal to expose an Ironic API that allows us to boot an arbitrary live-iso image, vs only specifying an image to be written to disk by IPA.

### DIFF
--- a/design/baremetal-operator/bmh_live_iso.md
+++ b/design/baremetal-operator/bmh_live_iso.md
@@ -1,0 +1,128 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add boot-iso API to BareMetalHost
+
+## Status
+
+implementable
+
+## Summary
+
+Add a new interface so that it is possible to boot arbitrary
+iso images via Ironic.
+
+## Motivation
+
+In some circumstances it is desirable to boot an existing iso image
+instead of having Ironic boot IPA, for example:
+
+To reduce boot time for ephemeral workloads, it may be possible to boot an iso
+and not deploy any image to disk (saving the time to write the image and reboot)
+
+Where an alternative installer exists that is distributed as a live-iso, it
+may be desirable to leverage that toolchain instead of the IPA deploy ramdisk,
+for example the [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/)
+
+An [Ironic API](https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html)
+exists that supports this but it's not currently accessible via metal3.
+
+### Goals
+
+Expose the Ironic API to boot iso images via metal3.
+
+### Non-Goals
+
+This only considers the ability to boot an iso image (which Ironic supports
+via both redfish-virtualmedia and iPXE), not live kernel/ramdisk images.
+
+Ironic does support booting kernel/ramdisk images though, so the interface
+we decide on shouldn't preclude adding that capability in future.
+
+## Proposal
+
+Add a option to the BareMetalHost DiskFormat spec which indicates that instead
+of deploying an image to disk, a live-iso image will be booted:
+
+```yaml
+  spec:
+    image:
+      url: http://1.2.3.4/image.iso
+      format: live-iso
+    online: true
+```
+
+Note that in this mode, `rootDeviceHints` and `userData` will not be used
+since the image won't be written to disk, and Ironic doesn't
+support passing user-data in addition to the iso attachment at present.
+
+At some future point it would be desirable for Ironic to add the ability
+to pass userData - then a generic iso could be booted with the ability
+to define customization via the userData field in the normal way.
+
+## Design Details
+
+When this mode is selected, we need to configure Ironic to use the [ramdisk
+deploy interface](https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html)
+which is not currently enabled in ironic-image.
+
+We will also need to write the provided ISO URL into the
+[instance_info boot_iso field](https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#virtual-media-ramdisk)
+
+### Risks and Mitigations
+
+- Currently there is no detachment API in ironic, so in the case where an installer
+  iso is booted, we don't have a way to detach the virtualmedia and will have to
+  rely e.g on efibootmgr in whatever iso gets booted to ensure the correct boot device.
+
+- The inspection of a BMH will still rely on booting IPA and there's no support for
+  fast-track provisioning in this workflow (since it's booting two different
+  iso images) so this adds a reboot into the process.  This can potentially be
+  avoided in some cases where we can provide a status annotation on creation of
+  the BMH such that inspection is not performed.
+
+- Similarly cleaning of a BMH will rely on IPA, so will require a reboot which may
+  not be desirable in some situations.  This can potentially be avoided if cleaning
+  is [made configurable](https://github.com/metal3-io/metal3-docs/pull/151) in future.
+
+### Work Items
+
+- Enable ramdisk deploy interface in ironic-image
+- Add `live-iso` image format option to the BMH schema
+- Update BMO deploy logic to set the `deploy_interface` and `instance_info` appropriately
+
+### Dependencies
+
+There are some Ironic roadmap items which can be considered soft-dependencies,
+these don't block the initial implementation but may be desirable to make
+this feature more flexible:
+
+[Allow config-drive with virtualmedia ramdisk deploy](https://storyboard.openstack.org/#!/story/2008380)
+
+[Method to detach virtualmedia after initial boot](https://storyboard.openstack.org/#!/story/2008363)
+
+### Test Plan
+
+This should be tested in CI - do we have any metal3 coverage for
+redfish-virtualmedia atm?
+
+### Upgrade / Downgrade Strategy
+
+This new format is added to the BMH API as an optional new interface, all existing
+BMH interfaces should continue to work as before.
+
+On upgrade this new interface will become available, and once in use it will not
+be possible to downgrade, which given the expected use in net-new deployments
+is probably reasonable.
+
+## Alternatives
+
+We could avoid exposing this interface and mandate all users rely on IPA to
+deploy disk images, but this doesn't provide a good solution for the
+emphemeral worload case, and requires testing/supporting two install paths
+where a platform (such as FCOS mentioned) provides existing iso based
+tooling to deploy to disk.


design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+PXE.
+
+## Proposal
+
+Add a new field to the BareMetalHost spec which indicates that instead of
+deploying an image to disk, a live-iso image will be booted:
+
+```
+  spec:
+    image:
```
dhellmann, line 50: The `image` struct has several other fields related to each other. Would we use those? For example, do we need to verify the checksum of the ISO?

Could we trigger this different behavior if the `status.image.diskFormat` field is set to `ISO`? Or do we already support writing `ISO` images to disk?

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+
+At some future point it would be desirable for Ironic to add the ability
+to pass userData - then a generic iso could be booted with the ability
+to define customization via the userData field in the normal way.
+
+## Design Details
+
+When this mode is selected, we need to configure Ironic to use the [ramdisk
+deploy interface](https://docs.openstack.org/ironic/stein/admin/interfaces/deploy.html#ramdisk-deploy)
+which is not currently enabled in ironic-image.
```
dhellmann, line 67: Does that feature use different states in the Ironic state machine? Do we need corresponding changes in the state machine here in metal3, or will all of the changes be within the Ironic provisioner?

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+fast-track provisioning in this workflow (since it's booting two different iso images)
+so this adds a reboot into the process.  This can potentially be avoided in some cases
+where we can provide a status annotation on creation of the BMH such that inspection
+is not performed.
+
+- Similarly cleaning of a BMH will rely on IPA, so will require a reboot which may
+not be desirable in some situations.  This can potentially be avoided if cleaning
+is globally disabled in the ironic-image configuration via the
+[IRONIC_AUTOMATED_CLEAN](https://github.com/metal3-io/ironic-image/blob/master/configure-ironic.sh#L39)
+variable but this doesn't provide any per-host control.
```
dhellmann, line 92: We have had other people ask for a way to disable cleaning on deprovisioning. I think we can consider dealing with that via a separate design proposal.

maelk: @fmuyassarov is working on a proposal for that already

fmuyassarov: Yes, I will open a proposal in metal3-docs for having a configurable disk cleaning, latest end of this week.

design/baremetal-operator/bmh_boot_iso.md
hardys: Currently the boot_iso interface Ironic exposes does not require a checksum and I've not checked if the redfish attachment API supports that @dtantsur any idea?  Given that Ironic doesn't require it and metal3 won't download the image to verify any checksum, I omitted that for now.

I don't believe that Ironic has any support for writing ISO images to disk, so we probably could use `status.image.diskFormat` provided we expect that to remain the case?

design/baremetal-operator/bmh_boot_iso.md
hardys: The state transition is different - as soon as the ISO gets booted, the node is marked ACTIVE, which I think should work without changes to the metal3 state machine?

I did wonder if we should also allow booting ramdisk images via PXE since https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html supports both modes, and the state machine behavior is the same (then we could support this "live image boot" functionality for all BMC protocols, at the expense of potentially another image type/field)

design/baremetal-operator/bmh_boot_iso.md
hardys: I think I saw a comment that we're cleaning on both provisioning (the ironic `provide` state transition) and deprovisioning, is the plan to either remove one or make both configurable?

My concern is that for this use-case, we want to boot the FCOS installer iso as fast as possible, and there is already code in the image to handle wiping disks etc.

design/baremetal-operator/bmh_boot_iso.md
hardys: If we support the PXE option, I'm not sure if the `status.image.diskFormat` approach will work well, since we'll need a way to pass both a kernel and ramdisk URL IIUC (that's not explicitly mentioned in the [ironic docs](https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html#configuration) but I assume we'll need to override the default IPA kernel/ramdisk in driver_info)

design/baremetal-operator/bmh_boot_iso.md
fmuyassarov: The plan is to be able to completely disable/enable disk cleaning from metal3 per node. This requires some changes in the Ironic code as well, because, once we enable disk cleaning for a node, there is no way back to disable it again.

So, currently if we disable automated cleaning globally, then we can enable it only for certain nodes that require disk cleaning as part of deprovisioning. However, from the security perspective the safer option would be to enable automated cleaning globally and then disable it for certain nodes that don't require disk cleaning. And this option is not supported currently. 

We had quick chat on: https://github.com/metal3-io/baremetal-operator/issues/626 and the issue is opened in Ironic https://storyboard.openstack.org/#!/story/2008113.

design/baremetal-operator/bmh_boot_iso.md
dhellmann: OK, what I'm thinking about here is the fact that the fields under `spec.image` are all currently closely related. This new field is sort of related, but it's also sort of standalone because it sounds like the user would either set `spec.image.iso` or all of the other fields (with `diskFormat` remaining optional).

So, if the ISO URL isn't going to use the other image settings, perhaps it should be in its own group. You mention "live boot" in another comment, so how about `spec.liveBoot.url`? That would be parallel to the `spec.image.url` field, and give us a place to put any other settings we might need related to managing the live booting process. For example, if we decide to support PXE booting then we could have `spec.liveBoot.kernelURL` and `spec.liveBoot.ramDiskURL`.

hardys: @dhellmann thanks for the feedback - `spec.liveBoot.*` sounds good to me - do you think it'd be reasonable to rework this proposal so we cover both the live-iso and PXE ramdisk cases supported by ironic?

dhellmann: I could go either way on the PXE support. If we actually need it or want to support it now, then I could see including it in this proposal. If we don't need it right now, we could mention it as future work to explain why the API change is designed as it is and then have a separate enhancement later when we're ready to actually support it.

design/baremetal-operator/bmh_boot_iso.md
hardys: @dhellmann actually @dtantsur mentioned that Ironic can already boot an ISO image via ipxe, so perhaps using diskFormat would be enough - I've not yet tried that though.

hardys: Ok to clarify we'd always specify an iso, but internally set a different boot_interface depending on whether it's virtualmedia or not https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html so if we don't currently see a requirement for separate kernel/ramdisk I can adjust this propoal to specify an iso that can be booted via ipxe or virtualmedia transparently :)

dhellmann: I've grown to like the separate `liveBoot` approach, because it feels a bit more explicit about expressing the intent for how to treat the image. I think `diskFormat` will work fine, too, though. I don't think I feel strongly enough about either approach to hold up progress over it.

@zaneb do you have an opinion?

hardys: I'm fine with `liveBoot` and agree it makes it a little clearer it's a different interface/behavior to the current `image` section.

If we're bikeshedding interfaces I'd perhaps prefer `liveImage` since it aligns slightly better with the existing API but I'm happy to go with the consensus :)

dhellmann: I don't know about a checksum but @hardys did mention additional URLs for the ramdisk and kernel. I wouldn't want to add those to `spec.image` because it would be easy to get mixed up about which combination of fields work together and which are ignored in each case. A separate section makes that obvious, even at the expense of possibly having to have "duplicate" fields for some things.

So, how about a `spec.liveImage.url` for now to make it easy to add those other fields later?

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+instead of having Ironic boot IPA, for example:
+
+To reduce boot time for ephemeral workloads, it may be possible to boot an iso
+and not deploy any image to disk (saving the time to write the image and reboot)
+
+Where an alternative installer exists that is distributed as a live-iso, it
+may be desirable to leverage that toolchain instead of the IPA deploy ramdisk,
+for example the [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/)
+
+An [Ironic API](https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#virtual-media-ramdisk) exists for this, but it's not currently accessible via metal3, so it would be desirable to optionally enable this workflow.
```
dtantsur, line 31: Probably a better link: https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Goals
+
+Expose the Ironic API to boot iso images via metal3.
+
+### Non-Goals
+
+This only considers the case where the redfish-virtualmedia BMC
+option is used, not any ability to boot arbitrary iso images via
```
dtantsur, line 40: Ironic supports booting ISOs with iPXE as well, I think we should include it too.

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+    online: true
+```
+
+Note that in this mode, `rootDeviceHints` and `userData` will not be used
+since the image won't be written to disk, and Ironic doesn't
+support passing user-data in addition to the iso attachment at present.
+
+At some future point it would be desirable for Ironic to add the ability
+to pass userData - then a generic iso could be booted with the ability
+to define customization via the userData field in the normal way.
```
dtantsur, line 61: We're planning on supporting user data with redfish-virtual-media ISO boot (but not iPXE).

design/baremetal-operator/bmh_boot_iso.md
dtantsur: I like `liveImage.url`.

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+
+This will only work when the redfish-virtual-media boot interface is selected
+which will be the case when the `redfish-virtualmedia+http://foo` bmc address
+is specified - we need to ensure validation exists to catch invalid combinations
+
+### Risks and Mitigations
+
+- Currently there is no detachment API in ironic, so in the case where an installer
+iso is booted, we don't have a way to detach the virtualmedia and will have to
+rely e.g on efibootmgr in whatever iso gets booted to ensure the correct boot device.
```
dtantsur, line 80: On the roadmap.

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+
+- The inspection of a BMH will still rely on booting IPA and there's no support for
+fast-track provisioning in this workflow (since it's booting two different iso images)
+so this adds a reboot into the process.  This can potentially be avoided in some cases
+where we can provide a status annotation on creation of the BMH such that inspection
+is not performed.
```
dtantsur, line 86: Unfortunately, currently fast-track does not really work with virtual media at all (reboots always happen). I'll be working on it in the coming weeks.

design/baremetal-operator/bmh_boot_iso.md
dtantsur: A patch is on review for this feature.

design/baremetal-operator/bmh_boot_iso.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Work Items
+
+- Enable ramdisk deploy interface in ironic-image
+- Add `iso` field to the BMH schema
+- Update BMO deploy logic to set the `deploy_interface` and `instance_info` appropriately
+
+### Dependencies
+
+
```
dtantsur, line 102: Add the discussed future enhancements as soft dependencies?

design/baremetal-operator/bmh_boot_iso.md
hardys: Ok I'll go with `spec.liveImage.url` as that seems to be the consensus, and it leaves the option of adding more fields in future if needed, thanks!

design/baremetal-operator/bmh_boot_iso.md
hardys: Thanks, is there an upstream ironic link I can include here to reference that work?

design/baremetal-operator/bmh_boot_iso.md
hardys: Do you have a link I can reference please @dtantsur ?

design/baremetal-operator/bmh_boot_iso.md
hardys: Yes if you can provide links to the Ironic planned work we can include them here, or I'll add links inline with the text?

design/baremetal-operator/bmh_boot_iso.md
hardys: Ok I've mentioned that we can boot an iso via iPXE but that this proposal doesn't consider kernel/ramdisk boot, only an iso

design/baremetal-operator/bmh_boot_iso.md
hardys: Added a link to https://github.com/metal3-io/metal3-docs/pull/151

hardys:
Ok thanks for the feedback all - I've pushed an update which I think addresses the feedback so far :)

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,121 @@
...
+We will also need to write the provided ISO URL into the
+[instance_info boot_iso field](https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#virtual-media-ramdisk)
+
+### Risks and Mitigations
+
+- Currently there is no detachment API in ironic, so in the case where an installer
+  iso is booted, we don't have a way to detach the virtualmedia and will have to
+  rely e.g on efibootmgr in whatever iso gets booted to ensure the correct boot device.
+
+- The inspection of a BMH will still rely on booting IPA and there's no support for
```
dhellmann, line 81: We wouldn't do it as part of this enhancement, but should we add a flag to disable inspection to support workflows that get the hardware details another way?

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Work Items
+
+- Enable ramdisk deploy interface in ironic-image
+- Add `iso` field to the BMH schema
+- Update BMO deploy logic to set the `deploy_interface` and `instance_info` appropriately
+
+### Dependencies
+
+TODO link Ironic roadmap items as soft dependencies
```
dhellmann, line 99: Is this still a TODO?

dhellmann (commented):
I'm happy with the state of this. There are a few details we could fill in as updates to this PR or as separate PRs later.

/approve

dhellmann (approved):

design/baremetal-operator/bmh_live_boot.md
hardys: Yes we could do, I was thinking we'd rely on the existing interface e.g pass some status annotation at the point of creating the BMH, but then we'd need to change behavior so it can be updated after the hardware details are collected.  

A flag may be clearer though, I can either include that here or in a follow-up proposal when we have the boot-iso part working?

design/baremetal-operator/bmh_live_boot.md
hardys: Yes I've asked for some links from @dtantsur in the comments above, we can merge without if needed though and I'll add them as a follow-up

design/baremetal-operator/bmh_live_boot.md
dhellmann: For the use case you're working on, the host resource needs to exist before the ISO with the alternative agent is booted, so there's no opportunity to collect valid data in advance. We could pass fake data, but that feels wrong.

How about if we say that by default if there is a live image on the host we would not perform inspection, under the assumption that we either don't care about hardware details in those cases or the details will come from somewhere else later. Then we can use other design docs to cover an API for setting hardware details after the first reconciliation and to add a flag like `spec.liveImage.enableInspection` to restore the ability to (optionally) perform inspection with IPA before attaching the ISO.

design/baremetal-operator/bmh_live_boot.md
dhellmann: Yeah, I'm trying to help us keep moving. These documents are meant to be kept up to date, so I think it's OK to add links here later if we know there is work planned but don't have artifacts for that work, yet.

design/baremetal-operator/bmh_boot_iso.md
dtantsur: Just created https://storyboard.openstack.org/#!/story/2008380

design/baremetal-operator/bmh_boot_iso.md
dtantsur: https://storyboard.openstack.org/#!/story/2008363

design/baremetal-operator/bmh_live_boot.md
dtantsur: The links have been provided, see above.

design/baremetal-operator/bmh_live_boot.md
hardys: I'm not sure making inspection conditional on the spec.liveImage will work, because we may create the BMH resource then set the image reference later, and in that case we want the BMH to enroll and be ready to boot the iso, but not perform the existing inspection steps?

Perhaps adding an explicit flag elsewhere in the spec would be clearer?

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,121 @@
...
+```yaml
+  spec:
+    liveImage:
+      url: http://1.2.3.4/image.iso
+    online: true
+```
+
+Note that in this mode, `rootDeviceHints` and `userData` will not be used
+since the image won't be written to disk, and Ironic doesn't
+support passing user-data in addition to the iso attachment at present.
```
zaneb, line 60: It's about time we added a webhook to validate this kind of stuff.

design/baremetal-operator/bmh_live_boot.md
zaneb: Let's just have an annotation that causes inspection to be skipped. That's simple to do, works for all use cases, and avoids making it part of the Spec which is a bit icky.
There is a proposal I can't find right now to add an annotation to force a new inspection. Maybe we could use the same key with different values (e.g. `never` to skip it even the first time vs. `refresh` to trigger a new one).

Note that to add the data externally later on you'll need RBAC write access to the Status, which is unusual.

zaneb (commented):
/approve

design/baremetal-operator/bmh_live_boot.md
hardys: @zaneb I think this deserves a separate enhancement proposal - I suspect we need a way to skip inspection, but also add the data via the existing status annotation (avoiding the status RBAC issue you mention), which beyond the scope of this PR - I'll start a new one if that sounds reasonable.

design/baremetal-operator/bmh_live_boot.md
hardys: I added a new commit with the links, thanks!

design/baremetal-operator/bmh_boot_iso.md
hardys: @dtantsur thanks, if you have a link for this please let me know, and I'll add it to the list of Ironic roadmap items

hardys:
Ok I added the Ironic links, and if folks agree I think the skipping inspection workflow would be best discussed in a new proposal - implementation of this doesn't directly depend on it, so I'll create a new PR where we can decide the best approach for that.

design/baremetal-operator/bmh_live_boot.md
dhellmann: I agree with moving the inspection discussion to another enhancement.

dhellmann (approved):

hardys:
I think this is now ready to merge?

We'll also need some way to disable cleaning as discussed via https://github.com/metal3-io/metal3-docs/pull/151 and I'll raise a new enhancement to discuss how we can skip inspection.

design/baremetal-operator/bmh_live_boot.md
dhellmann: There's a period of time when validation web hooks aren't applied, between the time the CRD is installed (and the API server supports the new resource type) and when the web hook server is registered and starts.

dhellmann:
> I think this is now ready to merge?

I think so. @zaneb approved an earlier version and I'm happy with this version. It would be good to have a thumbs-up from someone else, too.

Maybe we can start the lazy consensus process this week?

hardys:
Discussion on the implementation PR and slack indicates we may want to go with a different interface so I'll hold this until that direction is agreed.

/hold

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,127 @@
...
+When this mode is selected, we need to configure Ironic to use the [ramdisk
+deploy interface](https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html)
+which is not currently enabled in ironic-image.
+
+We will also need to write the provided ISO URL into the
+[instance_info boot_iso field](https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#virtual-media-ramdisk)
+
+### Risks and Mitigations
+
+- Currently there is no detachment API in ironic, so in the case where an installer
```
dtantsur, line 77: There is now! Not sure we have any documentation about it, but ping @bfournie or myself for pointers.

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,127 @@
...
+  iso is booted, we don't have a way to detach the virtualmedia and will have to
+  rely e.g on efibootmgr in whatever iso gets booted to ensure the correct boot device.
+
+- The inspection of a BMH will still rely on booting IPA and there's no support for
+  fast-track provisioning in this workflow (since it's booting two different
+  iso images) so this adds a reboot into the process.  This can potentially be
+  avoided in some cases where we can provide a status annotation on creation of
+  the BMH such that inspection is not performed.
+
+- Similarly cleaning of a BMH will rely on IPA, so will require a reboot which may
```
dtantsur, line 87: For clarity: in-band cleaning. There is also out-of-band cleaning that is done via talking to the BMC, and we'll keep using that. It is only available as manual cleaning now, so it's fine to disable the automatic one.

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Dependencies
+
+There are some Ironic roadmap items which can be considered soft-dependencies,
+these don't block the initial implementation but may be desirable to make
+this feature more flexible:
+
+[Allow config-drive with virtualmedia ramdisk deploy](https://storyboard.openstack.org/#!/story/2008380)
+
+[Method to detach virtualmedia after initial boot](https://storyboard.openstack.org/#!/story/2008363)
```
dtantsur, line 105: Merged

design/baremetal-operator/bmh_live_boot.md
```diff
@@ -0,0 +1,127 @@
...
+- Add `iso` field to the BMH schema
+- Update BMO deploy logic to set the `deploy_interface` and `instance_info` appropriately
+
+### Dependencies
+
+There are some Ironic roadmap items which can be considered soft-dependencies,
+these don't block the initial implementation but may be desirable to make
+this feature more flexible:
+
+[Allow config-drive with virtualmedia ramdisk deploy](https://storyboard.openstack.org/#!/story/2008380)
```
dtantsur, line 103: On review, will try to get it merged this week.

design/baremetal-operator/bmh_boot_iso.md
dtantsur: https://storyboard.openstack.org/#!/story/2008321

hardys:
Alternative BMH API proposed via https://github.com/metal3-io/baremetal-operator/pull/759 - if we're in agreement with that I'll rework this to reflect the new interface.

hardys:
Updated to reflect interface in https://github.com/metal3-io/baremetal-operator/pull/759

/hold cancel

hardys:
The implementation landed ref https://github.com/metal3-io/baremetal-operator/pull/759 so it'd be good to merge this - @zaneb could you revisit perhaps?  Thanks!

dhellmann:
/retest
/lgtm

Since the code has merged, let's land this and update it in another PR if we're missing any details.

zaneb:
Netlify tests are failing with `make: *** No rule to make target 'build'.  Stop.`, so I would infer that this needs to be rebased to pick up whatever tooling support was added in the repo for those jobs.
(I miss Zuul.)

hardys:
> Netlify tests are failing with `make: *** No rule to make target 'build'. Stop.`, so I would infer that this needs to be rebased to pick up whatever tooling support was added in the repo for those jobs.
> (I miss Zuul.)

Thanks, now rebased and the jobs completed OK

PR #151, opened 2020-11-16
Allow disabling disk cleaning

fmuyassarov:
I have created this proposal under /design/cluster-api-provider-metal3/.. path even though it touches Baremetal Operator. But I think it would be easier to keep track of discussion/comments/feedback on a single document rather than splitting it into two proposals.

### DIFF
--- a/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
+++ b/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
@@ -1,0 +1,234 @@
+<!--
+This work is licensed under a Creative Commons Attribution 3.0
+Unported License.
+
+http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# allow disabling disk cleaning
+
+## Status
+
+implementable
+
+## Summary
+
+Configurable disk cleaning interface in Metal³ to allow users to
+disable/enable disk cleaning for the nodes.
+
+## Motivation
+
+Currently, when provisioning & de-provisioning Ironic nodes go
+through the [automated-cleaning](https://docs.openstack.org/ironic/latest/admin/cleaning.html#automated-cleaning)
+operation by default. This operation wipes out all the available
+disks on every node. While upgrade operations in Kubernetes,
+we want to ensure we don’t lose the data on the disks attached to
+the nodes. As such, we need a way to disable automated cleaning per
+node so that we can avoid disk cleaning while upgrading the cluster node(s).
+
+### User Stories
+
+#### Story 1
+
+As a cluster admin, I would like my node’s disks data kept untouched
+while I'm upgrading my Kubernetes cluster node(s).
+
+#### Story 2
+
+As a cluster admin, I would like my node’s disks to be cleaned
+while provisioning/de-provisioning (i.e. not upgrade scenario)
+operations.
+
+### Goals
+
+- Add a filtering mechanism in Cluster-api-provider-metal3 (CAPM3) to find out Machines
+   that should/not experience disk cleaning
+
+- To Baremetal-operator (BMO), instruct whether a BareMetalHost should have its disk
+   cleaned or not
+
+- To Ironic, instruct whether an Ironic node should have its disk cleaned or not
+
+### Non-Goals
+
+During the upgrade, a mechanism to select the exact same nodes after de-provisioning.
+In other words, nodes re-use logic.
+
+## Proposal
+
+Ironic supports enabling & disabling automated cleaning globally
+and on the node level. As such, Metal³ needs to be able to identify the nodes
+for which the disk cleaning should be disabled and communicate to Ironic to request
+disabling the disk cleaning for those specific nodes.
+
+### Metal³
+
+In the Metal³, we will need changes both in CAPM3 and BMO as follows
+
+#### Cluster-api-provider-metal3
+
+##### Metal3MachineTemplate object
+
+We introduce a new field `disableAutomatedClean` in the spec of the
+Metal3MachineTemplate object. `disableAutomatedClean` can store a boolean type
+of value. Setting it to True means don’t perform automated cleaning, while
+False value enables automated cleaning. By default, `disableAutomatedClean` will
+be set to False (i.e. do perform automated cleaning as it is now).
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+kind: Metal3MachineTemplate
+metadata:
+spec:
+ template:
+ spec:
+  disableAutomatedClean: False   #default value. Perform automated cleaning.
+```
+
+A new Metal3MachineTemplate controller will be introduced to reconcile
+Metal3MachineTemplate objects. When a user modifies `disableAutomatedClean`
+field on a Metal3MachineTemplate, CAPM3 Metal3MachineTemplate controller
+will update `disableAutomatedClean` field value on all the respective
+Metal3Machines referenced by that Metal3MachineTemplate.
+
+##### Metal3Machine object
+
+We introduce a new field `disableAutomatedClean` in the spec of the
+Metal3Machine object. `disableAutomatedClean` can store a boolean type
+of value. Setting it to True means do not perform automated
+cleaning, while False value enables automated cleaning. By default
+`disableAutomatedClean` will be set to False (i.e. do perform automated
+cleaning as it is now).
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+kind: Metal3Machine
+metadata:
+spec:
+ template:
+ spec:
+  disableAutomatedClean: False   #default value. Perform automated cleaning.
+
+```
+
+#### Baremetal Operator
+
+We introduce a new field `disableAutomatedClean` in the spec of the
+BaremetalHost object. `disableAutomatedClean` can store a boolean type
+of value. If set to True, the Baremetal Operator instructs Ironic
+to disable automated cleaning for an Ironic node, while if
+set to False, the Baremetal Operator instructs Ironic to enable automated
+cleaning. By default `disableAutomatedClean` will be set to False
+(i.e. do perform automated cleaning as it is now).
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+ name: worker-0
+spec:
+ disableAutomatedClean: False   #default value. Perform automated cleaning.
+ ...
+```
+
+## Design Details
+
+**Example flow including CAPM3:**
+
+1. User wants to disable disk cleaning before upgrading the nodes that are part
+   of the same KCP/MD.
+
+2. User sets True for the `disableAutomatedClean` field on a
+   Metal3MachineTemplate which is referenced by `infrastructureTemplate` field
+   in the KCP/MD.
+
+3. Metal3MachineTemplate controller keeps reconciling the Metal3MachineTemplate objects.
+   Once the update is seen on the Metal3MachineTemplate, Metal3MachineTemplate controller
+   starts mapping all the Metal3Machines referenced by that particular Metla3MachineTemplate,
+   and updates the `disableAutomatedClean` field to True on all the referenced Metal3Machines.
+
+4. Metal3Machine controller keeps reconciling the Metal3Machine objects.
+   Once the update is seen on the Metal3Machine, Metal3Machine controller starts
+   mapping the BareMetalHosts referenced by the Metal3Machines, and updates the
+   `disableAutomatedClean` field to True on all the referenced BaremetalHosts.
+
+5. Since the `disableAutomatedClean` field is set to True on the BaremetalHosts,
+   Baremetal Operator instructs the Ironic to disable disk cleaning for Ironic
+   nodes referenced by the BaremetalHosts.
+
+**Example flow when running only BMO:**
+
+1. User wants to disable disk cleaning for a single host
+
+2. User sets True for the `disableAutomatedClean` field on a
+   BareMetalHost spec.
+
+3. Baremetal Operator keeps reconciling the BareMetalHost, and after the update on
+   the BareMetalHost, BMO instructs the Ironic to disable disk cleaning
+   for the node.
+
+### Implementation Details/Notes/Constraints
+
+- As a user, you can still modify `disableAutomatedClean` field directly
+   on a BareMetalHost to disable/enable disk cleaning for the host that
+   is not managed by the CAPM3.
+
+- `disableAutomatedClean` can be updated on a BareMetalHost either before the provisioning
+   (i.e. `Ready` state) or deprovisioning (i.e. `Provisioned` state) takes place.
+
+- Updating `disableAutomatedClean` field on a Metal3MachineTemplate will ensure to
+    have the same value on all the corresponding Metal3Machines and BareMetalHosts
+    no matter what the current value is.
+
+### Risks and Mitigations
+
+None
+
+### Work Items
+
+- Implement a new field `disableAutomatedClean` in the spec of the Metal3MachineTemplate
+- Implement a new field `disableAutomatedClean` in the spec of the Metal3Machine
+- Implement a new field `disableAutomatedClean` in the spec of BareMetalHost
+- Implement a new Metal3MachineTemplate controller in Cluster-api-provider-metal3
+
+### Dependencies
+
+None
+
+### Test Plan
+
+- Unit tests should be added in Baremetal Operator
+- Unit tests should be added in Cluster-api-provider-metal3
+
+### Upgrade / Downgrade Strategy
+
+None
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+It may present some security issues since anyone can edit the `disableAutomatedClean`
+field. However, these operations can be restricted by proper RBAC rules.
+
+## Alternatives
+
+Automated cleaning can be disabled globally (by setting `automated_clean=False` for
+the conductor in [ironic.conf](https://github.com/metal3-io/ironic-image/blob/168207c1c2bdfa8761269bdf6434883b05647036/ironic.conf.j2#L62)
+and then enabled on the node level for nodes that require disk
+cleaning. However, this option introduces a couple of issues:
+
+1. As mentioned by @dtanstur in [2008113](https://storyboard.openstack.org/#!/story/2008113)
+   there could be some security concerns if we disable automated cleaning globally.
+
+2. Upgrade operations do not happen as frequently as provisioning & de-provisioning.
+   As such, this would require admins to always enable node automated
+   cleaning whenever they perform normal provisioning/de-provisioning.
+
+## References
+
+- Configurable disk cleaning options issue in BMO: <https://github.com/metal3-io/baremetal-operator/issues/626>
+- Make `node.automated_clean` work in both directions - story in Ironic: <https://storyboard.openstack.org/#!/story/2008113>
+- Make `node.automated_clean` work in both directions commit in Ironic: <https://review.opendev.org/#/c/762343/>


design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+##### Metal3Machine Controller
+
+Currently Metal3Machine controller removes the consumerRef on a BaremetalHost
+when it goes to deprovisioning. However, we are proposing to set the name of
+KubeadmControlPlane(KCP)/MachineDeployment(MD) in the consumerRef when
+BaremetalHost is going for deprovisioning.This will avoid the BaremetalHost
+to be utilized by any other Metal3Machines apart from the ones that were used
+before starting the upgrade.
```
maelk, line 130: This is actually unrelated to the configurable disk cleaning. It is needed to achieve predictability of the hosts selected during the process (and the main reason why we are implementing this disk cleaning configurability) but not strictly linked to it. Could you move it somewhere else in this proposal or even in another proposal centered on node re-use during upgrades ?

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+    updates the `disableAutomatedClean` field to False on all the referenced
+    BaremetalHosts.
+
+5. Since the `disableAutomatedClean` field is set to False on BaremetalHosts,
+    Baremetal Operator instructs Ironic to disable disk cleaning for the Ironic
+    nodes referenced by the BaremetalHosts.
+
+6. User started upgrading the nodes. CAPM3 removes the name of the Metal3Machine
+    from the consumerRef field of the BareMetalHost and instead adds the name of
+    the KCP/MD.
```
maelk, line 182: Same here, this should be in the second part / proposal

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+
+7. Deprovisioning of BaremetalHosts started. Ironic moves the nodes directly
+    to a `Manageable` without cleaning.
+
+8. New Machines and Metal3Machines are created. CAPM3 filters BareMetalHosts
+    and if the name of the KCP/MD exists on them, it pickups up those BareMetalHosts.
+
+9. If step8 is yes, then CAPM3 replaces/sets the value
+    of the consumerRef field of BareMetalHost to the name of the Metal3Machine.
+    If step8 is no, then CAPM3 follows the relies on the filtering logic.
```
maelk, line 192: Same here

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Risks and Mitigations
+
+### Work Items
+
+- Implement [2008113](https://storyboard.openstack.org/#!/story/2008113) in Ironic
+- Implement a new field `disableAutomatedClean` in the spec of the Metal3Machine
+- Implement a new field `disableAutomatedClean` in the spec of the Metal3MachineTemplate
+- Implement a new field `disableAutomatedClean` in the spec of BareMetalHost
+- Modify CAPM3 controller to set the KCP/MD name in the consumerRef of BareMetalHost
```
maelk, line 212: Same here

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+## Alternatives
+
+Automated cleaning can be disabled globally (by setting `automated_clean=False`
+in ironic.conf) and then enabled for only those nodes that require disk
+cleaning. However, this option introduces a couple of issues:
+
+1. As mentioned by @dtanstur in [2008113](https://storyboard.openstack.org/#!/story/2008113)
+    there could be some security concerns if we disable automated cleaning globally.
+
+2. Upgrade operations are don't happen as frequent as provisoning/deprovisoning.
```
maelk, line 252:
```suggestion
2. Upgrade operations don't happen as frequent as provisoning/deprovisoning.
```

maelk (commented):
This looks good, however I would propose to split the automated cleaning part from the node re-use during upgrade part of the proposal. Those are two different things, that should be combined to achieve hosted storage, but can be discussed individually

jan-est:
Overall, this looks good. I am also in favor that the node reuse part is separated from cleaning disc in an another proposal.

furkatgofurov7 (commented):
Looks good. Small nits and questions inline.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+### Goals
+
+- Add a filtering mechanism in CAPM3 to find out Machines that should/not
+    experience disk cleaning
+
+- To BMO, instruct whether a BareMetalHost should have its disk cleaned or not
+
+- To Ironic, instruct whether an Ironic node should have its disk cleaned or not
+
+- Enable disabling automated cleaning for 1/N nodes in Ironic.
```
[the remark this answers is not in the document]
fmuyassarov: it just an example to say that you can disable cleaning for 1 or N nodes.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+cleaning).
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+kind: Metal3MachineTemplate
+metadata:
+spec:
+  template:
+  spec:
+   disableAutomatedClean: True
```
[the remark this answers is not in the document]
fmuyassarov: It just to show how you would ask automated cleaning to be disabled (True value).

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: Yes I removed node reuse part from this proposal, and will open a new one for that part. The current one includes only disk cleaning.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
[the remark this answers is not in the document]
fmuyassarov: Updated it.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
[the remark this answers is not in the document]
fmuyassarov: Thanks. Fixed.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,260 @@
...
+### Metal³
+
+In the Metal³, we will need changes both in Cluster-api-provider-metal3(CAPM3)
+and Baremetal Operator(BMO) as follows.
+
+#### Cluster-api-provider-metal3
+
+##### Metal3MachineTemplate object
+
+We introduce a new field `disableAutomatedClean` in Metal3MachineTemplate.
```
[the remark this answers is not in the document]
fmuyassarov: Fixed.Thanks

fmuyassarov:
Thanks all for the early reviews. I have updated the current proposal to keep only disk cleaning related part and for the node reuse part will open a separate proposal and add a reference to this one.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,234 @@
...
+1. User wants to disable disk cleaning before upgrading the nodes that are part
+    of the same KCP/MD.
+
+2. User sets True for the `disableAutomatedClean` field on a
+    Metal3MachineTemplate which is referenced by `infrastructureTemplate` field
+    in the KCP/MD.
+
+3. Metal3MachineTemplate controller keeps reconciling the Metal3MachineTemplate objects.
+    Once the update is seen on the Metal3MachineTemplate, Metal3MachineTemplate controller
+    starts mapping all the Meta3Machines referenced by that particular Metla3MachineTemplate,
```
jan-est, line 159:
```suggestion
    starts mapping all the Metal3Machines referenced by that particular Metal3MachineTemplate,
```

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,234 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# configurable disk cleaning
```
dhellmann, line 8: This title made me think we were talking about choosing individual disks. Maybe we can retitle this to something like "allow disabling disk cleaning"?

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,234 @@
...
+    Once the update is seen on the Metal3MachineTemplate, Metal3MachineTemplate controller
+    starts mapping all the Meta3Machines referenced by that particular Metla3MachineTemplate,
+    and updates the `disableAutomatedClean` field to True on all the referenced Metal3Machines.
+
+4. Metal3Machine controller keeps reconciling the Metal3Machine objects.
+    Once the update is seen on the Metal3Machine, Metal3Machine controller starts
+    mapping the BareMetalHosts referenced by the Metal3Machines, and updates the
+    `disableAutomatedClean` field to True on all the referenced BaremetalHosts.
+
+5. Since the `disableAutomatedClean` field is set to True on the BaremetalHost,
```
dhellmann, line 167: Does it matter what state the host is in when this happens, either in the operator or in Ironic?

If the flag is false, should the baremetal-operator ensure that the corresponding flag in Ironic is also false?

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
```diff
@@ -0,0 +1,234 @@
...
+    Baremetal Operator instructs Ironic to disable disk cleaning for the Ironic
+    nodes referenced by the BaremetalHosts.
+
+### Implementation Details/Notes/Constraints
+
+- As a user, you should not modify `disableAutomatedClean` field directly
+    on a Metal3Machine or BareMetalHost but rather do it through
+    Metal3MachineTemplate so that you can update the field on all the Metal3Machines
+    referenced by a single Metal3MachineTemplate at once.
```
dhellmann, line 176: This feels limiting. If I deploy a bunch of hosts but want to pull one out of the cluster without erasing it for some sort of maintenance, shouldn't I be able to do that?

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: I agree. Will change it.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: > Does it matter what state the host is in when this happens, either in the operator or in Ironic?

That's a good question. So, I would assume bmh will need to be either in `provisioned` or `ready` state, while an Ironic node is in `active` or `manageable`. I think I should mention this in the proposal right?

> If the flag is false, should the baremetal-operator ensure that the corresponding flag in Ironic is also false?

By default, an Ironic node will have `automated_clean` set to True/None If I'm not mistaken. And that would be equal to the default value (False) of the `disableAutomatedClean` field in bmh. And after that, whatever value is set, BMO will make sure to set the same in the Ironic node via the Ironic.

dhellmann: Right, that was my point: We need to tell Ironic what value to use every time it is important for it to have the right value. How often do we need to do that? Just before deprovisioning? Every time we reconcile the host? Somewhere in between those two?

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: It is a limitation exactly as you said and we discussed it internally with @maelk. The reason why we want to control it through Metal3MachineTemplate is to be able to successfully complete the upgrade operation.

In other words, we need to put a logic in Metal3 to know - to differentiate which nodes will be affected by the upgrade that user is about to start. So, normally that happens for a group of nodes under the same KCP/MD. 

And, we thought if we ask a user to express which nodes should have its cleaning disabled/enabled via Metal3MachineTemplate, then in Metal3 we would take care of passing that all the way down to Ironic nodes. 

User updates disableAutomatedClean on Metal3MachineTemplate
CAPM3 controllers will update Metal3Machines -> BareMetalHosts

So, Metal3MachineTemplate was the only solution to select nodes referenced by the same KCP/MD. And, as you can see the proposal is really focused on the upgrade.

If we update `disableAutomatedClean` on a BareMetalHost, then it is problematic, because - 

If that BareMetalHost is part of a KCP/MD that has more than one Machines/Metal3Machines/BareMetalHosts, then we will end up having N number of BMHs with different values in `disableAutomatedClean`. This is okay, unless you are not planning to perform the upgrade.
Even if the user updates disableAutomatedClean on a BareMetalHosts, we can’t simply update top layer objects (Metal3Machine, Metal3MachineTemplate) based on this change.  Because, there might be more than one BMH under the same KCP/MD (or Metal3MachineTemplate) and we can’t force disableAutomatedClean to be the same on others from bottom to top.

Of course, we can update `disableAutomatedClean` on a bmh, and BMO will communicate with Ironic to have the same value in the corresponding Ironic node. But if you think about what will happen when the user starts the upgrading - then I would say don’t modify it via BMH. 

But I would really appreciate it if you have some ideas/feedback to share on what we should do at this point. Should we make it clear that the user should not modify it via a BareMetalHost/Metal3Machine or still allow it but then what would we do when someone starts the upgrade?

One thing we can do, allow users to update that field on a BareMetalHost/Metal3Machine. And, if the user does the upgrade unexpectedly - CAPM3  will find  the Metal3Machines, and BareMetalHosts references by the same Metal3MachineTemplate and whatever value is set in Metal3MachineTemplate will be forcefully updated to be the same on all the BareMetalHosts. As such, if it’s not an upgrade, we are fine, and we don’t have to worry about updating Metal3Machine or Metal3MachineTemplate.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: TBH I'm not quite sure which option is better. For example, it make sense to update Ironic node just before the deprovisioning because upgrade will happen for already provisioned hosts, in other words before they go to deprovisioning. But on the other hand, maybe it is also good to update every time BMO reconciles the bmh so that we update the Ironic node asap.

fmuyassarov: Also, how often we do depends on what BMH states we allow/recommend to update `disableAutomatedClean`. 

If we allow the user to directly update `disableAutomatedClean` field on a bmh, then BMO needs to make sure to update the Ironic node asap, or whenever reconciliation occurs. Because, the user can update the field when the bmh is in `Ready` state. So, we can't wait until deprovisioning happens, because the intention of the user here is to disable cleaning even during provisioning.
If we allow/recommend the user to directly update `disableAutomatedClean` field on a bmh, then BMO can update the Ironic node before going to deprovisioning. 

But now I think the better option would be to talk to Ironic to update node.automated_clean whenever BMO reconciles. We can also do the check in bmh to see if the current value of `disableAutomatedClean`  is the same or not. If not then, then it communicates with Ironic.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
dhellmann: > It is a limitation exactly as you said and we discussed it internally with @maelk. The reason why we want to control it through Metal3MachineTemplate is to be able to successfully complete the upgrade operation.
> 
> In other words, we need to put a logic in Metal3 to know - to differentiate which nodes will be affected by the upgrade that user is about to start. So, normally that happens for a group of nodes under the same KCP/MD.
> 
> And, we thought if we ask a user to express which nodes should have its cleaning disabled/enabled via Metal3MachineTemplate, then in Metal3 we would take care of passing that all the way down to Ironic nodes.

If the template's version of `disableAutomatedClean` is a `*bool` then we can tell the difference between `false` and not set. When the value in the template is not set, CAPM3 doesn't need to update the machine. That would allow someone to modify a single machine independently, while still supporting the "top down" propagation described in the algorithm here for the upgrade case.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
dhellmann: I'm probably making a bigger deal out of this than necessary, but it's important to think clearly about how the value propagates through the system to ensure it has the right value when needed. We really only need to make sure we update the value before asking Ironic to do something that would rely on the value, right? What operations or state transitions in Ironic will have different behavior based on the setting? Maybe setting the value in Ironic only needs to be a step in the Deprovision() method of the provisioner?

dhellmann: I replied to your first comment before reading the second.

If we also want to use this flag to disable cleaning during provisioning, then we would need to set it then, too.

I guess the point is, we don't need a step in our state machine or a separate reconciliation phase to set the value. If we want it done all the time, then we could do it in ValidateManagementAccess(). It seems reasonable to just set it for operations that would use it, though, and that might make it more obvious to someone trying to understand the code later.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
maelk: I agree with Doug here. I think that whenever the value is set in the template, it should take precedence to what the user has configured directly on the BMH and overwrite it during the reconciliation. If unset, let's leave it up to the user, agreed.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
dtantsur: Choosing individual disks sounds interesting, but we're not quite there, +1 for renaming.

fmuyassarov: Fixed, thanks.

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: Updated the document. Thanks @dhellmann , @maelk

design/cluster-api-provider-metal3/configurable_disk_cleaning.md
fmuyassarov: You are right that we don't need an extra step. I'm not fully sure yet where to add the logic for setting the value in Ironic. But I think the latter option you proposed makes more sense. But of course, I will have a better idea during the implementation.

fmuyassarov:
The proposal is now updated. Do you think we can put a lazy consensus of one week if the current status is ok @dhellmann?

design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### Cluster-api-provider-metal3
+
+##### Metal3MachineTemplate object
+
+We introduce a new field `disableAutomatedClean` in the spec of the
+Metal3MachineTemplate object. `disableAutomatedClean` can store a boolean type
+of value. Setting it to True means don’t perform automated cleaning, while
+False value enables automated cleaning. By default, `disableAutomatedClean` will
+be set to False (i.e. do perform automated cleaning as it is now).
```
dhellmann, line 76: We were going to use a pointer field, right? So the default will be nil, which is treated the same as false in this case.

dhellmann (approved):
I'm happy with this draft. I have one nit (see below), but we can deal with that in the implementation if there's no other reason to update this PR.

I would like to get someone from the Ironic team to sign off on the general approach, since it requires work in Ironic, too. @dtantsur what do you think?

design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
fmuyassarov: Yes, that should work perfectly fine.

fmuyassarov:
Just a small update/comment that required change in Ironic side was merged recently. That's why I removed 762343 as dependency in the proposal since it landed already.  https://review.opendev.org/c/openstack/ironic/+/762343

fmuyassarov:
@dhellmann Can we put lazy consensus on this proposal too? I will also make sure to bring this up in the next community meeting,

dtantsur (approved):
Let's make sure we have a warning in the logs in case cleaning is disabled.

dtantsur:
A follow-up to the today's discussion: have you considered using the `rebuild` feature? It's the same API as provisioning, but using `rebuild` verb instead of `active`, providing a new image (and a configdrive).

fmuyassarov:
Thank you @dtantsur . I just checked Ironic's state machine and common.states [module](https://docs.openstack.org/ironic/latest/contributor/api/ironic.common.states.html#ironic.common.states.REBUILD) to get some idea. 

In this proposal,  the idea is to be able to differentiate the nodes in Metal3 that may or may not need to experience disk cleaning. So, I think `rebuild` could be something that I will need to consider during the implementation (I'm not yes quite familiar with the API though),  but as you mentioned today on the call this should not be probably blocker. And if so, I would like to move forward and start the implementation.

@dhellmann What do you think, can this be also merged, since it has already been reviewed and approved by you and Dmitry ?
As said above, I brought up this proposal in today's community [meeting ](https://youtu.be/Yh_XJ-d_8kM?t=1443)and there weren't any objections apart from a question that Dmitry mentioned above.

jan-est:
lgtm

maelk:
We cannot use rebuild because we do upgrades through de-provisioning and provisioning again, since our nodes are immutable. When upgrading, a new metal3machine is created while the existing one is deleted, so the BMH consumer changes (inc. metaData, networkData and userData). 
/lgtm

hardys:
@fmuyassarov is there any WIP code for this yet?  I'm interested in helping/testing as it's related to what I'm trying to do in https://github.com/metal3-io/baremetal-operator/pull/754 (when booting a live iso we may not need/want to do cleaning)

fmuyassarov:
> @fmuyassarov is there any WIP code for this yet? I'm interested in helping/testing as it's related to what I'm trying to do in [metal3-io/baremetal-operator#754](https://github.com/metal3-io/baremetal-operator/pull/754) (when booting a live iso we may not need/want to do cleaning)

@hardys I haven't opened a PR for this yet. But will do it very soon. I am working on it right now locally. Thanks for offering help.

PR #155, opened 2021-01-06
Add proposal for supporting external introspection

hardys:
Add a proposal for an API to optionally disable introspection - this is related to the LiveImage proposal, where we may wish to skip inspection either to boot an appliance image faster, or because the image booted does it's own inspection.

### DIFF
--- a/design/baremetal-operator/external-introspection.md
+++ b/design/baremetal-operator/external-introspection.md
@@ -1,0 +1,108 @@
+# Support external introspection
+
+## Status
+
+implementable
+
+## Summary
+
+A declarative API is proposed to disable inspection of a BareMetalHost
+and optionally allow external sources of inspection data to update the
+hardware status data.
+
+## Motivation
+
+Related to the work to enable booting of a
+[LiveImage](https://github.com/metal3-io/metal3-docs/pull/150),
+there is the requirement to optionally disable inspection on initial
+registration of a BareMetalHost (so that the live image can boot
+more quickly, ref user stories below).
+
+### Goals
+
+- A declarative API to disable inspection on BMH registration
+- Provide an interface to update hardware status data at an arbitrary time
+  after BMH creation.
+
+### Non-Goals
+
+- There are no plans for any mechanism to trigger or consume data from any out
+  of band inspection process, other than providing hardware data via an annotation.
+
+## Proposal
+
+### Disable inspection proposal
+
+To align with the [re-inspection API proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md),
+the `inspect.metal3.io` annotation will be reused, with the addition of a value.
+
+The optional `inspect.metal3.io: disabled` annotation will be used to describe
+the situation where we wish to disable the default inspection behavior.
+
+When the BMO finds this annotation, it will skip performing inspection
+during the
+[Inspecting state](https://github.com/metal3-io/baremetal-operator/blob/master/docs/BaremetalHost_ProvisioningState.png)
+
+### Hardware status update proposal
+
+In the current implementation, when `baremetalhost.metal3.io/status` is
+provided, it can set any status field, and thus is only evaluated on the
+very first reconcile (primarily to support externally provisioned hosts,
+where we collect the inspection data prior to creating the BMH resources).
+
+In the case where metal3 is booting a live-image that contains code that
+can collect hardware details, it's desirable to have a way to update the
+hardware status after the image is booted.
+
+To enable this safely, we can add a new `inspect.metal3.io/hardwaredetails`
+annotation, which will allow updating the status/hardware field:
+
+- At any time when inspect.metal3.io=disabled
+- When there is no existing HardwareDetails data in the Status
+
+In the latter case, it may be a potentially safer/more constrained interface
+than the current `baremetalhost.metal3.io/status` API.
+
+Given that the primary use-case for this is live-boot images (where no disk
+image is written to disk), and that profile matching is no longer the preferred
+interface for specifying root-disk hints, if `inspect.metal3.io/hardwaredetails`
+is updated and the BMH is in the `Ready` state, we will not attempt to match
+profiles based on this data.
+
+In the event that both `baremetalhost.metal3.io/status` and
+`inspect.metal3.io/hardwaredetails` are specified on BMH creation,
+`inspect.metal3.io/hardwaredetails` will take precedence and overwrite any
+hardware data specified via `baremetalhost.metal3.io/status`.
+
+### User stories
+
+#### Fast-boot appliance live-image
+
+Where the LiveImage booted is an appliance that must be running as quickly as
+possible, it may be desirable to skip the time taken for inspection
+(and also cleaning which is discussed in an [existing proposal](https://github.com/metal3-io/metal3-docs/pull/151)
+
+#### Live-image installer does inspection
+
+[Installer ISO images](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/#_installing_from_live_iso)
+may be booted which can include their own inspection tooling.
+
+In this case, it is desirable to avoid the extra reboot and have the live-iso
+collect the required data (and update the BMH via the status annotation).
+
+## Alternatives
+
+The main alternative is to provide a status annotation at the point of
+creating the BMH, which might be enough for the fast-boot appliance use-case,
+but isn't ideal for the case where there is data collected by the live-iso
+which can be used to subsequently update the hardware status.
+
+We could also enable evaluation of the existing annotation at any arbitrary time
+but this is potentially unsafe, given that the BMO stores data in some other
+status fields.
+
+## References
+
+- Re-Inspect API [proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/re-introspection-interface.md)
+- Live Image [proposal](https://github.com/metal3-io/metal3-docs/pull/150)
+- Live Image [implementation](https://github.com/metal3-io/baremetal-operator/pull/754)


design/baremetal-operator/disable-introspection-intereface.md
```diff
@@ -0,0 +1,96 @@
...
+[LiveImage](https://github.com/metal3-io/metal3-docs/pull/150),
+there is the requirement to optionally disable inspection on initial
+registration of a BareMetalHost (so that the live image can boot
+more quickly, ref user stories below).
+
+### Goals
+
+- A declarative API to disable inspection on BMH registration
+- The option to update status data with a status annotation when inspection is
+  performed by some means other than the regular BMH flow
```
dhellmann, line 23: We can already do this. I think the goal is really to be able to do it at any arbitrary point in time, right?

design/baremetal-operator/disable-introspection-intereface.md
```diff
@@ -0,0 +1,96 @@
...
+the situation where we wish to disable the default inspection behavior.
+
+When the BMO finds this annotation, it will move directly from `Registering`
+[state](https://github.com/metal3-io/baremetal-operator/blob/master/docs/BaremetalHost_ProvisioningState.png)
+to `Ready` and not perform any inspection.
+
+In this case (similar to the reboot interface), when the suffixed form of
+`inspect.metal3.io/disable` is found, it will never be removed by the BMO.
+
+If both inspect.metal3.io and inspect.metal3.io/disable forms are in specified,
```
dhellmann, line 49:
```suggestion
If both inspect.metal3.io and inspect.metal3.io/disable forms are specified,
```

design/baremetal-operator/disable-introspection-intereface.md
```diff
@@ -0,0 +1,96 @@
...
+the `inspect.metal3.io/disable` form will take priority and the
+`inspect.metal3.io` annotation will be ignored.
+
+When `baremetalhost.metal3.io/status` is updated, any data should be
+consumed even in the `Ready` or `Provisioned` state, so that in the
+case where inspection data is collected via a live-image we can still
+store that data in the BMH status field.
```
dhellmann, line 56: This will change how we manage deleting the annotation. Today we delete it if it is present and the status fields are not nil. If we're going to allow updates, we will need to be more careful.

Which status fields do we want to allow someone to update this way? The current annotation encodes the entire status block, including provisioning state. As a one-time thing, that's OK. I'm not sure we want to give anyone with permission to add an annotation to a host permission to overwrite its "internal" provisioning status.

If we only care about allowing updates to the hardware details, perhaps we want a new annotation for that? Or we want at least to specify that only the hardware details will be updated?

We should also specify whether the settings from the annotation are merged with or overwrite the existing settings. I think we're going to want them to overwrite?

design/baremetal-operator/disable-introspection-intereface.md
```diff
@@ -0,0 +1,96 @@
...
+collect the required data (and update the BMH via the status annotation).
+
+## Open Questions
+
+When `baremetalhost.metal3.io/status` is updated and the BMH is in the `Ready`
+state, should we attempt to match profiles based on that inspection data?
+
+Given that the primary use-case for this is live-boot images (where no disk
+image is written to disk), and that profile matching is no longer the preferred
+interface for specifying root-disk hints, it may be reasonable to skip this?
```
dhellmann, line 81: I think so, yes. We want to remove the profile part of the API, but don't want to make that a prerequisite to this work.

design/baremetal-operator/disable-introspection-intereface.md
hardys: @dhellmann Yes, IIRC currently the status annotation is only consumed on the very first reconcile?  So we'll need the ability to update it after provisioning in this case.

dhellmann: Right, so I think the phrasing of the goal should reflect that.

design/baremetal-operator/disable-introspection-intereface.md
hardys: Oh I didn't realize the entire status could be overwritten, so yes we'll need to be more careful and I think we only care about updating the hardware details.

Perhaps it should be a new suffix e.g a path like `baremetalhost.metal3.io/status/hardware` - and that is allowed to be updated (either unconditionally, or when inspection is disabled?)

dhellmann: I feel like disabling introspection and updating the hardware details are slightly orthogonal. So, I would use a new annotation (to be bikeshedded later) for the hardware details and apply that unconditionally. Then I would add the annotation to disable inspection to avoid that part of the workflow.

The implementation of the hardware details update probably belongs in the controller near the place where we update the status based on the annotation now.

For the implementation of the disable annotation, I think we could still go through the same states in the state machine, and just have the handler for the inspection state skip its work if the annotation is present. That seems simpler than adding a new transition? I could go either way on that part.

hardys: Ok so perhaps I should remove the status annotation content from this PR and create another separate proposal?

They're both related to the same use-cases, but I agree the implementation can be approached independently.

I also don't have a strong opinion on a new state vs skipping inspection in the existing state - happy to take your guidance on which is likely to be simpler as I've not looked that closely at the code in the state machine yet :)

dhellmann: > Ok so perhaps I should remove the status annotation content from this PR and create another separate proposal?

I think we can reframe this proposal as "support external hardware inspection" and talk about both here.

> 
> They're both related to the same use-cases, but I agree the implementation can be approached independently.
> 
> I also don't have a strong opinion on a new state vs skipping inspection in the existing state - happy to take your guidance on which is likely to be simpler as I've not looked that closely at the code in the state machine yet :)

@zaneb or @andfasano would be better judges these days, to be honest. :-)

andfasano: From the given description, I'd give my preference to the inspection skipping during the `Inspecting` state, rather than adding a new state - there's not enough "critical mass" to justify a new state IMHO, and the skip implementation would be pretty straightforward.

hardys:
/retitle Add proposal for supporting external introspection

design/baremetal-operator/disable-introspection-intereface.md
hardys: Ok I retitled the proposal, and split the disable/update parts in the proposal section.  Also mentioned that we'll skip inspecting in the `Inspecting` state.

hardys:
@dhellmann @andfasano updated based on your feedback, thanks!

design/baremetal-operator/external-introspection.md
```diff
@@ -0,0 +1,116 @@
...
+When the BMO finds this annotation, it will skip performing inspection
+during the
+[Inspecting state](https://github.com/metal3-io/baremetal-operator/blob/master/docs/BaremetalHost_ProvisioningState.png)
+
+In this case (similar to the reboot interface), when the suffixed form of
+`inspect.metal3.io/disable` is found, it will never be removed by the BMO.
+
+If both inspect.metal3.io and inspect.metal3.io/disable forms are specified,
+the `inspect.metal3.io/disable` form will take priority and the
+`inspect.metal3.io` annotation will be ignored.
```
zaneb, line 55: Possible alternative: use the same `inspect.metal3.io` annotation and decide what to do based on the content (this way you can never have a conflict), e.g.:
* (empty): re-inspection as proposed
* `disable`: don't inspect
* `{...}` (HardwareStatus in json format): replace HardwareStatus with this data

design/baremetal-operator/disable-introspection-intereface.md
zaneb: I think the discussion got garbled a little bit - Doug talked about adding a new state _transition_ (i.e. from Registering->MatchProfile, skipping Inspecting), not adding a new state (which I agree we don't want).
Either doing the new transition (check annotation in `NeedsInspection()`) or just handling it during the `Inspecting` state (check annotation in `actionInspecting()`) is easy enough, so I don't really care either.

design/baremetal-operator/external-introspection.md
dhellmann: I like the way this avoids conflicts. Are there any precedents of using annotation content of different types in other APIS?

design/baremetal-operator/external-introspection.md
```diff
@@ -0,0 +1,116 @@
...
+In the current implementation, when `baremetalhost.metal3.io/status` is
+provided, it can set any status field, and thus is only evaluated on the
+very first reconcile (primarily to support externally provisioned hosts,
+where we collect the inspection data prior to creating the BMH resources).
+
+In the case where metal3 is booting a live-image that contains code that
+can collect hardware details, it's desirable to have a way to update the
+hardware status after the image is booted.
+
+To enable this safely, we can add a new `baremetalhost.metal3.io/status/hardware`
```
dhellmann, line 68: I don't think annotation names can contain multiple `/` characters (https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set). If we don't go with @zaneb's proposal above, this could be `baremetalhost.metal3.io/hardware` or `baremetalhost.metal3.io/status-hardware`.

design/baremetal-operator/external-introspection.md
```diff
@@ -0,0 +1,116 @@
...
+and optionally allow external sources of inspection data to update the
+hardware status data.
+
+## Motivation
+
+Related to the work to enable booting of a
+[LiveImage](https://github.com/metal3-io/metal3-docs/pull/150),
+there is the requirement to optionally disable inspection on initial
+registration of a BareMetalHost (so that the live image can boot
+more quickly, ref user stories below).
```
andfasano, line 19: @hardys Do you know if there's any other scenario, rather than the live-iso one, requiring to disable the initial inspection? If it's the only one, couldn't be maybe enough to detect the presence of `spec.image.format == live-iso` as described in the [BMH live-iso proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/bmh_live_iso.md) to skip the inspection?

design/baremetal-operator/external-introspection.md
hardys: I couldn't find any examples of this kind of interface, but it seems reasonable to me - the only slightly weird thing is combining one key to influence both the action (re-inspect, disable) and the data, what about separating them, e.g:

```
inspect.metal3.io/disable
inspect.metal3.io/data: <data in json format>
```

Then the bare `inspect.metal3.io` can be interpreted as re-triggering inspection, and we've got the ability to expand the interface in future if needed?

design/baremetal-operator/external-introspection.md
hardys: @andfasano I don't think such an option is specific to live-iso deployments, this could be used any time deploy walltime is desired above collecting hardware details?  

Particularly for some repeated CI/test cases on VMs I can imagine disabling inspection to reduce the time for test iterations.

andfasano: Thanks, just wanted to understand if there was any known use case.

design/baremetal-operator/external-introspection.md
andfasano: Using two keys sounds cleaner. What about collapsing everything in just one key, and specify the action as part of the json payload (in additon to the data)? It could help with the unmarshalling and there will be just one key

hardys: @andfasano I was thinking it would be cleaner to keep the action and data separate, it'll be simpler to just unmashall whatever is in `inspect.metal3.io/data` and put that into `status->hardware`?

What is the benefit of having only one key, vs using the `/subkey` approach?

andfasano: Just to slightly reduce the complexity (by mistake, one could define the `/data` subkey without the master action key, maybe it could be more difficult when both the things are defined as part of the json payload), but maybe it could be too much overhead in some cases. So fine for me to have a couple of subkeys, one dedicated for just the data when required.

hardys: Ok thanks - personally I'd prefer to avoid merging the action with the json data, so IMO we should either use two keys, or the single-key approach described by @zaneb where:

```
inspect.metal3.io                                     # re-inspect
inspect.metal3.io: disabled                      # skip inspection
inspect.metal3.io: <data in json format>  # consume data to update status->hardware
```

One question is, do we only support consuming the data in the case where `inspect.metal3.io: disabled` was the previous state?  In that case I guess we'd need to store the state in the status field?

It'd be simpler to consume the data unconditionally, but obviously that does mean some risk a user could overwrite status data we previously collected via inspection (which is a good or bad thing I guess depending on your perspective) :)

hardys: I'm happy to go with the consensus, @fmuyassarov any preferences?

zaneb: @hardys slightly confused about this discussion, because IIUC what you're proposing is 3 separate keys, the same as in the doc already?

Thinking about this a bit deeper though... I don't think we want to change the semantics of the re-inspection thing, which are that it only does something (flips to the Inspecting state) in the Ready state. And we'll need to be able to set the data when the Host is Provisioned, because we're only going to be able to get it from something running on host and that thing will have to be provisioned. So we probably shouldn't share the same key for setting data and triggering re-inspection after all. Accordingly, I'm amending my suggestion to this:

* `inspect.metal3.io` - move to Inspecting state from Ready
* `inspect.metal3.io=disabled` - don't inspect even in Inspecting
* `inspect.metal3.io/hardwaredetails={...}` - set HardwareDetails if empty and delete annotation

Open question: should we allow users to replace existing HardwareDetails? At any time? Only in the inspecting state after triggering re-inspection?

fmuyassarov: > I'm happy to go with the consensus, @fmuyassarov any preferences?

Hi @hardys , maybe a week would be fine if everyone here is okay with the proposal already. I'm trying also to update the implementation PR for re-inspection. Very sorry for letting that PR to hang for a long time.

hardys: > @hardys slightly confused about this discussion, because IIUC what you're proposing is 3 separate keys, the same as in the doc already?

Right, personally I think three keys would be fine, but both you and @andfasano expressed a preference for minmising the number of keys, so I'm just trying to bikeshed an interface everyone can agree on ;)
 
> Thinking about this a bit deeper though... I don't think we want to change the semantics of the re-inspection thing, which are that it only does something (flips to the Inspecting state) in the Ready state. And we'll need to be able to set the data when the Host is Provisioned, because we're only going to be able to get it from something running on host and that thing will have to be provisioned. So we probably shouldn't share the same key for setting data and triggering re-inspection after all. Accordingly, I'm amending my suggestion to this:
> 
>     * `inspect.metal3.io` - move to Inspecting state from Ready
> 
>     * `inspect.metal3.io=disabled` - don't inspect even in Inspecting
> 
>     * `inspect.metal3.io/hardwaredetails={...}` - set HardwareDetails if empty and delete annotation

This works for me, my main concern with your initial proposal was conflating the disable/re-inspect action with setting the hardware details.

> Open question: should we allow users to replace existing HardwareDetails? At any time? Only in the inspecting state after triggering re-inspection?

For the use-case I'm currently interested in, we need to update HardwareDetails after the live-iso is booted (since that's the thing collecting inspection data), but in that case we'll set `inspect.metal3.io=disabled` to avoid the time hit of inspecting twice.

So I'd say we need to allow updating `inspect.metal3.io/hardwaredetails`:

  * At any time when `inspect.metal3.io=disabled`
  * On BMH creation - in this case inspection will be skipped, and it's a potentially safer/more constrained interface than the current `baremetalhost.metal3.io/status` ?

In other cases, it probably makes sense to not allow update, since it would mean potential loss of metal3-initiated inspection data, and the user told us to handle inspection?

andfasano: > This works for me, my main concern with your initial proposal was conflating the disable/re-inspect action with setting the hardware details.
:+1: 

> So I'd say we need to allow updating `inspect.metal3.io/hardwaredetails`:
> 
>     * At any time when `inspect.metal3.io=disabled`
> 
>     * On BMH creation - in this case inspection will be skipped, and it's a potentially safer/more constrained interface than the current `baremetalhost.metal3.io/status` ?
> 
Logically speaking since it seems to me that when `inspect.metal3.io=disabled` the inspection is delegated to someone else (responsible to provide the required details in the `hardwaredetails` subkey, I would strictly bind the replacement in the same point (state) where the skipping decision will be done

design/baremetal-operator/external-introspection.md
dtantsur: > Particularly for some repeated CI/test cases on VMs I can imagine disabling inspection to reduce the time for test iterations.

If this is the reason, let's not do this. This is a dangerous path to take which will necessarily lead to broken inspection in the end.

design/baremetal-operator/external-introspection.md
hardys: >  I would strictly bind the replacement in the same point (state) where the skipping decision will be done

As I mentioned, there is a requirement to update the hardwaredetails after the live-iso has booted, at which point the BMH will be in the Provisioned state, so we can't only consume the hardwaredetails in the Inspecting state.

That also won't work for externally provisioned hosts since they bypass the Inspecting state, so IMHO it would be simpler to either unconditionally allow `inspect.metal3.io/hardwaredetails` or consume it only when either `inspect.metal3.io=disabled` or there is no existing data in the Status?

hardys: Another option is we only consume the `inspect.metal3.io/hardwaredetails` annotation when `inspect.metal3.io=disabled` and don't consider the externally provisioned host case.

My motivation in mentioning that is to improve the somewhat awkward pattern where we have to create a BMH but pause reconciling until such time as the status details are added:

https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#pausing-reconciliation
https://github.com/openshift/installer/blob/master/pkg/asset/machines/baremetal/hosts.go#L97
https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/usr/local/bin/master-bmh-update.sh#L48

@zaneb @dhellmann what are your thoughts, should we consider this new `inspect.metal3.io/hardwaredetails` as a potential alternative for the `baremetalhost.metal3.io/status` interface, or do we need to ensure reconciliation is paused until the hardware details are added in the externally provisioned case?

zaneb: I vote for Consume it only when either inspect.metal3.io=disabled or there is no existing **_HardwareDetails_** data in the Status. That way we won't need the awkward pause thing in OpenShift any more, but we won't overwrite existing data that has come from an actual inspection.

dhellmann: Yes, we should use this in OpenShift instead of the mechanism we're using now.

I'm less sure we should apply any rules about when the annotation is consumed. It feels like every time we say "we're only going to want this when..." we find out we're wrong. If the only danger is that a user might overwrite "real" data, can we not just tell them "don't do that if you want things to work properly"? That's how APIs work, right?

If we do want to protect the user, then Zane's proposed handling makes sense as a minimal set of safeguards.

hardys: I can think of times an unconditional API to overwrite existing hardware status would be useful for debugging/workarounds - so I'm in agreement with @dhellmann that perhaps we should just make `inspect.metal3.io/hardwaredetails` work unconditionally and clearly document that it overwrites any existing data.

Unlike `baremetalhost.metal3.io/status` it only risks removal of the hardware details (at the users request) and can't corrupt other state we rely on that's stored in the Status fields?

andfasano: :+1: for @zaneb proposal, seems to be better as improved safety guards, and it does not look like it will limit the range of applications as previously discussed.

hardys: Ok it sounds like we've reached consensus on the API but there's still a split of opinion re the hardwaredetails update behavior :sweat_smile: 

Unless there are any further opinions on that, I'll go with the "safer" consume it only if `inspect.metal3.io=disabled` or there is no existing HardwareDetails, then if we find a specific reason to relax that we can consider doing so in future.

design/baremetal-operator/external-introspection.md
hardys: Perhaps walltime optimization was a bad example to use, and I agree we probably shouldn't (ab)use this sort of interface in CI, my point was really that I don't think we should couple the `inspect.metal3.io` annotation to `spec.image.format == live-iso `, since they're essentially two orthogonal interfaces.

As @dhellmann mentions above it's the responsibility of the user to consider the impact of any API use, and provided inspection is enabled by default, and we clearly document the ramifications of disabling it, I don't see much benefit in implementing hidden dependencies between two essentially independent BMH APIs?

dhellmann: I agree with keeping the APIs separate. Our current use case for the new feature has to do with live ISOs, but that's not the only one (we've discussed in another thread on this PR using it in the openshift installer instead of the current method of loading the entire status block).

design/baremetal-operator/external-introspection.md
dhellmann: I'm OK with following the safer approach.

design/baremetal-operator/external-introspection.md
```diff
@@ -0,0 +1,108 @@
...
+
+Given that the primary use-case for this is live-boot images (where no disk
+image is written to disk), and that profile matching is no longer the preferred
+interface for specifying root-disk hints, if `inspect.metal3.io/hardwaredetails`
+is updated and the BMH is in the `Ready` state, we will not attempt to match
+profiles based on this data.
+
+In the event that both `baremetalhost.metal3.io/status` and
+`inspect.metal3.io/hardwaredetails` are specified on BMH creation,
+`inspect.metal3.io/hardwaredetails`` will take precedence and overwrite any
```
andfasano, line 74: typo: additional single quote

hardys: Thanks, fixed!

dhellmann:
I'm happy with this draft.

/approve

hardys:
Ok the lazy consensus period completed and there were no objectsion - @andfasano perhaps you can add the lgtm please?

PR #156, opened 2021-01-06
Design proposal: early cleaning/resetting

dtantsur:


design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,125 @@
...
+// Skip resetting BMC on registration.
+SkipReset bool `json:"skipReset,omitempty"`
+```
+
+### Implementation Details/Notes/Constraints
+
+The `AccessDetails` interface will be extended with a new call:
+
+```go
+ResetSteps() []nodes.CleanStep
```
dhellmann, line 76: The `pkg/bmc` package doesn't currently depend on gophercloud. Since that package is used outside of the baremetal-operator it would be good to keep it that way. Can we express the clean steps in some other way? For example, do we need to be explicit about which steps are run, or can we return a bool to say whether the driver supports or requires resetting? Or, could the provisioner just always reset the host?

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,125 @@
...
+* Run any in-band (i.e. using a ramdisk) clean steps.
+* Allow a user to provide the steps to run.
+
+## Proposal
+
+## Design Details
+
+All clean steps run asynchronously in the ironic API, even out-of-band steps
+that are reasonable fast (an order of seconds), thus we need a new state
+`Resetting` to represent the new procedure. So instead of
```
zaneb, line 54: I'm not aware of a technical reason that this could not be done in the Registering state. Currently we don't actually [do anything](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_state_machine.go#L233-L244) in the Registering state that we don't do in every state, so it would be easy to add something.

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,125 @@
...
+## Motivation
+
+iDRAC will refuse to create a job if there is already a pending job of the same
+type. To make sure we start in a coherent state, we will reset iDRAC before
+doing anything else to it.
+
+### Goals
+
+* Provide a mechanism to run out-of-band clean steps before inspection.
+* Allow a user to opt out of this mechanism on per node basis.
```
zaneb, line 40: Why?

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,130 @@
...
+Registering -> Resetting [label="!externallyProvisioned && NeedsReset()"]
+Registering -> Inspecting [label="!externallyProvisioned && !NeedsReset() && NeedsHardwareInspection()"]
+Resetting -> Inspecting [label="done && NeedsHardwareInspection()"]
+```
+
+`BareMetalHostSpec` is extended with
+
+```go
+// Skip resetting BMC on registration.
+SkipReset bool `json:"skipReset,omitempty"`
```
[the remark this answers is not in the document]
dtantsur: It's a safety precaution in case of hardware that does something (?) wrong when resetting. I'm fine with dropping it.

design/baremetal-operator/early-cleaning.md
dtantsur: Clean steps depend on a driver... The only thing I can suggest is to copy the `CleanStep` definition in `pkg/bmc`. But is it really useful to avoid this dependency? Why would anyone use `pkg/bmc` but not use `gophercloud` directly or indirectly?

design/baremetal-operator/early-cleaning.md
dtantsur: I assumed a new state would be more obvious and debuggable, but I'm fine with dropping it.

design/baremetal-operator/early-cleaning.md
dtantsur: A safety precaution in case of a crazy BMC? Let's drop it for now.

design/baremetal-operator/early-cleaning.md
dhellmann: Tying `pkg/bmc` to gophercloud forces consumers of `pkg/bmc` to use the same version of gophercloud that we do here.

Perhaps we need a new type in a new package to handle the cleaning interface? Maybe that could live inside `pkg/provisioner/ironic`?

design/baremetal-operator/early-cleaning.md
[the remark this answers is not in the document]
dhellmann: Yeah, I think let's leave it out until we have a clear need for it.

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,111 @@
...
+### Implementation Details/Notes/Constraints
+
+The `AccessDetails` interface will be extended with a new call:
+
+```go
+ResetSteps() []CleanStep
+```
+
+where `CleanStep` is a copy of the corresponding type from gophercloud (to
+avoid adding a dependency on `gophercloud` to `pkg/bmc`).
```
dhellmann, line 67: If the information in the response depends only on the driver name, and not on the other BMC settings, we could hide all of this inside the ironic provisioner. That should be the only thing that ever asks for the values, and it would just be a `map[string][]nodes.CleanStep` static data structure, right?

dtantsur: Aaah, I think I understand now.

dtantsur: Although this will require adding driver-dependent code to `pkg/provisioner/ironic`, currently only `pkg/bmc` does it. I'm fine with that, but we need an agreement.

dtantsur: Talked on slack a bit, we should probably keep it as it is to avoid unnecessary complicating the provisioner.

dhellmann: Yeah, I think it's safe to use a copy of the CleanStep data structure in pkg/bmc. It would be good to describe the other approaches we rejected, and why, as part of this document.

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,128 @@
...
+## Motivation
+
+iDRAC will refuse to create a job if there is already a pending job of the same
+type. To make sure we start in a coherent state, we will reset iDRAC before
+doing anything else to it.
+
+### Goals
+
+* Provide a mechanism to run out-of-band clean steps before inspection.
+* Run `known_good_state` step for iDRAC.
```
elfosardo, line 40: nit: clean step

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,128 @@
...
+
+* Ironic needs an ability to skip booting the ramdisk:
+  [RFE 2008491](https://storyboard.openstack.org/#!/story/2008491).
+
+* iDRAC resetting clean steps:
+  [RFE 2007617](https://storyboard.openstack.org/#!/story/2007617).
+
+### Test Plan
+
+Unfortunately, this change can only be tested on real hardware.
```
elfosardo, line 108: and only on iDRAC supported one :/

elfosardo (approved):

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,128 @@
...
+* Allow a user to provide the steps to run.
+
+## Proposal
+
+## Design Details
+
+All clean steps run asynchronously in the ironic API, even out-of-band steps
+that are reasonable fast (an order of seconds). When entering the `Registering`
+state, we will check if the driver supplies any reset clean steps. If yes, we
+will start manual cleaning and wait. Otherwise we'll move on.
```
andfasano, line 54: Can we also specify the expected behavior in case of clean failure (I suppose just retry with exp backoff)?

dtantsur: Isn't it what we always do with all actions?

andfasano: I wouldn't say it happens in all cases. Anyhow from a documentation/design point of view I think it could help describing the expected behaviour in such case.

andfasano: Another point: in the current implementation, the `Registering` state is a simple pass-through (it just decides the transition to `Inspecting` or `ExternallyProvisioned`) - as the registration is revalidated in almost every state before entering them. So maybe it's not the right place for a manual cleaning. As @zaneb already suggested in [1], maybe we could try to converge both the proposals into the newly `Preparing` state (even though the original idea was to have it before `Ready`, while here it seems required even before `Inspecting`)

[1] https://github.com/metal3-io/baremetal-operator/pull/292#pullrequestreview-566740018

dtantsur: Preparing happens too late, we need to run before any other actions. Would you prefer we return back to the initial proposal that added a `Resetting` state?

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,128 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Early out-of-band cleaning
```
zaneb, line 8: Can I suggest we not mention cleaning up-front? It tends to confuse people, and ultimately is just an implementation detail for how we go about resetting the BMC through Ironic.

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,128 @@
...
+## Proposal
+
+## Design Details
+
+All clean steps run asynchronously in the ironic API, even out-of-band steps
+that are reasonable fast (an order of seconds). When entering the `Registering`
+state, we will check if the driver supplies any reset clean steps. If yes, we
+will start manual cleaning and wait. Otherwise we'll move on.
+
+Resetting won't happen for externally provisioned host.
```
zaneb, line 56: Do we also want to reset the BMC when moving from externally provisioned to ready/inspecting? That would be a reason to add a separate Resetting state.

design/baremetal-operator/early-cleaning.md
zaneb: If this is the only place we need to run it then I support re-using the existing registering state. The fact that it has no implementation now is actually an advantage because it means we we'll only have one thing to do.

A bigger question is: how can we tell that the reset is done? The node state will be manageable both before and after, right? AIUI we'll need to store some extra state either in ironic or in the BMH. (A state transition counts, but we can't really transition to a new state as soon as we start the cleaning because it's asynchronous and may fail.)

design/baremetal-operator/early-cleaning.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Rejected Ideas
+
+We could use `nodes.CleanStep`, but it would require `pkg/bmc` to depened on
+gophercloud. In turn, downstream users of `pkg/bmc` would have to depend on
+exactly the same version as we, which was deemed undesirable.
+
+We could move the logic of determining the right clean steps for a node
+completely to `pkg/provisioner/ironic`. That would require introducing
+driver-specific logic, which currenctly lives only in `pkg/bmc`.
```
zaneb, line 87: typo: currently

design/baremetal-operator/early-cleaning.md
andfasano: Fine for me then to stick in the `Registering` phase. And agree about the reset completion issue

design/baremetal-operator/early-cleaning.md
dtantsur: Is it a real transition? I assumed externally provisioned things remain like that?

design/baremetal-operator/early-cleaning.md
dtantsur: > A bigger question is: how can we tell that the reset is done? The node state will be manageable both before and after, right?

Right. It may be a reason to use a new state.

> we can't really transition to a new state as soon as we start the cleaning because it's asynchronous and may fail

Mmm, not sure I get it. Cleaning is asynchronous, but *starting cleaning* is a synchronous operation.

design/baremetal-operator/early-cleaning.md
zaneb: Yes, there's a Spec flag for ExternallyProvisioned and we can't stop the user from clearing it (at least until we implement a validating webhook). We could just ignore it, but [currently](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_state_machine.go#L271-L278) we do allow the host to transition to a normal state. If we didn't then users would have to delete the whole BMH to e.g. reprovision an externally provisioned host.

design/baremetal-operator/early-cleaning.md
zaneb: Right, but if it fails you'd prefer the failure to be recorded in the state which is supposed to be doing the resetting, not whichever state comes after that. The latter would be confusing for users, and would make retries difficult.

The RAID patch has the same problem and they appear to be solving it with a flag in the Status. (In older versions of the patch the cleaning started in the Preparing state and any errors were recorded in the Provisioning state, which was not so bad in that case but wouldn't work as well here.)

design/baremetal-operator/early-cleaning.md
rhjanders: done

design/baremetal-operator/early-cleaning.md
rhjanders: done

design/baremetal-operator/early-cleaning.md
rhjanders: done

design/baremetal-operator/early-cleaning.md
rhjanders: There were a number of options considered here, but I am not sure if we need changes to this section in the end? Please let me know and I will edit accordingly if needed. Thanks!

design/baremetal-operator/early-cleaning.md
rhjanders: I'll remove line 56.

rhjanders:
Let's continue this work in this new PR: https://github.com/metal3-io/metal3-docs/pull/186
I will take this one on for now - and it's easier to start a new PR than to fix permissions on this one.

design/baremetal-operator/early-cleaning.md
rhjanders: done

PR #154, opened 2020-12-21
Add user-guide structure & tooling

fmuyassarov:
This adds an initial structure for Metal3 user-guide and tooling
to build up the guide and do previews.

PR #158, opened 2021-01-09
Add Dockerfile for mdbook

fmuyassarov:
Add Dockerfile for mdbook

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,28 +1,58 @@
+MDBOOK_VERSION ?= 0.4.5
+MDBOOK_BIN_VERSION ?= v0.4.5
+SOURCE_PATH := docs/user-guide
+CONTAINER_RUNTIME ?= sudo docker
+IMAGE_NAME := quay.io/metal3-io/mdbook
+HOST_PORT ?= 3000
 BIN_DIR := hack
 MDBOOK_BIN := $(BIN_DIR)/mdbook
-MDBOOK_VERSION := v0.4.3
-MDBOOK_RELEASE_URL := https://github.com/rust-lang/mdBook/releases/download/$(MDBOOK_VERSION)/mdbook-$(MDBOOK_VERSION)-x86_64-unknown-linux-gnu.tar.gz
-SOURCE_PATH := docs/user-guide
+MDBOOK_RELEASE_URL := https://github.com/rust-lang/mdBook/releases/download/$(MDBOOK_BIN_VERSION)/mdbook-$(MDBOOK_BIN_VERSION)-x86_64-unknown-linux-gnu.tar.gz
 
-## ----------------------
-## Documentation tooling
-## ----------------------
+## ------------------------------------
+## Documentation tooling for Netlify
+## ------------------------------------
 
-$(MDBOOK_BIN):
+# This binary is used by Netlify. Because,
+# Netlify build image doesn't support docker/podman.
+
+$(MDBOOK_BIN): # Download the binary
 	curl -L $(MDBOOK_RELEASE_URL) | tar xvz -C $(BIN_DIR)
 
-.PHONY: serve
-serve: $(MDBOOK_BIN)
-	$(MDBOOK_BIN) serve --open $(SOURCE_PATH)
-
-.PHONY: build
-build: $(MDBOOK_BIN)
+.PHONY: netlify-build
+netlify-build: $(MDBOOK_BIN) # Build the user guide
 	$(MDBOOK_BIN) build $(SOURCE_PATH)
 
-.PHONY: watch
-watch: $(MDBOOK_BIN)
-	$(MDBOOK_BIN) watch --open $(SOURCE_PATH)
+
+## ------------------------------------
+## Documentation tooling for local dev
+## ------------------------------------
+
+.PHONY: container-image
+container-image: # Build mdbook local container image
+	$(CONTAINER_RUNTIME) build --build-arg MDBOOK_VERSION=$(MDBOOK_VERSION) \
+	./docs/ -t $(IMAGE_NAME):$(MDBOOK_VERSION)
+
+.PHONY: build
+build: # Build the user guide
+	$(CONTAINER_RUNTIME) run \
+	--rm -it --name metal3 \
+	-v "$$(pwd):/workdir" \
+	$(IMAGE_NAME):$(MDBOOK_VERSION) \
+	mdbook build $(SOURCE_PATH)
+
+.PHONY: serve
+serve: # Serve the user-guide on localhost:3000 (by default)
+	$(CONTAINER_RUNTIME) run \
+	--rm -it --init --name metal3 \
+	-v "$$(pwd):/workdir" \
+	-p $(HOST_PORT):3000 \
+	$(IMAGE_NAME):$(MDBOOK_VERSION) \
+	mdbook serve --open $(SOURCE_PATH) -p 3000 -n 0.0.0.0
 
 .PHONY: clean
-clean: $(MDBOOK_BIN)
-	$(MDBOOK_BIN) clean $(SOURCE_PATH)
+clean: # Clean mdbook generated content
+	$(CONTAINER_RUNTIME) run \
+	--rm -it --name metal3 \
+	-v "$$(pwd):/workdir" \
+	$(IMAGE_NAME):$(MDBOOK_VERSION) \
+	mdbook clean $(SOURCE_PATH)
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,0 +1,5 @@
+FROM rust:1-slim
+ARG MDBOOK_VERSION="0.4.5"
+RUN cargo install mdbook --vers ${MDBOOK_VERSION}
+RUN cp /usr/local/cargo/bin/mdbook /usr/bin/mdbook
+WORKDIR /workdir
--- a/docs/instructions.md
+++ b/docs/instructions.md
@@ -40,11 +40,7 @@ try to fetch those documents and parse them out.
 
 ## Preview your changes locally
 
-1. Download Mdbook binary. This will save the binary in /hack folder.
-
-    ```bash
-    $ make bin
-    ```
+All the commands below are executed within mdbook container.
 
 1. Build the user-guide.
 
@@ -52,18 +48,25 @@ try to fetch those documents and parse them out.
     $ make build
     ```
 
-1. Preview the book built. This will open the user-guide in your browser.
-    You can keep running `make watch` and continue making doc changes.
-    Mdbook will detect and render the local changes automatically. Refresh
-    the browser page to see the final changes.
+1. Preview the user-guide built before pushing your changes. This will open the
+    user-guide in your browser at `http://localhost:3000/`. Export `HOST_PORT`
+    environment variable with desired port number to serve the user-guide on another
+    port. You can keep running `make serve` and continue making doc changes. Mdbook
+    will detect your changes, render them and refresh your browser page automatically.
 
     ```bash
-    $ make watch
+    $ make serve
     ```
 
-1. Once you have finished local preview, clean Mdbook auto-generated
-    content from docs/user-guide/book path.
+1. Clean Mdbook auto-generated content from docs/user-guide/book path once you
+    have finished local preview.
 
     ```bash
     $ make clean
+    ```
+
+1. Build mdbook container image locally.
+
+    ```bash
+    $ make container-image
     ```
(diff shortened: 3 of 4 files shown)


hack/stop_container.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+$@
```
fmuyassarov, line 3: This is need to be able to stop process (`mdbook serve --open` process) running within mdbook container, from a shell. Otherwise, user will need to open another shell to kill the running container, delete it. I'm not sure if there is a better solution to this.

Makefile
```diff
@@ -0,0 +1,40 @@
...
+	$(CONTAINER_RUNTIME) build . -t $(IMAGE_NAME):$(MDBOOK_VERSION)
+
+.PHONY: build
+build: # Build the user guide
+	$(CONTAINER_RUNTIME) run \
+	--rm -it --name metal3 \
+	-v "$$(pwd):/workdir" \
+	-u "$$(id -u):$$(id -g)" \
+	$(IMAGE_NAME):$(MDBOOK_VERSION) \
+	mdbook build $(SOURCE_PATH)
```
dhellmann, line 22: This PR doesn't build for me

```
$ CONTAINER_RUNTIME=podman make -b build
podman run \
--rm -it --name metal3 \
-v "$(pwd):/workdir" \
-u "$(id -u):$(id -g)" \
quay.io/metal3-io/mdbook:v0.4.5 \
mdbook build docs/user-guide
2021-01-11 14:41:49 [ERROR] (mdbook::utils): Error: Couldn't open SUMMARY.md
2021-01-11 14:41:49 [ERROR] (mdbook::utils): 	Caused By: No such file or directory (os error 2)
make: *** [Makefile:17: build] Error 101
```

I don't see a `SUMMARY.md` in the branch for the PR. Does it need to be rebased?

Makefile
```diff
@@ -0,0 +1,40 @@
+MDBOOK_VERSION := v0.4.5
+SOURCE_PATH := docs/user-guide
+CONTAINER_RUNTIME := docker
```
dhellmann, line 3: In order to be able to override this, I had to change it to
```suggestion
CONTAINER_RUNTIME ?= docker
```

Makefile
fmuyassarov: Sorry I had to mention in the description that it's normal, because we will have the `SUMMARY.md` after https://github.com/metal3-io/metal3-docs/pull/154 is merged. Because that's the PR that introduces `SUMMARY.md`

fmuyassarov: If you want to try it out, you can copy [/docs/user-guide](https://github.com/Nordix/metal3-docs/tree/user-guide/feruz/docs/user-guide) directory which includes the `SUMMARY.md`

dhellmann: #154 merged, and this branch now has conflicts. I'll try again after you rebase to fix those conflicts and fix the variable definition.

fmuyassarov:
Based on the master branch, this is the initial view of the user-guide https://metal3-user-guide.netlify.app/.
Netlify is failing on this PR because it doesn't have docker/podman installed to run docker/podman based Makefile targets. 
We can add a shell script to install Docker, and then Netllify would run that script every time it gets triggered. Even with this option, it actually seems that Netlify doesn't allow to install some packages. I have tried with `sudo apt install ...` and it complained that `sudo` doesn't exist, and when as unprivileged user it was failing requiring root privileges  :) 

With sudo
```
12:11:26 AM: $ ./hack/install_docker.sh
12:11:26 AM: ./hack/install_docker.sh: line 3: sudo: command not found
12:11:26 AM: ./hack/install_docker.sh: line 5: sudo: command not found
12:11:26 AM: ./hack/install_docker.sh: line 12: sudo: command not found
12:11:27 AM: (23) Failed writing body
12:11:27 AM: ./hack/install_docker.sh: line 14: sudo: command not found
12:11:27 AM: ./hack/install_docker.sh: line 19: sudo: command not found
12:11:27 AM: ./hack/install_docker.sh: line 20: sudo: command not found
12:11:27 AM: ./hack/install_docker.sh: line 22: docker: command not found
```
As unprivileged user
```
12:09:21 AM: $ ./hack/install_docker.sh
12:09:22 AM: Reading package lists...
12:09:22 AM: W: chmod 0700 of directory /var/lib/apt/lists/partial failed - SetupAPTPartialDirectory (1: Operation not permitted)
12:09:22 AM: E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
12:09:22 AM: E: Unable to lock directory /var/lib/apt/lists/
12:09:22 AM: E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
12:09:22 AM: E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
12:09:22 AM: ./hack/install_docker.sh: line 12: sudo: command not found
12:09:23 AM: (23) Failed writing body
12:09:23 AM: Error: must run as root
12:09:24 AM: Reading package lists...
12:09:24 AM: W: chmod 0700 of directory /var/lib/apt/lists/partial failed - SetupAPTPartialDirectory (1: Operation not permitted)
12:09:24 AM: E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
12:09:24 AM: E: Unable to lock directory /var/lib/apt/lists/
12:09:24 AM: E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
12:09:24 AM: E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
12:09:24 AM: ./hack/install_docker.sh: line 22: docker: command not found
12:09:24 AM: ​
12:09:24 AM: ────────────────────────────────────────────────────────────────
12:09:24 AM:   "build.command" failed                                    
```

fmuyassarov:
@dhellmann , @maelk I think this PR is ready for the review. A couple of notes here:
1. I added `--init ` flag for `mdbook serve` command to able to cancel the process after we finish the local preview. Without this we will need to kill the container manually, because Ctrl+C is not accepted.  `--init` only way I could find. Initially I added a script which I was injecting into the container. But this option seems a bit easier and cleaner to me. 
2. I added netlify.toml file at the root directory, which is used by Netlify. So, Netlify will pick up the file and run whatever is given under ` command` section. In our case it is `make netlify-build`
3. In the Makefile we still download mdbook binary and use the binary to build the user-guide via Netlify. That's because, Netlify build image doesn't have docker/podman installed by default and we can't even install Ubuntu packages (by default Netlify uses Ubuntu 16 image). You can see the reference [here](https://community.netlify.com/t/cant-install-packages/29985).
As such, in the Makefile we have two sections. 

    1. For netlify (download the binary & build)
    2. For developers ( container based version)

Makefile
```diff
@@ -1,28 +1,57 @@
+MDBOOK_VERSION ?= v0.4.5
+SOURCE_PATH := docs/user-guide
+CONTAINER_RUNTIME ?= docker
+IMAGE_NAME := fmuyassarov/mdbook
```
maelk, line 4: That does not look correct :)

docs/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM rust:1-slim
+ARG MDBOOK_VERSION="0.4.5"
```
maelk, line 2: This should come from the makefile to make sure the two are in sync.

fmuyassarov: Are we expecting the quay to run Makefile target to build the image? Also, do I need to add a push or will it be configured within the quay?

Makefile
fmuyassarov: Ops, forgot to clean up my tests.

Makefile
dhellmann: Now I get

```
$ CONTAINER_RUNTIME=podman make build
podman run \
--rm -it --name metal3 \
-v "$(pwd):/workdir" \
-u "$(id -u):$(id -g)" \
fmuyassarov/mdbook:v0.4.5 \
mdbook build docs/user-guide
2021-01-12 15:24:56 [INFO] (mdbook::book): Book building has started
2021-01-12 15:24:56 [INFO] (mdbook::book): Running the html backend
2021-01-12 15:24:56 [ERROR] (mdbook::utils): Error: Rendering failed
2021-01-12 15:24:56 [ERROR] (mdbook::utils): 	Caused By: Unable to remove stale HTML output
2021-01-12 15:24:56 [ERROR] (mdbook::utils): 	Caused By: Permission denied (os error 13)
make: *** [Makefile:35: build] Error 101
```

Removing the `-u` option makes the build work. Is that option required for docker?

docs/Dockerfile
dhellmann: Quay uses the Dockerfile directly.

Makefile
fmuyassarov: That's weird. It works fine for me though
```
fmuyassarov@est:~/final/metal3-docs$ make build
docker run \
--rm -it --name metal3 \
-v "$(pwd):/workdir" \
-u "$(id -u):$(id -g)" \
fmuyassarov/mdbook:v0.4.5 \
mdbook build docs/user-guide
2021-01-12 16:11:27 [INFO] (mdbook::book): Book building has started
2021-01-12 16:11:27 [INFO] (mdbook::book): Running the html backend
```
If we don't use -u, then the generated book directory will have root:root ownership
```
~metal3-docs/docs/user-guide$ ls -lah
total 24K
drwxrwxr-x 5 fmuyassarov fmuyassarov 4.0K Jan 12 18:12 .
drwxrwxr-x 3 fmuyassarov fmuyassarov 4.0K Jan 11 22:24 ..
drwxr-xr-x 6 root        root        4.0K Jan 12 18:12 book
-rw-rw-r-- 1 fmuyassarov fmuyassarov  207 Jan 11 17:31 book.toml
drwxrwxr-x 3 fmuyassarov fmuyassarov 4.0K Jan 11 17:31 src
drwxrwxr-x 2 fmuyassarov fmuyassarov 4.0K Jan 11 17:31 theme
```
So, with `-u` we could make sure that the `book` directory is under `${USER}:${USER}`, which would be needed for `make serve`. But now, I think it would make sense to add `sudo` for docker/podman commands, so that we don't have to worry even if `book` directory is `${USER}:${USER}', because `make serve` also will run as root. 

@dhellmann can you please run `make clean & make build` ? I have removed `-u` flag.

docs/Dockerfile
fmuyassarov: If quay uses dockerfile directly, then we probably can control the var in Dcokerfile itself @maelk  ?

fmuyassarov:
@dhellmann  Once this PR is merged, can you please help to set up the image building in Quay? I don't have rights to configure it myself.

Makefile
dhellmann: sudo really shouldn't be needed. Do we need a script to wrap the call to mdbook to change the ownership of the output?

fmuyassarov: Actually no. Mdbook doesn't set the root ownership. I think it was misconfiguration in the docker flag. Agree with you. I am going to remove it now. Even without the sudo, -u I'm able to run successfully. Please let me know if that's the same for you too with podman.

dhellmann: Part of the difference may be in how podman and docker run the container. Maybe the `sudo` should be part of the default value of `CONTAINER_RUNTIME` so when someone overrides it with `podman` the `sudo` can be dropped.

fmuyassarov: True. By default docker requires you to run with sudo, which isn't the case for podman. So, I thought adding sudo won't even users running podman.

fmuyassarov: > Part of the difference may be in how podman and docker run the container. Maybe the `sudo` should be part of the default value of `CONTAINER_RUNTIME` so when someone overrides it with `podman` the `sudo` can be dropped.

Yes, I think this will make sense.

fmuyassarov: Updated it.

dhellmann:
This is working well for me now.

/approve

docs/Dockerfile
maelk: My point was not about the quay building, but the container-image makefile command, that should pass the variable coming from the makefile. Otherwise there is a risk that if they are out of sync, a version x image would be built when the user expects a version y (set in makefile). This could be fixed by using something like --build-arg in the command. 
I don't really expect this command to be used a lot, so I'm also fine leaving things as they are now.

fmuyassarov: True, if only for build local copy of the image, then controlling the var with` --build-arg` make sense. Update it, thanks.

maelk:
I have set up the quay.io building triggers

fmuyassarov:
> I have set up the quay.io building triggers

Thank you!

PR #162, opened 2021-02-04
✨ Proposal: node reuse

furkatgofurov7:
This PR proposes node reuse feature for the upgrade/remediation operations.

Related to: [allow disabling disk cleaning](https://github.com/metal3-io/metal3-docs/blob/master/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md)
Co-authored by: @fmuyassarov


dhellmann:
I'm shifting my focus a bit to an internal project, so it's probably best not to block designs like this on my review.

/unassign dhellmann
/assign zaneb

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,250 @@
...
+utilized by other M3M(s) after it has been deprovisioned, the `consumerRef`
+field will store the name of KCP/MD even when it is in `Ready` state. During the
+upgrade, after KCP/MD creates new Machines, CAPM3 will create new M3M(s) and will
+filter out BMH(s) that have the name of KCP/MD in the `consumerRef` field. If the
+name of the KCP/MD matches the name set in `spec.consumerRef.name` field of BMH,
+then CAPM3 Machine controller will associate that BMH to the M3M.
+
+As such, even after deprovisioning we will keep the `consumerRef` on a BMH and
+when M3M will be requesting a new BMH, CAPM3 will filter BMH(s) based on their
+consumerRef and will pick up those that match the name of KCP/MD.
```
hardys, line 86: Maintaining state like this in consumerRef doesn't seem ideal, we already have https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/api/v1alpha4/metal3machine_types.go#L49 that could be used to filter BMH resources based on some label - would that be sufficient to address this requirement?

fmuyassarov: We can rely on label/annotation, and we had a thought about using annotation first. But we thought relying on labels/annotation is not the recommended to manage a behavior. As such, we tried to avoid tuning the functionality via labels/annotation.

fmuyassarov: I agree that we can just leave a label on a BMH during de-provisioning, and pick that up while next provisioning by relying on `HostSelector` to find it out. TBH, I don't have a strong opinion about which option would be the ideal. But if you think labeling is good enough in our case, then we can update the proposal. 

WDYT @maelk

maelk: Setting the consumer ref would effectively reserve that BMH, by showing that it is currently consumed by the KCP / MD object. I think that this is the goal of that field of the API, so I am not sure why we should add a separate API to achieve the same thing.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Summary
+
+Node reuse during the upgrade/remediation operation.
+
+## Motivation
+
+While performing upgrade operation and potentially remediation, it is desirable
+to be able to reuse the same pool of hosts during re-provisioning (i.e. when
+BareMetalHost is going through deprovisioning -> provisioning). However, there
```
andfasano, line 22: What's the effective benefit of reusing the same pool? I think it could be useful to clarify better (ie in term of costs/time saving) such reasons, especially in relationship to the cited scenario (upgrade and/or remediation)

fmuyassarov: So the one case that can benefit from this is upgrade. During the upgrade of Kubeadm Control Plane (KCP) machines, for example we upgrade the K8s version from 1.19 to 1.20, KCP will delete old Machines, thus Metal3Machines will be deleted. As such, BMHs de-provisioning will start. Since this is an upgrade operation, we don't want to lose the date on the external disks that are attached to the hosts. For that we first need to make sure that we disable disk cleaning ([proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md) and WIP PRs  [BMO](https://github.com/metal3-io/baremetal-operator/pull/784), [CAPM3](https://github.com/metal3-io/cluster-api-provider-metal3/pull/156)) before de-provisioning, and then we try to pick those BMHs that are now in Ready state with external disks left untouched.

I agree that we should mention this explicitly. Will add a note.

furkatgofurov7: Agree, _data loss_ can be both costly and time-consuming in case it happens, we can definitely mention it as this will save both time/costs(for retrieving data)

fmuyassarov: @andfasano, we have updated the motivation part now, please take a look and let us know if you think we need more clarification. Thank you.

andfasano: Thanks @fmuyassarov motivations are clear now for me

design/cluster-api-provider-metal3/node_reuse.md
andfasano: I tend to agree to @hardys comment, as it seems more in line with the management of similar problems in other contexts, for example [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/). ConsumerRef may be seen more as a technical details and implicit, whereas the usage of a label could be more clear and explicit as a public API (to indicate the desire of reuse)

fmuyassarov: Hi @andfasano  @hardys ,
@furkatgofurov7 has now updated the proposal to use a new label `infrastructure.cluster.x-k8s.io/node-reuse` rather than consumerRef. However, @maelk  pointed out that we can't use HostSelector, but need to add something similar due to the following:

When nodeReuse is true, and if the user wants to scale out KCP/MD, then HostSelector will try to find a BMH with `infrastructure.cluster.x-k8s.io/node-reuse`. However, it is obvious that there won't be any BMH available with the label since it is scale out, so scale won't happen. That's why we need to add some sort of similar logic as HostSelector, which would be responsible to filter out BMHs only for non scale-out use cases. When it is scale out, we still would rely on HostSelector, as we do it now.

Does it make sense to you all ?

furkatgofurov7:
@hardys @andfasano proposal has been updated, PTAL. Thanks!

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+If the name of the KCP/MD matches the name set in
+`infrastructure.cluster.x-k8s.io/node-reuse` label of BMH, then CAPM3 Machine
+controller will associate that BMH to the M3M.
+
+As such, even after deprovisioning we will keep the
+`infrastructure.cluster.x-k8s.io/node-reuse` on a BMH and when M3M will be
+requesting a new BMH, CAPM3 will filter BMH(s) based on their label and will
+pick up those that match the name of KCP/MD.
+
+**Example flow:**
```
andfasano, line 104: More than "example flow" I think that below are reported the details of three specific scenarios (and maybe they could be moved in the `Implementation Details/Notes/Constraints` section?) - so that they have been also named properly. Are there any other relevant scenarios to be considered (for example, the simple creation)?

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+    At this point, CAPM3 Machine controller will look in the newly created M3M(s)
+    for annotation `clonedFrom` which will be always pointing to the M3MTemplate.
+    In the following step, it will find the `NodeReuse` field in the M3MTemplate
+    and:
+    * If it is set to **True:**
+      * CAPM3 Machine controller will filter out ready BMH(s) labelled with
+      `infrastructure.cluster.x-k8s.io/node-reuse` label set to KCP/MD name:
+        * If BMH is found in Ready state:
+          * Pick that BMH and attach it to newly created M3M;
+          * Clean up the `infrastructure.cluster.x-k8s.io/node-reuse` label from
```
andfasano, line 133: Why is it required the cleanup? Do you mean that the entire label will be removed or just its value?

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+2. be able reuse the same pool of `BareMetalHost` so that we get the storage
+  data back. Currently, there is no mechanism available in Metal³ to pick the
+  same pool of hosts that were released while upgrading/remediation - for the
+  next provisioning phase. And this proposal tries to solve it.
+
+**NOTE:** This proposal focuses on upgrade use cases but it isn't limited to
+only upgrade. Other use cases like remediation can also benefit from this
+feature as well.
+
+### User Stories
```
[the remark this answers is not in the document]
furkatgofurov7: Fixed, thanks for the heads-up.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+  generation: 1
+    labels:
+      cluster.x-k8s.io/cluster-name: test1
+      infrastructure.cluster.x-k8s.io/node-reuse: KCP/MD name # new label in BMH.
+      ...
+status:
+  provisioning:
+    state: ready
+    ...
+```
```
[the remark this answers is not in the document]
furkatgofurov7: Thanks, we updated those parts, please have a look.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+          * Clean up the `infrastructure.cluster.x-k8s.io/node-reuse` label from
+          BMH;
+        * If BMH is found and is not in Ready state (but is particularly in
+         Deprovisioning state):
+          * Requeue until that BMH becomes Ready;
+        * Else if BMH is not found:
+          * Pick any random BMH that's available (even though it has no label);
+    * If it is set to **False:**
+      * Normal operation continues, where random BMH will be picked up by CAPM3
+      Machine controller.
```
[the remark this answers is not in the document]
furkatgofurov7: That was a good point. We added a diagram that describes both scenarios in one go(differences properly highlighted as suggested) and tried to simplify the implementation details/notes as much as possible. Hopefully, it looks better than the previous version now.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+
+**Some special cases:**
```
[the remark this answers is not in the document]
furkatgofurov7: Added, thanks.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,278 @@
...
+
+* In case, when we found less BMHs with that specific label, then we choose BMH
```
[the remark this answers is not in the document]
furkatgofurov7: Fixed.

design/cluster-api-provider-metal3/node_reuse.md
furkatgofurov7: We have moved them under `Implementation Details/Notes/Constraints` section.

design/cluster-api-provider-metal3/node_reuse.md
furkatgofurov7: In case when the user changes from using `nodeReuse` to **not** using `nodeReuse` feature, we were thinking of removing the whole label so that node should be available for other pools. Would that make sense @andfasano ?

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,200 @@
...
+
+Step2 can be done as follows:
+
+- CAPM3 controller filters host labelled with
+  `infrastructure.cluster.x-k8s.io/node-reuse: md-pool1` label.
+
+  Next:
+  - If host is found in `Ready` state:
+    - Pick that host for newly created M3M;
+    - Once it is picked up, remove the whole label from the host.
```
andfasano, line 148: `whole` -> `infrastructure.cluster.x-k8s.io/node-reuse: md-pool1`?

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,200 @@
...
+  `infrastructure.cluster.x-k8s.io/node-reuse: md-pool1` label, while for md-pool2,
+  it selects any random host, but not those with
+  `infrastructure.cluster.x-k8s.io/node-reuse: md-pool1` label that belong to md-pool1.
+
+**What if we found N-1 hosts with label, while we expect N hosts?**
+  Then we select any other hosts, even though it has no label we expect.
+
+The following diagram shows the example flow of node reuse feature in details:
+
+![node_reuse diagram](../images/node_reuse.png)
```
andfasano, line 169: I've added some additional specific comments for the diagram directly in the source file. In general I think it could really help, but the current version needs to be worked out a little bit more to remove some ambiguities and simplify it.

design/cluster-api-provider-metal3/node_reuse.md
andfasano: I'm personally not a big fan of label removals (usually applied in one-shot events), as they could add a time dependency on the flow, but I think it could work fine in this situation

design/cluster-api-provider-metal3/node_reuse.md
fmuyassarov: Fixed

design/cluster-api-provider-metal3/node_reuse.md
fmuyassarov: Hi @andfasano 
Yes sorry for that. We also felt the same. And we got the review just before pushing the updated versions :).
Please take a look once more, and let us know if they need to be updated. Thanks

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,225 @@
...
+    - Requeue until that host becomes `Ready`;
+  - If no host is found, while it should be (i.e for some reason host
+    is not in the cluster anymore):
+    - Fall back to the current flow, which selects host randomly.
+
+<p align="center">
+    <img src="../images/node_reuse_provisioning.png"/>
+    <br>
+    <em>diagram-2: Example flow of provisioning process</em>
+</p>
```
andfasano, line 167: See above

andfasano (commented):
In general the main part of the proposal looks definitely in a better shape. I have added one question for a specific scenario that I think it could be worth to clarify, and some suggestions to improve the diagrams.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,225 @@
...
+field, or on `KubeadmControlPlane`; for example, a user modifies Kubernetes
+version via `spec.version` field. Once update takes place, owning controller
+(e.g. KCP controller) starts rolling upgrade. As such, CAPI `Machines`, CAPM3
+`Metal3Machines` will be re-created based on the KCP changes. And of course,
+`BareMetalHosts` (below will be referenced as host) which are owned by
+`Metal3Machines`, will be deprovisioned. Normally, when host is
+getting deprovisioned, Ironic cleans up root and externally attached disks on
+the host. However, while performing upgrade operation we don't want hosts
+external disk data to be cleaned so that when host becomes provisioned
+again, it has still the disk data untouched. To achieve this, we need to
```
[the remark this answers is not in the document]
fmuyassarov: Done.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Implementation Details/Notes/Constraints
+
+#### Node Reuse Approach
+
+We should perform two steps to re-use the same pool of hosts.
+
+1. During host deprovisioning, set the `infrastructure.cluster.x-k8s.io/node-reuse`
+  label;
+2. During next provisioning, filter out hosts with the label.
```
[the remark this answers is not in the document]
fmuyassarov: Done.

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,225 @@
...
+
+**What if user changes from node reuse to no node reuse in the middle of upgrade?**
+
+  In the middle of the upgrade process, when hosts (already labelled during
+  deprovisioning) are provisioning, user disables nodeReuse. Based on the
+  diagram-2 above, CAPM3 controller always picks labelled hosts (when found)
+  regardless of nodeReuse feature is disabled/enabled. The reason is, we want to
+  ensure that the controller doesn't miss labelled hosts. In short, labelled
+  hosts always take the presedence over unlabelled hosts during selection.
+
```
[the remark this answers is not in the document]
fmuyassarov: Controller will always try to select hosts with the label, no matter if node reuse is enabled/disabled. As such, during the next upgrade/re-provisioning, if there is a host with the label, then yes controller will pick that host (even though it was labelled before the last deprovisioning).
Why do you think that host won't be selected? Do you mean in some corner cases, due to some issues that while selection, that host with the label wasn't available (for some reason), but became available after controller has picked another host?

design/cluster-api-provider-metal3/node_reuse.md
```diff
@@ -0,0 +1,225 @@
...
+Step1 can be done as follows:
+
+- CAPM3 controller sets the `infrastructure.cluster.x-k8s.io/node-reuse: md-pool1`
+  label on the hosts belonging to the same MD/KCP.
+
+<p align="center">
+    <img src="../images/node_reuse_deprovisioning.png"/>
+    <br>
+    <em>diagram-1: Example flow of deprovisioning process</em>
+</p>
```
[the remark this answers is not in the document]
fmuyassarov: @andfasano how about this diagram ?
xy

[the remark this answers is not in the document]
fmuyassarov: I can replace the order of provisioning and deprovisioning if needed. But it is easier to understand this why I think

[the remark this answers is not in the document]
fmuyassarov: If you think above diagram is okay, we will update the proposal as well. Thanks.

design/cluster-api-provider-metal3/node_reuse.md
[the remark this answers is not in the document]
andfasano: Yes, I was guessing if there was any corner case such that, and if it could had any drawbacks, as it means that the reuse procedure may have not succeeded (since an host marked for reuse wasn't picked up). I guess that probably it would be sufficient to log such scenario (ie, when nodeReuse is set but no hosts were found with the `infrastructure.cluster.x-k8s.io/node-reuse` during the provisioning)

design/cluster-api-provider-metal3/node_reuse.md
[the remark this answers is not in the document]
andfasano: Very nice, much better :+1: (I think that you could just add the label removal step and it's ok)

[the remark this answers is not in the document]
fmuyassarov: Ah true. Will add it and update it in the proposal. Thank you.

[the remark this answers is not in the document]
fmuyassarov: Updated.

design/cluster-api-provider-metal3/node_reuse.md
fmuyassarov: Fixed.

design/cluster-api-provider-metal3/node_reuse.md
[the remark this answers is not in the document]
furkatgofurov7: @andfasano sure thing, we will tackle the logging of such scenario in the implementation.

furkatgofurov7:
@andfasano the proposal has been updated and we addressed the latest comments(added the label removal step in the diagram). PTAL. Thank you!

andfasano (approved):
The current proposal looks fine to me. Leaving to @hardys the final comments

maelk (approved):
You should specify that md-pool1 is for the machine deployment named pool1, and if it was  KCP, it would be kcp-pool1 . This is not clear, and could be confusing if md and KCP have same names otherwise

furkatgofurov7:
> You should specify that md-pool1 is for the machine deployment named pool1, and if it was KCP, it would be kcp-pool1 . This is not clear, and could be confusing if md and KCP have same names otherwise

Thanks for review, updated.

design/images/node_reuse_flow.svg
```diff
@@ -0,0 +1 @@
...="matrix(0.043827160493827164,0,0,0.043827160493827164,15.733950617283952,0)" xlink:href="#X"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,24.499382716049386,0)" xlink:href="#ad"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,32.38827160493827,0)" xlink:href="#U"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,35.85061728395061,0)" xlink:href="#ae"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,43.7395061728395,0)" xlink:href="#U"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,47.20185185185184,0)" xlink:href="#X"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,55.967283950617265,0)" xlink:href="#V"/></g><g id="P"><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,0,0)" xlink:href="#T"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,8.765432098765432,0)" xlink:href="#X"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,17.530864197530864,0)" xlink:href="#ae"/><use transform="matrix(0.043827160493827164,0,0,0.043827160493827164,25.419753086419753,0)" xlink:href="#Y"/></g></defs></g></svg>
```
hardys, line 1: It'd probably be better to put the SVG into a separate file?

hardys:
> The current proposal looks fine to me. Leaving to @hardys the final comments

This version looks good to me, thanks!  One nit re the inlined SVG, that would perhaps be better as a separate file?

Otherwise happy for this to merge, but we agreed a few days of lazy consensus in the meeting yesterday.

design/images/node_reuse_flow.svg
furkatgofurov7: Hi @hardys, thanks for the review! Not fully sure, what you meant by putting the SVG into a separate file. Should we move _node_reuse_flow.svg_  from design/images to another folder or we better make it .png to have it under images?

fmuyassarov:
> > The current proposal looks fine to me. Leaving to @hardys the final comments
> 
> This version looks good to me, thanks! One nit re the inlined SVG, that would perhaps be better as a separate file?
> 
> Otherwise happy for this to merge, but we agreed a few days of lazy consensus in the meeting yesterday.

Thanks. As agreed yesterday in community meeting we will have lazy consensus until next community meeting on Wednesday (10.03) to give time for others too.

hardys:
Lazy consensus period completed and agreed to merge in the community meeting today.

/lgtm

PR #165, opened 2021-02-12
Update Remediation Controller proposal 

jan-est:
This PR updates Remediation Controller proposal with more detailed controller logic and Metal3Remediation CRD specification.

### DIFF
--- a/design/capm3-remediation-controller-proposal.md
+++ b/design/capm3-remediation-controller-proposal.md
@@ -95,7 +95,7 @@ own remediation controller to reconcile External Machine Remediation (EMR) CRs.
 EMRs are generated using infrastructure provider defined External Remediation
 Template.
 
-CAPM3 needs a new controller, rmediation controller to watch and reconcile
+CAPM3 needs a new controller, remediation controller to watch and reconcile
 Metal3Remediation objects. New controller does not have effect on existing APIs
 and adding new remediation features does not include changes in other CAPM3
 controllers. This design adds two new CRDs, namely Metal3Remediation and
@@ -104,24 +104,28 @@ Metal3RemediationTemplate.
 #### Remediation Controller
 
 * MRC watches for the presence of Metal3Remediation CR.
-* Based on the remediation strategy defined in ```.spec.strategy``` in
+* Based on the remediation strategy defined in ```.spec.strategy.type``` in
   Metal3Remediation, RC uses BMO APIs to get hosts back into a healthy or
   manageable state.
 * RC uses ```.status.phase``` to save the states of the remediation. Available
-  states are running, rebooting, power off, failed.
+  states are running, waiting, deleting machine.
 * After RC have finished its remediation, it will wait for the Metal3Remediation
   CR to be removed. (When using CAPI MachineHealthCheck controller,
   MHC will noticed the Node becomes healthy and deletes the instantiated
   MachineRemediation CR.).
-* If RCs timeout for Node to become healthy expires, it annotates BareMetalHost
-  with ```capi.metal3.io/unhealthy``` annotation, sets BMH Online field
-  to False and removes the Machine object.
+* If RCs timeout for Node to become healthy expires, it sets
+  `capi.MachineOwnerRemediatedCondition` to False on Machine object
+  to start deletion of the unhealthy Machine and the corresponding Metal3Remediation.
+* If RCs timeout for Node to become healthy expires, it annotates BareMetalHost with
+  `capi.metal3.io/unhealthy` annotation, sets BMH Online field to False and
+  removes the Machine object.
 
 #### Remediation operations
 
 * BMO reboot annotation: ```reboot.metal3.io```
 * BMO power off/on annotation: ```reboot.metal3.io/{key}```
 * CAPM3 unhealthy annotation: ```capi.metal3.io/unhealthy```
+* CAPI condition: ```capi.MachineOwnerRemediatedCondition```
 
 #### Watch
 
@@ -132,6 +136,7 @@ Metal3RemediationTemplate.
 
 * Fetch a Machine matching the name of the Metal3Remediation and operate over
   BareMetalHost objects.
+* Sets ```capi.MachineOwnerRemediatedCondition``` to False on Machine objects.
 
 #### Metal3Remediation CRD
 
@@ -143,15 +148,17 @@ metadata:
   namespace: NAMESPACE_OF_UNHEALTHY_MACHINE
   ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1alphaX
-     kind: MachineHealthCheck
-     name: REMEDIATION_GROUP
+     kind: Machine
+     name: MACHINE_NAME
 spec:
-  type: reboot
-  rebootLimit: 3
-  timeout: 300s
+  strategy:
+    type: reboot
+    retryLimit: 3
+    timeout: 300s
 status:
-  phase: rebooting
+  phase: running
   retryCount: 1
+  lastRemediated: LAST_REMEDIATION_TIMESTAMP
 ```
 
 #### Metal3RemediationTemplate CRD


dhellmann:
/cc @n1r1 @beekhof 

Nir & Andrew, both of you are more familiar with this area than I am. Could you take a look at this update, please?

design/capm3-remediation-controller-proposal.md
```diff
@@ -104,24 +104,28 @@ Metal3RemediationTemplate.
 #### Remediation Controller
 
 * MRC watches for the presence of Metal3Remediation CR.
-* Based on the remediation strategy defined in ```.spec.strategy``` in
+* Based on the remediation strategy defined in ```.spec.strategy.type``` in
   Metal3Remediation, RC uses BMO APIs to get hosts back into a healthy or
   manageable state.
 * RC uses ```.status.phase``` to save the states of the remediation. Available
-  states are running, rebooting, power off, failed.
+  states are running, waiting, deleting machine.
```
n1r1, line 111: what does `waiting` mean? waiting for what?
I'm also not sure about `running`. maybe it worth to describe each state?

n1r1, line 111: what does `waiting` mean? waiting for what?
I'm also not sure about `running`. maybe it worth to describe each state?

design/capm3-remediation-controller-proposal.md
```diff
@@ -104,24 +104,28 @@ Metal3RemediationTemplate.
...
   MachineRemediation CR.).
-* If RCs timeout for Node to become healthy expires, it annotates BareMetalHost
-  with ```capi.metal3.io/unhealthy``` annotation, sets BMH Online field
-  to False and removes the Machine object.
+* If RCs timeout for Node to become healthy expires, it sets
+  `capi.MachineOwnerRemediatedCondition` to False on Machine object
+  to start deletion of the unhealthy Machine and the corresponding Metal3Remediation.
+* If RCs timeout for Node to become healthy expires, it annotates BareMetalHost with
+  `capi.metal3.io/unhealthy` annotation, sets BMH Online field to False and
+  removes the Machine object.
```
n1r1, line 121: So after machine deletion, we expect a new machine to be created.
does this new machine will be able to use that powered off host in case other hosts are not available? will BMO change the online state?

design/capm3-remediation-controller-proposal.md
```diff
@@ -143,15 +148,16 @@ metadata:
   namespace: NAMESPACE_OF_UNHEALTHY_MACHINE
   ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1alphaX
-     kind: MachineHealthCheck
-     name: REMEDIATION_GROUP
+     kind: Machine
```
n1r1, line 151: I'm wondering if we should should add two ownerRefs. One for `Machine` and one for `MachineHealthCheck`.

If we don't add ownerRef to `MachineHealthCheck`, the following scenario may occur:

1. Machine goes unhealthy
2. MHC creates a `Metal3Remediation` CR
3. User deletes corresponding `MachineHealthCheck` CR
4. RC remediates the machine, but MHC won't delete the `Metal3Remediation` CR, since there's no `MachineHealthCheck` associated with the machine. RC will think that the remediation failed, although it succeeded, and will delete the machine eventually.

design/capm3-remediation-controller-proposal.md
beekhof: Perhaps `in-progress` or `active` instead of `running`
IIRC `waiting` meant that we'd performed an action (like reboot) and we're waiting to see if the node returns before the timeout.

design/capm3-remediation-controller-proposal.md
```diff
@@ -104,24 +104,28 @@ Metal3RemediationTemplate.
...
+  states are running, waiting, deleting machine.
 * After RC have finished its remediation, it will wait for the Metal3Remediation
   CR to be removed. (When using CAPI MachineHealthCheck controller,
   MHC will noticed the Node becomes healthy and deletes the instantiated
   MachineRemediation CR.).
-* If RCs timeout for Node to become healthy expires, it annotates BareMetalHost
-  with ```capi.metal3.io/unhealthy``` annotation, sets BMH Online field
-  to False and removes the Machine object.
+* If RCs timeout for Node to become healthy expires, it sets
+  `capi.MachineOwnerRemediatedCondition` to False on Machine object
```
beekhof, line 117: Whats the motivation here if we immediately remove the Machine anyway?

design/capm3-remediation-controller-proposal.md
```diff
@@ -143,15 +148,16 @@ metadata:
...
-     name: REMEDIATION_GROUP
+     kind: Machine
+     name: MACHINE_NAME
 spec:
   type: reboot
-  rebootLimit: 3
+  retryLimit: 3
   timeout: 300s
 status:
   phase: rebooting
```
beekhof, line 158: We said above that the `reboot` phase doesn't exist anymore right?

design/capm3-remediation-controller-proposal.md
```diff
@@ -143,15 +148,16 @@ metadata:
...
    - apiVersion: cluster.x-k8s.io/v1alphaX
-     kind: MachineHealthCheck
-     name: REMEDIATION_GROUP
+     kind: Machine
+     name: MACHINE_NAME
 spec:
   type: reboot
-  rebootLimit: 3
+  retryLimit: 3
   timeout: 300s
```
beekhof, line 156: These need to belong to the `spec.strategy` field don't they?

design/capm3-remediation-controller-proposal.md
```diff
@@ -143,15 +148,16 @@ metadata:
   namespace: NAMESPACE_OF_UNHEALTHY_MACHINE
   ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1alphaX
-     kind: MachineHealthCheck
-     name: REMEDIATION_GROUP
+     kind: Machine
+     name: MACHINE_NAME
 spec:
   type: reboot
```
beekhof, line 154: An enumeration of all the `spec.type`  values that we plan to implement would be a good idea.

beekhof (changes requested):

design/capm3-remediation-controller-proposal.md
jan-est: Yes, `waiting` is state where we have reboot the host and we wait the timeout. `Running` is the state where our controller has detected the remediation request but the first reboot is not done yet.

design/capm3-remediation-controller-proposal.md
jan-est: This is more for the future work to support multiple remediation strategies. I think `rebootLimit` naming bound us to reboot strategy only.

design/capm3-remediation-controller-proposal.md
jan-est: I would use Machine OwnerRef only. What happens if the RC sets a condition and the `Machine` object gets deleted, will the Remediation request be deleted if it still has a `MachineHealthCheck `OwnerRef?

design/capm3-remediation-controller-proposal.md
jan-est: We need to use `capi.MachineOwnerRemediatedCondition` to be able to owning machine controller to do the deletion of the Machine object. For example, if our Machine is control plane CAPI will handle preflight checks, keep etcd quorum healthy and deletes Machine after needed healtchecks. If we just set the deletion timestamp on Machine we are able to break the controlplane and we have high risk to lose the whole cluster.

jan-est: So we are not removing the Machine ourselves we let CAPI to handle the removing of the Machines.

design/capm3-remediation-controller-proposal.md
jan-est: Yes, fixed!

design/capm3-remediation-controller-proposal.md
jan-est: I am not quite sure how BMO handles this? But if it is not able to power on the host that should be issued and take care in BMO.

jan-est:
@beekhof @n1r1 let's try to get this PR to be merged. Are there still points where we should discuss?

We discussed in Slack that we may need a new alternative Remediation strategy? I suggest that lets merge this one and you can add a new proposal on top of this.

design/capm3-remediation-controller-proposal.md
beekhof: Fair enough. Although realistically, if we're deleting the machine then chances are it is already /not/ contributing towards the etcd quorum calculation.

furkatgofurov7:
As agreed during the community meeting, let's have a lazy consensus of a week from today and if no objections, we can merge it next week.

n1r1:
looks like more than a week past since the lazy consensus. 
were there any objections to get this merged?

jan-est:
> looks like more than a week past since the lazy consensus.
> were there any objections to get this merged?

No, we are planning to take this in today!

fmuyassarov:
> As agreed during the community meeting, let's have a lazy consensus of a week from today and if no objections, we can merge it next week.

Merging it since today is end of consensus and there were no objections on this change. Also @jan-est brought this topic in the community meeting to verify if there are more comments/objections.
/lgtm

PR #166, opened 2021-03-04
Update disabling automated cleaning proposal

fmuyassarov:
Update disabling automated cleaning proposal as per https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r585629005

### DIFF
--- a/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
+++ b/design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
@@ -5,7 +5,7 @@ Unported License.
 http://creativecommons.org/licenses/by/3.0/legalcode
 -->
 
-# allow disabling disk cleaning
+# allow disabling automated cleaning
 
 ## Status
 
@@ -13,8 +13,8 @@ implementable
 
 ## Summary
 
-Configurable disk cleaning interface in Metal³ to allow users to
-disable/enable disk cleaning for the nodes.
+Configurable automated cleaning interface in Metal³ to allow users to
+disable/enable automated cleaning for nodes.
 
 ## Motivation
 
@@ -69,11 +69,11 @@ In the Metal³, we will need changes both in CAPM3 and BMO as follows
 
 ##### Metal3MachineTemplate object
 
-We introduce a new field `disableAutomatedClean` in the spec of the
-Metal3MachineTemplate object. `disableAutomatedClean` can store a boolean type
-of value. Setting it to True means don’t perform automated cleaning, while
-False value enables automated cleaning. By default, `disableAutomatedClean` will
-be set to False (i.e. do perform automated cleaning as it is now).
+We introduce a new field `automatedCleaningMode` in the spec of the
+Metal3MachineTemplate object. `automatedCleaningMode` store a enum type
+of value. Setting it to Disabled means don’t perform automated cleaning, while
+Metadata value enables automated cleaning. By default, `automatedCleaningMode` will
+be set to Metadata (i.e. do perform automated cleaning as it is now).
 
 ```yaml
 apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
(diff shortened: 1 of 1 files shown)


dhellmann:
This looks like what we discussed over on the code review. Thanks!

/approve

design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
```diff
@@ -82,22 +82,22 @@ metadata:
 spec:
  template:
  spec:
-  disableAutomatedClean: False   #default value. Perform automated cleaning.
+  automatedCleaningMode: None   # default value is Full
```
hardys, line 85: Coming from python looks wrong at first glance :)

This will be a literal string "None", not a yaml `null`  - I guess the enum validation will catch any wrong values, but this might be a bit clearer if it was e.g `automatedCleaningMode: disabled` ?

hardys:
lgtm, just one minor nit re the choice of "None" vs perhaps "disabled/Disabled" - @dhellmann what are your thoughts on that?

design/cluster-api-provider-metal3/allow_disabling_node_disk_cleaning.md
dhellmann: Heh, yeah, that's fair. I complain about Python APIs that are hard to use from golang all the time, so we probably should change that to set a good example. ;-)

I like `Disabled` (the style guide calls for title caps).

hardys: > the style guide calls for title caps

Do you happen to have a link to that guideline?  I'm wondering if I did the wrong thing using lowercase for the inspect annotation, or if that's just for the structural schema defined for the spec?

dhellmann: I think it mostly applies to the API fields, not necessarily annotations. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

fmuyassarov: @dhellmann @hardys so you suggest `Enabled/Disabled` instead of `Full/None`, or `Full/Disabled`? I would personally go with `Enabled/Disabled` but okay with other too.

fmuyassarov: I have update the proposal. If you meant `Full/Disabled`, then please let me know and I'll update the proposal & implementation.

dhellmann: I think if we're considering some version of a "partial" cleaning implementation, then we want to say "Full" now so there isn't confusion about the difference between "Enabled" and "Partial" later.

fmuyassarov: Thanks, make sense. Updated.

dhellmann:
I'm happy with this draft, thanks @fmuyassarov. I'll leave it for @hardys to lgtm in case he has any other feedback.

/approve

hardys:
This looks good except for an unresolved comment from @dtantsur on the implementation PR https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r589422010

He mentions that "Full" could imply (or be used in future for) a full disk erase, which is not the current behavior (the disk metadata/partition-table is removed).

I don't have a strong opinion on what the alternative value should be (`Metadata` ?), but I wanted to ensure we have that discussion so that this PR matches the implementation.

fmuyassarov:
> This looks good except for an unresolved comment from @dtantsur on the implementation PR [metal3-io/baremetal-operator#784 (comment)](https://github.com/metal3-io/baremetal-operator/pull/784#discussion_r589422010)
> 
> He mentions that "Full" could imply (or be used in future for) a full disk erase, which is not the current behavior (the disk metadata/partition-table is removed).
> 
> I don't have a strong opinion on what the alternative value should be (`Metadata` ?), but I wanted to ensure we have that discussion so that this PR matches the implementation.

replaced `Full` -> `Metadata` both in here and in the implementation.

dhellmann:
Looks good.

/approve

dtantsur (approved):

PR #168, opened 2021-03-11
Add detached annotation proposal

hardys:
First draft of a proposal to enable control of maintenance mode via the BMH API.

### DIFF
--- a/design/baremetal-operator/detached-annotation.md
+++ b/design/baremetal-operator/detached-annotation.md
@@ -1,0 +1,112 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add support for Detached annotation
+
+## Status
+
+implementable
+
+## Summary
+
+Provide a way to prevent management of BaremetalHost resources
+after provisioning is completed, to facilitate the pivot of
+BaremetalHost resources in a multi-tier deployment.
+
+## Motivation
+
+In a multi-tier deployment where one cluster deploys another, the "parent"
+cluster will contain BMH resources for initial provisioning,
+but the "child" cluster will later contain BMH resources that reference the
+same physical hosts.
+
+In this scenario it's necessary to prevent management operations from the
+parent cluster such as asserting power state, or BMH actions on the child
+cluster such as the [reboot annotation](reboot-interface.md) may fail due
+to unwanted BMC interactions from the parent cluster.
+
+There is an existing
+[pause annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#pausing-reconciliation)
+which pauses reconciliation of the BMH resources, but this does not remove
+the underlying Ironic host, so power management actions may be performed
+even when the BMH is marked as `paused`.
+
+### Goals
+
+* Add an API to disable management of a BMH on the parent cluster, including
+  all power management actions
+* Make it possible to delete a BMH resource in this "detached" without
+  triggering deprovisioning
+* Ensure it is possible to restart management of BMH resources (in the case
+  where they are not deleted from the parent cluster)
+* Avoid changing behavior of the existing `paused` annotation since that
+  behavior is necessary as part of the CAPI pivot process.
+
+### Non-Goals
+
+* Any coordination between the tiers of BMH resources, that has to be handled externally
+
+## Proposal
+
+### Expected workflow
+
+* User creates BMH resource(s) in parent cluster
+* Provisioning of BMH triggered, which results in a running child cluster
+* Parent cluster BMH resources annotated as detached
+* Child cluster BMH resources created, with BMC credentials, but marked
+  externallyProvisioned: true
+
+At this point, the physical hosts are owned by the child cluster BMH, but the
+inventory still exists in the parent cluster.
+
+In the event that all of the child cluster hosts require reprovisioning, it
+would be necessary to remove the detached annotation on the parent BMH resources,
+so that management of those resources can resume, e.g for reprovisioning.
+
+### API changes
+
+Add support for a new annotation, where the key is `baremetalhost.metal3.io/detached`
+
+```yaml
+baremetalhost.metal3.io/detached: ""
+```
+
+The value is ignored, similar to the `paused` annotation and could optionally
+include context from the system/user which applies the annotation.
+
+This annotation will only be consumed while the BMH is in either `Provisioned`
+or `ExternallyProvisioned` state, in all other cases it is ignored.
+
+This annotation will be evaluated early in the `Reconcile()` loop, but after the
+`paused` `status` and `hardwaredetails` annotations are evaluated.
+
+When the `detached` annotation is set, we will check the `status.provisioning.ID`
+and if necessary delete the corresponding host from Ironic (without triggering
+deprovisioning)
+
+When the `detached` annotation is removed, we will re-create the host in Ironic
+via the existing `ensureRegistered` state machine logic.
+
+If a BMH resource is deleted while the `detached` annotation is set, we will
+move directly to the `Deleting` state, without performing any `Deprovisioning`.
+
+## Alternatives
+
+It would be possible to modify the behavior of the
+[pause annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#pausing-reconciliation)
+such that Ironic hosts are removed while paused, however
+this means that we cannot reflect any error via the status
+or increment the errorCount for the retry backoff.
+
+We could add an API that sets the Ironic
+[maintenance mode flag](https://docs.openstack.org/api-ref/baremetal/?expanded=set-maintenance-flag-detail#set-maintenance-flag)
+but this means hosts could potentially permanently be in this state
+and there are concerns about corner-cases such as adoption when an
+ephemeral Ironic is used and a rechedule occurs.
+
+## References
+


design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,93 @@
...
+metadata:
+  name: worker-0
+spec:
+  online: true
+  maintenance: true
+ ...
+```
+
+TODO - we don't expose any reason here, is it enough to have
+a standard reason string e.g "disabled by BaremetalHost maintenance field"?
```
dhellmann, line 71: I assume the caller knows why the host is in maintenance mode. We could ask for a string reason, but I'm not sure what we would do with it. A UI or something could always use an annotation to store a message. Unless maybe we would show the reason as part of the default `kubectl` output for hosts?

design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,93 @@
...
+
+TODO - what happens if a BMH is deleted with `maintenance: true` - do we
+skip deprovisioning or should that be an independent annotation/field?
```
dhellmann, line 74: We should look at what Ironic does and see if we want that to be our default behavior.

@dtantsur thoughts?

design/baremetal-operator/maintenance-mode.md
hardys: The underlying Ironic API does accept a reason, but I guess the user isn't exposed to that - I was thinking maybe generating an event with the reason could be useful, but it could also just be a generic event saying the host got put into maintenance mode (which simplifies the API as it's just a boolean flag)

dhellmann: Ah, yeah, I wasn't really concerned with the value in Ironic so much as what we ask for or show on the host resource. Let's leave the reason out for now, under the theory that it's easier to add the field later if we decide we need it.

design/baremetal-operator/maintenance-mode.md
hardys: From local testing I'm fairly sure that when you've set the maintenance flag you can just delete the node without any deprovisioning, so aligning with that behavior seems reasonable to me.

dhellmann: That matches what I remember.

design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,87 @@
...
+## Summary
+
+Currently there is no way to prevent management of BaremetalHost resources
+after provisioning is completed.  The underlying Ironic API contains a
+maintenance-mode flag which could be used to disable management of
+BMH resources.
+
+## Motivation
+
+In a multi-tier deployment where one cluster deploys another, the "parent"
```
romfreiman, line 23: I think in general - transfer ownership between clusters, right?

dhellmann: There have been some discussions about moving hosts between clusters as part of auto-scaling, but there are some different implications when doing that (we would absolutely need to deprovision those hosts as part of the transfer, to avoid leaking data from one cluster to another, for example). So, I think for this proposal let's stay focused on the multi-tier deployment case.

kashifest (commented):
+1 for this. This will be really helpful. I think also as a non-goal it should be made clear that when ironic is putting a node on maintenance mode, we won't want to sync that back to BMH resource, right or would we? 
/cc @maelk

hardys:
> +1 for this. This will be really helpful. I think also as a non-goal it should be made clear that when ironic is putting a node on maintenance mode, we won't want to sync that back to BMH resource, right or would we?

This is a good point - I was thinking that the BMO sync would remain active (since reconciliation is not paused), but since maintenance mode flag in Ironic pauses all periodic tasks inside Ironic, it's unlikely anything in the BMH will change from Ironic->BMH while in this state.

Another option (mentioned in the alternatives) would be to modify the behavior of the `baremetalhost.metal3.io/paused` annotation to set the Ironic host to maintenance mode (or even remove it from Ironic completely), but just exposing the maintenance mode flag seems a little cleaner, and this approach would prevent some desirable workflows (in particular the ability to delete a BMH in maintenance mode, skipping deprovisioning)

design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,87 @@
...
+
+In this scenario it's necessary to prevent management operations from the
+parent cluster such as asserting power state, or BMH actions on the child
+cluster such as the [reboot annotation](reboot-interface.md) may fail due
+to unwanted BMC interactions from the parent cluster.
+
+This may also be useful for temporary maintenance operations
+where you may wish for power management operations and polling
+to be stopped for specific hosts (this is the intended use if the
+underlying [Ironic API](https://docs.openstack.org/api-ref/baremetal/?expanded=set-maintenance-flag-detail#set-maintenance-flag)
```
dtantsur, line 36: It is, as is written here, the intended use case of maintenance, but only if we're ready to avoid touching ironic API while in maintenance. E.g. trying to do adoption in maintenance mode is a huge stretch already, I'm not even sure it works (even less sure if it should work).

The alternative suggestion: removes ironic nodes (not BMHs!) that the current cluster should not be managing. This way there is only one ironic responsible for each node, and we don't have any split-brain issues.

WDYT?

dtantsur (changes requested):

design/baremetal-operator/maintenance-mode.md
hardys: I'm primarily trying to solve the "cluster deploys a cluster" problem, where we need to pivot ownership of the BMH from the parent->child cluster (but maintain the BMH in the parent for inventory management and so redeployment could be possible)

In that case, some way to remove the nodes from Ironic on the parent makes sense, but I'm not sure the `spec.maintenance: true` API really describes this behavior.

Maybe we should add an annotation e.g `baremetalhost.metal3.io/unmanage` which triggers removal from Ironic, and makes all subsequent management operations a no-op?

@kashifest @maelk @dhellmann what are your thoughts?

dhellmann: > It is, as is written here, the intended use case of maintenance, but only if we're ready to avoid touching ironic API while in maintenance. E.g. trying to do adoption in maintenance mode is a huge stretch already, I'm not even sure it works (even less sure if it should work).

What sequence of events would lead to trying to adopt a host with the maintenance flag on?

Should we treat hosts with their maintenance flag set as paused, except to allow delete?

> The alternative suggestion: removes ironic nodes (not BMHs!) that the current cluster should not be managing. This way there is only one ironic responsible for each node, and we don't have any split-brain issues.

Are you suggesting that the maintenance flag should cause the BMO to remove the host from Ironic completely? Or that we should describe another way to do that instead of (or in addition to) adding a maintenance flag?

> Maybe we should add an annotation e.g `baremetalhost.metal3.io/unmanage` which triggers removal from Ironic, and makes all subsequent management operations a no-op?

I think that's more or less what the existing pause annotation was meant to be, but we didn't think through the fact that by not removing the host from Ironic we leave it's periodic tasks operating. So we could extend what the existing annotation does, too, if we decide we need to go this route.

Does anyone else run into this problem of wanting to deploy repeatedly from a persistent cluster? Or do most users run the deployment from a minikube or other disposable cluster?

celebdor: Is there some reason we'd want the systems to remain in Ironic? In case we kept them in BMO but removed them from Ironic with `baremetalhost.metal3.io/unmanage`, what would happen when the target cluster is done with the node and we remove the `baremetalhost.metal3.io/unmanage` and we add the system to Ironic again? Would that trigger BMC actions?

design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,87 @@
...
+### Non-Goals
+
+TBC
+
+## Proposal
+
+Introduce a new BMH field `maintenance` in the spec of the
+BaremetalHost object. This will be a boolean flag, which
+when set to `true` will put the underlying Ironic host
+into maintenance mode.
```
avishayt, line 52: It's not clear to me what actions are possible on a BMH that has maintenance=true.  Can I still provision/deprovision?
Can you please describe (either in comment or in the proposal) what the expected flow would be in the multi-tier deployment case?  Specifically, when should the BMH on the parent cluster have maintenance set to true and when to false?

design/baremetal-operator/maintenance-mode.md
hardys: > What sequence of events would lead to trying to adopt a host with the maintenance flag on?

That will happen any time the metal3 pod gets rescheduled, assuming we leave the hosts in Ironic.
 
> Should we treat hosts with their maintenance flag set as paused, except to allow delete?

We could, I was thinking it might be useful to enable some actions, in particular to update the hardwaredetails via the [new annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/inspectAnnotation.md) - so that the details of each node are visible on the parent cluster.

Pausing would also been reasonable I guess, it'd just mean e.g UX on the parent cluster would have to interact with the child-cluster BMH (which is probably necessary for other reasons).

Pausing is also probably easier to implement than skipping all states except delete?

> Are you suggesting that the maintenance flag should cause the BMO to remove the host from Ironic completely? Or that we should describe another way to do that instead of (or in addition to) adding a maintenance flag?

I'm suggesting we could describe another API that does that (such as an "unmanage" annotation), instead of the maintenance flag (although that could also be added if there's a requirement for it)

hardys: > Is there some reason we'd want the systems to remain in Ironic? 

I don't think so, the main reason I proposed the maintenance flag is it looks like the simplest option to implement.

> In case we kept them in BMO but removed them from Ironic with `baremetalhost.metal3.io/unmanage`, what would happen when the target cluster is done with the node and we remove the `baremetalhost.metal3.io/unmanage` and we add the system to Ironic again? Would that trigger BMC actions?

If the parent cluster removes the annotation (or any other flag we add here), I would expect the host to be added to Ironic again, and that will trigger BMC actions.

This may be a necessary option in the case where the parent cluster has to redeploy a child cluster, but in all other cases probably won't happen, since the child cluster BMH resources are managing the hardware?

design/baremetal-operator/maintenance-mode.md
hardys: > It's not clear to me what actions are possible on a BMH that has maintenance=true. Can I still provision/deprovision?

No, the Ironic API causes all periodic tasks for that host to stop, and it's not a candidate for provisioning/deprovisioning (but you can delete it, without deprovisioning)

> Can you please describe (either in comment or in the proposal) what the expected flow would be in the multi-tier deployment case? Specifically, when should the BMH on the parent cluster have maintenance set to true and when to false?

Sure, lets see if we can discuss and agree the flow here, then I'll sync the result back into the doc, I'm thinking:

- User creates BMH resource(s) in parent cluster
- Provisioning of BMH triggered, which results in a running child cluster
- Parent cluster BMH resources all marked as `maintenance:  true` (or unmanaged, or paused, depending on the outcome of the other discussion threads)
- Child cluster BMH resources created, with BMC credentials, but marked `externallyProvisioned: true`

At this point, the physical hosts are owned by the child cluster BMH, but the inventory still exists in the parent cluster.

In the event that all of the child cluster hosts require reprovisioning, it would be necessary to remove the maintenance/unmanaged flag, or delete/re-create the BMH resource so that the process starts again at the beginning.

avishayt: OK that makes sense.  Basically it's up to the orchestration on the Hub to ensure that the Hub BMH is not in maintenance only when no Spoke BMH exists for that host.

hardys: > OK that makes sense. Basically it's up to the orchestration on the Hub to ensure that the Hub BMH is not in maintenance only when no Spoke BMH exists for that host.

Right, that may be worth calling out explicity, e.g a non-goal is any coordination between the tiers of BMH resources, that has to be handled externally

avishayt: +1

hardys:
Ok so to summarize, we have three options which could satisfy the underlying requirement:

## Add `spec.maintenance` flag

This would set the Ironic maintenance mode flag, as described in the current draft proposal

Advantages:
- Simple to implement
- Well understood behavior based on the underlying Ironic API

Disadvantages:
- Node still exists in Ironic, which could be confusing, and may not be desirable for the multi-tier use-case (where in reality it's not "maintenance" but a potentially permanent change in ownership of the physical server management)
- In the case where the parent cluster deploys $many child clusters, this solution might not scale as well as some of the others?

## Modify `baremetalhost.metal3.io/paused` behavior

In this case we'd modify the current "paused" behavior, such that:
- While paused, the host is removed from Ironic
- It is possible to delete a BMH in paused state (without any deprovisioning actions)

Advantages:
- Should be fairly simple to implement
- Avoids adding a net-new API

Disadvantages:
- Modifies existing API behavior (are we aware of any users of this other than the [openshift installer](https://github.com/openshift/installer/blob/master/pkg/asset/machines/baremetal/hosts.go#L97) atm?)

## Add new "unmanaged" annotation

In this case we'd add a new annotation which either behaves as described above for the modifed-pause approach, or we add logic to every state which skips management operations.

Advantages:
- Avoids changing existing API behavior

Disadvantages:
- Potentially more complex to implement (in the "every state" case)
- Could be confusing since it's similar in behavior to the existing "paused" annotation


Currently my preference is to go with modifying the behavior of the paused annotation - what do others think about that?

celebdor:
> Currently my preference is to go with modifying the behavior of the paused annotation - what do others think about that?

I also think that's a good way to go.

Can you expand on _Add new "unmanaged" annotation_'s disadvantage of "Potentially more complex to implement (in the "every state" case)"

hardys:
> Can you expand on _Add new "unmanaged" annotation_'s disadvantage of "Potentially more complex to implement (in the "every state" case)"

With the `paused` annotation approach, we just [exit from the Reconcile() early](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L118)

If we introduced some new annotation which didn't do that, it would likely mean we need to update many/all states in the controller to look for the new annotation and do nothing when it's set - just exiting the Reconcile early (but adding some new logic to deal with the removal from Ironic, and to enable deletion) seems like a simpler and less error-prone approach to me.

dhellmann:
> +1 for this. This will be really helpful. I think also as a non-goal it should be made clear that when ironic is putting a node on maintenance mode, we won't want to sync that back to BMH resource, right or would we?
> /cc @maelk

This is an interesting point. Does Ironic set this flag on hosts by itself under some conditions, @dtantsur?

dhellmann:
> ## Modify `baremetalhost.metal3.io/paused` behavior
> In this case we'd modify the current "paused" behavior, such that:
> 
> * While paused, the host is removed from Ironic
> * It is possible to delete a BMH in paused state (without any deprovisioning actions)
> 
> Advantages:
> 
> * Should be fairly simple to implement
> * Avoids adding a net-new API
> 
> Disadvantages:
> 
> * Modifies existing API behavior (are we aware of any users of this other than the [openshift installer](https://github.com/openshift/installer/blob/master/pkg/asset/machines/baremetal/hosts.go#L97) atm?)

I believe the feature was added to make it easier for upstream to safely pivot and copy CR content from the provisioning cluster to the new cluster. I don't think the change in behavior will conflict with that use case, but we should confirm that. @kashifest, @fmuyassarov, @maelk what do you think?

> Currently my preference is to go with modifying the behavior of the paused annotation - what do others think about that?

+1 for extending the behavior of the annotation we already have.

andfasano:
> 
> Currently my preference is to go with modifying the behavior of the paused annotation - what do others think about that?

:+1: for that. I think that the potential ambiguity of a new - similar but slightly different - annotation is worse than modifying the existing behavior of  `baremetalhost.metal3.io/paused`.

The only part that I guess still requires some more clarification it's about the "unpausing" step: what is the expected behavior when the annotation will be required, considering that the underlying Ironic node has been previously deleted?

hardys:
> The only part that I guess still requires some more clarification it's about the "unpausing" step: what is the expected behavior when the annotation will be required, considering that the underlying Ironic node has been previously deleted?

I'm thinking it's essentially the same process as recovering from a pod reschedule where the Ironic DB is lost - we'd create the Ironic host again and resume management of the host?

Provided we're confident that process is robust, I think we should be OK to delete the Ironic host, but we'll need to check there aren't unhandled corner cases.

design/baremetal-operator/maintenance-mode.md
dhellmann: > > What sequence of events would lead to trying to adopt a host with the maintenance flag on?
> 
> That will happen any time the metal3 pod gets rescheduled, assuming we leave the hosts in Ironic?

Sure. But in that case, Ironic doesn't think the host is in maintenance mode and we can avoid putting it into maintenance mode until we get to a steady state. I think this point is probably moot, though, if we're going to reuse the pause annotation.

> > Are you suggesting that the maintenance flag should cause the BMO to remove the host from Ironic completely? Or that we should describe another way to do that instead of (or in addition to) adding a maintenance flag?
> 
> I'm suggesting we could describe another API that does that (such as an "unmanage" annotation), instead of the maintenance flag (although that could also be added if there's a requirement for it)

I still view Ironic as an implementation detail of metal3. It's one that would be extremely difficult to change, but I don't think we want to treat metal3 as "Ironic for k8s". Exposing the state of whether we have synced a host into Ironic's database leaks all sorts of things through our API that are best kept hidden so we can change them -- the fact that Ironic has a separate view of the state of the world, the fact that we deploy Ironic without stable storage, the fact that we have to restore Ironic's knowledge of a host before we can do virtually anything with it, etc.

hardys:
> > +1 for this. This will be really helpful. I think also as a non-goal it should be made clear that when ironic is putting a node on maintenance mode, we won't want to sync that back to BMH resource, right or would we?
> > /cc @maelk
> 
> This is an interesting point. Does Ironic set this flag on hosts by itself under some conditions, @dtantsur?

Oh, I think I misinterpreted the question from @kashifest in my previous reply, apologies!

Yes, in some circumstances Ironic does put a host into maintenance mode - IIRC it happens when there are issues connecting to the BMC, but @dtantsur can perhaps advise if there are other conditions we should be aware of.

Regardless IMO that shouldn't be exposed directly via the BMH, it's an internal state which we might reflect via some error status or event, but not expose directly (if we added a `maintenance` flag, which it seems is not the likely direction for the requirement I'm trying to solve)

hardys:
Ok thanks all for the feedback, it seems that extending the `paused` annotation is the preferred approach, so I'll rework this proposal to reflect that.

design/baremetal-operator/maintenance-mode.md
hardys: > Sure. But in that case, Ironic doesn't think the host is in maintenance mode and we can avoid putting it into maintenance mode until we get to a steady state. I think this point is probably moot, though, if we're going to reuse the pause annotation.

I was thinking we'd have to support reschedule (and thus adopt) of hosts that are in maintenance mode, if we went with that option, but as you say it seems like reworking the paused behavior is the preferred option which conveniently avoids dealing with that problem ;)

andfasano:
> > The only part that I guess still requires some more clarification it's about the "unpausing" step: what is the expected behavior when the annotation will be required, considering that the underlying Ironic node has been previously deleted?
> 
> I'm thinking it's essentially the same process as recovering from a pod reschedule where the Ironic DB is lost - we'd create the Ironic host again and resume management of the host?
> 
> Provided we're confident that process is robust, I think we should be OK to delete the Ironic host, but we'll need to check there aren't unhandled corner cases.

Agree, I think it could be useful to have such behavior tracked down in the proposal for clarity

design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,87 @@
...
+TODO - what happens if a BMH is deleted with `maintenance: true` - do we
+skip deprovisioning or should that be an independent annotation/field?
+
+## Alternatives
+
+It would be possible to modify the behavior of the
+[pause annotation](https://docs.openstack.org/api-ref/baremetal/?expanded=set-maintenance-flag-detail#set-maintenance-flag)
+such that Ironic maintenance mode is set when paused (or the host is completely
+removed from Ironic), however this could pose a risk of regression to
+any users expecting power-management to continue while BMO reconciliation
```
zaneb, line 77: I don't think any users have that expectation. While Ironic continues to poll the power state, any changes will not show up in the BMH when it is paused. And no changes to the power state are actioned unless they started before the BMH is paused.

design/baremetal-operator/maintenance-mode.md
```diff
@@ -0,0 +1,87 @@
...
+is paused, and it also prevents any other BMH APIs being used while
+management is disabled.
```
zaneb, line 79: For instance?

zaneb (commented):
Reusing the paused annotation seems like the tidiest user interface.
My only worry would be that it makes the implementation more complex, since the pause annotation is checked very early but we will now need to make calls to the provisioner when it is present (an obvious place to put the code for e.g. an `unmanaged` annotation would be in the vicinity of [`ensureRegistered()`](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_state_machine.go#L229-L266)).
A key point about the paused annotation is that it is used in pivoting to make sure the Status subresource is not written to (to make sure that it can be regenerated from the status annotation). So that precludes us ever including in the Status any indication that the Host is in maintenance mode and that all of the data (power state &c.) is meaningless.

dhellmann:
> Reusing the paused annotation seems like the tidiest user interface.
> My only worry would be that it makes the implementation more complex, since the pause annotation is checked very early but we will now need to make calls to the provisioner when it is present (an obvious place to put the code for e.g. an `unmanaged` annotation would be in the vicinity of [`ensureRegistered()`](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_state_machine.go#L229-L266)).
> A key point about the paused annotation is that it is used in pivoting to make sure the Status subresource is not written to (to make sure that it can be regenerated from the status annotation). So that precludes us ever including in the Status any indication that the Host is in maintenance mode and that all of the data (power state &c.) is meaningless.

These are good points. I especially hadn't considered the fact that we couldn't write to the status fields. I am starting to think that adding a new annotation, while expanding the UI, would make the implementation easier to follow and verify. Given the complexity of managing the host state, I would err on the side of simplifying the implementation in this instance.

design/baremetal-operator/maintenance-mode.md
hardys: I was thinking of the recently added [hardwaredetails annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/inspectAnnotation.md) but that's probably not an issue, it can just be updated prior to pausing the parent cluster BMH.

hardys:
> These are good points. I especially hadn't considered the fact that we couldn't write to the status fields

Do we need to write to the status fields?  The only time I can think that might be necessary is if we encounter an unrecoverable error deleting the underlying host from Ironic (when first encountering the `paused` annotation), but we could generate an event without updating the host status in that case?

dhellmann:
> > These are good points. I especially hadn't considered the fact that we couldn't write to the status fields
> 
> Do we need to write to the status fields? The only time I can think that might be necessary is if we encounter an unrecoverable error deleting the underlying host from Ironic (when first encountering the `paused` annotation), but we could generate an event without updating the host status in that case?

The error handling relies on a counter in the status field to manage the retry interval backoff. We wouldn't be able to update that field. Maybe we don't need to? Talking to Ironic can result in errors, even if they are just 409s, though. If we can ensure that all of the pause handling happens outside of the state machine, we should be able to avoid the backoff logic and maybe that would be OK.

It would be useful to see if the amount of change in `Reconcile()` matches what @zaneb describes, too. The pause annotation happens very early. Between that point and where we have a Provisioner, we handle deletion, the status annotation, and the hardware annotation. I guess we're talking about allowing delete operations while paused? So maybe we only need to update the logic around the other annotations to skip them if we're paused. And then when we get to the point where we have a provisioner, if we're paused we would handle that instead of entering the state machine. We wouldn't be able to update the status fields on the host to record that we have handled it (by clearing the provisioning ID field), though, so every time the host is reconciled we would look at Ironic to verify that the host is not present. I guess once that's done, we don't need to keep reconciling the host, so it wouldn't be too expensive.

So, I think we could still do it with the pause annotation. It's going to complicate `Reconcile()`, but I don't think there's a way to do this that won't add some complexity there.

hardys:
Ok this was discussed today in the community meeting, in summary:

* @maelk pointed out that the CAPI pivot relies on us not updating the status, and as mentioned above by @dhellmann that may not be ideal in the case where there is some error deleting the host from Ironic
* @zaneb mentioned it may be simpler to implement if we go with a new annotation
* Some other use-cases for maintenance mode were discussed, but I think for the multi-tier use-case we're agreed that removing the host from Ironic is probably preferable as it's simpler to implement and avoids potential corner-cases around adoption.

Accordingly I'll update this proposal to describe a new annotation and attempt to capture the feedback so far, thanks!

design/baremetal-operator/unmanaged-annotation.md
```diff
@@ -0,0 +1,106 @@
...
+inventory still exists in the parent cluster.
+
+In the event that all of the child cluster hosts require reprovisioning, it
+would be necessary to remove the maintenance/unmanaged flag, or
+delete/re-create the BMH resource so that the process starts again at the
+beginning.
+
+### API changes
+
+Add support for a new annotation, where the key is `baremetalhost.metal3.io/unmanaged`
```
zaneb, line 72: A requirement I thought I heard on the call the other day (@maelk correct me if this is wrong) is that for pivoting we need a way of knowing when the host has become unmanaged, so that the paused annotation can then be added as well afterwards. I'm unsure if this is true though, because I would have assumed the pause was mainly for the Host in the target cluster and the unmanage is for the Host in the bootstrap cluster.

If this is a requirement then we will need to figure out how to reflect progress in the Host status. If that ends up being a separate field then it probably makes sense to add this to the Spec rather than an annotation. If not then I prefer keeping it as an annotation.

hardys: Thanks @zaneb I've also been thinking about this - and the fact that we already have a BMH [Unmanaged State](https://github.com/metal3-io/baremetal-operator/blob/master/apis/metal3.io/v1alpha1/baremetalhost_types.go#L160)

This is kind of a weird non-state though, because we don't expose this to the user at all via `status.provisioning.state`

It's also not possible to transition from any state other than [Created](https://github.com/metal3-io/baremetal-operator/blob/master/docs/BaremetalHost_ProvisioningState.png) into this state (if you remove BMC creds from an existing BMH in `Provisioned` state, we go into `registration error`)

When discussing this with @dhellmann I think we said it would be best to avoid adding any transition from `$state->Unmanaged`, but I'm wondering if we should revisit that idea?

If we modified the `Unmanaged` behavior so that:

- We move to `Unmanaged` state whenever there are no BMC details, or when the `unmanaged` annotation is set
- When moving to `Unmanaged` state we remove any corresponding host in Ironic
- Set `status.provisioning.state` to reflect when we're in `Unmanaged` state (either on create, BMC details are removed, or when the annotation is added)
- When moving from the Unmanaged State, always go to `Registering`, except when an existing Image exists in the status - then we'd add a new transition to `Provisioned`

WDYT?

design/baremetal-operator/unmanaged-annotation.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Goals
+
+* Add an API to disable management of a BMH on the parent cluster, including
+  all power management actions
+* Make it possible to delete a BMH resource in this "unmanaged" without
+  triggering deprovisioning
+* Ensure it is possible to restart management of BMH resources (in the case
+  where they are not deleted from the parent cluster)
+* Avoid changing behavior of the existing `paused` annotation since
```
kirankt, line 46: Dangling sentence? Maybe?

design/baremetal-operator/unmanaged-annotation.md
```diff
@@ -0,0 +1,106 @@
...
+* Any coordination between the tiers of BMH resources, that has to be handled externally
+
+## Proposal
+
+### Expected workflow
+
+* User creates BMH resource(s) in parent cluster
+* Provisioning of BMH triggered, which results in a running child cluster
+* Parent cluster BMH resources marked as unmanaged
+* Child cluster BMH resources created, with BMC credentials, but marked
```
kirankt, line 59: Who is responsible for this step?

design/baremetal-operator/unmanaged-annotation.md
zaneb: > Thanks @zaneb I've also been thinking about this - and the fact that we already have a BMH [Unmanaged State](https://github.com/metal3-io/baremetal-operator/blob/master/apis/metal3.io/v1alpha1/baremetalhost_types.go#L160)
> 
> This is kind of a weird non-state though, because we don't expose this to the user at all via `status.provisioning.state`

Yes we do: https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_state_machine.go#L275
Proof: https://bugzilla.redhat.com/show_bug.cgi?id=1936966#c0

> It's also not possible to transition from any state other than [Created](https://github.com/metal3-io/baremetal-operator/blob/master/docs/BaremetalHost_ProvisioningState.png) into this state (if you remove BMC creds from an existing BMH in `Provisioned` state, we go into `registration error`)

We stay in Provisioned and the operational status goes to error with error type "registration error". (Being pedantic because prior to metal3-io/baremetal-operator#388 there was a RegistrationError provisioning state.)

> When discussing this with @dhellmann I think we said it would be best to avoid adding any transition from `$state->Unmanaged`, but I'm wondering if we should revisit that idea?
> 
> If we modified the `Unmanaged` behavior so that:
> 
> * We move to `Unmanaged` state whenever there are no BMC details, or when the `unmanaged` annotation is set
> * When moving to `Unmanaged` state we remove any corresponding host in Ironic
>  * Set `status.provisioning.state` to reflect when we're in `Unmanaged` state (either on create, BMC details are removed, or when the annotation is added)
> * When moving from the Unmanaged State, always go to `Registering`, except when an existing Image exists in the status - then we'd add a new transition to `Provisioned`
> 
> WDYT?

The main problem with changing the provisioning state is that you then have to figure out which state you came from in order to go back. Basically we'd have to resurrect [this abomination](https://github.com/metal3-io/baremetal-operator/pull/388/commits/0e0d2af24c901b2933b8ff0e2d4f1d97f67e2056#diff-eeed3703d04e4c23a7d7af8cd0b7931b6b7990f23d826c49bdbc31c5f0a50291L214-L230). Limiting it to only returning to either Registration or Provisioned definitely makes it less bad than it could be. Whether the host has been provisioned can currently be determined reliably by looking at the existence of an image in the status (although I feel that this may be a mistake, and we possibly should put it in the status as soon as we start provisioning).

I guess the other concern would be for the user does it make sense that they could have got into this state (and therefore have to get out of it) by either of two completely different methods. Currently the OperationalStatus in the Unmanaged state is `discovered`; we probably want to change that to `inactive` or something for this case. Maybe that would be enough to distinguish them, although we did recently hide the operational status from the default `kubectl get` output.

Another option would be to change *only* the operational status (and leave the provisioning state where it was).

I'm certainly in favour of not having to add yet another status field.

hardys: > Yes we do: https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/host_state_machine.go#L275

Thanks @zaneb - turns out that only works when `externallyProvisioned: true` is set, which is probably a bug - if you create a BMH without that set (and without any BMC details), the Operational State ends up empty.

hardys: So yeah, this is a bug because we try to [create the provisioner](https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L216) even when there are no BMC details - this results in an error, and we never get as far as setting any status.

I'm not yet sure why adding `externallyProvisioned` works around that though.

zaneb: It's because of #748.

hardys: Resolved via https://github.com/metal3-io/baremetal-operator/pull/824

dtantsur (approved):
👍🏽

hardys:
WIP implementation ref https://github.com/metal3-io/baremetal-operator/pull/827 - feedback welcome

design/baremetal-operator/unmanaged-annotation.md
```diff
@@ -0,0 +1,106 @@
...
+[pause annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#pausing-reconciliation)
+which pauses reconciliation of the BMH resources, but this does not remove
+the underlying Ironic host, so power management actions may be performed
+even when the BMH is marked as `paused`.
+
+### Goals
+
+* Add an API to disable management of a BMH on the parent cluster, including
+  all power management actions
+* Make it possible to delete a BMH resource in this "unmanaged" without
```
dhellmann, line 42:
```suggestion
* Make it possible to delete a BMH resource with this "unmanaged" setting, without
```

design/baremetal-operator/unmanaged-annotation.md
dhellmann, line 46: This sentence wasn't completed.

design/baremetal-operator/unmanaged-annotation.md
```diff
@@ -0,0 +1,106 @@
...
+
+```yaml
+baremetalhost.metal3.io/unmanaged: ""
+```
+
+The value is ignored, similar to the `paused` annotation and could optionally
+include context from the system/user which applies the annotation.
+
+This annotation will be evaluated early in the `Reconcile()` loop, but after the
+`paused` `status` and `hardwaredetails` annotations are evaluated.
```
dhellmann, line 82: Why do we want to do it after those other annotations are handled? Does it actually matter, or is that just what we're doing?

I think we want to do it after the hardware details and status annotations are handled to avoid triggering inspection when the unmanaged annotation is set?

It seems like want to handle it before the paused annotation, since unmanaged is a more comprehensive way to say "stop paying attention to this host" than paused is.

Would it be safer for this new annotation to only apply to hosts in a steady state? What happens if provisioning starts and this annotation is added, for example? Or deprovisioning?

zaneb: It definitely must be handled after paused and status. I think in the case of hardwaredetails that might just be where we're planning on doing it rather than a strict requirement.

The question of what happens in a non-steady state is a good one. This might be a good reason to use the Unmanaged provisioning state as we discussed, since it means the host will always return to the Registering or Provisoned states afterwards. We might want to ignore the annotation if we are in the Deprovisioning state though, because AFAIR there's no way to distinguish whether the host was Provisioned or Deprovisioning before going to Unmanaged.

hardys: >  I think in the case of hardwaredetails that might just be where we're planning on doing it rather than a strict requirement

Yes it's not a strict requirement, but I was thinking there's not really any reason to prevent consuming hardwaredetails while in the unmanaged mode, thus what I proposed in https://github.com/metal3-io/baremetal-operator/pull/827

> The question of what happens in a non-steady state is a good one. This might be a good reason to use the Unmanaged provisioning state as we discussed, since it means the host will always return to the Registering or Provisoned states afterwards

Agreed I'd not really considered this case - previous feedback from @dhellmann was that we wouldn't want to add a new transition to the `Unmanaged` state, but it seems like the cleanest way to enforce that is to handle this mode via the state machine?

The other option is to only allow this annotation to function when in a specific steady state (e.g for the pivot use-case, only `Provisioned` would be enough)

kashifest:
@hardys if I understand correctly, this proposal then wouldn't sync BMH api with ironic's maintenance mode? I mean there wouldn't be any BMH spec field from which we would be able to set maintenance mode in ironic?

design/baremetal-operator/unmanaged-annotation.md
dhellmann: > It definitely must be handled after paused and status. I think in the case of hardwaredetails that might just be where we're planning on doing it rather than a strict requirement.

If we handle this after the paused annotation, that means pausing has higher precedence than marking it as unmanaged. I'm not sure that makes sense.


> > The question of what happens in a non-steady state is a good one. This might be a good reason to use the Unmanaged provisioning state as we discussed, since it means the host will always return to the Registering or Provisoned states afterwards
> 
> Agreed I'd not really considered this case - previous feedback from @dhellmann was that we wouldn't want to add a new transition to the `Unmanaged` state, but it seems like the cleanest way to enforce that is to handle this mode via the state machine?
> 
> The other option is to only allow this annotation to function when in a specific steady state (e.g for the pivot use-case, only `Provisioned` would be enough)

Can we handle it in `actionManageSteadyState` in the controller?

zaneb: > If we handle this after the paused annotation, that means pausing has higher precedence than marking it as unmanaged. I'm not sure that makes sense.

Pausing is meant to appear to stop reconciliation, so the only way it _does_ make sense is if it has higher precedence than everything.

> Can we handle it in `actionManageSteadyState` in the controller?

No, because it would get re-registered first.

design/baremetal-operator/unmanaged-annotation.md
```diff
@@ -0,0 +1,105 @@
...
+* Child cluster BMH resources created, with BMC credentials, but marked
+  externallyProvisioned: true
+
+At this point, the physical hosts are owned by the child cluster BMH, but the
+inventory still exists in the parent cluster.
+
+In the event that all of the child cluster hosts require reprovisioning, it
+would be necessary to remove the maintenance/unmanaged flag, or
+delete/re-create the BMH resource so that the process starts again at the
+beginning.
```
sadasu, line 68: This statement could be made clearer. Do these operations "remove the maintenance/unmanaged flag" and "delete/re-create the BMH resource" both happen on the parent cluster when the child cluster is being reprovisioned?

hardys:
> @hardys if I understand correctly, this proposal then wouldn't sync BMH api with ironic's maintenance mode? I mean there wouldn't be any BMH spec field from which we would be able to set maintenance mode in ironic?

@kashifest right, that was my initial proposal, but now the host is removed completely from Ironic ref https://github.com/metal3-io/baremetal-operator/pull/827 - we could still add a `spec.maintenance` field, but based on the discussion here it doesn't sound like the best fit for the use-case I'm currently trying to solve.

design/baremetal-operator/unmanaged-annotation.md
maelk: Yes, it is really a requirement for pivoting as done with CAPI and CAPM3, that the pausing annotation is the first thing handled and has the highest precendence

design/baremetal-operator/unmanaged-annotation.md
hardys: Thanks, I clarified this in the latest version.

hardys:
Updated to align with https://github.com/metal3-io/baremetal-operator/pull/827

design/baremetal-operator/detached-annotation.md
```diff
@@ -0,0 +1,112 @@
...
+
+This annotation will only be consumed while the BMH is in either `Provisioned`
+or `ExternallyProvisioned` state, in all other cases it is ignored.
+
+This annotation will be evaluated early in the `Reconcile()` loop, but after the
+`paused` `status` and `hardwaredetails` annotations are evaluated.
+
+When the `detached` annotation is set, we will check the `status.provisioning.ID`
+and if necessary delete the corresponding host from Ironic (without triggering
+deprovisioning)
```
andfasano, line 89: Maybe it could help to add that until the annotation remains then BMO will ignore any event for that resource?

andfasano:
@hardys the current version LGTM, anything else missing?

hardys:
> @hardys the current version LGTM, anything else missing?

@andfasano no I think this can merge now, particularly since the related BMO implementation landed some time ago ;)

andfasano:
Ok thanks

/lgtm

PR #169, opened 2021-03-16
Add update strategy to Metal3DataTemplate

kashifest:
Due to the immutability of the Metal3DataTemplate, it was not possible
to update properly the metadata and network data for a
Metal3MachineTemplate. This commits adds an update strategy to allow
the user to create a new template and transition to it via rolling
upgrade of the CAPI machine deployments or KCP.

### DIFF
--- a/design/metadata-handling.md
+++ b/design/metadata-handling.md
@@ -96,6 +96,11 @@ and be re-used during upgrade.
 As a user, I want to give my own network configuration for cloud-init as a field
 on Metal3Machine
 
+#### Story 5
+
+As a user using a template for metaData and networkData, I want to be able to
+update the template when updating the Metal3MachineTemplate.
+
 ### Implementation Details/Notes/Constraints
 
 - The metadata field should contain some required element, such as UUID, that is
@@ -112,6 +117,11 @@ on Metal3Machine
 - Documentation of this feature would be very important.
 - The controllers must be able to recreate all status objects, i.e. no
   necessary information should be stored in the status without being recoverable
+- Since the BareMetalHost references the secrets and they are used at provisioning
+  time, the secrets cannot be updated to be able to reprovision the node in the
+  exact same state. This means that the data template parts containing the
+  metadata and network data must be immutable. Updates have to be done by creating
+  a new template and referencing it in the new / updated Metal3MachineTemplate.
 
 ### Risks and Mitigations
 
@@ -273,6 +283,7 @@ metadata:
     name: cluster-1
 spec:
   clusterName: cluster-1
+  templateReference: SharedRef
   metaData:
     strings:
       - key: abc
@@ -597,15 +608,21 @@ metadata:
     name: nodepool-1
 spec:
   index: 0
+  templateReference: SharedRef
   metaData:
     name: machine-1-metadata
     namespace: default
   networkData:
     name: machine-1-metadata
     namespace: default
-  Metal3DataClaim:
+ claim:
     name: machine-1
     namespace: default
+  template:
+    name: nodepool-1
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    namespace: default
+    kind: Metal3DataTemplate
 status:
   ready: true
   error: false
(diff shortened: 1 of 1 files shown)


maelk:
Is this still WIP ?

design/metadata-handling.md
```diff
@@ -597,15 +608,21 @@ metadata:
...
     namespace: default
-  Metal3DataClaim:
+ claim:
     name: machine-1
     namespace: default
+  template:
+    name: nodepool-1
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    namespace: default
+    kind: Metal3DataTemplate
```
maelk, line 625: to be clear, those changes are not related to the proposal, they only fix the deviation between the implementation and the proposal.

design/metadata-handling.md
```diff
@@ -112,6 +117,11 @@ on Metal3Machine
 - Documentation of this feature would be very important.
 - The controllers must be able to recreate all status objects, i.e. no
   necessary information should be stored in the status without being recoverable
+- Since the BareMetalHost references the secrets and they are used at provisioning
+  time, the secrets cannot be updated to be able to reprovision the node in the
+  exact same state. This means that the data template parts containing the
+  metadata and network data must be immutable. Updates have to be done by creating
+  a new template and referencing it in the new / updated Metal3MachineTemplate.
```
maelk, line 124: maybe it could be listed as an alternative that the M3DT could be mutable and reconciled so that the secrets would be updated, and reason why this would be problematic. For example, that the reprovisioning of the node would then lead to a different configuration, and that might be unexpected. And if we allow the mutability without updating the secrets,there is no way to know which version of the template they were created with and what they contain.

design/metadata-handling.md
```diff
@@ -630,16 +654,22 @@ object will be reconciled. There will be two cases:
...
-existing Metal3Data object linked to this Metal3DataTemplate and to get the
-unavailable indexes. The indexes always start from 0 and increment by 1. The
-lowest available index is to be used next. The `dataNames` field contains the
-map of Metal3Machine to Metal3Data and the `indexes` contains the map of
-allocated indexes and claims.
+existing Metal3Data object linked to this Metal3DataTemplate and get the
+unavailable indexes. The Metal3Data objects are linked to a Metal3DataTemplate
+by three ways. Either they directly reference the template in the `template`
+field of the `spec`, or they have the same `templateReference` key as the
+template, or the template's `templateReference` matches the `name` of the
```
maelk, line 667: If I understood correctly, the third way is to bridge the gap between the old and new implementations, right ? Maybe that should be specified here, that it's a workaround for the migration ?

design/metadata-handling.md
```diff
@@ -597,15 +608,21 @@ metadata:
     name: nodepool-1
 spec:
   index: 0
+  templateReference: SharedRef
```
maelk, line 611: Is there a need to add that to the metal3dataclaim too ? My guess would be no, considering that the claims are not shared between M3DTs as the Metal3Data are, right ?

maelk (commented):
This looks good to me overall. Maybe the naming "templateReference" is not optimal since it can be quite obscure when in the m3dt itself, what about just "reference" ? or any other idea ?

design/metadata-handling.md
kashifest: Yes. Spot on

design/metadata-handling.md
kashifest: Just to be clear, do you mean to say I should point out why M3DT mutability is not a viable option?

design/metadata-handling.md
kashifest: True, I can repharse it.

design/metadata-handling.md
kashifest: Yes, no need to add it to dataclaim as you correctly pointed out.

kashifest:
> This looks good to me overall. Maybe the naming "templateReference" is not optimal since it can be quite obscure when in the m3dt itself, what about just "reference" ? or any other idea ?

I would rather keep it as templateReference just to be crystal clear since there can be other types of references in future, ofcourse if no strong opinions against it.

fmuyassarov (commented):
LGTM

design/metadata-handling.md
maelk: yes, in the alternatives section

design/metadata-handling.md
```diff
@@ -96,6 +96,11 @@ and be re-used during upgrade.
 As a user, I want to give my own network configuration for cloud-init as a field
 on Metal3Machine
 
+#### Story 5
+
+As a user using a template for metadata and networking, I want to be able to update
```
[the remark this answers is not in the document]
kashifest: Thanks. Addressed

design/metadata-handling.md
kashifest: Addressed

design/metadata-handling.md
kashifest: Addressed

kashifest:
@maelk @fmuyassarov PTAL again. I have addressed the comments

fmuyassarov (commented):
/lgtm

maelk:
/approve
Since there does not seem to be interest on this from others than our team (it was opened more than two weeks ago), let's proceed to merge it, it's a relative small change.

PR #170, opened 2021-03-16
Add design document proposal for noop management interface support

rhjanders:
This change adds a design document requested in [PR811](https://github.com/metal3-io/baremetal-operator/pull/811) for further discussion

### DIFF
--- a/design/baremetal-operator/noop-management.md
+++ b/design/baremetal-operator/noop-management.md
@@ -1,0 +1,171 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add support for Ironic noop management driver in metal3
+
+## Status
+
+implementable
+
+## Summary
+
+This design aims to provide the operator with a metal3 interface
+to disable Ironic BIOS settings management for a specific baremetal node.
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/noop-management.md
```diff
@@ -0,0 +1,160 @@
...
+settings that suit my needs.
+
+## Design Details
+
+Add a new `NoopManagement` attribute to the definitions of BMCs which support
+noop management interface. Similar changes will be required in the Custom
+Resource Definition, `baremetalhost_types`, `make-bm-worker` and BMO
+templates so that the request to create the node with `noop`
+`management_interface` is passed to Ironic.
+
```
hardys, line 80: This section covers the implementation approach, but it'd be useful to add an example to clearly illustrate the new BMH interface.

hardys:
Overall this LGTM - on the implementation PR one of the concerns was the BMH API (e.g new bmc field vs additional BMC URL types) - so I think it would be useful to add a section/example that describes the new API and behavior, in addition to the implementation details already covered.

design/baremetal-operator/noop-management.md
rhjanders: Thank you for your feedback Steve and apologies for the delay in responding. Is this example sufficient, or would you like me to provide further information?

PR #171, opened 2021-03-17
Add proposal for retry interface

andfasano:
Draft proposal to address the problem discussed originally in https://github.com/metal3-io/baremetal-operator/issues/798

### DIFF
--- a/design/baremetal-operator/resume-interface.md
+++ b/design/baremetal-operator/resume-interface.md
@@ -1,0 +1,90 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Title
+
+Resume from error state
+
+## Status
+
+provisional
+
+## Summary
+
+This proposal aims to introduce a new declarative API
+to forcibly resume a waiting BareMetalHost resource from an error state
+when a fix is immediately available, so that it could be possible
+to reduce the waiting time required
+
+## Motivation
+
+The current BMO error retry mechanism implements an
+exponential backoff with jittering: this means that when a BMH
+goes into an error state its reconciliation loop gap increases
+as long as the error conditions persists (based on BMH `Status.ErrorCount`).
+
+In some cases it's necessary a human (or more generally an external)
+intervention to fix the issue. If the fix involves amending the
+BMH resource then an immediate reconciliation is triggered and the
+new state is evaluated again.
+Instead, if the fix does not generate a relevant event (for example,
+a firmware update on the host) captured by BMO, then it's necessary
+to wait for the next scheduled reconciliation loop for a new
+state evaluation, and the waiting time could be longer for higher
+values of `Status.ErrorCount`.
+
+### Goals
+
+* A declarative API to force resuming a BMH currently waiting in error state
+
+### Non-Goals
+
+* Clear or remove the current BMH error state
+
+## Proposal
+
+## Design Details
+
+A new annotation `resume.metal3.io` will be used to trigger a new
+reconciliation loop, and to reset the `Status.ErrorCount` field to 1. In this
+way, the BMH resource will restart reconciliation with a narrowed loops
+
+### Implementation Details/Notes/Constraints
+
+The annotation must be consumed as soon as possible within the BMO reconcile
+loop, and it should be removed as soon as the `Status.ErrorCount` is set to 1.
+If the BMH resource is not in an error state (`Status.ErrorCount` already
+set to zero) then the annotation must be removed and the event properly logged.
+
+### Risks and Mitigations
+
+None
+
+### Work Items
+
+* Modify the BMO reconcile loop to handle the new annotation
+* Unit tests for the above points
+
+### Test Plan
+
+Verify through unit tests that a reconcile loop remove the
+`resume.metal3.io` annotation and sets the ErrorCount field to 1 (if the BMH
+was in error)
+
+### Upgrade / Downgrade Strategy
+
+None
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+Reducing the backoff gaps (or upper limit) could mitigate somehow as the
+waiting time will be reduced
+
--- a/design/baremetal-operator/retry-interface.md
+++ b/design/baremetal-operator/retry-interface.md
@@ -7,7 +7,7 @@
 
 # Title
 
-Resume from error state
+Retry from error state
 
 ## Status
 
@@ -16,9 +16,9 @@ provisional
 ## Summary
 
 This proposal aims to introduce a new declarative API
-to forcibly resume a waiting BareMetalHost resource from an error state
-when a fix is immediately available, so that it could be possible
-to reduce the waiting time required
+to forcibly retry the reconciliation of a waiting BareMetalHost resource
+(being in error state) when a fix is immediately available, so that it
+could be possible to reduce the waiting time required
 
 ## Motivation
 
@@ -32,14 +32,16 @@ intervention to fix the issue. If the fix involves amending the
 BMH resource then an immediate reconciliation is triggered and the
 new state is evaluated again.
 Instead, if the fix does not generate a relevant event (for example,
-a firmware update on the host) captured by BMO, then it's necessary
+a firmware update on the host or a temporary connectivity loss to the
+BMC network) captured by BMO, then it's necessary
 to wait for the next scheduled reconciliation loop for a new
 state evaluation, and the waiting time could be longer for higher
 values of `Status.ErrorCount`.
 
 ### Goals
 
-* A declarative API to force resuming a BMH currently waiting in error state
+* A declarative API to force retrying the evaluation of a BMH currently
+  waiting in error state
 
 ### Non-Goals
 
@@ -49,7 +51,7 @@ values of `Status.ErrorCount`.
 
 ## Design Details
 
-A new annotation `resume.metal3.io` will be used to trigger a new
+A new annotation `retry.metal3.io` will be used to trigger a new
 reconciliation loop, and to reset the `Status.ErrorCount` field to 1. In this
 way, the BMH resource will restart reconciliation with a narrowed loops
 
@@ -62,7 +64,7 @@ set to zero) then the annotation must be removed and the event properly logged.
 
 ### Risks and Mitigations
 
-None
+-
 
 ### Work Items
 
@@ -72,7 +74,7 @@ None
 ### Test Plan
 
 Verify through unit tests that a reconcile loop remove the
-`resume.metal3.io` annotation and sets the ErrorCount field to 1 (if the BMH
+`retry.metal3.io` annotation and sets the ErrorCount field to 1 (if the BMH
 was in error)
 
 ### Upgrade / Downgrade Strategy


andfasano:
cc @s3rj1k @dhellmann

s3rj1k:
@andfasano IMO naming annotation `resume.metal3.io` could bring some with pause annotation.
Assuming someone does not have deep knowledge of docs, he would assume that adding resume annotation would remove pause annotation as well.

If both of annotations does not have intended dependencies, does not affect each other, I would recommend renaming `resume.metal3.io` to something else.

Other then naming issue looks good and should solve related issue without removing error context as you mentioned before.

design/baremetal-operator/resume-interface.md
```diff
@@ -0,0 +1,90 @@
...
+exponential backoff with jittering: this means that when a BMH
+goes into an error state its reconciliation loop gap increases
+as long as the error conditions persists (based on BMH `Status.ErrorCount`).
+
+In some cases it's necessary a human (or more generally an external)
+intervention to fix the issue. If the fix involves amending the
+BMH resource then an immediate reconciliation is triggered and the
+new state is evaluated again.
+Instead, if the fix does not generate a relevant event (for example,
+a firmware update on the host) captured by BMO, then it's necessary
```
zaneb, line 35: Yeah, another example could be losing network connectivity to the BMC network. This will put all of the BMHs into error backoffs, but once you restore connectivity you might want to retry immediately.

design/baremetal-operator/resume-interface.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Non-Goals
+
+* Clear or remove the current BMH error state
+
+## Proposal
+
+## Design Details
+
+A new annotation `resume.metal3.io` will be used to trigger a new
```
zaneb, line 52: What do you think about `retry` instead of `resume`?

design/baremetal-operator/resume-interface.md
```diff
@@ -0,0 +1,90 @@
...
+reconciliation loop, and to reset the `Status.ErrorCount` field to 1. In this
+way, the BMH resource will restart reconciliation with a narrowed loops
+
+### Implementation Details/Notes/Constraints
+
+The annotation must be consumed as soon as possible within the BMO reconcile
+loop, and it should be removed as soon as the `Status.ErrorCount` is set to 1.
+If the BMH resource is not in an error state (`Status.ErrorCount` already
+set to zero) then the annotation must be removed and the event properly logged.
```
zaneb, line 61: Is there a way to avoid a thundering herd bearing down on ironic if somebody does a bulk apply of this to all BMHs in a system at once?

design/baremetal-operator/resume-interface.md
s3rj1k: IMO: retry is fine

design/baremetal-operator/resume-interface.md
s3rj1k: @zaneb doesn't PROVISIONING_LIMIT solves this ?

design/baremetal-operator/resume-interface.md
andfasano: Sounds fine

design/baremetal-operator/resume-interface.md
andfasano: That check is limited only to a subset of states, while the annotation trigger in the current proposal is not bound to anyone. So I don't think currently we have a complete mechanism to prevent that, and moreover this could be a more general problem for any annotation that may request an Ironic action (see for example the [`inspect` annotation,](https://github.com/metal3-io/baremetal-operator/pull/607) even though in such case it could be triggered only from the `Ready` state).
If it's ok, I could add this point in the risks paragraph.

design/baremetal-operator/resume-interface.md
andfasano: Good case, I'll add it in the description

design/baremetal-operator/retry-interface.md
```diff
@@ -0,0 +1,92 @@
...
+## Status
+
+provisional
+
+## Summary
+
+This proposal aims to introduce a new declarative API
+to forcibly retry the reconciliation of a waiting BareMetalHost resource
+(being in error state) when a fix is immediately available, so that it
+could be possible to reduce the waiting time required
```
dhellmann, line 21: I'm still not clear why we need a special API for this. Any edit to the host, including adding a temporary label, would trigger reconciliation.

design/baremetal-operator/retry-interface.md
```diff
@@ -0,0 +1,92 @@
...
+None
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+Reducing the backoff gaps (or upper limit) could mitigate somehow as the
+waiting time will be reduced
```
dhellmann, line 91: Why not reduce the upper bound for the backoff?

design/baremetal-operator/retry-interface.md
zaneb: I guess the difference this makes is that if it fails again for a different (perhaps more easily recoverable) reason then the time to retry is shorter.
That said, I must confess to being [somewhat sceptical](https://github.com/metal3-io/baremetal-operator/issues/798#issuecomment-806048253) of the value also.

design/baremetal-operator/retry-interface.md
zaneb: Another alternative: build a CLI util that resets the error count without making it part of the BMH API.

design/baremetal-operator/retry-interface.md
andfasano: Yes, the key point of the proposal it's about resetting the ErrorCount to 1 - otherwise any edit will be fine fo retriggering the reconciliation.

design/baremetal-operator/retry-interface.md
andfasano: > Why not reduce the upper bound for the backoff?

I think it's listed as a an alternative. `maxBackOffCount` could be reduced to cap the max waiting time after the host reached that number of (failed) iterations. Current gap values can be checked [here](https://github.com/metal3-io/baremetal-operator/blob/3f79aae54bcead50cee67f8f2b19ecd605e457f2/controllers/metal3.io/action_result.go#L127-L137)

dhellmann: Yes, it is listed as an alternative, but there is no reason given about why it was rejected.

design/baremetal-operator/retry-interface.md
dhellmann: Why do we want to artificially reset the error count? Just to shorten retry intervals if there's another failure?

andfasano: Yes (or another set of failures)

design/baremetal-operator/retry-interface.md
andfasano: Maybe I misunderstood the goal of the `Alternatives` section, but my intention was to list another approach so that it could have been discussed. Personally I think it's still a potentially viable alternative - as not requiring explicit user intervention

andfasano:
Hi all,
I'd like to summarise the current situation in order to continue if possible the discussion and reach a conclusion. It looks like the consensus is not to change the API with an additional annotation, but eventually to improve the retry mechanism where needed or modify the upper limit to reduce the waiting time. The latter could slightly help in those situations where a resource having an high ErrorCount value receives an update request (triggered by a normal timeout or due an artificial change request on the resource) and fails again - this because ErrorCount is usually reset on a successful state change. The pro of such approach is that it wouldn't require any change in the public API, and the user that would request an immediate re-evaluation could force it with a normal update.

s3rj1k:
@andfasano I we go this way, maybe we could have some kind of CLI utility to patch-up spec just in case then fix is needed ASAP?

https://github.com/metal3-io/metal3-docs/pull/171#discussion_r604984609

cc: @zaneb

andfasano:
> @andfasano I we go this way, maybe we could have some kind of CLI utility to patch-up spec just in case then fix is needed ASAP?
> 
> [#171 (comment)](https://github.com/metal3-io/metal3-docs/pull/171#discussion_r604984609)
> 
> cc: @zaneb

As already mentioned [here](https://github.com/metal3-io/baremetal-operator/issues/798#issuecomment-785820755) I think that the ErrorCount should be a field managed by BMO only without any external intervention. If an intervention is required immediately then there is still the possibility to manually force the update.

s3rj1k:
@andfasano Well, this is not community friendly, this effectively will force patching operator `in field` when fixing error is not possible.

This is what triggered all of this discussion, I everybody is fine with that, let's close this and hope that no body gets hurt by broken production cluster.

andfasano:
> @andfasano Well, this is not community friendly, this effectively will force patching operator `in field` when fixing error is not possible.
> 
> This is what triggered all of this discussion, I everybody is fine with that, let's close this and hope that no body gets hurt by broken production cluster.

TBH I'm not sure to understand exactly the point, given that I was originally fine about the proposal of lowering the ErrorCount (to 1) via annotation, and actively trying to reach a solution. From a design point of view I simply don't agree moving such action outside from the operator scope (the manual update has been mentioned just as a current workaround to avoid waiting the next reconcile loop). I'm also fine in reducing the ErrorCount cap limit as an alternative approach, as it could allow the operator to re-evaluate the resource currently in error slightly more frequently.

andfasano:
> > @andfasano Well, this is not community friendly, this effectively will force patching operator `in field` when fixing error is not possible.
> > This is what triggered all of this discussion, I everybody is fine with that, let's close this and hope that no body gets hurt by broken production cluster.
> 
> TBH I'm not sure to understand exactly the point, given that I was originally fine about the proposal of lowering the ErrorCount (to 1) via annotation, and actively trying to reach a solution. From a design point of view I simply don't agree moving such action outside from the operator scope (the manual update has been mentioned just as a current workaround to avoid waiting the next reconcile loop). I'm also fine in reducing the ErrorCount cap limit as an alternative approach, as it could allow the operator to re-evaluate the resource currently in error slightly more frequently.

Just to be clear, after reviewing Zane's comment, if the consensus is about writing a very tiny CLI tool that only resets the ErrorCount to 1 (but the error status is maintained untouched as mentioned in the current proposal), just as a sort of "safety valve" for unplanned situations, then it would be ok for me.

zaneb:
I'd have thought the field recovery procedure would go like this:

1. fix whatever external issue is causing the problem
2. make literally any change to the BMH to retrigger reconciliation
3. wait to see if it works
4. if not, go to 1.

Resetting the error count can help if you want to skip step 3, but that hardly seems like an emergency worthy of recompiling a new bmo (which will in most cases take longer than waiting to see if it worked). Nevertheless, while I don't quite see the point, having a CLI tool that just does `if errorCount > 1 { errorCount = 1}` seems mostly harmless. If somebody wrote that and wanted to submit it upstream I wouldn't object.

s3rj1k:
@zaneb `3.` I would need to wait for ~5h for reconcile re-trigger, as by the time `1.` and `2.` are finished, error back-off is triggered

dhellmann:
> @zaneb `3.` I would need to wait for ~5h for reconcile re-trigger, as by the time `1.` and `2.` are finished, error back-off is triggered

The error back-off is only applied when the operator tells controller-runtime to try again after a delay. It is not applied when reconciliation is triggered because the CR is edited. So, step 2 should trigger reconciliation immediately, without waiting for the error back-off. Is that not the behavior you see?

s3rj1k:
@dhellmann No. Last time I checked changing something in BMH CR re-triggers single reconcile loop that is again stuck at back-off timer at the and.

Only thing that forced a reset on error count was removing provisioning image URL, and that triggered deprovisioning.

Deprovisioning a perfectly healthy (working kubeworker but broken BMH object) host in a bit harsh solution.

So the idea emerged, after discussion in several different threads, to somehow fix long back-off delay, either by resetting errorCount to 1 or by changing back-off time.

dhellmann:
> @dhellmann No. Last time I checked changing something in BMH CR re-triggers single reconcile loop that is again stuck at back-off timer at the and.

This feels like a bug. If we successfully reconcile the host, we should reset that back-off. @andfasano, what do you think?

> So the idea emerged, after discussion in several different threads, to somehow fix long back-off delay, either by resetting errorCount to 1 or by changing back-off time.

We need to make the BMO manage that value correctly.

s3rj1k:
@dhellmann 
> This feels like a bug. If we successfully reconcile the host, we should reset that back-off.

I do not thing that reconcile was successful, it was just stuck at back-off with no way of re-triggering successful reconcile. (only via deprovisioning).

I do believe that a root cause of that bug was fixed by https://github.com/metal3-io/baremetal-operator/commit/c45ecfb94c8a1912c621755f216ec3a6896a7189.

But back-off part is still the same, so this could re-emerge in some other unexpected place.

zaneb:
After you trigger the reconcile, if another error happens it will indeed not retry a third time until the delay has expired (which will be a long time if the error count is high). If the problem is not fixed then it doesn't matter if you retry fast or slow because it will always fail - in fact slow is better because it wastes less resources. If this is caused by a bug in BMO, there is no substitute for fixing the bug or at least hacking around in ironic. If it is caused by some external condition (e.g. network unavailable) then that needs to be resolved before trying again.

However, if there is no error on the retry then it will proceed and eventually clear the error and the error count. If this doesn't happen it's a bug.

andfasano:
> After you trigger the reconcile, if another error happens it will indeed not retry a third time until the delay has expired (which will be a long time if the error count is high). If the problem is not fixed then it doesn't matter if you retry fast or slow because it will always fail - in fact slow is better because it wastes less resources. If this is caused by a bug in BMO, there is no substitute for fixing the bug or at least hacking around in ironic. If it is caused by some external condition (e.g. network unavailable) then that needs to be resolved before trying again.
> 
> However, if there is no error on the retry then it will proceed and eventually clear the error and the error count. If this doesn't happen it's a bug.

I think that @Zane summarized very precisly the situation and the expected behavior - when BMO reacts to a resource related event. For that, if there is a specific execution flow in BMO that doesn't reset properly the ErrorCount (upon success) it's definitely a bug and it should be addressed case by case. 

Instead, a potentially interesting scenario addressed by the current proposal (or the CLI) was mainly related when the fix for a resource (currently in error state) _does not_ generate an event (as for example the firmware update on the host). In such case it is necessary to wait the natural timeout of the previously enqueued request for that host or, if you don't want to wait for such period, generate artificially a new one to force BMO reaction (and fallback in the previously described situation).

dhellmann:
> @dhellmann
> 
> > This feels like a bug. If we successfully reconcile the host, we should reset that back-off.
> 
> I do not thing that reconcile was successful, it was just stuck at back-off with no way of re-triggering successful reconcile. (only via deprovisioning).
> 
> I do believe that a root cause of that bug was fixed by [metal3-io/baremetal-operator@c45ecfb](https://github.com/metal3-io/baremetal-operator/commit/c45ecfb94c8a1912c621755f216ec3a6896a7189).
> 
> But back-off part is still the same, so this could re-emerge in some other unexpected place.

So the BMO correctly responded to an update event and reconciled. That reconciliation resulted in an error. The error count on the host was not decreased, and the BMO went back to waiting for the error back-off timeout.

That's the behavior I would expect when an error occurs. Did you expect something different?

zaneb:
> @dhellmann
> 
> > That's the behavior I would expect when an error occurs. Did you expect something different?
> 
> I expected that baremetal-operator would allow me to try to fix the issue without waiting ~5h for next reconcile :)

But you hadn't fixed the issue yet. That's why it failed again.

Once you had, you can trigger another reconcile at any time by writing any change at all to the BMH.

_You_ are _always_ free to try to fix the issue at _any_ time. The backoff time is only the amount of time that the baremetal-operator will wait before trying to fix the issue _itself_ - in the dumbest way possible (blindly trying again). The time increases with each failure because you don't want it doing that in a tight loop.

> It is not always obvious why error is occurring, so to try different fixes one would have to try different changes.

That's fair, but you can always retrigger a reconcile after each try. It goes without saying that you are interactively debugging it at this point, so this is always an option and there's no real need to have it automatically retrying on a quicker schedule (and in fact it might be better if it doesn't, because that could interfere with your debugging attempts).

> In current situation the only way to re-trigger single reconcile (without deprovisioning) is to use pause annotation :)
> Seems like a hack.

You can use any annotation you want to trigger a reconcile. It doesn't have to be one recognised by the baremetal-operator.

Furthermore, by bringing up deprovisioning you are conflating the original issue, which prevented the BareMetalHost running inspection and therefore left it totally unusable, with the provisioned state in which only power management errors can occur. The ErrorCount is cleared on any successful state transition, or any successful power management action in a provisioned/ready state, not only on deprovisioning.


We're going around in circles at this point. The exact nature of the problem in metal3-io/baremetal-operator#798 is lost in the mists of time. The root cause was addressed by metal3-io/baremetal-operator#817. Meanwhile, we have also merged metal3-io/baremetal-operator#840 which fixed a real bug with error handling during inspection that could easily have resulted in very confusing behaviour.

As far as we can establish, the error backoff is now working the way one would expect, and we have no reason to think that retriggering a reconcile would not allow a complete recovery from an error condition that has been corrected. If you see any further issues that are caused by unintended behaviour, please open another bug.

/close

s3rj1k:
@zaneb Fair, probably it is impossible to reproduce that state at this point.

Can we tweak back-off time to make it not so long?

cc: @andfasano

PR #173, opened 2021-03-30
Add proposal to do bulk set of BIOS configuration

bfournie:
Draft of a proposal to do bulk set of BIOS Configuration across
similar vendors.

### DIFF
--- a/design/baremetal-operator/bulk-set-bios-config.md
+++ b/design/baremetal-operator/bulk-set-bios-config.md
@@ -1,0 +1,445 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# bulk-set-bios-config
+
+## Status
+
+implementable
+
+## Summary
+
+Provide the ability to get and set the BIOS Configuration. Retrieve the current
+BIOS Configuration settings from the host via Ironic and use that as a template.
+Allow a user to set new values for the BIOS Configuration. Validate the values
+before sending to Ironic via the clean-steps API. The values from the CRD on
+one host can be manually copied to other hosts of the same vendor and model,
+effectively doing a `bulk set`.
+
+## Motivation
+
+When provisioning a large amount of machines with identical settings, our
+customers want to provide a known, validated payload with all the desired BIOS
+configuration. This ensures all the machines will have the identical settings,
+reducing issues caused by unexpected configuration differences.
+
+The challenge that BIOS configuration presents is that no two hardware vendors
+use the same names for settings that control the same behavior, and the
+available settings often overlap but have significant differences. There
+currently is an approach that would abstract these differences for individual
+vendors where the common names are converted to vendor names by the BMC driver
+[individual set](https://github.com/metal3-io/baremetal-operator/pull/302).
+This new proposal would not require BareMetalHost to map the vendor specific
+names to a common set of names to manage the settings. It will be entirely
+transparent and inherently support all vendors. This new proposal could also
+work in conjunction with the individual settings approach. A user could
+initially set up a machine using that method and then copy the configuration
+to multiple machines as described here.
+
+Most hardware vendors support a way to apply bulk settings, often by manually
+configuring one host, exporting those settings in a vendor-specific file
+format, and then importing that file on the BMCs of other hosts. Accepting that
+the format for these profiles is vendor-specific eliminates a lot of the
+implementation complexity, at the risk of moving that complexity to the
+end-user. This requirement is similar to something Dell has proposed in the
+Ironic community
+([hw config](https://review.opendev.org/c/openstack/ironic-specs/+/740721))
+to export and import an entire system iDRAC inventory including RAID and
+networking requirements. Compared to that approach, this proposal attempts
+a more lightweight approach just for BIOS configuration.
+
+### Goals
+
+- Allow updating BIOS configuration using the vendor's BIOS names and not
+  require a mapping layer to a common set of names
+- Create a template for the current BIOS configurations settings cached on a
+  server
+- Give access to the BIOS registry which can be used as documentation so a
+  user can easily update the BIOS settings to desired values
+- Validate the BIOS Configuration values based on the vendor's BIOS Registry
+  before sending to the BMC
+- Provide ability to manually transfer the settings in a bulk manner across
+  similar hardware in the cluster using `kubectl` commands. Its possible to
+  use scripts to help automate this `bulk set`
+
+### Non-Goals
+
+This proposal does not attempt to fulfill the following goals:
+
+- Allow the transfer of BIOS configuration between different vendor’s
+  servers or different vendor models
+- Support bulk set of other configurations besides BIOS
+- Provide storage mechanism for BIOS settings
+- Keep the BIOS settings on similar hardware in sync automatically
+  using an additional operator, although that functionality is planned
+  for a follow-on
+- A future feature is planned to make use of this funtionality to
+  automatically keep BIOS settings in sync on similar hardware using
+  a new operator. This future will require a separate proposal and is out
+  of scope for this proposal.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+When provisioning a large amount of machines with identical settings, our
+customers want to provide a known, validated payload with all the desired
+firmware configuration.
+
+This ensures all the machines will have the identical settings, reducing issues
+caused by unexpected configuration differences.
+
+Since the name/value pair mapping is vendor specific and not intuitive, it
+doesn’t lend itself to creating the input by hand using existing vendor
+documentation, except for potentially a small subset of names. It’s desirable
+to take an existing BIOS config set and modify the current settings to the
+desired settings. The BIOS Registry can be used as a source of documentation
+to modify the settings.
+
+## Design Details
+
+This proposes a new Custom Resource Definition (CRD) to store the settings
+read from Ironic along with the modified settings to write to Ironic. Since
+each vendor provides 100 or more BIOS Settings, its not practical to store
+these in the BareMetalHost CRD. The new CRD will be named
+`HostFirmwareSettings` and will consist of the following:
+
+- `settingsData` - the current BIOS Settings (names and values) retrieved
+  from Ironic via the BIOS API (`v1/nodes/{node_ident}/bios?detail=True`)
+  [BIOS API](https://docs.openstack.org/api-ref/baremetal/?expanded=#node-bios-nodes)
+  will be stored in the Status section along with the BIOS Registry
+  information which provides a schema for the settings.
+- `vendorSettings` - BIOS Settings to write to the host via Ironic will be
+  stored in the Spec section. It will be initially populated with the
+  name/value pairs from `settingsData` that do not have the `read-only` or
+  `unique` schema flag set. Existing values will not be overwritten.
+
+A second CRD `FirmwareSchema` will be added to store schema data (aka BIOS
+Registry) received from Ironic via the BIOS API. This schema provides
+additional info about each BIOS setting along with limits that can be used
+both by a user when changing the settings or as a part of input validation.
+The BIOS Registry support was recently added to Ironic (see the Ironic proposal
+[BIOS Registry](https://review.opendev.org/c/openstack/ironic-specs/+/774681).
+This schema is the same for each vendor's model so it will not need to be
+updated when stored. More info for the BIOS registry data can be found in the
+[Redfish Registry](https://redfish.dmtf.org/schemas/v1/AttributeRegistry.v1_3_5.json)
+
+The schema fields returned via the BIOS API include:
+
+- Type (e.g. enum, string, integer, boolean, or password)
+- ReadOnly (a boolean indicating whether the attribute is read only)
+- Unique (indicating the value is specific to the host where it was
+  retrieved)
+- Allowable values for enum type attributes
+- Minimum and maximum values (for numeric attributes)
+- Minimum and maximum character lengths (for string attributes)
+
+The BIOS Settings are retrieved from the BMC by Ironic and cached
+whenever the node moves to `manageable` or `cleaning`, or when the settings
+are updated. The baremetal-operator will manage the data as follows:
+
+- The node first transitions to manageable during the `Registering` state, so
+  at the end of that state `settingsData` will be populated.
+- Settings can be updated during the `Preparing` state, so at the end of that
+  state the settings will also be retrieved and used to update
+  `settingsData`.
+
+Upon retrieving the BIOS Settings from Ironic, the baremetal-operator will
+copy the name/value pairs to the `vendorSettings` field in
+`HostFirwareSettingsSpec` if that field is empty. Settings that have the
+`read-only` flag set will not be copied as these settings cannot be written
+to the BMC. Likewise, settings that have the `unique` flag set will also not
+be copied, these are settings like serial numbers or product ids for example,
+that should not be written to other servers.
+
+A user can update `vendorSettings` to the desired values for the
+settings using the limits from the `FirmwareSchema` section as a guide.
+
+The baremetal-operator will detect when `vendorSettings` changes by comparing
+the name/value pairs to `settingsData`. When a change is detected, the BMO
+will add the new values to the Ironic clean-steps API in the Preparing state,
+when building the manual clean steps. Before adding it though, the BMO will
+do validation checks on the new values using the schema. For example, integer
+types will checked against minimum and maximum parameters and enumeration types
+will be compared against allowable values. A failure will be returned from the
+clean steps stage if any errors are detected in order to avoid pushing a
+partial configuration to Ironic.
+
+As a follow-on, a validation webhook for the HostFirmwareSettings CRD can be
+leveraged to add the validations. This webhook would reject the update before
+it is stored in the resource, removing the need to do the validations in
+manual clean steps. Since there could be some scenarios where the webhook would
+not be available, the same validation checks will be done in the controller
+before adding the settings to the clean steps.
+
+When valid settings are added to the Ironic clean-steps API, Ironic will set
+the BIOS Configuration in the BMC as part of the manual cleaning process which
+requires a reboot. Manual clean occurs in the Preparing state of the
+BaremetalHost, the Host will re-enter this state from Ready/Available state
+whenever its config differs from the last one it stored.
+
+After applying the new BIOS Configuration, Ironic will then read and cache the
+new BIOS Settings, which can be retrieved by the BMO and used to update
+`settingsData`.
+
+Once a known good set of values are stored in the `vendorSettings` it can be
+used to update the spec section of other hosts using the `kubectl` commands,
+effectively cloning the settings from a reference host and ensuring that all
+similar systems have the same BIOS configuration. External scripts can be
+used with `kubectl` to help automate this manual configuration.
+
+### Implementation Details/Notes/Constraints
+
+One of the primary drivers of his design was that each vendor uses different
+names and values for similar BIOS Configuration settings. This design does not
+rely on vendor specific checks but instead uses common mechanisms to ensure
+the BIOS Settings are valid:
+
+- copying the names and values from `settingsData` to
+  `vendorSettings` ensures initial values are correct
+- when updating the values in `vendorSettings`, the user can use the
+  limitations in the `settingsData`, e.g. minimum and maximum values
+- before writing to the Ironic clean-steps API, the BMO runs validation checks
+  against the `settingsData` to ensure updated values are valid
+
+In order to interoperate with the individual BIOS Configuration set
+approach the BMO must ensure that the same settings don't get added twice
+to the Ironic clean-steps API. In addition, as each clean step requires a
+reboot the number of clean steps that are added must be managed.
+
+The use of a second set of CRs to store schema data instead of storing
+the schema in `HostFirmwareSettings` is to reduce storing duplicate
+schema information. As hosts of the same vendor and model will use
+the same schema, its not necessary to cache it each time for similar hardware.
+When BIOS data is read from the Ironic API for each host the following
+actions will be taken:
+
+- If the `HostFirmwareSettings` already has a reference to a  `FirmwareSchema`
+  resource, then the schema data in the BIOS API will be ignored.
+- If there is no reference, the schema data for all BIOS settings for the
+  host will be stored locally.
+- As the same schema names and values will be used for systems of the
+  same vendor and model number, there is an opportunity to conserve
+  creating multiple copies of the schema.
+- A hash will be generated from names in the schema, if a corresponding hash
+  does not exist in the `FirmwareSchema` CRs, a new CR will be created and
+  the schema data will be copied to it.
+- The reference to the `FirmwareSchema` resource in `HostFirmwareSettings`
+  will be updated with the hash.
+- Using the hash will ensure that the schemas match. In the case of
+  hardware from the same vendor and model, but with a different firmware
+  version, it is possible that the schemas would differ. In this case a new
+  `FirmwareSchema` will be created.
+- The name of the `FirmwareSchema` resource will be added to the
+  `HostFirmwareSettings` resource so that it can be used to delete the
+  `FirmwareSchema` resource when it is no longer being used.
+- It is possible that the hosts will be in different namespaces (and the
+  `FirmwareSchema` will be in same namespace as host). For this case,
+  a `namespace` hint will be added to the BMO to specify the namespace to
+  use for the schema. These BMO changes will be described in a follow-on
+  doc.
+- It is recommended that the refererence to the schema not use
+  corev1.ObjectReference, but instead define a separate struct, see
+  [ObjectReference](https://github.com/kubernetes-sigs/cluster-api/issues/2318)
+
+An example of the proposed CRD for `HostFirmwareSettings` is as follows:
+
+```yaml
+---
+kind: CustomResourceDefinition
+metadata:
+  name: hostfirmwaresettings.metal3.io
+spec:
+  group: metal3.io
+  names:
+    kind: HostFirmwareSettings
+    listKind: HostFirmwareSettingsList
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+          spec:
+            description: HostFirmwareSettingsSpec defines the desired state
+            properties:
+              vendorSettings:
+                items:
+                  description: describes a setting whose value can be changed
+                  properties:
+                    name:
+                      description: Identifier for the Firmware setting.
+                      type: string
+                    value:
+                      description: Value of the Firmware setting.
+                      type: string
+                  required:
+                  - name
+                  - value
+                  type: object
+                type: array
+            type: object
+          status:
+            description: defines the observed state of HostFirmwareSettings
+            properties:
+              schemaReference:
+                  description: Reference to resource containing schema
+                  type: string
+              schemaNamespace:
+                  description: The namespace where the schema resource resides
+                  type: string
+              settingsData:
+                items:
+                  description: describes current settings and their schema
+                  properties:
+                    name:
+                      description: Identifier for the Firmware setting.
+                      type: string
+                    value:
+                      description: Value of the Firmware setting.
+                      type: string
+                type: array
+            type: object
+```
+
+An example of the proposed CRD for `FirmwareSchema` is as follows:
+
+```yaml
+---
+kind: CustomResourceDefinition
+metadata:
+  name: firmwareschema.metal3.io
+spec:
+  group: metal3.io
+  names:
+    kind: FirmwareSchema
+    listKind: FirmwareSchemaList
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+          status:
+            description: The schema to use for HostFirmwareSettings
+            properties:
+                schemaIdentifier:
+                  description: Unique identifier for this schema, derived
+                               from a hash of the schema contents
+                  type: string
+                schemaVendor:
+                  description: The vendor that corresponds to this schema
+                  type: string
+                schemaModel:
+                  description: The model that corresponds to this schema
+                  type: string
+                schema:
+                      description: Vendor specific data describing each setting
+                  items:
+                    properties:
+                      name:
+                        description: Identifier for the Firmware setting.
+                        type: string
+                      allowable_values:
+                        description: The allowable value for Enumeration type
+                        items:
+                          type: string
+                          type: array
+                      attribute_type:
+                        description: The type of setting.
+                        enum:
+                        - Enumeration
+                        - String
+                        - Integer
+                        - Boolean
+                        - Password
+                        type: string
+                      lower_bound:
+                        description: The lowest value for an Integer type
+                        type: integer
+                      max_length:
+                        description: Maximum length for a String type
+                        type: integer
+                      min_length:
+                        description: Minimum length for a String type
+                        type: integer
+                      read_only:
+                        description: Whether or not this setting is read only
+                        type: boolean
+                      reset_required:
+                        description: Whether or not a reset is required
+                        type: boolean
+                      unique:
+                        description: Whether or not setting's value is unique
+                        type: boolean
+                      upper_bound:
+                        description: The highest value for an Integer type
+                        type: integer
+                    type: object
+```
+
+### Risks and Mitigations
+
+There is no risk in exposing the majority of the BIOS settings even if these
+values are ReadOnly, such as `NumCores` or `CoreSpeed`. However, some vendors
+use the `Password` AttributeType defined in
+[Registry](https://redfish.dmtf.org/schemas/v1/AttributeRegistry.v1_3_5.json)
+and include settings like `SysPassword`. Although the current values are not
+included, or set to "", for these types, the values should not be stored in
+the status section and they should not be allowed to be set via the spec
+section.
+
+### Work Items
+
+Ironic
+
+- Cache BIOS Registry
+- Add support for BIOS Registry information in the Ironic API
+
+Gophercloud
+
+- Add support for nodes/{node}/bios endpoint.  The microversion
+  should be 1.74 in order to use the ``?detail=True`` field to
+  get the bios registry information.
+
+Baremetal-operator
+
+- Add a new CRD for `HostFirmwareSettings`
+- Add a new CRD for `FirwmareSchema``
+- Get BIOS Settings via the Ironic API at the end of the Registration state and
+  store the name/value pairs in `settingsData`
+- Determine if a schema already exists for the BIOS registry data received
+  in the Ironic API. Create a new `FirmwareSchema` CR if one doesn't exist
+  and update the reference in `settingsData`
+- Copy name/value pairs from `settingsData` to
+ `vendorSettings` for settings that are not `read-only` or `unique`
+- Check for changes to `vendorSettings` and when detected, validate
+  values and add updated name/value pairs to Ironic clean-steps API
+
+### Test Plan
+
+- Unit test for changes to provisioner and baremetal-controller
+- metal3-dev-env integration test
+- Test on running cluster
+- Verify that new Custom Resources for HostFirmwareSettings and
+  schema are created and match the Ironic settings
+- Verify that the settings can be updated
+- Verify that after the host goes through cleaning the settings have
+    been updated
+
+## Alternatives
+
+- Rely on the granular approach for setting BIOS configuration for a few
+  individual values in a vendor-specific way as proposed.
+- Support the Configuration Mold approach in the installer as proposed in Ironic
+
+## References
+
+[individual set](https://github.com/metal3-io/baremetal-operator/pull/302)
+[ironic hw config](https://review.opendev.org/c/openstack/ironic-specs/+/74071)
+[BIOS API](https://docs.openstack.org/api-ref/baremetal/?expanded=#node-bios-nodes)
+[BIOS Registry](https://review.opendev.org/c/openstack/ironic-specs/+/774681)
+[Redfish Registry](https://redfish.dmtf.org/schemas/v1/AttributeRegistry.v1_3_5.json)
+[ObjectReference](https://github.com/kubernetes-sigs/cluster-api/issues/2318)


design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,215 @@
...
+          }
+        ],
+        "WarningText": null,
+        "WriteOnly": false
+    }
+
+The cached BIOS registry from Ironic will also be made available in the
+BaremetalHost CRD. This can be a ConfigMap referenced by the
+`BaremetalHostStatus`. Since the BIOSRegistry can be quite large this should be
+stored gzip compressed so as to not exceed the 1Mb data limit in the ConfigMap.
```
dhellmann, line 154: These are big, and likely to be the same for a lot of hosts in a homogeneous environment. Instead of always collecting them, what would an API look like for requesting a registry explicitly?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,215 @@
...
+identified in the BIOS Registry, when building the JSON input. It’s not
+expected that the entire set read from Ironic will be passed down as:
+1. many of the values read from Ironic are ReadOnly
+2. setting ReadOnly values to their current settings increases the size of the
+input and may potentially affect the Redfish request to the BMC
+
+The BIOS Settings are changed during the node cleaning step which requires a
+reboot.  Manual clean occurs in the Preparing state of the BaremetalHost, the
+Host will re-enter this state from Ready/Available state whenever its config
+differs from the last one it stored.
```
dhellmann, line 185: I think we've previously said we only wanted to apply BIOS settings when metal3 thinks it needs to reprovision the host. I can't tell from the way this is phrased if you're suggesting that we include BIOS settings in the triggers for reprovisioning, or if we would still only apply BIOS settings when a new image is going to be written.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Good point, we only need the BIOS registry on the reference hosts, there is no reason to store it on all hosts in particular, especially considering the size of this registry.  I've updated this section although left the API definition general for the design phase.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,215 @@
...
+stored gzip compressed so as to not exceed the 1Mb data limit in the ConfigMap.
+
+Using the BIOS settings and BIOS registry, the user will be able create a JSON
+list of bulk BIOS settings (aka “opaque blob”). This same JSON blob can be set
+on multiple machines to load and apply, effectively cloning all the settings
+from another machine. This user data will also be stored in a ConfigMap
+referenced in the BareMetalHostStatus. The BMO will send the data to Ironic via
+the clean-steps API without validating or interpreting it. Ironic will convert
+the JSON to a set of clean steps and run cleaning to configure the BIOS
+settings.
```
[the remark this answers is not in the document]
bfournie: OK, update this section to indicate that it needs a CRD definition.  Again I'd like to push the definition off to the design phase after we have general agreement.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Implementation Details/Notes/Constraints
+
+One of the primary drivers of this design was the desire to not have the
+Installer validate the user inputted bulk BIOS settings as it requires vendor
```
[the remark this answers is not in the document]
bfournie: Rewrote this to applicable to BMO and indicated that we should have definition of CRD although the individual fields are vendor specific and don't require definition.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Yes, rewrote this to indicate that a change to BIOS settings will not require reprovisioning.  The settings will be written during the provisioning stage if the desired (bulk) settings don't match the current settings.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,217 @@
...
+            "ValueName": "Enabled"
+          },
+          {
+            "ValueDisplayName": "Disabled",
+            "ValueName": "Disabled"
+          }
+        ],
+        "WarningText": null,
+        "WriteOnly": false
+    }
```
dhellmann, line 151: How firm is this part of the standard? If most vendors support s similar schema, it seems like we could expose a subset of the information in a CRD using a similar format. The `Name`, `CurrentValue`, `Type`, and `ReadOnly` fields look especially useful. Maybe the `Value`, too, since that seems to be related to the enum?

Do integer values have min and max settings?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,217 @@
...
+whenever the node moves to `manageable` or `cleaning`. These BIOS settings
+can be read via the Ironic API (`/v1/nodes/{node_ident}/bios`) as a set of
+name:value pairs. Ironic can set the BIOS Configuration values through the use
+of clean_steps as part of the manual cleaning process as described in the Ironic
+documentation. BIOS settings are configured as name/value pairs and can include
+from one to multiple pairs.
+
+What we are proposing is to make the cached BIOS settings from Ironic available
+to the user. This can be either via a ConfigMap or special purpose CRD
+referenced in the BareMetalHostStatus.  As there are no fields in BIOS
```
dhellmann, line 111: I think what we'll want is a new CRD with a name like `BIOSRegistry` that has a host reference in the spec fields. When one is created, a controller would wait for ironic to have the information it needs and then fill in information in the status fields of `BIOSRegistry`. That means the main reconcile loop of the baremetal-operator doesn't have to deal with this new feature at all, and there can be separate RBAC settings for being able to ask for BIOS info.

We could (eventually) do things like filter out read-only values, or even let the user ask for specific settings by name.

What do you think, @zaneb?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,217 @@
...
+1. many of the values read from Ironic are ReadOnly
+2. setting ReadOnly values to their current settings increases the size of the
+   input and may potentially affect the Redfish request to the BMC
+
+The BIOS Settings are changed during the node cleaning step which requires a
+reboot. The change to BIOS settings will not be a trigger for reprovisioning a
+host, the BIOS settings will only be applied when a new image is going to be
+written. In order to prevent redundant BIOS settings to the BMC, the current
+BIOS settings should be compared to the bulk BIOS settings and the settings
+only written to Ironic when there is a difference.
```
dhellmann, line 189: The BMO would need to do this filtering work?

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Got it :-)  I'll work on adding details here.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Yes, the current BIOS settings and new BIOS settings should be fairly small (10's of attributes), especially if ReadOnly fields are stripped out (compared to the BIOSRegistry which is quite large with descriptions, enum values etc.) per attribute. 

Or we could do something like every time the new BIOS settings are received send it to Ironic on the next provisioning.

design/baremetal-operator/bulk-set-bios-config.md
zaneb: I hadn't really thought about a separate CRD, but I like the idea. We ended up kinda regretting not doing that with HardwareDetails, so it might make sense to get ahead of that.
The relationship is 1:1 so I think it should be linked just by having the same name so that the lookup is O(1) in both directions, rather than having to list and filter all of them to find the one that corresponds to a particular host. That's one thing a CRD makes possible that you can't really do with a ConfigMap, since the user can create their own ConfigMaps with arbitrary names for other reasons.
This could be done by a separate controller, but I'm not sure we want two things talking to ironic. A new CRD could just as easily be created by the existing BMH controller, and this would ensure it got updated at the right points in the workflow.
For the Registry, having it in a CRD also allows us to have a(n OpenAPI) schema for the schema.

On the flip side:
This paragraph is talking about exposing the actual settings, but the name `BIOSRegistry` implies that you're talking about the schema. Are you suggesting to combine them into a single CRD?
We know that the registry (i.e. schema) is very large - on the same order of magnitude as the limit on k8s object size - and we talked about compressing it before storing in a ConfigMap. That wouldn't really be an option for a CRD.

dhellmann: I was suggesting 2 CRDs. One to ask for the current settings, and another to provide a set of values to change for a particular host. The current settings CRD could have a subset of the info from the registry. The change CRD would only need a list of name/value pairs.

design/baremetal-operator/bulk-set-bios-config.md
zaneb: Interesting that they have an OpenAPI specification for their custom non-OpenAPI schema :D
I wonder how hard it would be to slim this down and transform the format itself into OpenAPI, for consistency with how k8s APIs are specified, since afaik it's not actually our goal for someone to be able to build a GUI on top of this.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,217 @@
...
+identified in the BIOS Registry, when building the JSON input. It’s not
+expected that the entire set read from Ironic will be passed down as:
+
+1. many of the values read from Ironic are ReadOnly
+2. setting ReadOnly values to their current settings increases the size of the
+   input and may potentially affect the Redfish request to the BMC
+
+The BIOS Settings are changed during the node cleaning step which requires a
+reboot. The change to BIOS settings will not be a trigger for reprovisioning a
+host, the BIOS settings will only be applied when a new image is going to be
```
zaneb, line 186: TBH I thought this was clearer before you rewrote it. How it will works is:
* if the Host is in the Ready state (i.e. provisioning has not started)
* if the settings in the status differ from the ones in the spec
* then the Host will re-enter the Preparing state
* in the Preparing state the settings get copied from the status to the spec
* in the Preparing state, manual cleaning takes place with the new settings
* after Preparing, the Host returns to Ready
* when provisioning, the node is already configured with the latest setttings

The previous statement captured that at a high level, including the fact that it only happens when the Host is Ready:

> Manual clean occurs in the Preparing state of the BaremetalHost, the
> Host will re-enter this state from Ready/Available state whenever its config
> differs from the last one it stored.

Whereas this one is actually incorrect - the settings are not applied when a new image is going to be written, they're applied before that. They'll be applied even if the host _never_ has an image written.

dhellmann: I understand that the timing in ironic depends on specific states that may not line up with what we have in our API. I think we should limit ourselves to making changes to the host "during provisioning" which I interpret as "taking all of the settings given in the metal3 API and applying them to the host through ironic". So, I don't think we want Ironic to make BIOS settings until we know we're going to be asked to write an image to the host.

Is that possible?

bfournie: The BIOS settings are written to Ironic as a set of clean steps during manual cleaning, e.g. https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L1318

design/baremetal-operator/bulk-set-bios-config.md
bfournie: The current settings will come directly from Ironic via the /v1/nodes/{node_ident}/bios API and will only include name/value pairs (https://docs.openstack.org/ironic/rocky/admin/bios.html) including RO attributes.  We could modify this based on the contents of BIOSRegistry (including filtering to remove RO atrributes) but I thought otherwise we'd keep it as is, copy it to the change CRD, and update the values there based on the allowed values in the BIOSRegistry.  This change CRD would provide the set for values to use for hosts.

design/baremetal-operator/bulk-set-bios-config.md
zaneb: The manual cleaning flow has already been implemented (that's what the Preparing state is for - see metal3-io/baremetal-operator#763) and at least as far as we know works correctly with the Ironic states, so implementation is just a matter of adding new steps to the list we send. The current implementation took 17 months to land, so we're not going to start again from scratch :)

bfournie:
Updated to reflect changes made to Ironic BIOS Registry API (see https://review.opendev.org/c/openstack/ironic-specs/+/774681). Instead of a separate endpoint a subset of the particular registry information is now included with each BIOS Setting in the v1/nodes/{node_ident}/bios?detail=True API. This reduces complexity and size of the BIOS data. Also added text regarding validations of the BIOS settings using the registry before adding to clean steps.

In addition made changes regarding sequencing of clean-steps from the recent discussion regarding the granular approach to BIOS config.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+in 'currentBIOSSettings'. For example, integer types will checked against
+minimum and maximum parameters and enumeration types will be compared against
+allowable values. If the validation fails a message will be logged and the
+BIOS Setting will not be added to clean-steps.
+
+When settings are added to the Ironic clean-steps API Ironic, Ironic will set
+the BIOS Configuration in the BMC as part of the manual cleaning process which
+requires a reboot. Manual clean occurs in the Preparing state of the
+BaremetalHost, the Host will re-enter this state from Ready/Available state
+whenever its config differs from the last one it stored.
```
zaneb, line 163: I guess this means we'll have a second copy of the settings in the status (this one is a copy of the settings that we applied, whereas the other one contains all the settings we read from ironic, and they're not guaranteed to be the same).
It might be theoretically possible to collapse this into one, but not without significant and risky changes to the way we trigger manual cleaning now. We'll have to be careful with naming to minimise the confusion.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+
+The baremetal-operator will detect when 'updateBIOSSettings' changes by
+comparing the 'Name'/'Value' pairs in 'updateBIOSSettings' to the current BIOS
+configuration in 'currentBIOSSettings'. When a change is detected, the BMO will
+add the new values to the Ironic clean-steps API. Before adding it though, the
+BMO will do validation checks on the new values using the Registry information
+in 'currentBIOSSettings'. For example, integer types will checked against
+minimum and maximum parameters and enumeration types will be compared against
+allowable values. If the validation fails a message will be logged and the
+BIOS Setting will not be added to clean-steps.
```
zaneb, line 157: Arda is in the process of adding a validating webhook, maybe that would be the place to verify the data so that the user is prevented from setting invalid values, rather than just dropping them on the floor.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+Since the name/value pair mapping is vendor specific and not intuitive, it
+doesn’t lend itself to creating the input by hand using existing vendor
+documentation, except for potentially a small subset of names. It’s desirable
+to take an existing BIOS config set and modify the current settings to the
+desired settings. The BIOS Registry can be used as a source of documentation
+to modify the settings.
+
+## Design Details
+
+This proposes three new data elements in the BareMetalHost:
```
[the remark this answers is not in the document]
bfournie: Yes, updated and added that info

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+
+- 'currentBIOSSettings' - the current BIOS Settings retrieved from Ironic via
+  the bios API. This dataset includes BIOS Registry information which provides
+  a schema for the settings.
+- 'updateBIOSSettingsTemplate' - a list of writeable BIOS Setting name/value
+  pairs from 'currentBIOSSettings', suitable for setting to a host via Ironic
+- `updateBIOSSettings` - user-defined BIOS Settings to write to host via Ironic.
```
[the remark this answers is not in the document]
bfournie: :-)  Changed the use firmwareConfig as that seems consistent, but open to suggestions

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+  It's assumed that this will be generated from 'updateBIOSSettingsTemplate'
+
+The BIOS Settings are retrieved from the BMC by Ironic and cached
+whenever the node moves to `manageable` or `cleaning`, or when the settings
+are updated. Recent changes to Ironic (see the Ironic proposal
+[BIOS Registry](https://review.opendev.org/c/openstack/ironic-specs/+/774681)
+added an additional query of the BIOS Registry.
```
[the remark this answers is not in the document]
bfournie: Updated the doc to add this info. However I don't think that the data will need to be retrieved at the end of Deprovisioning.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+  - ReadOnly (a boolean indicating whether the attribute is read only)
+  - Immutable (a boolean indicating whether or not the attribute is changeable)
+  - Help text (localized)
+  - Allowable values for enum type attributes
+  - Minimum and maximum values (for numeric attributes)
+  - Minimum and maximum character lengths (for string attributes)
+
+Upon retrieving the BIOS Settigs from Ironic, the baremetal-operator will
+populate the 'currentBIOSSettings` field for each host, represented as a
+ConfigMap with an annotation in BareMetalHostStatus. As there are no fields
```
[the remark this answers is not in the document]
bfournie: Rewrote this section as we discussed to use a separate CRD for the BIOS settings info read from Ironic.  Depending on the vendor this can be up to 100K.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,218 @@
...
+in BIOS configuration that are secret (i.e. no passwords etc.) this new entry
+does not need to be secret-protected. 
+
+From `currentBIOSSettings` the BMO will also generate the
+`updateBIOSSettingsTemplate` which is a list consisting only of `Name`/`Value`
+pairs per Setting. The BMO will use the `ReadOnly` field in the Registry
+information to build this list. Settings with `ReadOnly` set to `True` will be
+ignored to avoid adding these settings via Ironic as they can't be written to
+a BMC.  'updateBIOSettingsTemplate' will be stored in the CRD, referenced in
+the BareMetalHostStatus.
```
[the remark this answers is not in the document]
bfournie: Fixed

design/baremetal-operator/bulk-set-bios-config.md
bfournie: The way I revised it is to have one copy of the data from Ironic and one another of the settings to apply. If reduces complexity to add a 2nd copy of the settings in the status to save in case of write failures we can do that, but yeah the naming will be important.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: OK great, I added a note that we can leverage the validating webhook if possible.

furkatgofurov7:
Hi!
As agreed during the community meeting, let's have a lazy consensus of a week from today and if there are no further objections/comments on the proposal, we can go ahead with merging it next week.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+these in the BareMetalHost CRD. The new CRD will be named
+`HostFirmwareSettings` and will consist of the following:
+
+- `currentFirmwareSettings` - the current BIOS Settings retrieved from Ironic
+  via the bios API will be stored in the Status section. This dataset includes
+  BIOS Registry information which provides a schema for the settings.
+- `writeableFirmwareSettings` - BIOS Settings to write to the host via Ironic
+  will be stored in the Spec section. If it is empty, it will be initially
+  populated with the name/value pairs from `currentFirmwareSettings` that do
+  not have the `read-only` or `unique` flag set.
```
zaneb, line 106: I think we're ready for the bikeshed-on-naming portion of the discussion :grin:
Some thoughts:

* "desired" might be better than "writeable", but...
* it's kind of implied by the Spec/Status split that the Spec is the desired state and the Status is the current state
* the `status.currentFirmwareSettings` contains not only the settings but the whole schema. Calling them both "settings" might be confusing as it  implies they have the same structure.

What would folks think about something like `spec.vendorSettings` and `status.firmwareRegistry`?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+  against the `currentFirmwareSettings` to ensure updated values are valid
+
+In order to interoperate with the individual BIOS Configuration set
+approach the BMO must ensure that the same settings don't get added twice
+to the Ironic clean-steps API. In addition, as each clean step requires a
+reboot the number of clean steps that are added must be managed.
+
+### Risks and Mitigations
+
+N/A
```
zaneb, line 193: I remember at some point we said there was no sensitive data being exposed, but I can't remember if we were proposing to show the read-only setting values at that time. Can you confirm that this is still the case?

zaneb (commented):
I'm very happy with the simplicity of this approach. I did start a little bikeshedding ;)
/approve

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+  state the settings will also be retrieved and used to update
+  `currentFirmwareSettings`.
+
+Upon retrieving the BIOS Settigs from Ironic, the baremetal-operator will
+copy the name/value pairs to the `writeableFirmwareSettings` field in
+`HostFirwareSettingsSpec` if that field is empty. Settings that have the
+`read-only` flag set will not be copied as these settings cannot be written
+to the BMC. Likewise, settings that have the `unique` flag set will also not
+be copied, these are settings like serial numbers or product ids for example,
+that should not be written to other servers.
```
zaneb, line 144: Shouldn't we also copy missing values? e.g. if there are r/w settings A. B, and C available, but only a value for B specified, I think we should copy values for A and C into the Spec. It follows that we need to do this continuously, whenever they get out of sync.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Thanks Zane! Agree that we can take "Firmware" out of the name since its redundant with that name in CRD itself, and also that using "settings" in both places is confusing.

For the spec section I think "vendorSettings" is good.
For the status section, this includes both the name+value and the registry section so I don't think calling it firmwareRegistry is adequate.  Maybe "currentValues"?

design/baremetal-operator/bulk-set-bios-config.md
bfournie: I think we should show the readOnly data as its not sensitive, its normally things like NumCores or CoreSpeed which can't be changed. However some vendors, Dell in particular, make use of the "Password" AttributeType defined in https://redfish.dmtf.org/schemas/v1/AttributeRegistry.v1_3_5.json and include settings like "SysPassword" and "SetupPassword" with this AttributeType. In this case it looks like the values returned are either not included or set to the empty string but we should not include settings of this type in the currentValues and disallow any sets of this AttributeType.

I will add a note to this section

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Yes good point.  I will note that settings that aren't in the spec but are in the status will be copied to the spec. I think this only needs to be done when the status is updated, e,g. after moving the node to manageable and getting the data from Ironic.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+## Design Details
+
+This proposes a new Custom Resource Definition (CRD) to store the settings
+read from Ironic along with the modified settings to write to Ironic. Since
+each vendor provides 100 or more BIOS Settings, its not practical to store
+these in the BareMetalHost CRD. The new CRD will be named
+`HostFirmwareSettings` and will consist of the following:
+
+- `currentFirmwareSettings` - the current BIOS Settings retrieved from Ironic
+  via the bios API will be stored in the Status section. This dataset includes
```
andfasano, line 101: A link to the `bios` API here could be useful

design/baremetal-operator/bulk-set-bios-config.md
andfasano: Given that the name of the CRD is pretty specific and clear `HostFirmwareSettings`, what about reducing even further to `spec.vendorSettings` and `status.schema`?

@bfournie I think that in any case putting an example of the CRD help would be both useful as for documentation and to visualize it better for the ongoing discussion. For example, will the CRD have an `id` field to identify it? Do you plan to have additional fields for host selection?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+are updated. Recent changes to Ironic (see the Ironic proposal
+[BIOS Registry](https://review.opendev.org/c/openstack/ironic-specs/+/774681)
+added an additional query of the BIOS Registry.
+
+The baremetal-operator can get the BIOS Settings from Ironic after the node
+has been moved to manageable via the `v1/nodes/{node_ident}/bios?detail=True`
+API.  This API provides a list of BIOS Settings (aka `Attributes`) in JSON and
+each setting includes the following data:
+
+- Name
```
andfasano, line 119: For the sake of clarity, the `Name` will be vendor-specific

design/baremetal-operator/bulk-set-bios-config.md
andfasano: At any given time, how many CRD instances are expected to be present in the cluster (e.g. one for BMH, one for distinct vendor/server type, etc)? 
Also, considering that BMHs could be deployed in multiple namespaces, what would be the `scope` of the CRD?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+`read-only` flag set will not be copied as these settings cannot be written
+to the BMC. Likewise, settings that have the `unique` flag set will also not
+be copied, these are settings like serial numbers or product ids for example,
+that should not be written to other servers.
+
+A user can update `writeableFirmwareSettings` to the desired values for the
+settings.  Likewise, `writeableFirmwareSettings` can be copied to multiple hosts
+of the same type to effectively clone the settings from a reference host and
+ensuring that all similar systems have the same BIOS configuration,
+i.e. "bulk set".
```
andfasano, line 150: Can you please elaborate a little bit more how the bulk copy process will work? From the description so far I wasn't able to find any templating feature so far, as mentioned at the beginning of the proposal

design/baremetal-operator/bulk-set-bios-config.md
zaneb: This resource is 1:1 with BMHs. They share the same name + namespace.

zaneb: I agree with y'all that the status one is tricky - we need to somehow convey that it contains both the current settings and the schema, ideally without literally typing currentSettingsAndSchema. In ironic 'registry' is taken to mean the schema, although in other contexts (e.g. Windows registry) it's the opposite, so it's confusing (or cunningly ambiguous, depending on your point of view).'values' has the opposite problem where it implies it's just the values.

I'm trying to think of words that describe values along with their schema, and the only thing coming to mind is 'database'. Maybe we should go with something generic like `settingsData` (or s/settings/configuration/ in both).

However, I also wonder if the fact that we are having difficulty coming up with a good name is a hint that we are trying too hard to squeeze two different things into one field? Given that we are in a separate CRD here I don't feel like there is the same premium on adding fields as there is in the BMH. Maybe we could afford to have separate `status.values` and `status.schema` fields? (A downside of this is that it makes it less clear that you should copy the `spec.vendorSettings`, not the `status.values`, when you want to clone settings to another Host.)

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+configuration. This ensures all the machines will have the identical settings,
+reducing issues caused by unexpected configuration differences.
+
+The challenge that BIOS configuration presents is that no two hardware vendors
+use the same names for settings that control the same behavior, and the
+available settings often overlap but have significant differences. There
+currently is an approach that would abstract these differences for individual
+vendors
+[individual set](https://github.com/metal3-io/baremetal-operator/pull/302).
+This new proposal would not require Metal3 to
```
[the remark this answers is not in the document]
bfournie: Yes BareMetalHost is really the correct term.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,228 @@
...
+configuring one host, exporting those settings in a vendor-specific file
+format, and then importing that file on the BMCs of other hosts. Accepting that
+the format for these profiles is vendor-specific eliminates a lot of the
+implementation complexity, at the risk of moving that complexity to the
+end-user. This requirement is similar to something Dell has proposed in the
+Ironic community
+([hw config](https://review.opendev.org/c/openstack/ironic-specs/+/740721))
+to export and import an entire system iDRAC inventory including RAID and
+networking requirements. Compared to that approach, this proposal attempts
+a more lightweight, vendor-neutral approach just for BIOS configuration.
```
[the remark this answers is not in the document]
bfournie: Yes this could be clearer. It really is not vendor-neutral per-se but it uses the vendor specific names and doesn't require a mapping layer that would convert the names to a common set of names. To transfer the settings to other hosts using k8s commands, only servers from the same vendors can be used as the names are specific to that vendor.  I will rewrite this.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Added.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Sure I can an example of the CRD here, especially now that https://github.com/metal3-io/baremetal-operator/pull/901 is posted. I'll get agreement here and update 901 with the results.

And yes the name will be vendor specific.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: There is one CRD instance per host, the same as BMH and in the same namespace. Each CRD instance will hold multiple BIOS settings, from a survey of common vendors this seems to be 100-200 settings.

As for the bulk copy, the plan was to just use the kubectl (oc) edit and apply commands to update the settings on a host from a known good host. I'm not aware if there is a simpler way to do this. I'l add that to the spec

design/baremetal-operator/bulk-set-bios-config.md
bfournie: I had thought about having a separate struct in the status.currentValues for the schema section, I do think it needs to be broken out as this schema doesn't (or shouldn't) change while the values will change when modified. The downside of having status.values and status.schema in separate fields though is we need the name field in both, if have something like this then you don't need to duplicate the name, but that doesn't really solve the naming problem:
currentValues {
string name
string value
struct schema {
string attributeType
etc.
}
}

bfournie: Names of spec and status fields have been updated and the CRD section has been included, let me know if this is reasonable.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Updated spec to add CRD and reduce naming. Also indicated that the name is vendor specifc.

design/baremetal-operator/bulk-set-bios-config.md
andfasano: I don't want to drag away from scope this proposal, but I really see a potential opportunity here, so throwing in the idea just for the sake of the discussion. In order to really push for an (automated) bulk apply of the settings, the CRD could be enhanced by adding new selector fields for:

* The source - the host from where the settings will be copied
* The destination hosts - the targets where the settings will be copied to

(Selector fields could be labels or hints). In this case a dedicated controller would monitor the namespace for new CRD tagged with the _source_ and, in case, it will generate new CRD for the selected targets, by copying the allowed settings (at this point the BMO will operate as described in the proposal, going on with the communication to Ironic).
The controller could also take care of additional checks, for example to generate the CRD only for the select targets that also matches the hw vendors.
In this way we could have a quick way to effectively apply in bulk the settings without a tedious/error prone manual operation - especially for a large number of hosts.

bfournie: Andrea - I think this is a great idea and, as it turns out, a feature some customers have been asking for per discussion with our PMs.

To summarize our offline conversation regarding this:
- this current proposal is really for a minimal BIOS set feature that requires manual configuration for bulk set of the spec fields to other systems of the same type
- we should indicate in this proposal that there is a future feature to automate BIOS setting using a new operator that would take as input a list of systems of the same type and attempt to keep the settings in sync. This feature will require a new doc and potentially some new fields to the CRD although it will not be a breaking change to the existing CRD as it will not add mandatory fields
 
I'll update the doc to reflect this discussion.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,389 @@
...
+- `settingsData` - the current BIOS Settings retrieved from Ironic
+  via the bios API
+  [BIOS API](https://docs.openstack.org/api-ref/baremetal/?expanded=#node-bios-nodes)
+  will be stored in the Status section along with the BIOS Registry
+  information which provides a schema for the settings.
+- `vendorSettings` - BIOS Settings to write to the host via Ironic will be
+  stored in the Spec section. It will be initially populated with the
+  name/value pairs from `settingsData` that do not have the `read-only` or
+  `unique` flag set, existing values will not be overwritten.
+
```
andfasano, line 122: Do you maybe have a small, concrete sample yaml for this CRD to be shown here as an example?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,389 @@
...
+                  - value
+                  type: object
+                type: array
+            type: object
+          status:
+            description: defines the observed state of HostFirmwareSettings
+            properties:
+              settingsData:
+                items:
+                  description: describes fields read from Ironic
```
andfasano, line 268: Minor: _"describes the current host firmware settings along with their schema_" (I think the reference to Ironic could also be safely removed, as it is not relevant in this context)

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,389 @@
...
+                  properties:
+                    schema:
```
andfasano, line 270: On a second thought, and somehow continuing the discussion on a point previously analysed, I'm starting to think if the schema should really go into a separate, dedicated CRD (with the granularity of at least one per vendor/model). I could envision a couple of benefits with such design:

* Smaller `HostFirmwareSettings` footprint. In scenario with thousands of BMH, currently the same schema will be duplicated thousands time, with a waste of space for a lot of data that will be identical.
* Future support for bulk copy/value drifting. Having a single schema resource for vendor/model (with potentially a golden set of value) could make it easier the work for the future controller responsible to apply the bulk copy

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,389 @@
...
+current BIOS configurations settings cached on a server. Give access to the
+BIOS registry which can be used as documentation so a user can easily update
+the BIOS settings to desired values. Validate the BIOS Configuration values
+based on the vendor's BIOS Registry before sending to the BMC.
+
+Because the names are in the vendor format, they can be manually transferred
+in a bulk manner across similar hardware in the cluster using `kubectl`
+commands. Its possible to use scripts to help automate this `bulk set`.
+
+A future feature is planned to make use of this funtionality to
```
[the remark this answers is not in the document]
bfournie: Agreed, done.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,389 @@
...
+### Goals
+
+Allow updating BIOS configuration using the vendor's BIOS names and not
+require a mapping layer to a common set of names. Create a template for the
+current BIOS configurations settings cached on a server. Give access to the
+BIOS registry which can be used as documentation so a user can easily update
+the BIOS settings to desired values. Validate the BIOS Configuration values
+based on the vendor's BIOS Registry before sending to the BMC.
+
+Because the names are in the vendor format, they can be manually transferred
```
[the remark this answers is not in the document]
bfournie: Done.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Agree, done.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: This seems like a reasonable change to reduce the memory consumption as yes, the schema is largely fixed per vendor model. We could create a separate CRD referenced from HostFirmwareSettings to contain the schema. When retrieving the data from Ironic the vendor info in BMH can be used to create a new entry in the schema CRD if there is not a reference in HFS.

I'd like to get Zane's thoughts on this. The most recent update I made to the spec did not include any mention of a new CRD but I can add it on next update if we get agreement.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Agree, I have stripped down the CRD example to include only relevant sections.

design/baremetal-operator/bulk-set-bios-config.md
zaneb: I'm not sure about the mechanics here...

> the CRD could be enhanced by adding new selector fields for:
> 
>     * The source - the host from where the settings will be copied

I don't think this works as a selector - if there's more than one matching host, what do you choose? I could see it working as just a straight link, although it would be inefficient to implement.

>     * The destination hosts - the targets where the settings will be copied to

Why both? Doing it in this direction feels a bit wrong - another Host can opt you in to receiving settings changes. There may even be some RBAC implications, although I'm not sure.

> In this way we could have a quick way to effectively apply in bulk the settings without a tedious/error prone manual operation - especially for a large number of hosts.

That would certainly be a nice improvement for the future.

design/baremetal-operator/bulk-set-bios-config.md
zaneb: How would you do the de-duplication of the schema?

bfournie: I could see it working like this, please correct any misconceptions I have about CRD access:
- We'd add a reference to `settingsData` (in the HFS CRD) to point to a entry in the new schema CRD. 
- The first time the BMO gets BIOS data from Ironic it would use the `manufacturer` and `model` to either access an existing schema entry (and update the reference) or create a new entry and copy the BIOS registry data in API to it, instead of storing this in the HFS CRD as currently proposed (and also update the reference)
- Any future access to Ironic API will not need to update the schema once the schema reference is set, in fact we can even reduce the API bandwidth to Ironic by not use the `?detail=True` flag (which causes Ironic to include BIOS registry info)

zaneb: I don't think we can guarantee that the schemas for a given manufacture+model are all the same, given the potential for different firmware versions and peripherals.
A safe way would be for the name of the Schema resource to contain a hash of a canonical form of the schema (sans actual settings values, obviously). In practice I think that would probably yield good deduplication. Maybe we can do something like `<vendor>-<model>-<first 64? chars of hash>` so the user experience won't be too hideous.

The other challenge is reference counting. Maybe we can also label the HFS CRs with the name of the Schema resource, so we can clean them up when there's nothing left referencing it.

bfournie: Thanks Zane, good suggestions. I've updated the doc to add the hash in that format. I haven't addressed the counting issue but will in a follow-on.

Andrea and Zane - I've added the second CRD to the doc, thanks for all the input!

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,414 @@
...
+            type: object
+```
+
+An example of the proposed CRD for `FirmwareSchema` is as follows:
+
+```yaml
+---
+kind: CustomResourceDefinition
+metadata:
+  name: hostfirmwaresettings.metal3.io
```
andfasano, line 295: I guess it's typo here (and below): hostfirmwareschema?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,414 @@
...
+                items:
+                  description: describes current settings and their schema
+                  properties:
+                    name:
+                      description: Identifier for the Firmware setting.
+                      type: string
+                    value:
+                      description: Value of the Firmware setting.
+                      type: string
+                    schemaReference:
```
andfasano, line 282: Wouldn't be enough to specify the schema reference per host (instead that per item)? Not sure if also would be required to specify the namespace (thinking to the case were the BMH could be spread across different namespaces, but I guess the schema will continue to stay in just one)

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Yes. Fixed the typo

design/baremetal-operator/bulk-set-bios-config.md
zaneb: That's an interesting point about the namespace, actually...

If we don't specify the namespace, it's implied that we're looking in the same namespace as the Host/Settings. In principle, that makes sense to me - it's not really safe to have Hosts in different namespaces referencing the same Schema CR for RBAC reasons.

_However_ in the use case we're thinking about (and trying to save memory in), there are hundreds of Hosts, each in its own namespace. If we always reference the schema in the local namespace, then we'll end up back with 1 per Host and we won't save any memory at all by splitting the CRDs (in fact, it'll be slightly worse).

The "correct" solution to this is [API decomposition](https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit#heading=h.lmpo60bkt5pm) but this is not on the cards any time soon.

The best hack I can think of right now would be to have an optional CLI flag for the controller to use a particular namespace for the Schema instead of the local one. So I think do we need to store a namespace in the CR after all.

andfasano: Or assuming that anyhow BMO is the only actor responsible to create/update (and maybe delete) the Firmware Schema resources, then they could be kept in the current [POD_NAMESPACE](https://github.com/metal3-io/baremetal-operator/blob/399f5ef7ee3831014c1425250bc4fa49641a8709/main.go#L106) configured - this could simplifies a little bit the RBAC settings and wouldn't require any additional namespace field in the Schema CR, as I originally thought

zaneb: Hardware admins need to be able to read it though (we're creating it for their benefit, not so much BMO's), and I don't think we should assume that in general they'll have RBAC rights to the same namespace that the BMO pod is running in.
We also need to be pretty clear about telling them where to look for it.

bfournie: OK, I added a `schemaNamespace` hint to the HFS CRD for this case. I also changed the FirmwareSchema CRD to be scope `Namespaced`.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,421 @@
...
+                  required:
+                  - name
+                  - value
+                  type: object
+                type: array
+            type: object
+          status:
+            description: defines the observed state of HostFirmwareSettings
+            properties:
+              schemaReference:
```
dhellmann, line 277: https://github.com/kubernetes-sigs/cluster-api/issues/2318 has some useful tips on object reference types.

design/baremetal-operator/bulk-set-bios-config.md
andfasano: Ok, it seems fine to keep the namespace reference. Do we already have a definition for the "hardware admin" role somewhere? Or do we need to clarify it better here?

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Thanks Doug. IIUC from that we should define a separate struct for the reference such as in https://github.com/kubernetes/api/blob/release-1.17/admissionregistration/v1/types.go#L533, and not use corev1.ObjectReference.  I'll make a note of that in the doc.

design/baremetal-operator/bulk-set-bios-config.md
zaneb: It might be helpful if we wrote up a doc with a bunch of personas that we expect to interact with metal³ (it would certainly make writing user stories easier) but I don't think we have one yet.

In any moderately sized organisation I'd expect there to be different people responsible for maintaining the hardware fleet vs. deploying apps on k8s. In a multitenant situation with multiple spoke clusters being provisioned from a central hub, those are _definitely_ different people.

andfasano: Even though not strictly related, while looking for some materials on the multi-tenancy I stumbled upon https://github.com/kubernetes-sigs/multi-tenancy/blob/master/docs/profiles/profile-soft-multitenancy-s1.md, which has an interesting personas definition

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,421 @@
...
+`read-only` flag set will not be copied as these settings cannot be written
+to the BMC. Likewise, settings that have the `unique` flag set will also not
+be copied, these are settings like serial numbers or product ids for example,
+that should not be written to other servers.
+
+A user can update `vendorSettings` to the desired values for the
+settings using the limits from the `FirmwareSchema` section as a guide.
+
+The baremetal-operator will detect when `vendorSettings` changes by comparing
+the name/value pairs to `settingsData`. When a change is detected, the BMO
```
andfasano, line 165: Do we need to restrict in which BMO states this will be possible? If so, it would be useful to list them here.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,421 @@
...
+  store the name/value pairs in `settingsData`
+- Determine if a schema already exists for the BIOS registry data received
+  in the Ironic API. Create a new `FirmwareSchema` CR if one doesn't exist
+  and update the reference in `settingsData`
+- Copy name/value pairs from `settingsData` to
+ `vendorSettings` for settings that are not `read-only` or `unique`
+- Check for changes to `vendorSettings` and when detected, validate
+  values and add updated name/value pairs to Ironic clean-steps API
+
+## Alternatives
```
andfasano, line 409: Is there any suggestion for the [`Tests Plan`](https://github.com/metal3-io/metal3-docs/blob/master/design/_template.md#test-plan)?

andfasano (approved):

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Clean steps will be updated at the same time its currently done, when building the ManualCleanSteps https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L935 in actionPreparing:
https://github.com/metal3-io/baremetal-operator/blob/master/controllers/metal3.io/baremetalhost_controller.go#L764.

I'll update this section to make it clear.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,438 @@
...
+
+The baremetal-operator will detect when `vendorSettings` changes by comparing
+the name/value pairs to `settingsData`. When a change is detected, the BMO
+will add the new values to the Ironic clean-steps API in the Preparing state,
+when building the manual clean steps. Before adding it though,
+the BMO will do validation checks on the new values using the schema. For
+example, integer types will checked against minimum and maximum parameters and
+enumeration types will be compared against allowable values. The validation
+webhook can be leveraged here to log a message regarding the failure; the
+invalid BIOS Setting will not be added to clean-steps.
```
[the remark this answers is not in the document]
zaneb: I think this paragraph might be based on a misunderstanding of what a validating webhook does.

The webhook sits in the API pipeline between the input and the data store, so it allow us to reject an invalid update before it is stored in the resource. Ideally we should do this so that the user knows immediately if they submit an invalid change.

Without a webhook we end up with invalid config stored in the object, and then the controller has to decide what to do with it given that it's too late to reject it. Since we'd have to put the Host in an error state anyway, and retry after the user has fixed it, I agree there'd probably not be much point in doing a partial application.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Added a section for `Test Plan`

design/baremetal-operator/bulk-set-bios-config.md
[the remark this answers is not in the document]
bfournie: Updated to state that validity checks will be done when building the manual clean steps and will return a failure for any failures, preventing partial configurations.  

Also indicated that a validation webhook will be done as a follow-on, to catch validation errors before storing the data in a resource.

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,443 @@
...
+  from Ironic via the BIOS API (`v1/nodes/{node_ident}/bios?detail=True`)
+  [BIOS API](https://docs.openstack.org/api-ref/baremetal/?expanded=#node-bios-nodes)
+  will be stored in the Status section along with the BIOS Registry
+  information which provides a schema for the settings.
+- `vendorSettings` - BIOS Settings to write to the host via Ironic will be
+  stored in the Spec section. It will be initially populated with the
+  name/value pairs from `settingsData` that do not have the `read-only` or
+  `unique` schema flag set. Existing values will not be overwritten.
+
+A second CRD `FirmwareSchema` will added to store schema data (aka BIOS
```
lshilin-r, line 123: will be added?

design/baremetal-operator/bulk-set-bios-config.md
```diff
@@ -0,0 +1,443 @@
...
+do validation checks on the new values using the schema. For example, integer
+types will checked against minimum and maximum parameters and enumeration types
+will be compared against allowable values. A failure will be returned from the
+clean steps stage if any errors are detected in order to avoid pushing a
+partial configuration to Ironic.
+
+As a follow-on, a validation webhook for the HostFirmwareSettings CRD can be
+leveraged to add the validations. This webhook would reject the update before
+it is stored in the resource, removing the need to do the validations in
+manual clean steps.
```
andfasano, line 177: As per webhook best practices (see ["Validating a resource in the controller"](https://www.openshift.com/blog/kubernetes-operators-best-practices)), there could be some scenario where the webhook could not be available, so it'd be recommended to apply the same validation rules still in the controller itself (using some shared code)

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Fixed.

design/baremetal-operator/bulk-set-bios-config.md
bfournie: Updated text to indicate that the same checks will be done before adding clean steps.

andfasano:
Thanks @bfournie for the proposal, the current state looks fine to me.
/lgtm
/hold In case @zaneb would want to do a final pass

PR #178, opened 2021-04-21
Design: propose reproducible metal3-dev-env

macaptain:


design/reproducible-metal3-dev-env.md
```diff
@@ -0,0 +1,144 @@
...
+
+* Minikube v1.19 was released and automatically pulled into our environment. It
+  appears to have a flaky issue with instantiating an IP address for the
+  minikube VM in our environment, which we did not see in v1.18.
+* A kubelet bug in Kubernetes v1.20.4 occasionally breaks builds with a
+  multi-node control plane
+* A change to tripleo-repos script was pulled in automatically by the
+  Dockerfile which builds vbmc, and broke all builds and integration tests
+* A network problem between the CI infrastructure and images.rdoproject.org led
+  to build timeouts when the Ironic Python Agent was downloaded
```
hardys, line 51: This last one doesn't seem to be relevant, IIUC the proposal is to pin versions, and implement a way to automate bumping those versions - which doesn't solve the issue that sometimes network outages happen?

macaptain: @hardys, thanks for taking a look at the proposal. Yes, you are right. The network issue that we had drew my attention to the fact that we're downloading from https://images.rdoproject.org/centos8/master/rdo_trunk/current-tripleo/, where fresh tarballs are pushed and not versioned, but the network troubles themselves would not have been ameliorated by any of the suggestions in this proposal. I've removed this line.

(As an aside, for CI builds, I'm now downloading the IPA tarball from a mirror closer to the infrastructure so that we don't run into this problem again. Sweden -> US West was quite a way to go for the IPA! But the IPA is still not versioned.)

design/reproducible-metal3-dev-env.md
```diff
@@ -0,0 +1,151 @@
...
+
+A comprehensive list of software (TODO: along with where it is used):
+* Kubernetes
+* Minikube
+* docker kvm2 machine drived
+* vbmv image: tripleo-repos
+* sushy-tools image: libvirt-dev and libvirt-python (we could also track
+  updates to sushy-tools itself)
+* ironic-python-agent version downloaded with the ipa-downloader image
+* TODO: Ansible? Ubuntu? CentOS? More?
```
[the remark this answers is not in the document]
macaptain: Thanks, I added these.

design/reproducible-metal3-dev-env.md
hardys: Right, we have a CACHEURL variable which can point to a locally mirrored copy https://github.com/metal3-io/ironic-ipa-downloader/blob/master/get-resource.sh#L26 so perhaps doing that in CI would be a good idea, ideally with some automated way to test any new version before updating the cached copy.

fmuyassarov (commented):
This looks good to me. +1 for trying Renovate in Metal3-dev-env.
@hardys @dtantsur WDYT about creating a new repo for storing vmbc and sushy-tools docker files? That would avoid building local container images on every run of integration test in Metal3-dev-env.

namnx228:
LGTM

namnx228 (approved):

macaptain:
Let's lazy consensus this for Wednesday 2nd June. Please do leave more feedback on this PR if you have it, but if there's no objections, on this date, let's merge. Then I'll post to the Metal3 mailing list and ask an owner of the metal3 GitHub organization to add the Renovate GitHub app to the metal3-dev-env repo and we'll go from there.

fmuyassarov:
Let's have a lazy consensus of a week from today and if no objections, we can merge it next week.

fmuyassarov:
@maelk @hardys @dhellmann 
Last meeting we agreed on lazy consensus and as I can see there were no further comments to be addressed on this proposal. As such, can we merge the proposal and move on with its implementation ?

PR #180, opened 2021-04-27
Design: customizable deploy procedure

dtantsur:


design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Customizable deployment procedure
+
+## Status
+
+One of: implementable
```
dhellmann, line 12:
```suggestion
implementable
```

design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+This design proposes customizing the deployment procedure by providing a high
+level abstraction around Ironic [deploy steps][deploy-steps], as well as
+allowing per-host deployment (IPA) images.
+
+[deploy-steps]: https://docs.openstack.org/ironic/latest/admin/node-deployment.html#deploy-steps
+
+## Motivation
+
+By default the Ironic agent downloads a provided image via HTTP(s), converts it
+(if needed) and writes directly to the disk. This procedure works for any
```
dhellmann, line 25:
```suggestion
(if needed) and writes directly to the disk. This procedure works for many
```

design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+As an operator I want to execute site-specific actions during deployment. I'm
+ready to provide them as plugins to the Ironic agent inside my custom agent
+image.
+
+## Design Details
+
+Additions to `BareMetalHost`:
+
+- New field `AgentImageURL`. It will accept an optional link to a custom agent
+  ISO image.
```
dhellmann, line 62: We have an `Image` type [1]. We try to be consistent when we can elsewhere in the API. Would it make sense to reuse that here? So, an `agentImage` field in the API would have all of the sub-fields that `image` has today? Or maybe we would define a new type with a subset of those fields, if we don't need something like `diskFormat` for now? Even if that just gives us `image` with `url` as a sub-field, that would open the way to more flexibility later.

[1] https://github.com/metal3-io/baremetal-operator/blob/master/apis/metal3.io/v1alpha1/baremetalhost_types.go#L396

design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+
+NOTE: custom kernel/initramfs are out of scope for this proposal.
+
+- New field `DeploySteps` with a list of deploy steps to apply. Each step
+  is a structure with fields `Name` (string) and `Args` (JSON).
+- New boolean field `SkipDefaultDeploySteps` defaulting to `false`. If `true`,
+  the execution of the default deployment procedure will be supressed.
+- The `Image.URL` field will become optional if `SkipDefaultDeploySteps` is
+  `true` since we don't know if the deploy method actually requires a separate
+  image.
```
dhellmann, line 72: How will the operator know when it is time to try to provision if the URL is optional? That's the 1 trigger we have today.

design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+
+- New field `AgentImageURL`. It will accept an optional link to a custom agent
+  ISO image.
+
+NOTE: custom kernel/initramfs are out of scope for this proposal.
+
+- New field `DeploySteps` with a list of deploy steps to apply. Each step
+  is a structure with fields `Name` (string) and `Args` (JSON).
+- New boolean field `SkipDefaultDeploySteps` defaulting to `false`. If `true`,
+  the execution of the default deployment procedure will be supressed.
```
dhellmann, line 69: I would be more comfortable if we could express this as a non-boolean enum. We get a lot of pushback from k8s/OpenShift folks when we use bool APIs.

design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+
+Additions to `BareMetalHost`:
+
+- New field `AgentImageURL`. It will accept an optional link to a custom agent
+  ISO image.
+
+NOTE: custom kernel/initramfs are out of scope for this proposal.
+
+- New field `DeploySteps` with a list of deploy steps to apply. Each step
+  is a structure with fields `Name` (string) and `Args` (JSON).
```
dhellmann, line 67: One of the arguments we will need to provide is the network data. Is that merged with the map from the args parameter or do we expect the user to provide it in that field?

design/deploy-steps.md
```diff
@@ -0,0 +1,134 @@
...
+  required, even if the deploy interface does not use it.
+
+### Dependencies
+
+- [custom-agent deploy interface in Ironic][custom-agent]
+
+### Test Plan
+
+We can potentially add CI jobs with CoreOS, although it may require some
+OpenShift bits.
```
dhellmann, line 115: A minimal test could deploy an FCOS image and ensure that it boots, without worrying about whether it can be a node.

design/deploy-steps.md
dtantsur: Makes sense, especially since we may add kernel/initramfs later.

design/deploy-steps.md
dtantsur: By setting deploy steps?

design/deploy-steps.md
dtantsur: If you need to pass explicit network data to `coreos-installer`, it will be one of the arguments, like in my prototype: https://github.com/dtantsur/ironic-agent-image/blob/main/hardware_manager/ironic_coreos_install.py#L24

design/deploy-steps.md
dhellmann: Let's call that change out in the design, then.

dtantsur: Yep, already did - see the latest updates.

design/deploy-steps.md
dhellmann: That means the user needs to provide the data 2 places. I suppose it may need to be in 2 formats? The existing field is a reference to a Secret resource, but posting it inline in the step would mean (a) adding potentially large data to the host resource and (b) it couldn't be protected.

What if instead of a JSON blob field, we ask for a Secret reference, and pass the keys and values from that Secret through? That preserves the security, lets us use a type that accepts arbitrary data in key/value pairs, and keeps the host resource itself small.

dhellmann: It would also be useful to look at what the template controller generates. Maybe we want to use a reference to one of those resources instead, to tie in with the templating capabilities.

dtantsur: > That means the user needs to provide the data 2 places.

I'm not sure I'm following. With this approach there is only one place in BMO: this one. BMH.NetworkData will be unused.

> What if instead of a JSON blob field, we ask for a Secret reference

What you're suggestion requires explicit knowledge of CoreOS. We've agreed not to do that, thus we cannot offer special handling of any arguments.

Or have I misunderstood the proposal?

dhellmann: > > That means the user needs to provide the data 2 places.
> 
> I'm not sure I'm following. With this approach there is only one place in BMO: this one. BMH.NetworkData will be unused.

I guess the user will have to know not to use the networkData field when their custom steps need that information passed directly.

It feels like it would be cleaner if we defined an API for the step arguments that we will pass to the backend. For example, if we said we were going to assemble a standard set of data to pass to the steps, including user data, network data, etc., and that the step-specific arguments would be added to that payload then all steps would receive the same info and could ignore what they don't need/want.

> > What if instead of a JSON blob field, we ask for a Secret reference
> 
> What you're suggestion requires explicit knowledge of CoreOS. We've agreed not to do that, thus we cannot offer special handling of any arguments.
> 
> Or have I misunderstood the proposal?

I expect the user to need to understand how the custom IPA image they're using works. What I don't want is to have the BMO need to understand it.

I'm suggesting that we follow the pattern established elsewhere in the API for handling "a large blob of data we can't validate" and put it into a Secret, which gives us a key/value pair data type that has no expectations for validation and that kubernetes users are used to manipulating. We could also use a ConfigMap, but I think we're using Secret everywhere else and consistency is nice.

If we combine the 2 ideas in this thread, then we could say that all steps will receive a payload with a `networkData` key containing the content of the `networkData` secret, a `userData` key containing the content of the `userData` secret, and a `stepData` key containing the step-specific data from that secret. We could also include a `hostData` key with some additional info, like the host resource name or other values that might be useful during provisioning.

dhellmann: > > could ignore what they don't need/want
> 
> Ironic steps do have validation, you cannot pass arbitrary data to them.

What defines the validation?

> > I'm suggesting that we follow the pattern established elsewhere in the API for handling "a large blob of data we can't validate" and put it into a Secret, which gives us a key/value pair data type that has no expectations for validation and that kubernetes users are used to manipulating.
> 
> Ah, so pass all arguments as one secret?

Yes, one per step.

> > all steps will receive a payload with a networkData key containing the content of the networkData secret, a userData key containing the content of the userData secret, and a stepData key containing the step-specific data from that secret
> 
> Will work nicely for us, but see above. This will limit this proposal to our use case.

I'm not sure why you see it as limited. It would use a bunch of details from other fields on the host as well as the step-specific arguments to construct a well-defined payload. The user would have complete control over what goes into the payload, but wouldn't have to construct the full payload themselves. Step authors would know what to expect to receive as arguments. Where is the missing flexibility?

dtantsur: > What defines the validation?

The step itself. Like here https://github.com/dtantsur/ironic-agent-image/blob/main/hardware_manager/ironic_coreos_install.py#L23-L48

> I'm not sure why you see it as limited.

We won't be able to execute any steps that are not written specifically for metal3. E.g. ironic ships a step to inject files into the instance, we won't able to use it, it accepts a different set of arguments.

> The user would have complete control over what goes into the payload, but wouldn't have to construct the full payload themselves. 

But there will be network_data/user_data/step_data arguments, no? Or have I misunderstood it again?

dhellmann: > > I'm not sure why you see it as limited.
> 
> We won't be able to execute any steps that are not written specifically for metal3. E.g. ironic ships a step to inject files into the instance, we won't able to use it, it accepts a different set of arguments.

The step-wise validation was the piece I was missing to understand this. Thanks for explaining that. I guess users are going to have to understand which steps are available and what arguments those steps take.

Aside from the arguments passed to the steps, how much do we think the step list will vary between hosts? Would it be useful to define an API with a set of steps that the BMO could use as a template to build the instructions to give to Ironic?

dtantsur: > how much do we think the step list will vary between hosts?

Steps come from two sources: the driver and the agent. The former can vary quite a bit, the latter can in theory, but very rarely in practice.

> Would it be useful to define an API with a set of steps

You mean, in Ironic? Would love to, but there is a huge chicken-and-egg problem with booting the agent. We don't know which steps are in the agent until we boot it. 

Now, in our situation it's fine because we boot the agent early for inspection/cleaning (unless these are disabled). But in a general Ironic context it's normal when the agent is only boot for deployment itself, too late to return anything to a user. For this reason we haven't implemented any "get clean/deploy steps" API.

dhellmann: > > Would it be useful to define an API with a set of steps
> 
> You mean, in Ironic? Would love to, but there is a huge chicken-and-egg problem with booting the agent. We don't know which steps are in the agent until we boot it.
> 
> Now, in our situation it's fine because we boot the agent early for inspection/cleaning (unless these are disabled). But in a general Ironic context it's normal when the agent is only boot for deployment itself, too late to return anything to a user. For this reason we haven't implemented any "get clean/deploy steps" API.

Well, I meant here in metal3. So that instead of listing the steps to provision each host individually on that host, we would have a separate API resource that lists the steps and the host would have a reference to that resource. That does make passing per-host settings a bit more complex, though, so maybe that's not ideal.

dtantsur: > separate API resource that lists the steps

Where will this information come from given that we cannot ask Ironic for that?

design/deploy-steps.md
```diff
@@ -0,0 +1,137 @@
...
+- New structure field `CustomDeploy` with two field `DefaultSteps` and `Steps`.
+- `Steps` is a list of deploy steps to apply. Each step is a structure with
+  fields `Name` (string) and `Args` (a reference to a secret).
+- `DefaultSteps` is an enumeration defaulting to `Enabled`. If `Disabled`,
+  the execution of the default deployment procedure will be supressed.
+- The `Image.URL` field will become optional if `DefaultSteps` is `Disabled`
+  since we don't know if the deploy method actually requires a separate image.
+
+Setting `CustomDeploy.Steps` to a non-empty value and `DefaultSteps` to
+`Disabled` will trigger deployment even if `Image.URL` is empty.
```
zaneb, line 75: This seems very... general, and it worries me because it is exposing a _lot_ of ironic internals. I was hoping for something much simpler along the lines of setting something like:

```
  image:
      url: live-iso:<provision_step_name>[:<optional_suffix_that_is_ignored>]
```

and that the existing user/meta/networkData secrets would provide sufficient additional data. Effectively we dictate an API to the author of the custom provision step in IPA, rather than the other way around. In principle that seems like that should be enough for CoreOS, which just needs an ignition file AFAIK?

Perhaps I'm being naïve about how little customisation we can get away with here. We had another proposal (#181) already for a similarly-general interface, although without any use cases provided so it's difficult to evaluate. I'm willing to be educated on this topic, but my strong preference is to expose this functionality in a much more circumscribed way.

zaneb (commented):
While #177 felt a little too specific, this feels too general. I'm confident this means we're approaching the Goldilocks Zone.

design/deploy-steps.md
```diff
@@ -0,0 +1,137 @@
...
+As an operator I want to execute site-specific actions during deployment. I'm
+ready to provide them as plugins to the Ironic agent inside my custom agent
+image.
+
+## Design Details
+
+Additions to `BareMetalHost`:
+
+- New structure field `AgentImage` with the only field `URL`. It will accept
+  an optional link to a custom agent ISO image.
```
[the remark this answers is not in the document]
dtantsur: This API is needed if you generate the ISO per host, which is what we're planning on for network configuration.

> If the user modifies this field, do you anticipate the host being rebooted to use the new image if it hasn't been provisioned yet? Reprovisioned if it was currently provisioned?

What do we do if a user modifies `Image.URL` in-flight? I guess we do the same here. For the first iteration we can ignore the case of the Agent URL updated in the middle of, say, inspection (leave it as a TODO for better times).

design/deploy-steps.md
dtantsur: Well, a non-standard URL schema is definitely not simpler for any definition of "simple" I'm aware of :) I'd prefer not to mix the notion of custom deploy and the image. A custom deploy *may* have an image (actually, even coreos-installer supports using a different image). If we want this to be *actually* usable for anything except for CoreOS, we need to allow at least some degree of flexibility.

I guess you main concern is stemming from providing arbitrary arguments? What I can do is to create one simple field `CustomDeploy` that will unconditionally override the deploy process. This makes this proposal twice less generic (and thus twice less usable FWIW).

design/deploy-steps.md
```diff
@@ -0,0 +1,153 @@
...
+
+As a CoreOS user I want to use `coreos-installer` to write the CoreOS image
+without losing any functionality that Ironic offers (unlike the live ISO
+workflow).
+
+## Design Details
+
+Additions to `BareMetalHost`:
+
+- New structure field `AgentImage` with the only field `URL`. It will accept
```
kirankt, line 52: Instead of a new field, can we instead use the existing `image` alongside `image.format` to mimic this field's behavior? Perhaps we can even add a sub-field under `image` to reflect that this is an custom iso?

dtantsur: I'm very explicitly avoiding to mix the Image with the Agent image. They're not the same thing in a normal situation, they don't even have to be based on the same technology (as it's the case now).

design/deploy-steps.md
```diff
@@ -0,0 +1,158 @@
...
+allowing per-host deployment (IPA) images.
+
+[deploy-steps]: https://docs.openstack.org/ironic/latest/admin/node-deployment.html#deploy-steps
+
+## Motivation
+
+By default the Ironic agent downloads a provided image via HTTP(s), converts it
+(if needed) and writes directly to the disk. This procedure works for many
+distributions, but may not suit some cases well. A prominent example is CoreOS,
+which has its own installer (`coreos-installer`).
```
hardys, line 27: We could link to https://github.com/coreos/coreos-installer here

design/deploy-steps.md
```diff
@@ -0,0 +1,158 @@
...
+- The `Image.URL` field will become optional if `CustomDeploy` is set
+  since we don't know if the deploy method actually requires a separate image.
+
+Setting `CustomDeploy` to a non-empty value will trigger deployment even if
+`Image.URL` is empty.
+
+### Implementation Details: AgentImage
+
+We will accept a custom agent ISO image to be used for each host in the new
+`AgentImage.URL` field (custom kernel/initramfs are out of scope for this
```
hardys, line 65: If we're decoupling AgentImage and CustomDeploy APIs, do we expect AgentImage to also enable passing custom IPA images in the non-CustomDeploy case?

If not then I wonder if it should be one new spec field with two subfields, e.g something like:

```
  customDeploy:
    image: http://foo.iso
    method: custom-agent
```

dhellmann: I think we do want to support custom IPA images for standard deployment to enable IPv4 static IPs and virtual media.

We could still put the agentImage field under the customDeploy section of the API.

kirankt: I like this option. IMO this makes a better user experience and explicit that the image being used as part of the custom deploy steps.

design/deploy-steps.md
```diff
@@ -0,0 +1,158 @@
...
+interface][custom-agent] so that the default deploy steps do not apply. Then we
+request a deploy step with the name provided in the `CustomDeploy.Method`
+field. This step has to be provided in the agent image and is responsible for
+the whole deployment process.
+
+NOTE: `CustomDeploy` is a single-field structure to allow potential extensions
+in the future.
+
+Any required customization is expected to be provided via `UserData` and
+`NetworkData` or be built into the image itself.
```
dhellmann, line 88: We should specify here how we are going to pass that data (and any other arguments) to the deploy step so that someone writing a deploy step knows what to expect.

design/deploy-steps.md
dtantsur: > do we expect AgentImage to also enable passing custom IPA images in the non-CustomDeploy case

Why not? E.g. per-host network customization is not limited to CoreOS and may apply to a regular deployment the same way.

design/deploy-steps.md
dtantsur: Ok

design/deploy-steps.md
dtantsur: > explicit that the image being used as part of the custom deploy steps

Exactly this impression I'm trying to avoid since it's not correct. These features are related, but are not mutually dependent.

design/deploy-steps.md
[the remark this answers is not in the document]
zaneb: > This API is needed if you generate the ISO per host, which is what we're planning on for network configuration.

It's one of several options (currently my third preference).

> > If the user modifies this field, do you anticipate the host being rebooted to use the new image if it hasn't been provisioned yet? Reprovisioned if it was currently provisioned?
> 
> What do we do if a user modifies `Image.URL` in-flight? I guess we do the same here.

We deprovision and reprovision.

design/deploy-steps.md
zaneb: > Well, a non-standard URL schema is definitely not simpler for any definition of "simple" I'm aware of :)

It's matching a fixed string prefix in an existing API field, difficult to see how it could be much simpler tbh.

> I'd prefer not to mix the notion of custom deploy and the image. A custom deploy _may_ have an image (actually, even coreos-installer supports using a different image). If we want this to be _actually_ usable for anything except for CoreOS, we need to allow at least some degree of flexibility.

What advantage is there to using coreos-installer to install a _different_ image to the one in the ISO, as opposed to just installing the image we want through the regular clean steps?

> I guess you main concern is stemming from providing arbitrary arguments? What I can do is to create one simple field `CustomDeploy` that will unconditionally override the deploy process.

This is definitely an improvement IMHO, thanks.

> This makes this proposal twice less generic (and thus twice less usable FWIW).

Making the most generic solution is not a goal from my perspective, and in fact exposing all of the potential functionality of ironic and all custom deploy steps that could be written is an anti-goal. To my mind the goal here should be to support provisioning methods that don't require a separate provisioning image, without hardcoding the name of any particular technology (e.g. CoreOS) into the API.

design/deploy-steps.md
```diff
@@ -0,0 +1,159 @@
...
+  an optional link to a custom agent ISO image.
+- New structure field `CustomDeploy` with the only field `Method` to engage
+  a non-default deployment procedure as described below.
+- The `Image.URL` field will become optional if `CustomDeploy` is set
+  since we don't know if the deploy method actually requires a separate image.
+
+Setting `CustomDeploy` to a non-empty value will trigger deployment even if
+`Image.URL` is empty.
+
+### Implementation Details: AgentImage
```
zaneb, line 62: This is a big topic and it's _almost_ orthogonal to this one - having the ability to specify the deploy image will make testing the custom deploy procedure easier in the early stages of development, but other than that it's intended to solve a future requirement for networking customisation that is unrelated. I opened a separate review (#183) so we can discuss it in more depth.

design/deploy-steps.md
dtantsur: > It's matching a fixed string prefix in an existing API field, difficult to see how it could be much simpler tbh.

Simpler: a field per feature.

> What advantage is there to using coreos-installer to install a different image to the one in the ISO, as opposed to just installing the image we want through the regular clean steps?

I'm not sure, but they do it for some reason. Non-CoreOS cases may need this as well.

design/deploy-steps.md
dtantsur: Okay, let's move this part there.

design/deploy-steps.md
```diff
@@ -0,0 +1,142 @@
...
+  a non-default deployment procedure as described below.
+- The `Image.URL` field will become optional if `CustomDeploy` is set
+  since we don't know if the deploy method actually requires a separate image.
+
+Setting `CustomDeploy` to a non-empty value will trigger deployment even if
+`Image.URL` is empty.
+
+### Implementation Details: CustomDeploy
+
+When `CustomDeploy` is set, we will use the new [custom-agent deploy
```
andfasano, line 63: For the sake of clarity, I'd suggest to replace `we` with the actual actor (ie `BMO`) and the state affected

design/deploy-steps.md
```diff
@@ -0,0 +1,142 @@
...
+field. This step has to be provided in the agent image and is responsible for
+the whole deployment process.
+
+NOTE: `CustomDeploy` is a single-field structure to allow potential extensions
+in the future. The `UserData` and `NetworkData` will be available to a deploy
+step implementation as part of the Node's `instance_info["configdrive"]`.
+
+Any required customization is expected to be provided via `UserData` and
+`NetworkData` or be built into the image itself.
+
```
andfasano, line 75: nit: it seems it's missing a description introducing the below steps?

design/deploy-steps.md
```diff
@@ -0,0 +1,142 @@
...
+1. A user sets `CustomDeploy.Method = "step_name"` and optionally `Image.URL`,
+   `UserData` and `NetworkData`.
+2. BMO sets `Node.DeployInterface = "custom-deploy"` and requests deployment
+   with a custom deploy step `step_name`.
+3. Ironic boots the agent (if not booted for inspection/cleaning already).
+4. Ironic requests the agent to execute the deploy step `step_name` (it is
+   an error to provide a non-existing step), providing it the complete Node
+   object, including `configdrive`.
+5. Ironic waits for the step to finish.
```
andfasano, line 84: What is the expected flow in case of failure? Is there any special way to facilitate the troubleshooting? Also, being a completely custom step I don't think it's possible to perform a preliminary kind of validation for an earlier detection of issues, so I'd expect that most of the problems will be discovered until a later moment (ie a malformed `UserData`)

design/deploy-steps.md
```diff
@@ -0,0 +1,142 @@
...
+the whole deployment process.
+
+NOTE: `CustomDeploy` is a single-field structure to allow potential extensions
+in the future. The `UserData` and `NetworkData` will be available to a deploy
+step implementation as part of the Node's `instance_info["configdrive"]`.
+
+Any required customization is expected to be provided via `UserData` and
+`NetworkData` or be built into the image itself.
+
+1. A user sets `CustomDeploy.Method = "step_name"` and optionally `Image.URL`,
```
andfasano, line 76: Is there a way to prevalidate the availability of `step_name`?

design/deploy-steps.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Implementation Details: CustomDeploy
+
+When `CustomDeploy` is set, we will use the new [custom-agent deploy
+interface][custom-agent] so that the default deploy steps do not apply. Then we
+request a deploy step with the name provided in the `CustomDeploy.Method`
+field. This step has to be provided in the agent image and is responsible for
+the whole deployment process.
+
+NOTE: `CustomDeploy` is a single-field structure to allow potential extensions
```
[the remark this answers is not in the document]
dtantsur: Yeah, what Zane said. Having one very generic `Method` may also allow us to accommodate things like `ansible` deploy.

design/deploy-steps.md
dtantsur: Malformed `UserData` is always discovered very late.

The mode of failure will be pretty much the same as now: the node gets into `deploy failed`. The only new issue is providing a non-existing deploy step, which will also end up in `deploy failed`.

design/deploy-steps.md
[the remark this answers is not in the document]
stbenjam: Is there a reason to not make it a `[]string` -- sort of in the middle between the two approaches? I understand supporting arguments complicates things, but limiting ourselves to just one method makes it harder to change the API later.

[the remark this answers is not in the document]
dtantsur: It is normal, yes. But without arguments we cannot reuse upstream steps anyway. If we write our own steps, it does not matter if there is one or many.

Having a list of steps exposes substantially more about the underlying implementation, I guess that's Zane's concern.

design/deploy-steps.md
```diff
@@ -0,0 +1,149 @@
...
+
+Additions to `BareMetalHost`:
+
+- New structure field `CustomDeploy` with the only field `Method` to engage
+  a non-default deployment procedure as described below.
+- The `Image.URL` field will become optional if `CustomDeploy` is set
+  since we don't know if the deploy method actually requires a separate image.
+
+Setting `CustomDeploy` to a non-empty value will trigger deployment even if
+`Image.URL` is empty.
```
dhellmann, line 59: Today, if the ironic database is lost we re-register the host and adopt it if we think it has been provisioned. We use the image URL to keep track of whether we have provisioned the host. If that's going to become optional, we will need to add similar status fields for the `CustomDeploy` values that also trigger provisioning.

design/deploy-steps.md
```diff
@@ -0,0 +1,149 @@
...
+and is responsible for the whole deployment process.
+
+The `Image.URL` will no longer be required since some images (such as CoreOS)
+already include the root filesystem.
+
+NOTE: `CustomDeploy` is a single-field structure to allow potential extensions
+in the future. The `UserData` and `NetworkData` will be available to a deploy
+step implementation as part of the Node's `instance_info["configdrive"]`.
+Any required customization is expected to be provided via `UserData` and
+`NetworkData` or be built into the image itself.
```
dhellmann, line 76: Since we're customizing the interface anyway, would it make sense for us to pass any metal3-specific information? Do we have any data in the host that isn't available to an ironic deploy step through ironic's API?

dhellmann:
I'm happy with the general direction this is going. After the details are updated to include the status fields, I think it's ready for final review and approval.

design/deploy-steps.md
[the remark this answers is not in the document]
andfasano: I'd be more inclined to keep the feature limited to one (deploy) step only, to avoid any over-complication and limiting its misuse, since the pre-requisites are already non-trivial for just a single step (the user must known in advance the correct step name, how to pass/format any required data via network/meta/userData, and I suppose the step itself must be available somehow within Ironic)

design/deploy-steps.md
```diff
@@ -0,0 +1,149 @@
...
+
+The expected workflow is as follows:
+
+1. A user sets `CustomDeploy.Method = "step_name"` and optionally `Image.URL`,
+   `UserData` and `NetworkData`.
+2. BMO sets `Node.DeployInterface = "custom-deploy"` and requests deployment
+   with a custom deploy step `step_name`.
```
andfasano, line 83: In the `Provisioning` state I guess

design/deploy-steps.md
```diff
@@ -0,0 +1,149 @@
...
+4. Ironic requests the agent to execute the deploy step `step_name` (it is
+   an error to provide a non-existing step), providing it the complete Node
+   object, including `configdrive`.
+5. Ironic waits for the step to finish.
+6. Ironic configures the Node's boot device and clean ups the deployment
+   environment.
+7. Ironic reboots the node into the instance.
+
+On failure (including unsupported deploy step) the node will be moved to the
+`deploy failed` state, and the `LastError` field will be populated.
```
andfasano, line 94: That's the Ironic behavior. From a BMO point of view, the BareMetalHost will remain in the `Provisioning` state, with the error fields set

design/deploy-steps.md
dtantsur: Yeah, same place where the deployment logic (incl. live ISO) lives now.

design/deploy-steps.md
dtantsur: I don't think we have anything that ironic does not have.

design/deploy-steps.md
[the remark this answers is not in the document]
stbenjam: Ok you've convinced me

design/deploy-steps.md
andfasano: Right, I think it's useful in general in such kind of design documents to hilight  the expected BMO/Ironic behavior, considering that they have two different state machines.

andfasano (approved):

andfasano:
As discussed in the latest meeting, let's have a week of lazy consensus for the current proposal.

design/deploy-steps.md
```diff
@@ -0,0 +1,150 @@
...
+
+As a CoreOS user I want to use `coreos-installer` to write the CoreOS image
+without losing any functionality that Ironic offers (unlike the live ISO
+workflow).
+
+## Design Details
+
+Additions to `BareMetalHost`:
+
+- New structure field `CustomDeploy` with the only field `Method` to engage
```
dhellmann, line 53: bike-shed

```suggestion
- New structure field `Deploy` with the only field `Method` to engage
```

design/deploy-steps.md
dhellmann: If we decide later that we want to do something like passing the host CR metadata in, so that a user could use an annotation to pass custom data to their custom image and custom step, would that break existing users?

dhellmann (commented):
I'm generally happy with this approach. I have one future-proofing concern and one naming bike-shed, but I don't think we need to hold this up over either question.

/approve

design/deploy-steps.md
dtantsur: I like that the `Custom` prefix highlights that this is an advanced feature, rather than something that people normally set.

design/deploy-steps.md
dtantsur: If we start passing something unconditionally to a deploy step, it may break deploy steps that were written without it in mind. If we keep in conditional or use `MetaData` to pass it (it's pretty much free-form), we're fine.

fmuyassarov:
As agreed during last week's community meeting, we should be okay to merge this proposal if there no more objections/comments to be addressed.

elfosardo (approved):

design/deploy-steps.md
```diff
@@ -0,0 +1,150 @@
...
+## Drawbacks
+
+- A potential footgun.
+
+## Alternatives
+
+- [Explicit CoreOS support](https://github.com/metal3-io/metal3-docs/pull/177)
+- Expose deploy steps more fully. Would allow greater customization at the
+  expense of a more complex API.
+- Do nothing, keep using live ISO.
```
hardys, line 148: We could add more detail about why using the live-iso approach isn't sufficient, e.g call out the metal3/Ironic features lost by choosing that approach.

hardys:
One comment about additional detail in the alternative section but that could be added as a followup if needed, otherwise lgtm, thanks!

/lgtm

PR #181, opened 2021-04-29
Propose BareMetalHost.Spec.ProvisionerPassthrough

redixin:


design/baremetal-operator/provisioner-passthrough.md
```diff
@@ -0,0 +1,64 @@
...
+provisioner:
+
+```yaml
+  spec:
+    privisionerPassthrough:
+      ironic:
+        driver: ansible
+        ansible_extra:
+          foo: bar
+```
```
hardys, line 45: I personally don't think we should add this kind of API to the BMH - it was deliberately designed to abstract details of the provisioner, so that Ironic (or potentially some other provisioner backend) is an implementation detail.

If we require such customization, IMO it would make more sense to enable it only globally (e.g via a BMO environment variable), or via the ironic config (variable given to ironic-image, if the desired config can be achieved via ironic.conf)

/cc @dhellmann @zaneb @maelk who were involved in the initial BMH/provisioner API design to comment further

dhellmann: I generally agree. We want to be careful how much we ask API consumers to understand about the implementation details of the provisioning system. Over on https://github.com/metal3-io/baremetal-operator/pull/867 I linked to #180 which does start to expose some customization of this sort in a more limited way. It would be useful to consider the two approaches together.

/cc @dtantsur

dtantsur: I think https://github.com/metal3-io/metal3-docs/pull/180 may solve this problem in a more structured way.

design/baremetal-operator/provisioner-passthrough.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Summary
+
+Add a new interface so that it is possible to configure any provisioner specific options.
+
+## Motivation
+
+In some circumstances it is desirable to use some provisioner specific features that
+can't be exposed to generic operator interface, e.g. ironic ansible deploy driver,
+ironic vendor passthrough options etc.
```
dtantsur, line 22: This seems very generic. I'd like to hear about specific use cases.

design/baremetal-operator/provisioner-passthrough.md
```diff
@@ -0,0 +1,64 @@
...
+      ironic:
+        driver: ansible
+        ansible_extra:
+          foo: bar
+```
+
+### Work Items
+
+- Add generic passthrough parameter to BareMetalHost.Spec.
+- Implement any provisioner specific options as needed.
```
dtantsur, line 50: I'd rather have this proposal define its complete scope rather than relying on "as needed". Otherwise we cannot even determine if this is the best way to address the use cases.

dtantsur (changes requested):

design/baremetal-operator/provisioner-passthrough.md
redixin: @hardys 

> enable it only globally

Do we always need all the same parameters for all BareMetalHost objects?

redixin: > I think #180 may solve this problem in a more structured way.

Pretty interesting. I need to dig into `DeploySteps` maybe.

zaneb: Yeah, I too am very opposed to exposing all sorts of ironic implementation details in the BMH API.

hardys: > Do we always need all the same parameters for all BareMetalHost objects?

Without more details on your use-case I can't really say :)

My point is that it *might* be OK to expose this kind of customization to the owner of the BMO deployment, IMO it's probably not OK to expose the user of the BMH API to this kind of implementation detail.

redixin: > Yeah, I too am very opposed to exposing all sorts of ironic implementation details in the BMH API.

So what do you think about exposing deploy steps in #180 ?

zaneb: > > Yeah, I too am very opposed to exposing all sorts of ironic implementation details in the BMH API.
> 
> So what do you think about exposing deploy steps in #180 ?

I also think that's too much TBH.

dtantsur: #180 has been made much more generic, and you may be able to reuse its interface for your goals. E.g. you allow `CustomDeploy.Method = "ansible"` to designate the ansible deploy. Then you can use metaData for passing additional information.

PR #183, opened 2021-05-05
Propose deploy ramdisk Image build API and integration

zaneb:
Enable a custom image containing the ironic-python-agent to be built for
each BareMetalHost, instead of requiring all hosts to use the same agent
image. The interface to the image builder will be defined by a new
RamDiskImageRequest CR in metal³, with users able to substitute their
own implementations of image building (in most cases this is expected to
involve light customisation of an existing image) by creating a
controller. The baremetal-operator will create and use
RamDiskImageRequests as required when configured to do so.

This allows users to customise the ramdisk for each Host to contain
network settings specific to that host.

### DIFF
--- a/design/baremetal-operator/image-builder-integration.md
+++ b/design/baremetal-operator/image-builder-integration.md
@@ -1,0 +1,284 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Custom agent image controller
+
+## Status
+
+implementable
+
+## Summary
+
+Enable a custom image containing the ironic-python-agent to be built for each
+BareMetalHost, instead of requiring all hosts to use the same agent image. The
+interface to the image builder will be defined by a new PreprovisioningImage CR
+in metal³, with users able to substitute their own implementations of image
+building (in most cases this is expected to involve light customisation of an
+existing image) by creating a controller. The baremetal-operator will create
+and use PreprovisioningImages as required when configured to do so.
+
+## Motivation
+
+When running the ironic-python-agent (IPA) deploy ramdisk, the agent is reliant
+on information baked in to the ramdisk image or available to discover from the
+immediate environment in order to be able to contact Ironic. In most cases this
+involves DHCP supplying an IP address and default route.
+
+Some users would prefer not to run DHCP in their infrastructure, but rather use
+static IP addresses instead. In some cases, particularly when provisioning
+using virtualmedia and not using PXE, they might prefer that the network on
+which to contact Ironic is not an untagged VLAN. In many cases, servers will be
+connected to switches that use static link aggregators, and thus should set up
+NIC bonding to enable reliable two-way communication.
+
+Currently there is no way to use static IPs or configure any of the other
+options on a per-host basis because all hosts are booted using the same ramdisk
+image.
+
+### Goals
+
+- Allow each host to use its own deploy ramdisk image customised with an
+  appropriate network configuration.
+
+### Non-Goals
+
+- Allow users who expect to provision a host to specify their own custom image
+  containing IPA.
+- Use Ironic with images that don't contain IPA.
+- Implement custom provisioners not based on Ironic.
+
+## Proposal
+
+### User Stories
+
+As a data center operator, I would like to manage hosts with metal³ and deploy
+them via virtualmedia on networks where there is no DHCP-like configuration
+service.
+
+## Design Details
+
+A new Custom Resource type, `PreprovisioningImage` will be created in the
+metal3.io group. The Spec will contain the following (optional) fields:
+
+- Name of networkData Secret
+- Architecture
+
+The Status will contain the following fields:
+
+- Image URL
+- Image format (ISO or initrd)
+- Image checksum
+- SecretReference + Version of networkData
+- Architecture
+- Error message
+- Ready condition
+- Error condition
+
+A default controller for this resource type will be implemented that simply
+sets a fixed URL (supplied in its own environment or command line) as the
+output. However, provision will be made for interested parties to implement
+alternate controllers to allow image customisation by making it easy to disable
+this default controller.
+
+A command-line switch (defaulting to off) will enable integration between the
+baremetal-operator and the image request controller.
+
+A new field, `preprovisioningNetworkData` will be added to the `BareMetalHost`
+Spec for the purposes of image building. If the regular `networkData` field is
+not specified, the `preprovisioningNetworkData` will be used in its place for
+the provisioned image also.
+
+### Implementation Details/Notes/Constraints
+
+When integration is enabled and the driver is virtualmedia based (i.e. supports
+an ISO ramdisk format), the controller portion of the baremetal-operator will
+update the `PreprovisioningImage` with the same name and namespace as the
+`BareMetalHost`, creating it if it doesn't exist, to point to a networkData
+Secret.
+
+The networkData Secret to be used will be obtained from the new
+`preprovisioningNetworkData` field in the `BareMetalHost` Spec.
+
+The Architecture will be obtained from the Host's hardware profile. Currently
+we don't ratify the choice of hardware profile until after inspection, but we
+could move this state earlier. In any event, we don't support multiple
+architectures today, so we could use a default value in the interim.
+
+In some provisioning states, the ironic provisioner will refuse to report
+registration complete in `ValidateMangementAccess()` until the Ready condition
+is true in the corresponding `PreprovisioningImage` Status (indicating that the
+image is available at the linked URL), and the Secret reference + Architecture
+in the Status matches what is expected. It will then configure the Node to use
+the ramdisk Image from the URL in the Status. This must be done before
+inspection or cleaning starts, but can be after the Ironic node reaches the
+Manageable state. Completion of `ValidateManagementAccess()` will only be
+blocked due to the image not being ready in the Registering, Inspecting, and
+Deprovisioning states, or when required by Ironic.
+
+The contents of the networkData Secret are generally assumed to be in the same
+format output by OpenStack and consumable by cloud-init. However, for Metal³'s
+purposes this is not a requirement as the data is passed through directly.
+Authors of image customisation controllers are free to interpret the data in
+any way they see fit.
+
+Since PXE-booted hosts generally must be connected to the provisioning network
+on an untagged VLAN using LACP for any NIC bonding, and on the provisioning
+network ironic will supply an IP address via DHCP, there should be no need for
+this feature other than when using virtualmedia drivers. An API option can
+always be added later if this proves not to be the case.
+
+The `BareMetalHost` should be set as the Owner of the `PreprovisioningImage` so
+that the image gets deleted when the Host is deleted. This allows controllers
+to use the resource to perform cache management.
+
+If the image cache is lost, the controller must clear the Ready condition on
+all existing `PreprovisioningImage` objects.
+
+### Risks and Mitigations
+
+If the consumer of the BareMetalHost resource is able to modify the
+networkData, they will be able to influence the configuration of the Host after
+they have deprovisioned and potentially released it. This is largely mitigated
+by specifying the preprovisioningNetworkData separately from the networkData
+that is potentially controlled by the user. A user with write access to the
+`BareMetalHost` already has almost unlimited control over it anyway; if their
+access is moderated through another controller then they will only be able to
+control the networkData and not the preprovisioningNetworkData.
+
+### Work Items
+
+- Implement the new `PreprovisioningImage` CRD
+- Implement a basic controller to reconcile the `PreprovisioningImage`
+- Add runtime flag to configure image building integration
+- Add the `preprovisioningNetworkData` field to the `BareMetalHost` Spec
+- Add a watch on the `PreprovisioningImage` for the `BareMetalHost` reconciler
+- Create/update the `PreprovisioningImage` when it is missing/outdated
+- Pause registration until the image URL is available
+- Pass the ramdisk URL to Ironic
+
+### Dependencies
+
+None
+
+### Test Plan
+
+Enable the functionality in the integration test, using the default (trivial)
+controller implementation for the `PreprovisioningImage`. This should produce
+the same result as today, but exercise the integration between the the two
+resource types.
+
+### Upgrade / Downgrade Strategy
+
+This functionality will be disabled by default, so that existing users don't
+need to do anything special on upgrade.
+
+### Version Skew Strategy
+
+The API surface area is very small, so we are unlikely to encounter version
+skew issues.
+
+## Drawbacks
+
+This adds complexity inside the baremetal-operator, for something that is not
+needed by many users.
+
+## Alternatives
+
+### Add the ramdisk image URL to the BareMetalHost API
+
+Instead of trying to integrate the two controllers, add a field to the
+BareMetalHost specifying the URL of the ramdisk image.
+
+To avoid a race, this must be done as part of creating the BaremetalHost CR or
+the baremetal-operator must block until the image is available.
+
+The former would be a significant limitation as it requires running a little
+workflow in the process of creating a BareMetalHost - something that often
+might be done manually by a user.
+
+Since this functionality would be disabled by default, however, the latter
+would not prevent backwards compatibility. We could automatically fill in the
+field in the Spec with the known value whenever it is blank if the
+functionality is disabled.
+
+This puts a lot of power in the hands of any user with write access to the
+BareMetalHost, but this is largely the case anyway pending the implementation
+of API decomposition (which would also eliminate the concern here).
+
+Arguably this couples the BareMetalHost API more tightly to how today's Ironic
+works, with the assumption that pre-provisioning we will boot a ramdisk agent
+now baked in to the API.
+
+Overall this is more flexible, since the image URL can be added in any way (not
+necessarily by adding a separate controller). It also has fewer moving parts.
+On the other hand, the more opinionated proposal above provides a more
+stuctured way for associating the network data with the BareMetalHost it
+applies to.
+
+### Have Ironic customise the image
+
+Currently Ironic just boots the Node with a given kernel and initrd. However,
+when deploying using virtual media it has the capability to [customise the ISO
+image to include the
+networkData](https://docs.openstack.org/ironic/latest/admin/dhcp-less.html#configuring-network-data).
+(Specifically, this is implemented in the iLO and redfish drivers, and
+inherited by idrac-redfish-virtualmedia.)
+
+Unfortunately, this requires that the image already contain built-in capability
+to decode the networkData and apply the network config. (In practice, this
+means images using cloud-init and network data in the OpenStack format only.)
+This excludes others, such as CoreOS. Although CoreOS has the ability to pull
+in a container to process data, it cannot do so without network access, so
+processing the networkData other than with something built in to the operating
+system presents a chicken-and-egg problem.
+
+Another downside of this is that there is no access from outside for a user to
+request their own image build (e.g. to manually boot a server into the deploy
+ramdisk for discovery).
+
+Ironic verifies the data against a JSON Schema, whereas currently the
+networkData field in the BareMetalHost is free-form JSON. However, the choice
+of schema is customisable by a config option, so it could be set to free-form
+to avoid breaking backward compatibility for existing resources.
+
+With this strategy we would always have to update the network data when it
+changes; we have no way of distinguishing between changes made by an
+administrator or other user.
+
+### Configure the IP address from within a fixed image
+
+Using IPv6 Neighbour Discovery, we can determine the network prefix for each
+interface. Using the network prefix we can statelessly choose a static IP
+address without fear of collison (by using the Modified EUI-64 method based on
+the MAC address, and/or a stable privacy address).
+
+If the host is not directly attached to the network Ironic is on, Neighbour
+Discovery can also locate routers on the attached networks and identify which
+are suitable to be used as a default route.
+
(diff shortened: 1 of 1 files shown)


zaneb:
/hold
I actually prefer one of the first two alternatives, but since the third alternative was already proposed as part of #180, it seems like we should have a more in-depth discussion of the options.
/cc @dtantsur @hardys @dhellmann

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,215 @@
...
+A command-line switch (defaulting to off) will enable integration between the
+baremetal-operator and the image request controller.
+
+### Implementation Details/Notes/Constraints
+
+When integration is enabled, the controller portion of the baremetal-operator
+will update the `RamDiskImageRequest` with the same name and namespace as the
+`BareMetalHost`, creating it if it doesn't exist, to point to the metaData and
+networkData Secrets referenced in the `BareMetalHost` Spec. The Architecture
+will be obtained from the Host's hardware profile.
```
dtantsur, line 94: There is no hardware profile, nor inspection data, when the ramdisk is needed.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,215 @@
...
+
+In the Registration state, the ironic provisioner will avoid attempting to
+Manage the Node in Ironic until the Ready condition is true in the
```
dtantsur, line 97: I'd prefer we move to `manageable` first. It involves validating the credentials and does not involve any actions with a ramdisk just yet.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,215 @@
...
+interpret the data in any way they see fit.
+
+The `BareMetalHost` should be set as the Owner of the `RamDiskImageRequest` so
+that the request gets deleted when the Host is deleted. This allows controllers
+to use the resource to perform cache management.
+
+Once the image URL is set, it shouldn't be changed. If the user wishes to
+rebuild the ramdisk with newer settings, they can do so by deleting the image
+request. The baremetal-operator should watch the image request and recreate it
+when this occurs.
```
dtantsur, line 114: What happens to the provisioned node? What happens if the node is in flight?

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,215 @@
...
+A new Custom Resource type, `RamDiskImageRequest` will be created in the
+metal3.io group. The Spec will contain the following fields:
+
+- Name of metaData Secret
+- Name of networkData Secret
+- Architecture
+
+The Status will contain the following fields:
+
+- Image URL
```
dtantsur, line 73: ISO URL. Let's not preclude a future possibility to build a kernel/ramdisk pair.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,215 @@
...
+As a data center operator, I would like to manage hosts with metal³ and deploy
+them via virtualmedia on networks where there is no DHCP-like configuration
+service.
+
+## Design Details
+
+A new Custom Resource type, `RamDiskImageRequest` will be created in the
+metal3.io group. The Spec will contain the following fields:
+
+- Name of metaData Secret
```
dtantsur, line 67: To make it usable for CoreOS we need userData. MetaData uses the OpenStack format and is related to the instance, which may not exist at this point.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,215 @@
...
+alternate controllers to allow image customisation.
+
+A command-line switch (defaulting to off) will enable integration between the
+baremetal-operator and the image request controller.
+
+### Implementation Details/Notes/Constraints
+
+When integration is enabled, the controller portion of the baremetal-operator
+will update the `RamDiskImageRequest` with the same name and namespace as the
+`BareMetalHost`, creating it if it doesn't exist, to point to the metaData and
```
dtantsur, line 92: s/metaData/userData/

How do you handle situation when a user needs different userData for the ramdisk and the instance, e.g. when a user needs to do something on first boot on the instance (build RAID, add users), but not on the ramdisk?

dtantsur (changes requested):

design/baremetal-operator/image-builder-integration.md
zaneb: Good point (although the hardware profile may be available from the spec, and in practice the default will give us what we want anyway for now, so we can get it). If we ever want to do multi-arch we will have to move the hardware profile earlier (before registration), or find another solution for specifying the architecture.

design/baremetal-operator/image-builder-integration.md
zaneb: We'll update the settings in the node if they don't match, and pause during registration if the image is not available during registration. The next time ironic boots into the deploy image it'll use the new one.

design/baremetal-operator/image-builder-integration.md
zaneb: I think we can drop the metaData secret - I had been thinking of it as we'll pass all of the stuff supplied by the cloud, as opposed to the stuff supplied by the user (userData), but that's the wrong way of thinking about it. We should pass the stuff that is not specific to a Nova instance, which is only the networkData.

As far as the CoreOS Ignition goes, I was expecting that we would build a fixed (not host-specific) Ignition into the image that knows how to interpret the network data (and loads the IPA container).

design/baremetal-operator/image-builder-integration.md
zaneb: See above, I don't think we should pass userData to the ramdisk, it belongs to the user for use with the actual provisioned host.

zaneb:
@maelk I'd be interested to know if y'all have any requirement for something like this... from the fact that the CAPM3 is integrated with an IPAM I'd guess that there are some similar use cases around. Is this something that would help?

dtantsur:
Note that to make this work with CoreOS we may need to pass Ignition somewhere, not just the networking configuration (e.g. we need to supply the registry to download IPA from).

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+- Name of networkData Secret
+- Architecture
+
+The Status will contain the following fields:
+
+- Image URL
+- Name + Version of networkData Secret
+- Error message
+- Ready condition
+- Error condition
```
dhellmann, line 76: Why are `Ready` and `Error` separate fields?

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+them via virtualmedia on networks where there is no DHCP-like configuration
+service.
+
+## Design Details
+
+A new Custom Resource type, `RamDiskImageRequest` will be created in the
+metal3.io group. The Spec will contain the following fields:
+
+- Name of networkData Secret
+- Architecture
```
dhellmann, line 68: Are both fields required or optional?

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+A command-line switch (defaulting to off) will enable integration between the
+baremetal-operator and the image request controller.
+
+### Implementation Details/Notes/Constraints
+
+When integration is enabled, the controller portion of the baremetal-operator
+will update the `RamDiskImageRequest` with the same name and namespace as the
+`BareMetalHost`, creating it if it doesn't exist, to point to the networkData
+Secret referenced in the `BareMetalHost` Spec. The Architecture will be
+obtained from the Host's hardware profile. (Currently we don't ratify the
```
dhellmann, line 92: The hardware profile isn't set via automation until after inspection, so we won't be able to use that to determine the architecture unless we start requiring the user to provide a good value. We've previously said we wanted to deprecate that user-provided input entirely, in favor of more detailed fields like `rootDeviceHints`.

We chatted offline about having a different CRD to define how to acquire an image, so that different implementations could have different recipes that accept different inputs. If we do that, architecture could go into that CRD.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+choice of hardware profile until after inspection, but we could move this state
+earlier. In any event, we don't support multiple architectures today, so we
+could use a default value in the interim.)
+
+The ironic provisioner will refuse to report registration complete in
+`ValidateMangementAccess()` until the Ready condition is true in the
+corresponding `RamDiskImageRequest` Status. It will then configure the Node to
+use the ramdisk Image from the URL in the Status. This must be done before
+inspection or cleaning starts, but can be after the Ironic node reaches the
+Manageable state.
```
dhellmann, line 102: Based on lines 110-126, I think this only applies to virtual media drivers. Is that right?

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+feature except when using virtualmedia drivers. An API option can always be
+added later if this proves not to be the case.
+
+The `BareMetalHost` should be set as the Owner of the `RamDiskImageRequest` so
+that the request gets deleted when the Host is deleted. This allows controllers
+to use the resource to perform cache management.
+
+Once the image URL is set, it shouldn't be changed. If the user wishes to
+rebuild the ramdisk with newer settings, they can do so by deleting the image
+request. The baremetal-operator should watch the image request and recreate it
```
dhellmann, line 134: Are the spec field immutable, then? Do they need to be?

Does the BMO actually need to watch the image request, or would it just make one when it sees that it needs one based on host state?

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+be enough to accomplish our goals here. It is implemented in the iLO and
+redfish drivers, and inherited by idrac-redfish-virtualmedia.
+
+One downside of this is that there is no access from outside for a user to
+request their own image build (e.g. to manually boot a server into the deploy
+ramdisk for discovery).
+
+With this strategy we would always have to update the network data when it
+changes; we have no way of distinguishing between changes made by an
+administrator or other user.
```
dhellmann, line 203: We should point out that the customization approaches for different agent image types (especially coreos) aren't necessarily compatible with ironic.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+are suitable to be used as a default route.
+
+Information about VLANs and static bonds can be obtained from LLDP, and
+information about dynamic bonds from LACP or LLDP. This information could be
+used to automatically configure the interfaces on the ramdisk.
+
+This only works for IPv6, but it would be pretty unusual for an IPv4-only
+environment not to have DHCP. It also assumes that LLDP is enabled on the local
+switch, although that is common (and also relied upon by our hardware details
+introspection).
```
dhellmann, line 223: This wouldn't give us the ability to use bonded NICs, right?

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Add the ramdisk image URL to the BareMetalHost API
+
+Instead of trying to integrate the two controllers, add a field to the
+BareMetalHost specifying the URL of the ramdisk image.
+
+This must be done at the time the BareMetalHost is created; at any later time
+it may or may not take effect due to a race. This is a significant limitation
```
dhellmann, line 231: We could avoid the race by having the BMO wait to perform inspection until it has that URL, either always or only when some other flag is set. Then another controller could ensure the URL is provided.

design/baremetal-operator/image-builder-integration.md
zaneb: They're Conditions.

design/baremetal-operator/image-builder-integration.md
dhellmann: It's easier to add fields than change or drop them. If we're not going to use it right now, I would just leave architecture out.

design/baremetal-operator/image-builder-integration.md
zaneb: It would work fine if the switch is configured with LACP (which is apparently the common case). It wouldn't work reliably if the switch is configured with a static link aggregation.

design/baremetal-operator/image-builder-integration.md
zaneb: How would we handle backward compatibility?

dhellmann: The flag could help with backwards compatibility. Default to "no wait" and deploy with "wait" when that's the desired behavior. In OpenShift, the CBO could update hosts to set the URLs as part of the upgrade process.

design/baremetal-operator/image-builder-integration.md
```diff
@@ -0,0 +1,252 @@
...
+`BareMetalHost`, creating it if it doesn't exist, to point to the networkData
+Secret referenced in the `BareMetalHost` Spec. The Architecture will be
+obtained from the Host's hardware profile. (Currently we don't ratify the
+choice of hardware profile until after inspection, but we could move this state
+earlier. In any event, we don't support multiple architectures today, so we
+could use a default value in the interim.)
+
+The ironic provisioner will refuse to report registration complete in
+`ValidateMangementAccess()` until the Ready condition is true in the
+corresponding `RamDiskImageRequest` Status. It will then configure the Node to
```
andfasano, line 101: It's not clear under which circumstances the RamDiskImageRequest Status Ready condition will become true. I guess it would be useful to clarify this point, at least for the default controller (I'd assume after it verified and set the Image URL)

maelk:
@zaneb I think this addresses quite well the requirement of networks without DHCP in the virtual-media case. However, there is no guarantee that the networkData as desired for the target node allows communication with Ironic. I think the networkData for provisioning should be different than the networkData of the provisioned node (you might not want to be able to reach the provisioning network on  a running K8S worker for example).

zaneb:
> @zaneb I think this addresses quite well the requirement of networks without DHCP in the virtual-media case. However, there is no guarantee that the networkData as desired for the target node allows communication with Ironic. I think the networkData for provisioning should be different than the networkData of the provisioned node (you might not want to be able to reach the provisioning network on a running K8S worker for example).

Thanks for the feedback, this a is a good point. I proposed using the existing network data field on the basis that this is how ironic does it in an OpenStack cloud, and also to make sure it mostly Just Works without having to do an extra step of configuring the Host. For stuff like static IPs, they will need to be the same so having two fields is an invitation for them to get out of sync. Theoretically the network topology doesn't change between boot time and provisioning time, so they should be able to have the same config. (While you may not want to be able to reach the provisioning network from a worker node, simply not telling the worker node about that network when it is nevertheless connected is a pretty limited form of isolation.)

We have talked about integrating metal³ with something to control the network switches, and in that case it's quite possible there will be different configs for boot and provisioning. But if necessary the networkData can change over time (i.e. you can write new network data at provisioning time, and that will be used for provisioning - the annoying part is changing it back).

Some options:

1. Use the same field for now, and wait to see if we encounter the need for another
2. Add an additional field for ramdiskNetworkData. Fall back to the regular networkData if it is not present.
3. In the [proposed API decomposition](https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit?usp=sharing), have networkData fields in both the BareMetalAllocation and BareMetalImage, and have the controller write whichever is appropriate for the current phase into the single field in the BareMetalHost.

What do you think?

design/baremetal-operator/image-builder-integration.md
zaneb: That could work.

design/baremetal-operator/image-builder-integration.md
zaneb: > Do we need the extra layer of complexity, though?  I think we already know we're going to need multiple ways to build images. Perhaps not _simultaneously in a single cluster_, though. Can we put the cluster-specific configuration for the image building into the settings for the controller that will respond to the RamDiskImageBuilder API, so we can keep that API simple and avoid having too much info in the BMO? And, if we take that approach, can we expand to something like what I described here later if we need to?

I'm not seeing a lot of value in creating the RamDiskImageBuilder type - unlike the PersistentVolume case it's not clear to me what data would actually be stored in that. The obvious way to select among multiple controllers is to have a label that they can filter on. Perhaps we could copy the label from the BMH to avoid adding another field to the BMH API. And you can always run different controllers in different namespaces. So I think there are paths forward if we find later that we need that extra complexity.

> bike-shed: We should call the API RamDiskImage instead of RamDiskImageRequest.

Agreed.

maelk:
Yes, we do have two different switch configurations in the network configuration design proposal, one for PXE boot, and one for running hosts. It is still being discussed however, since when provisioning, we need the change to happen after IPA was successfully provisioning the node, but before that node boots up. But in this case we definitely have two different configurations, and we should support this in the proposal. 
Using the same field might be complicated considering that CAPM3 would not be able to just configure the BMH and wait for it to be provisioned, it would need to handle the extra logic around creating the image (and recreation in case the cache is lost), that might end up quite complex.
Point 2 seems the most straightforward right now, until we can move to point 3 when that gets implemented.

hardys:
> Add an additional field for ramdiskNetworkData. Fall back to the regular networkData if it is not present.

This sounds reasonable if we expect to support the scenario described by @maelk now, and it avoids additional complexity for users where the deploy-time and post-deploy configuration is the same.

Nit re naming `deployNetworkData` might fit better with the naming discussed in #180

dtantsur:
> Nit re naming deployNetworkData might fit better with the naming discussed in #180

Unlike #180, this applies during inspection and cleaning as well, I'd keep ramdisk.

design/baremetal-operator/image-builder-integration.md
dtantsur: > I guess eventually we'll likely want to add an Architecture field to BMH

Quite likely. E.g. ironic is capable of serving different PXE options for different architectures, but it has to know the architecture.

hardys:
> > Nit re naming deployNetworkData might fit better with the naming discussed in #180
> 
> Unlike #180, this applies during inspection and cleaning as well, I'd keep ramdisk.

Perhaps we should divert this discussion to #180 but AFAIK we'd want to avoid an additional reboot in the coreos case, so IPA running on coreos would still handle inspection/cleaning?

But fair enough, I'm not blocking on the naming "ramdisk" just implies PXE to me, even though we know there is a ramdisk baked into any virtualmedia image as well.

dtantsur:
> we'd want to avoid an additional reboot in the coreos case, so IPA running on coreos would still handle inspection/cleaning?

This is how it will normally work. My point is, if we call it `deployNetworkData`, won't it be confusing that is also applies to inspection and cleaning?

hardys:
> > we'd want to avoid an additional reboot in the coreos case, so IPA running on coreos would still handle inspection/cleaning?
> 
> This is how it will normally work. My point is, if we call it `deployNetworkData`, won't it be confusing that is also applies to inspection and cleaning?

Maybe?  I'm not sure tbh, "ramdisk" doesn't really seem to address that potential confusion either though, but I don't want to trigger undue bikeshedding :)  I was just thinking "deploy" as in all operations associated with deployment (vs boot of the deployed image).

zaneb:
Updated based on the latest comments. Chose `preprovisioningNetworkData` for now, it's a bit of a mouthful but shouldn't introduce any confusion.

I'm actually not hating the idea of just adding an image field (given that we're adding a field anyway) now that Doug has suggested a way it could remain backwards-compatible without races. That would certainly be simpler. If there's an advantage to having the builder integrated as in the main proposal, it's probably that it provides a more structured way to associate the network data with the host, which should lead to more consistency across different deployments of metal³.

zaneb:
Made a couple more changes:

networkData now falls back to preprovisioningNetworkData instead of the other way around. I think this is conceptually better: the infrastructure owner always controls the preprovisioningNetworkData, and often you'll want to deploy with the same network configuration, but sometimes either the infrastructure owner or the user will want to provide a different network config for the provisioned image.

This means we can drop the constraint of making the Image object immutable, and just rebuild whenever the secret changes. That takes out a lot of the ugliness and complexity that was creeping in to the previous version.

dtantsur:
> networkData now falls back to preprovisioningNetworkData

Just keep in mind that with our CoreOS IPA `networkData` will be no-op. Unless somebody tells me how to wire it in.

zaneb:
> > networkData now falls back to preprovisioningNetworkData
> 
> Just keep in mind that with our CoreOS IPA `networkData` will be no-op. Unless somebody tells me how to wire it in.

Yes, although once we have booted with the preprovisioningNetworkData (baked in to the image) and connected to the network, we could in theory pull in a container to interpret the provision-time networkData (possibly with some changes to the coreos-installer deploy step?). Or we can do network configuration in ignition (passed through userData), or something else. I think it's yet to be determined how we want to pass the config in OpenShift, since we might also want to change it on day 2, but there are a number of options.

I think conceptually this makes sense as a generic interface for the BMH.

zaneb:
I think stuff like that should be completely up to the controller implementation. As long as it gets a URL it can pass to Ironic that will work, BMO shouldn't care what it looks like.

fmuyassarov:
Let's have a lazy consensus of a week from today and if no objections, we can merge it next week.

fmuyassarov:
Since there were on objections/further comments within the last week (after we agreed on lazy consensus), I am going to merge this proposal.
/lgtm

furkatgofurov7:
Hi @zaneb, unless I am missing it, I have not found a implementation patch in BMO related to this proposal, is it correct? If that is the case, any timeline when this will land. Thanks.

zaneb:
It's metal3-io/baremetal-operator#936. I'll flip it out of Draft once I've actually been able to verify that it mostly works (made a lot of progress last week).

PR #196, opened 2021-09-23
Hardware RAID Extension - Physical disks and RAID controller

MahnoorAsghar:
Added a spec for extending hardware RAID parameters, allowing the user to specify physical disks and RAID controller names for a hardware RAID volume
Also, corrected a spelling error in design/baremetal-operator/bios-config.md

This is to replace an old spec PR ([PR 148](https://github.com/metal3-io/metal3-docs/pull/148)), that proposed a generic (vendor agnostic) naming convention for physical disks and controllers. This approach was deemed infeasible after discussion with the metal3 and ironic communities. Therefore, the proposal is now to extend hardware RAID using exact disk and controller names.

### DIFF
--- a/design/baremetal-operator/raid-disk-controller.md
+++ b/design/baremetal-operator/raid-disk-controller.md
@@ -1,0 +1,194 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# raid-extension-for-physical-disks-and-controller
+
+## Status
+
+provisional
+
+## Summary
+
+This document provide a way to express the physical disks and controller sections
+for the (hardware) RAID configuration of a baremetal host.
+
+## Motivation
+
+The user should be able to specify exactly which physical disks and RAID
+controllers are to be used to construct the hardware RAID volume(s).
+
+### Goals
+
+The primary goal is to depict the physical disk and RAID controller names in
+the baremetal-host ``hardware-raid`` section.
+
+Afterwards, implementation of the same will be done within the
+``baremetal-operator`` by extending the ``baremetal-host`` specification.
+
+### Non-Goals
+
+- This specification does not deal with ``software-raid`` and extension for it.
+- It does not attempt to cover any generic (vendor agnostic)  naming convention
+  for disks or controllers.
+- It does not cover testing for hardware from all the vendors. It will only be
+  tested on Dell EMC hardware, and other vendors will have to test it on their
+hardware.
+
+## Proposal
+
+The proposal is to allow specifying the physical disks and RAID controller
+names in the hardware RAID section. For this purpose, new YAML attributes are
+to added in the BareMetalHost spec, namely `physicalDisks` and `controller`.
+
+In terms of the CRD, the current implementation of the RAID has the following
+sections in it, for hardware RAID:
+
+```yaml
+spec:
+  raid:
+    hardwareRAIDVolumes:
+      - sizeGibiBytes: 1000
+        level: 1
+        name: <volume-name>
+        rotational: true
+        numberOfPhysicalDisks: 2
+```
+
+We propose to add fields into the CRD, in the following manner:
+
+```yaml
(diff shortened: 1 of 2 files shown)


design/baremetal-operator/raid-disk-controller.md
```diff
@@ -0,0 +1,193 @@
...
+  raid:
+    hardwareRAIDVolumes:
+      - sizeGibiBytes: 1000
+        level: 1
+        name: <volume-1-name>
+        rotational: true
+        numberOfPhysicalDisks: 2
+        physicalDisks:
+          - <disk-1-name>
+          - <disk-2-name>
```
macaptain, line 73: Just thinking aloud on the decision to make the API for `physicalDisks` a list of strings here.

In gophercloud, the `physicalDisks` is of type `[]interface{}` (https://github.com/gophercloud/gophercloud/blob/ad8066cf841fa16a5497835d6df808bf7aadcbe4/openstack/baremetal/v1/nodes/requests.go#L605-L606), not `[]string`.

That's because in Ironic the `physical_disks` parameter can either be a list of device names (like you're using here) or a list of pairs for device hints: https://github.com/openstack/ironic/blob/af0e5ee096fa237290776969a37f3bced96b7456/ironic/drivers/raid_config_schema.json#L59-L63

I think the hints are more for software RAID. I guess if there was a hardware RAID manager that used hints, we could have `physicalDiskHints` here in the API and then handle passing that to gophercloud later, so I _think_ it's all good for `physicalDisks` here to be a list of strings for the names of the disks.

design/baremetal-operator/raid-disk-controller.md
```diff
@@ -0,0 +1,193 @@
...
+
+```yaml
+spec:
+  raid:
+    hardwareRAIDVolumes:
+      - sizeGibiBytes: 1000
+        level: 1
+        name: <volume-1-name>
+        rotational: true
+        numberOfPhysicalDisks: 2
```
macaptain, line 70: If `numberOfPhysicalDisks` didn't match the length of `physicalDisks`, Ironic would throw an error at runtime:

https://docs.openstack.org/ironic/latest/admin/raid.html#backing-physical-disks

> If properties from both “Backing physical disk hints” or “Backing physical disks” are specified, they should be consistent with each other. If they are not consistent, then the RAID configuration will fail (because the appropriate backing physical disks could not be found).

I don't think there's an easy way to validate with kubebuilder this scenario, and I don't think it needs validation, but maybe there is a way.

macaptain (approved):
I haven't worked a huge amount on this area, but this proposal makes complete sense to me. The additions to the API map to supported fields in gophercloud and Ironic, where the behavior is already implemented, so it seems like it would be straightforward and useful to add this functionality.

The proposed implementation seems good. I don't think any changes would be needed to the disk cleaning, for example.

My two comments are just thinking aloud on the API design before it gets fixed. In my view, it's all good.

design/baremetal-operator/raid-disk-controller.md
```diff
@@ -0,0 +1,193 @@
...
+
+# raid-extension-for-physical-disks-and-controller
+
+## Status
+
+provisional
+
+## Summary
+
+implementation for the same.
```
zaneb, line 16: ahem

zaneb (commented):
/approve

design/baremetal-operator/raid-disk-controller.md
MahnoorAsghar: Good observation! I had to convert the list of strings to a list of interfaces while implementing this. Since the exact physical disk names needed to be used in our use case, using strings made more sense to me.

design/baremetal-operator/raid-disk-controller.md
MahnoorAsghar: Okay, shall I add a note of that here?

design/baremetal-operator/raid-disk-controller.md
MahnoorAsghar: fixed, thanks for pointing that out

PR #208, opened 2021-10-31
HostData CR for host inspection data

fmuyassarov:
Instead of writing hardware details into the status of
 BareMetalHost, write it to a HostData Custom Resource.
 Baremetalhost and HostData will be parent and child objects
 respectively, and this relationship is controlled via ownerReferences
 chain.
    
 Having quite big inspection data is problematic when pivoting
 BareMetalHost objects to a target cluster. Storing the data in a
 CR will avoid the issue regardless of the inspection data
 size.

### DIFF
--- a/design/baremetal-operator/hostdata_crd.md
+++ b/design/baremetal-operator/hostdata_crd.md
@@ -1,0 +1,235 @@
+# hostData custom resource for host inspection data
+
+## Status
+
+implementable
+
+## Summary
+
+Introduce HostData Custom Resource (CR) to store host inspection data.
+
+## Motivation
+
+Currently inspection data is written to baremetalhost status subresource. When
+pivoting to a target cluster, with *`clusterctl move`*, status subresource is
+not copied with the host, because from Cluster-API perspective, it's controllers'
+job to rebuild the status in the target cluster. Also, according to the [Kubernetes API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), status is expected to be reconstructed by controllers if
+it's lost for any reason. Baremetal Operator (BMO) is capable of reconstructing
+host status by re-inspecting the host, but it will force the already provisioned
+host to go through another round of provisioning, which eventually reboots the
+host after IPA steps are completed. To have the inspection data available in the
+target cluster and to avoid unnecessary provisioning after pivoting, we currently
+copy inspection data to baremetalhost.metal3.io/status annotation so that BMO in
+the target cluster can reconstruct the BareMetalHost from the annotation with the
+appropriate inspection data and as such the reconciliation of the BMH can continue
+from the same state as it was in the source cluster.
+
+However, there is a limitation with the current approach, because when inspection
+data is very large (e.g. when the host has 1000 multipath disks), it is easy to hit
+the global limit of `metadata.annotations` [262144](https://github.com/kubernetes/kubernetes/blob/master/test/integration/controlplane/synthetic_controlplane_test.go#L313) bytes and once the limit is hit
+the move process will fail to pivot BareMetalHosts successfully.
+
+### Goals
+
+- Store host inspection data into Spec of HostData CR
+- Adjust pivoting workflow to restore inspection data from HostData
+- Make the HostData CR immutable
+- Don't write `status.hardware` data to the status annotation
+
+### Non-Goals
+
+- Remove `status.hardware` from BareMetalHost CR
+- Drop status annotation
+
+## Proposal
+
+### User Stories
+
+As an operator, from an ephemeral cluster I would like to pivot/move all the Cluster
+API, CAPM3 and BMO objects to a target cluster successfully.
+
+## Design Details
+
+Apart from writing hardware details into the status of BareMetalHost, write it to a
+HostData CR spec. BareMetalHost and HostData will be parent and child objects
+respectively, and this relationship is controlled via ownerReferences chain. As such,
+the same name and namespace will be used for HostData as its owning object BareMetalHost.
+The spec of HostData will contain the same fields & hardware information as currently
+under `status.hardware` of BareMetalHost.
+
+OwnerReferences refers to the linked BareMetalHost and is added by the operator.
+To see all the available fields of ObjectReference type, check [core/v1](https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference) package.
+
+There will be duplication of inspection data, i.e.,in the HostData spec and BareMetalHost
+status, for the purpose of avoiding breaking API changes. However, we plan to drop
+[hardware](https://github.com/metal3-io/baremetal-operator/blob/05d12b6768a9989a9a4e61dad6cd1f9e84a6e078/apis/metal3.io/v1alpha1/baremetalhost_types.go#L721)
+field from the status of BareMetalHost in the first version bump of the API.
+
+### Pivoting
+
+To pivot objects excluded from Cluster API chain to a target cluster, one can set
+`clusterctl.cluster.x-k8s.io=""` label on HostData and BareMetalHost CRDs.
+Current status annotation based pivoting workflow will stay as it is. However,
+CAPM3 machine controller will no longer copy `status.hardware` of BareMetalHost
+to the annotation. Because that part will be avialable in the form of HostData CR
+in the target cluster.
+
+### Implementation Details/Notes/Constraints
+
+The life of HostData is dependent on BareMetalHost. When BareMetalHost is entering
+inspection state, the HostData will be created. Failure in reaching inspection state
+or having `inspect.metal3.io: disabled` (without `inspect.metal3.io/hardwaredetails`
+annotation) annotation on BareMetalHost will force the operator to not create the
+HostData. Deleting a BareMetalHost will result in deletion of a HostData respectively.
+
+HostData should be immutable and this will be controlled via [Validating Webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) and with RBAC permissions.
+In case, when [re-inspection](https://github.com/metal3-io/baremetal-operator/blob/master/docs/inspectAnnotation.md) of a `available` BareMetalHost is requested,
+new HostData will be re-created, because updating the old HostData will be
+blocked by the Validating Webhook.
+
+Deprovisioning of the BareMetalHost must not delete the HostData. It can
+be deleted either when BareMetalHost is being deleted or when re-inspection
+is requested. To ensure HostData availability while BareMetalHost is provisioned,
+a finalizer should be set on the HostData.
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData is missing, operator acts differently depending on the current
+state of the BareMetalHost.
+
+- host is provisioned: we expect that HostData never existed. Even if it was existed
+    but user deleted it, then it is user's fault. Normally finalizer blocks HostData
+    deletion until BareMetalHost exists.
+- host is available:
+    1. copy inspection data from the status of BareMetalHost if exists;
+    2. re-inspect the host if inspection data doesn't exist in the BareMetalHost status.
+        This will be the first and only option in the future when `status.hardware` is removed;
+
+As mentioned above, we will have two copies of inspection data available for the
+time being. One in the Spec of HostData and the other in the Status of BareMetalHost.
+In the future, when new API version of BareMetalHost is introduced, `status.hardware`
+will be removed from BareMetalHost. As such, missing HostData for available BareMetalHost
+will result in another inspection before starting provisioning.
+
+**Note:** [cross-namespace owner references are not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications). Thus, HostData
+and BareMetalHost must be in the same namespace.
+
+#### Example of HostData lifecycle
+
+- BareMetalHost is created
+- BareMetalHost is in inspecting state
+- Operator creates a HostData with the same name and namespace as BareMetalHost, fills out
+    the HostData with the host inspection data and sets ownerReferences
+- Provisioning is requested
+- Pivoted to a target cluster, operator restores host inspection data from HostData and
+    the rest from status annotation (e.g. `status.Provisioning`, `status.poweredOn` and etc)
+- Deprovisioning is requested
+- BareMetalHost deletion requested, operator deletes the HostData and BareMetalHost
+
+The spec of HostData:
+
+```go
+// Adding HardwareDetails struct here for visibility. During implementation,
+// we don't have to duplicate it, but use from baremetalhost_types.go.
+type HardwareDetails struct {
+
+    SystemVendor HardwareSystemVendor `json:"systemVendor,omitempty"`
+    Firmware     Firmware             `json:"firmware,omitempty"`
+    RAMMebibytes int                  `json:"ramMebibytes,omitempty"`
+    NIC          []NIC                `json:"nics,omitempty"`
+    Storage      []Storage            `json:"storage,omitempty"`
+    CPU          CPU                  `json:"cpu,omitempty"`
+    Hostname     string               `json:"hostname,omitempty"`
+
+}
+
+type HostDataSpec struct {
+    // The hardware discovered on the host.
+    HardwareDetails *HardwareDetails `json:"hardware,omitempty"`
+}
+```
+
+Example CR:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: HostData
+metadata:
+  finalizers:
+  - baremetalhost.metal3.io
+  name: hostdata-sample
+  ownerReferences:
+  - apiVersion: metal3.io/v1alpha1
+    controller: true
+    kind: BareMetalHost
+    name: bmh-sample
+    uid: 546392e0-66b1-45c0-8d3d-6994ff82b477
+spec:
+    # inspection data
+    cpu:
+    firmware:
+    hostname:
+    nics:
+    ramMebibytes:
+    storage:
+    systemVendor:
+```
+
+### Risks and Mitigations
+
+Existing deployments will require installation of the HostData CRD and necessary
+RBAC permissions when bumping the operator version. This needs to be clearly
+documented.
+
+### Work Items
+
+- Add HostData API
+- Modify the operator code
+- Adjust CAPM3 machine controller to exlude writing `status.hardware` into the statusAnnotation
+- Add unit tests
+- Update documentation
+
+### Dependencies
+
+None
+
+### Test Plan
+
+Unit tests should be added. Integration test doesn't require any change.
+
+### Upgrade / Downgrade Strategy
+
+From a user perspective, this feature doesn't introduce any breaking changes
+to existing deployments, because `status.hardware` field will stay until is
+decided otherwise and this feature is only an addition to the current APIs.
+For now, there will be two copies of inspection data available, in the
+`spec.hardware` of HostData and `status.hardware` of BareMetalHost.
+
+### Version Skew Strategy
+
+To mitigate version skew issues, we plan to keep BaremetalHost and HostData
+synchronized by bi-directional copying of inspection data.
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+- Use k8s built in objects like ConfigMap/Secret. From security point of view, although,
+    it is possible to restrict permissions to edit ConfigMaps (only to those
+    storing inspection data) based on the name (i.e. `resourceNames`), but it will add
+    complexity to the operator.
+- Increase metadata global limit. However, this does not always work as it is hard
+    to know the maximum length of inspection data.
+- Add filters to filter-out unn inspection data?
+
+## References
+
+- [Discussion issue CAPM3](https://github.com/metal3-io/cluster-api-provider-metal3/issues/266)
+- [Discussion issue BMO](https://github.com/metal3-io/baremetal-operator/issues/952)
+- [Status annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/statusAnnotation.md)
+- [clusterctl move](https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html#clusterctl-move)
+- [Pause annotation](https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#pausing-reconciliation)
+- [Disable inspection](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/external-introspection.md#disable-inspection-proposal)
+- [ObjectReference type](https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference)


design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+Currently inspection data is written to baremetalhost status subresource. When
+pivoting to a target cluster, with *`clusterctl move`*, status subresource is not
+copied with the host, because from Cluster-API perspective, it's controllers'
+job to rebuild the status in the target cluster. Also, according to the [Kubernetes API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), status
+is expected to be reconstructed by controllers if it's lost for some reason.
+Baremetal Operator (BMO) is capable of reconstructing host status by re-inspecting
+the host, but it will force the already provisioned host to go through another
+round of provisioning, which eventually reboots the host after IPA steps are
+completed. To have the inspection data available in the target cluster and to
+avoid unnecessary provisioning after pivoting, we normally copy inspection data
```
namnx228, line 22:
```suggestion
avoid unnecessary provisioning after pivoting, we currently copy inspection data
```

fmuyassarov:
As @macaptain suggested offline, we could also use [immutable ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable), because we don't expect normally to patch inspection data. But, this means when going from `Provisioned` -> `Available`, operator will not be able to clean up the inspection data, so then the only way to get ready for the next provisioning is to delete and re-create the ConfigMap. This might work when a user deletes the BareMetalHost, but not when only deprovisions. Because, when deprovisioning we don't delete the ConfigMap, but only update it (that's how I was expecting, but perhaps this is not necessarily).

Or maybe we should create ConfigMap only when provisioning started ( `Available` -> `Provisioned`), and delete it when deprovisioning kicks off. However, I'm not sure if we would be missing something in between. Maybe someone can share some ideas here?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+
+The life of a ConfigMap is tightly coupled to the corresponding BareMetalHost.
+Creating or deleting a BareMetalHost will result in creation or deletion of a
+ConfigMap respectively.
+
+**Note:** [cross-namespace owner references are not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications). Thus, ConfigMap and BareMetalHost
+must be in the same namespace.
+
+- BareMetalHost is created
+- Operator creates a ConfigMap with the same name and namespace as BareMetalHost
```
zaneb, line 73: I don't think it's safe to require the same name here, because ConfigMaps can be used for anything; this could easily collide with something the user is doing.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+    and sets ownerReferences
+- Operator writes name of the ConfigMap to `spec.hardwareInfo` of the BareMetalHost
```
zaneb, line 75: This kind of contradicts the previous step. If the name is always the same there is no need to write it in the Spec.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+- Provisioning requested, operator fills out the ConfigMap with the host
+    inspection data
+- Deprovisioning requested, operator cleans up the inspection data from the
+    ConfigMap
+- BareMetalHost deletion requested, operator deletes the ConfigMap and BareMetalHost
+
+#### BareMetalHost API changes
+
+- `status.hardware` field should be removed ⚠️
```
zaneb, line 84: We cannot do this without bumping the BMH CRD version.
We need to get to a point soon where we can do that, but in the meantime it would be good to have a plan for what to do in the meantime (populate both?).

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+### Test Plan
+
+Unit tests should be added for new functions and integration test scripts
+should be updated.
+
+### Upgrade / Downgrade Strategy
+
+From a user perspective, this feature doesn't introduce any breaking changes
+to existing deployments, except when there is reliance on the `status.hardware`
+field from external code base.
```
zaneb, line 120: I don't agree here. Removing the HardwareDetails field is definitely a breaking change, and not populating it any more would be a breaking change for consumers of the API.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Alternatives
+
+- Use Secret instead of ConfigMap. However, inspection data doesn't contain
+    confidential information, so there is no benefit we would get by having
+    the data encoded.
+- Create a new Custom Resource type, and write the data into its spec. Since
+    there is no control loop specifically on the inspection data and we only
+    write, read or delete it, it seems redundant to introduce a new CRD for
+    this purpose when Kubernetes offers built in objects like ConfigMap/Secret.
```
zaneb, line 138: I disagree here. The value of a CRD is not that you can write a controller to reconcile it (you can write a controller to reconcile anything, including ConfigMaps), it's that the schema is documented, verified, and versioned, and that specific RBAC rules can be applied to it.

I think defining a new CRD that puts the hardware details in the Spec instead of the Status is the correct solution. (This also solves the name collision issue, which also makes it easier to roll out in v1alpha1.)

I documented how I think this should work in [the API decomposition proposal](https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit#heading=h.vb3zlo2h9t7).

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,148 @@
...
+to existing deployments, except when there is reliance on the `status.hardware`
+field from external code base.
+
+### Version Skew Strategy
+
+I can not foresee any version skew issues.
+
+## Drawbacks
+
+None. Is there anything you can imagine?
```
zaneb, line 128: Loss of schema information, inability to set RBAC independently of other ConfigMaps.

zaneb (commented):
This needs to be a new CRD imho.

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: @zaneb thanks a lot for the quick review. I agree now that it would make more sense to have a new CRD instead of ConfigMap, which was something I didn't have in my mind while designing. 
If we go with a CRD and keep the hardwareDetails for some time (not to break current users and give enough time for others to prepare), is there some other blocker you can see? I can re-write the proposal, and request another review and after addressing comments, can we right away start implementation or do we have some major plans with regard to API change that this work should hold on?

zaneb: I don't think there's a blocker. We have added a bunch of new CRDs without bumping the version. Removing the existing HardwareDetails would be a blocker, but I think the answer is just to copy them from the CRD if it is available into the Status, and vice-versa, but exclude it from the status annotation. That way it survives the pivot (via the CRD), doesn't break the status annotation size limit, and presents the same behaviour for existing API users.

fmuyassarov: Awesome. Thanks again, I will ping once I updated the proposal.

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: thanks, fixed

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Updated accordingly.

fmuyassarov:
@zaneb Please take a look. Updated the proposal.

Is the naming of the CRD `InspectionData` okay ? I didn't want to name it HardwareDetails to avoid messing up with existing the struct of it. Any other suggestions are welcome:)

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
...
+type InspectionDataStatus struct {
+    // Last time the status was updated/observed.
+    // +optional
+    LastUpdated *metav1.Time `json:"lastUpdated,omitempty"`
+}
+```
+
+### Risks and Mitigations
+
+Any suggestions?
```
fmuyassarov, line 135: maybe I should mention potential security issue here and the need for some RBAC or securing the cluster?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
...
+
+Unit tests should be added. Integration test doesn't require any change.
+
+### Upgrade / Downgrade Strategy
+
+From a user perspective, this feature doesn't introduce any breaking changes
+to existing deployments, because `status.hardware` field will stay for the
+time being and this feature is only an addition to the current APIs. For now,
+there will be two copies of inspection data available, in the `spec.hardware`
+of InspectionData and `status.hardware` of BareMetalHost.
```
zaneb, line 159: We should specify somewhere above how we plan to keep them in sync: i.e. copy one to the other if only one exists, new resource takes precedence if they are different.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Version Skew Strategy
+
+I can not foresee any version skew issues.
```
zaneb, line 163: I think the strategy is doing a bi-directional copy to keep the BaremetalHost and the new CR in sync. That will prevent any skew issues.

design/baremetal-operator/configmap_for_host_inspection_data.md
zaneb: I don't know that there's any change in the security implications (if anything it's better).

One risk could be the fact that anybody upgrading an existing metal³ deployment will now need to have installed this CRD and the relevant RBAC permissions.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
+# inspectionData custom resource for host inspection data
+
+## Status
+
+implementable
+
+## Summary
+
+Introduce InspectionData Custom Resource (CR) to store host inspection data.
```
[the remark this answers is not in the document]
fmuyassarov: yup, sounds good to me :+1:

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
...
+respectively. Deprovisioning of the BareMetalHost will result in the inspection
+data removal from the spec of InspectionData.
+
+**Note:** [cross-namespace owner references are not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications). Thus, InspectionData
+and BareMetalHost must be in the same namespace.
+
+- BareMetalHost is created
+- Operator creates a InspectionData with the same name and namespace as
+    BareMetalHost and sets ownerReferences
+- Provisioning is requested, operator fills out the InspectionData with the
```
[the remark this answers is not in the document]
fmuyassarov: My bad, :+1:

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
...
+
+3. `clusterctl.cluster.x-k8s.io/move` label on InspectionData CRD. This will
+    not move child objects as in option 1.
+
+### Implementation Details/Notes/Constraints
+
+The life of an InspectionData is tightly coupled to a BareMetalHost. Creating or
+deleting a BareMetalHost will result in creation or deletion of a InspectionData
+respectively. Deprovisioning of the BareMetalHost will result in the inspection
+data removal from the spec of InspectionData.
```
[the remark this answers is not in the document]
fmuyassarov: > I don't think we should bother to create the CR until we do inspection. 

When we create a BareMetalHost, it automatically goes through inspection state until it reaches available right? If yes, why should not we create the CR whenever BareMetalHost is created? 

> We definitely must not remove the inspection data on deprovisioning. 

Sorry my bad, I had a wrong impression that we delete the inspection data when deprovisioning. Though, related to immutability point above, if the CR is immutable, and if someone requests inspection (due to some hardware change) of `Available` BMH, we can't update the CR. So, then we delete and re-create it ?
 
> We should add a finalizer on the CR to prevent it being deleted while a Host is provisioned. More details on how I think the lifecycle should work: https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit#heading=h.vb3zlo2h9t7

+1

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,186 @@
...
+}
+```
+
+The status of InspectionData:
+
+```go
+type InspectionDataStatus struct {
+    // Last time the status was updated/observed.
+    // +optional
+    LastUpdated *metav1.Time `json:"lastUpdated,omitempty"`
```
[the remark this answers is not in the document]
fmuyassarov: I'm +1 on immutability point and based on that it makes sense to drop `LastUpdated`, because there is no possibility anyways updating the CR due to immutability.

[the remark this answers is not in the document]
fmuyassarov: ow you meant with RBAC, I was rather thinking to do it with ValidationgWebhook.

design/baremetal-operator/configmap_for_host_inspection_data.md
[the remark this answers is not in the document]
zaneb: > > I don't think we should bother to create the CR until we do inspection.
> 
> When we create a BareMetalHost, it automatically goes through inspection state until it reaches available right? If yes, why should not we create the CR whenever BareMetalHost is created?

Eventually, but there are a couple of states to go through first. If we get stuck in e.g. Registration there is probably no reason to make it look like we've already done inspection by creating the CR. Also, there's an annotation you can add to skip inspection.

> > We definitely must not remove the inspection data on deprovisioning.
> 
> Sorry my bad, I had a wrong impression that we delete the inspection data when deprovisioning. Though, related to immutability point above, if the CR is immutable, and if someone requests inspection (due to some hardware change) of `Available` BMH, we can't update the CR. So, then we delete and re-create it ?

Yes. I think this will make implementation easier too.

> > We should add a finalizer on the CR to prevent it being deleted while a Host is provisioned. More details on how I think the lifecycle should work: https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit#heading=h.vb3zlo2h9t7
> 
> +1

design/baremetal-operator/configmap_for_host_inspection_data.md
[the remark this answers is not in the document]
zaneb: We can do both :)

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Yup, thank you very much for suggestion

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Good point. I shall include it.

fmuyassarov:
@zaneb thank you very much for suggestions and reviews.
I think all the comments are now addressed, please take a look 
once more and let me know if there is something else need to be
clarified.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+host status by re-inspecting the host, but it will force the already provisioned
+host to go through another round of provisioning, which eventually reboots the
+host after IPA steps are completed. To have the inspection data available in the
+target cluster and to avoid unnecessary provisioning after pivoting, we currently
+copy inspection data to baremetalhost.metal3.io/status annotation so that BMO in
+the target cluster can continue reconciliation from a point it was before pivoting,
+avoiding redundant provisioning.
+
+However, there is a limitation with the current approach, because when inspection
+data is quite big (e.g. when the host has 1000 multipath disks), it is easy to hit
```
Xenwar, line 27:
```suggestion
data is very large (e.g. when the host has 1000 multipath disks), it is easy to hit
```

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+Introduce HostData Custom Resource (CR) to store host inspection data.
+
+## Motivation
+
+Currently inspection data is written to baremetalhost status subresource. When
+pivoting to a target cluster, with *`clusterctl move`*, status subresource is
+not copied with the host, because from Cluster-API perspective, it's controllers'
+job to rebuild the status in the target cluster. Also, according to the [Kubernetes API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), status is expected to be reconstructed by controllers if
+it's lost for some reason. Baremetal Operator (BMO) is capable of reconstructing
```
Xenwar, line 17:
```suggestion
it's lost for any reason. Baremetal Operator (BMO) is capable of reconstructing
```

fmuyassarov:
Thanks for the review @Xenwar , addressed comments.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Motivation
+
+Currently inspection data is written to baremetalhost status subresource. When
+pivoting to a target cluster, with *`clusterctl move`*, status subresource is
+not copied with the host, because from Cluster-API perspective, it's controllers'
+job to rebuild the status in the target cluster. Also, according to the [Kubernetes API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), status is expected to be reconstructed by controllers if
+it's lost for any reason. Baremetal Operator (BMO) is capable of reconstructing
+host status by re-inspecting the host, but it will force the already provisioned
+host to go through another round of provisioning, which eventually reboots the
```
kashifest, line 19: I would say another round of inspection

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+host after IPA steps are completed. To have the inspection data available in the
+target cluster and to avoid unnecessary provisioning after pivoting, we currently
+copy inspection data to baremetalhost.metal3.io/status annotation so that BMO in
+the target cluster can continue reconciliation from a point it was before pivoting,
```
kashifest, line 23: BMO can reconstruct the BMH in the target cluster from the annotation with the appropriate inspection data and as such the reconciliation of the BMH can continue from the same status as it was in the source cluster.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+avoiding redundant provisioning.
+
+However, there is a limitation with the current approach, because when inspection
+data is very large (e.g. when the host has 1000 multipath disks), it is easy to hit
+the global limit of `metadata.annotations` [262144](https://github.com/kubernetes/kubernetes/blob/master/test/integration/controlplane/synthetic_controlplane_test.go#L313) bytes and fail to pivot successfully.
```
kashifest, line 28:
```suggestion
the global limit of `metadata.annotations` [262144](https://github.com/kubernetes/kubernetes/blob/master/test/integration/controlplane/synthetic_controlplane_test.go#L313) bytes and once the limit is hit the move process will fail to pivot BMH successfully.
```

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Goals
+
+- Store host inspection data into Spec HostData CR
```
kashifest, line 32: this line is not clear to me.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+- Adjust pivoting workflow to restore inspection data from HostData
+- Make the HostData CR immutable
+
+### Non-Goals
+
+Remove `status.hardware` from BareMetalHost CR
```
kashifest, line 38: Until this point I am a bit confused between hostData and inspectionData. If we are only moving `status.hardware` then we are not moving the other parts of baremetalhost.Status data and would that mean that move process has not to take care of the status.hardware from the new CRD and then the rest of the status from annotation as before?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Proposal
+
+### User Stories
+
+As an operator, from an ephemeral cluster I would like to pivot/move to a target
+cluster successfully.
```
kashifest, line 45: pivot what? Shall we be more specific since this proposal isn't really targeting the whole of the pivot/move process

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+However, there is a limitation with the current approach, because when inspection
+data is very large (e.g. when the host has 1000 multipath disks), it is easy to hit
+the global limit of `metadata.annotations` [262144](https://github.com/kubernetes/kubernetes/blob/master/test/integration/controlplane/synthetic_controlplane_test.go#L313) bytes and fail to pivot successfully.
+
+### Goals
+
+- Store host inspection data into Spec HostData CR
+- Adjust pivoting workflow to restore inspection data from HostData
+- Make the HostData CR immutable
```
kashifest, line 34: How would it work in case of re-introspection when some new hardware is added for example?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+under `status.hardware` of BareMetalHost.
+
+### Pivoting
+
+There are three options one can choose to pivot, objects excluded from Cluster
+API chain, to a target cluster:
+
+1. [`clusterctl.cluster.x-k8s.io/move-hierarchy`](https://github.com/kubernetes-sigs/cluster-api/blob/0e6428151f2a388b98d1e220a698b85940a5f290/cmd/clusterctl/api/v1alpha3/labels.go#L38) label should be set
+    on BareMetalHost CRD before issuing *`clusterctl move`* so that the entire hierarchy
+    (i.e. including HostData, Secrets and ConfigMaps) is pivoted.
```
kashifest, line 63: I am a bit unsure still if thats the right label although it seems from the link you shared it is. This would need some testing

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+2. `clusterctl.cluster.x-k8s.io=""` and `cluster.x-k8s.io/provider="metal3"`
+    labels on HostData CRD to extend the list of provider objects.
```
kashifest, line 66: My initial testing suggests only `clusterctl.cluster.x-k8s.io=""` label is enough but there isn't any harm on having `cluster.x-k8s.io/provider` label. But then it should be set to `infrastructure-metal3` instead of `metal3`.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+3. `clusterctl.cluster.x-k8s.io/move` label on HostData CRD. This will
```
kashifest, line 68: what are the child objects hostData could have?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+Deprovisioning of the BareMetalHost must not delete the HostData. It can
+be deleted either when BareMetalHost is being deleted or when re-inspection
+is requested. To ensure HostData availability while BareMetalHost is provisioned,
+a finalizer should be set on the HostData to prevent its accidental deletion.
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData object doesn't exist, operator will first try to copy it from the
+BareMetalHost status (if available) and vice versa. If not available in neither
```
kashifest, line 92: This point is not clear. Once hostData exists  will it be available in BMH status anymore?

design/baremetal-operator/configmap_for_host_inspection_data.md
kashifest: I think  I got the answer down below.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+of them, new inspection will be enforced.
+
+#### Keeping inspection data in sync
+
+As mentioned above, we will have two copies of inspection data available. One in the Spec
+of HostData and the other in the Status of BareMetalHost. One of them will get its inspection
+data copied from the other. Thus, inspection data can be copied from HostData (if available)
+to BareMetalHost, or the vice versa. If there is delta in the inspection data while
+copying, HostData should take precedence and get copied to the Status of BareMetalHost.
```
kashifest, line 101: I would really make a sync between them whenever the pause annotation is applied so that we dont have any mismatched status before and after pivot.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+- Add unit tests
+- Update documentation
+
+### Dependencies
+
+None
+
+### Test Plan
+
+Unit tests should be added. Integration test doesn't require any change.
```
kashifest, line 158: The pivot test which is a part of the integration test should have a change right?

kashifest (commented):
Thanks for the writeup. Few questions inline. I am still a bit unsure about how are you planning to do the pivot? Through annotation or through HostData. At the end you have mentioned, there will be just the same two copies one in status and one in HostData. But what would be stored in status.annotation in BMH and would it not affect the move process since as I have mentioned the move of bmh has to know reconstruct bmh status where one piece is in annotation and the other in HostData CR?

fmuyassarov:
> Thanks for the writeup. Few questions inline. I am still a bit unsure about how are you planning to do the pivot? Through annotation or through HostData. At the end you have mentioned, there will be just the same two copies one in status and one in HostData. But what would be stored in status.annotation in BMH and would it not affect the move process since as I have mentioned the move of bmh has to know reconstruct bmh status where one piece is in annotation and the other in HostData CR?

1. To have successful pivot we will use HostData
2. Status annotation will be dropped. 
2. Inspection data will not be reconstructed from the status,annotation but from HostData

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: nope, it doesn't change the way clusterctl pivot works nor the way we trigger the pivoting

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: if the BMH status doesn't have inspection data, but hostData has, then we will copy it from hostData to the status

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: hostData itself is a child object to BMH. So, if you set it to BMH, hostData will not be moved.

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: how did it work for us until now where we were using `metal3` in metal3-dev-env ? Or was it still ok because it was not necessary ?

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: sorry, but what exactly ?

fmuyassarov: ah I see, I am missing `of`

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: I think it is also important to let people know that the host will be rebooted, which is I think when host is provisioning?

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Ah, sorry if you got confused from the PR description, we decided to rename inspectionData to HostData

design/baremetal-operator/configmap_for_host_inspection_data.md
kashifest: yes true

kashifest:
One more point, I would like to make is once HostData is introduced, is there any need to replicate the same information in BMH status anymore? Would that break anything for the existing workflow? I am concerned because whenever you have  two copies of the same information, there is always a possibility that one will lag behind. May be @zaneb you will know more how this will affect if we keep the hardwareDetails in HostData CR only?

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: So, apparently neither infrastructure-metal3 nor metal3 is needed, because of the ownerRefrence chain. So, I will remove the second label.

fmuyassarov:
> One more point, I would like to make is once HostData is introduced, is there any need to replicate the same information in BMH status anymore? Would that break anything for the existing workflow? I am concerned because whenever you have two copies of the same information, there is always a possibility that one will lag behind. May be @zaneb you will know more how this will affect if we keep the hardwareDetails in HostData CR only?

As we discussed offline, the reason for still keeping that info in the status is for backwards compatibility.

fmuyassarov:
So I have tested it in https://github.com/metal3-io/baremetal-operator/pull/1022, where I restored only `status.hardware` & `status.Provisioning` and that was enough, but only `status.hardware` wasn't :(.

fmuyassarov:
> So I have tested it in [metal3-io/baremetal-operator#1022](https://github.com/metal3-io/baremetal-operator/pull/1022), where I restored only `status.hardware` & `status.Provisioning` and that was enough, but only `status.hardware` wasn't :(.

So, it seems that we need to add more staff into the hostData, `status.Provisioning` to be exact. Any thoughts or objections on this @kashifest @zaneb ?

zaneb:
I don't think this change will be sufficient to get rid of the status annotation altogether, the goal should be just to strip out the hardware details from it.
Moving any of the Provisioning status stuff to a new CRD is problematic since so much relies on the Status subresource being updated atomically.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData object doesn't exist, operator will first try to copy it from the
+BareMetalHost status (if available) and vice versa. If not available in neither
+of them, new inspection will be enforced.
+
+#### Keeping inspection data in sync
+
+As mentioned above, we will have two copies of inspection data available. One in the Spec
```
dtantsur, line 97: Maybe I missed it above, but I don't quite understand why keep two copies of a quite large structure.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+None
+
+### Test Plan
+
+Unit tests should be added. Integration test doesn't require any change.
+
+### Upgrade / Downgrade Strategy
+
+From a user perspective, this feature doesn't introduce any breaking changes
+to existing deployments, because `status.hardware` field will stay until is
```
dtantsur, line 163: Okay, this is why we may want to keep the old location... Do we need a plan to deprecate it? We should do it before the BMH API reaches beta.

zaneb: Yes, our next (first) API version bump should delete this field from the BMH.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Implementation Details/Notes/Constraints
+
+The life of HostData is dependent on BareMetalHost. When BareMetalHost is entering
+inspection state, the HostData will be created. Failure in reaching inspection state
+or having `inspect.metal3.io: disabled` annotation on BareMetalHost will force the
+operator to not create the HostData. Deleting a BareMetalHost will result in deletion
+of a HostData respectively.
+
+HostData should be immutable and this will be controlled via [Validating Webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) and with RBAC permissions.
```
s3rj1k, line 79: Can this part be optional?
Running without `Validating Webhook` or with custom `Validating Webhook` should not break HostData CR code.

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+
+None
+
+## Alternatives
+
+- Use k8s built in objects like ConfigMap/Secret. From security point of view, although,
+    it is possible to restrict permissions to edit ConfigMaps (only to those
+    storing inspection data) based on the name (i.e. `resourceNames`), but it will add
+    complexity to the operator.
+- Increase metadata global limit. However, this does not always work as it is hard
```
s3rj1k, line 183: Adding filters to filter-out unneeded inspection data?

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: You are right, and I removed other alternatives not to complicate it here.

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Is it so that you want to see the CR to be mutable and if yes may I ask why?

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Thanks for suggestion, but I'm afraid that even if we add a filter to ignore unnecessary fields we still will hit the issue, because the size of the unnecessary fields is relatively small and instead it is `status.hardware` that we care about and which might grow a lot depending on the hardware.

fmuyassarov:
@kashifest @zaneb @dtantsur @s3rj1k 
I would appreciate if you can take a look once more. I have update some bits. Mainly:
- we don't drop the status annotation
- status annotation will store all the status data of BMH as now, but excluding `status.hardware ` which will be in the new CR
- when pivoted, BMO will reconstruct the host from new CR (here it will get inspection data or `status.hardware` part) and the rest (`status.Provisioning`, `status.poweredOn` and etc) from the status annotation it [currently does](https://github.com/metal3-io/baremetal-operator/blob/c22cb7ba6df61e7a4b3bae751f3b36b0ae825f1c/controllers/metal3.io/baremetalhost_controller.go#L142).

design/baremetal-operator/configmap_for_host_inspection_data.md
s3rj1k: No, was pointing out that someone could have custom `Validating Webhook` already in place, this could break immutability assumption.
One use case for custom web-hook is mutating webhook, to filter out some of unnecessary data from HostData CR.

So we should probably rephrase `HostData should be immutable` to something like `HostData should be protected from unintentional change by Validating Webhook`

design/baremetal-operator/configmap_for_host_inspection_data.md
s3rj1k: valid point

fmuyassarov:
@dtantsur @zaneb 
If you don't any more comments on this, can we merge the proposal?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+Deprovisioning of the BareMetalHost must not delete the HostData. It can
+be deleted either when BareMetalHost is being deleted or when re-inspection
+is requested. To ensure HostData availability while BareMetalHost is provisioned,
+a finalizer should be set on the HostData to prevent its accidental deletion.
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData object doesn't exist, operator will first try to copy it from the
+BareMetalHost status (if available) and vice versa. If not available in neither
+of them, new inspection will be enforced.
```
andfasano, line 91: A finalizer can always be manually patched to force a deletion at any time. This could generate a situation where an already provisioned (or any other state after inspection) host could suddenly loose the inspection data. Forcing re-inspection could become tricky (and this could be more problematic when the field will be removed from the host status)

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+- BareMetalHost is in inspecting state
+- Operator creates a HostData with the same name and namespace as BareMetalHost, fills out
+    the HostData with the host inspection data and sets ownerReferences
+- Provisioning is requested
+- Pivoted to a target cluster, operator restores host inspection data from HostData and
+    the rest from status annotation (e.g. `status.Provisioning`, `status.poweredOn` and etc)
+- Deprovisioning is requested
+- BareMetalHost deletion requested, operator deletes the HostData and BareMetalHost
+
+The spec of HostData:
```
andfasano, line 114: It could be useful to have the complete to CR, ie to see also the ownerRef

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+- Add unit tests
+- Update documentation
+
+### Dependencies
+
+None
+
+### Test Plan
+
+Unit tests should be added. Integration test doesn't require any change.
```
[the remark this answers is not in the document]
fmuyassarov: AFAIK we don't yet have e2e in BMO. If there is one in the future I can add it, but perhaps for now this can be continued to be tested in daily master jobs (Jenkins) that use metal3-dev-env code base?

design/baremetal-operator/configmap_for_host_inspection_data.md
```diff
@@ -0,0 +1,193 @@
...
+CAPM3 machine controller will no longer copy `status.hardware` of BareMetalHost
+to the annotation. Because that part will be avialable in the form of HostData CR
+in the target cluster.
+
+### Implementation Details/Notes/Constraints
+
+The life of HostData is dependent on BareMetalHost. When BareMetalHost is entering
+inspection state, the HostData will be created. Failure in reaching inspection state
+or having `inspect.metal3.io: disabled` annotation on BareMetalHost will force the
+operator to not create the HostData. Deleting a BareMetalHost will result in deletion
```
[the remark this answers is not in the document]
fmuyassarov: Good point, I will also mention it in the proposal so that I don't forget it during implementation.

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: That's right. However, I think in general we give cluster access to someone we trust. For example, accidental deletion of BMH can also happen if the cluster access is not secured. So, perhaps we will need to clearly mention that in the user document as potential cause of not securing the cluster?

Also, that's how Cluster API workflow is also designed. For example, if someone intentionally/accidentally deletes Cluster object, or Machine, then it will result in other objects deletion/deprovisioning on the chain.

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: @andfasano yes that's right. This double source of inspection data management will get much easier in the future when we drop the `status.hardware` from BMH.

fmuyassarov:
@andfasano thank you for your review. I've addressed/replied to your comments and added example CR + mentioned about future removal of `status.hardware` from the BMH.

design/baremetal-operator/configmap_for_host_inspection_data.md
andfasano: More than a security problem I think that probably the `new inspection will be enforced` phrase could be detailed a little bit more. In general, if the HostData deletion will happen after the `Inspecting` and within the `Available` state, the host could be forced back to the `Inspecting` without any particular issue. Instead, if it happens during a `Provisioned` state, as you also already mentioned, this could  trigger a deprovisioning first

design/baremetal-operator/configmap_for_host_inspection_data.md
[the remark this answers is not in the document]
andfasano: That's right, in any case I think there could be some value to define, even in a pure natural language, which e2e test could make sense (or it's reasonable) to have for validating/accepting the new feature - so that maybe it could be easier in future to implement them having previously defined in the design doc. I wasn't pointing to an exhaustive or super-detailed list, just a few of them

design/baremetal-operator/configmap_for_host_inspection_data.md
andfasano: I'm wondering at this point if it could be better to remove the `vice versa` part, in order to also to prepare the ground for the future drop of `status.hardware` - I understand that this could mean a more complex implementation as of now.

andfasano:
> @andfasano thank you for your review. I've addressed/replied to your comments and added example CR + mentioned about future removal of `status.hardware` from the BMH.

Thanks @fmuyassarov please see my reply on the `status.hardware` point [here](https://github.com/metal3-io/metal3-docs/pull/208#discussion_r773191847)

design/baremetal-operator/configmap_for_host_inspection_data.md
fmuyassarov: Sure, I'm more than happy to consider hostData as the main source of truth for the inspection data. As you mentioned, while `status.hardware` still exists, it can serve as a backup for the hostData only when hostData needs to be recreated. 

I remember @zaneb suggesting making it clear how we keep hostData and `status.hardware` of BMH in sync. So, I hope @zaneb would also agree on this?
To clarify the idea is:
- HostData is the main source of inspection data;
- Inspection data in the hostData can be populated/copied from BMH status only if hostData is missing
- Until we remove the `status.hardware` from BMH, inspection data can be populated/copied from the hostData to the status of BMH in case it is missing. This part is temporary, and will be removed during the API bump.

andfasano: As also mentioned in a later [comment](https://github.com/metal3-io/metal3-docs/pull/208#discussion_r773191847), a new problem to handle is how to manage the case where HD suddenly disappears (maybe due a deletion). In any case I would not use the BMH status to regenerate it, either because HD will be the single source of truth, either because it be removed in future

fmuyassarov:
@dtantsur @zaneb @andfasano PTAL. I've updated the proposal slightly based on the discussion on [slack discussion](https://kubernetes.slack.com/archives/CHD49TLE7/p1640186190294700).

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+
+As an operator, from an ephemeral cluster I would like to pivot/move all the Cluster
+API, CAPM3 and BMO objects to a target cluster successfully.
+
+## Design Details
+
+Apart from writing hardware details into the status of BareMetalHost, write it to a
+HostData CR spec. BareMetalHost and HostData will be parent and child objects
+respectively, and this relationship is controlled via ownerReferences chain. As such,
+the same name & namespace will be used for HostData as its owning object BareMetalHost.
```
andfasano, line 56: nit: `and` instead of `&`

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+be deleted either when BareMetalHost is being deleted or when re-inspection
+is requested. To ensure HostData availability while BareMetalHost is provisioned,
+a finalizer should be set on the HostData to prevent its accidental deletion.
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData is missing, operator acts differently depending on the current
+state of the BareMetalHost.
+
+- host is provisoned: enter error state
```
andfasano, line 100: typo: `provisioned`

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+- host is available:
+    1. copy inspection data from the status of BareMetalHost if exists;
```
andfasano, line 102: I'm still not convinced about point 1), and think we should keep just 2). It could have a couple of benefits:
a) HD cr will be effectively the only trusted source for haedware details
b) When in future the `status.hardware` field will be removed, the user will not observe a different behavior for that scenario.
Anyhow, I'd like to hear the other opinions on that.

fmuyassarov: @zaneb I would like to hear your opinion too on this since you also suggested earlier about bi-directional copy in https://github.com/metal3-io/metal3-docs/pull/208#discussion_r742196355.

zaneb: I'd have thought we'd need this for backward compatibility, at least initially. CAPM3 currently stores the hardwaredetails in the status annotation during a pivot. OpenShift installer does something similar (for control plane nodes). I don't know if there are any other third parties relying on the status annotation, but if we don't use it when it is available and the resource is not, we're requiring all of them to implement support for the new CRD before we have even landed the code to use it.

andfasano: Ok for backward compatibility. @fmuyassarov maybe in the implementation PR we could add a specific log trace when performing such back copy (bmh.status -> hd) in case of missing resource for warning the user, for clarity (even though here in the design doc it's mentioned that in future only re-inspection will be the viable option).

fmuyassarov: @andfasano +1 for that. And when we are going remove the status completely, we can clean up the tracking also accordingly.

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+[hardware](https://github.com/metal3-io/baremetal-operator/blob/05d12b6768a9989a9a4e61dad6cd1f9e84a6e078/apis/metal3.io/v1alpha1/baremetalhost_types.go#L721)
+field from the status of BareMetalHost in the first version bump of the API.
+
+### Pivoting
+
+To pivot objects excluded from Cluster API chain to a target cluster, one can set
+`clusterctl.cluster.x-k8s.io=""` label on HostData and BareMetalHost CRDs.
+Current status annotation based pivoting workflow will stay as it is. However,
+CAPM3 machine controller will no longer copy `status.hardware` of BareMetalHost
+to the annotation. Because that part will be avialable in the form of HostData CR
```
zaneb, line 74: nit: typo in 'available'

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData is missing, operator acts differently depending on the current
+state of the BareMetalHost.
+
+- host is provisioned: enter error state
+- host is available:
+    1. copy inspection data from the status of BareMetalHost if exists;
+    2. re-inspect the host but only when provisioning is requested if inspection
```
zaneb, line 103: Do we wait for provisioning to be requested for reinspection now? I would have thought that if reinspection is requested and we are in an allowed state for it (i.e. available) we should do it straight away.

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+As mentioned above, we will have two copies of inspection data available for the
+time being. One in the Spec of HostData and the other in the Status of BareMetalHost.
+In the future, when new API version of BareMetalHost is introduced, `status.hardware`
+will be removed from BareMetalHost. As such, missing HostData for available BareMetalHost
+will result in another inspection before starting provisioning.
+
+**Note:** [cross-namespace owner references are not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications). Thus, HostData
+and BareMetalHost must be in the same namespace.
+
+- BareMetalHost is created
```
zaneb, line 116: This list could use some sort of heading/intro

andfasano:
Thanks @fmuyassarov for the latest updates/comments, which clarified my previous points. I think that now the proposal looks in good shape.

/approve

zaneb (commented):
/approve

kashifest:
/hold

design/baremetal-operator/hostdata_crd.md
zaneb: See my other comment: https://github.com/metal3-io/metal3-docs/pull/208#discussion_r783177703

If the host is in available, what does it matter if we start inspection? The worst that it can do is cause a delay until you can provision again, but since we're re-inspecting before provisioning anyway then that's still going to happen. (In fact it's better to inspect sooner, because it will make provisioning faster later.)

In practice it will be handled by step 1 (copy hardwaredetails from the host status) anyway.

design/baremetal-operator/hostdata_crd.md
```diff
@@ -0,0 +1,232 @@
...
+be deleted either when BareMetalHost is being deleted or when re-inspection
+is requested. To ensure HostData availability while BareMetalHost is provisioned,
+a finalizer should be set on the HostData to prevent its accidental deletion.
+
+Only operator will be granted with create permissions for HostData and everyone
+else will be given no write permissions. If the BareMetalHost was inspected but
+HostData is missing, operator acts differently depending on the current
+state of the BareMetalHost.
+
+- host is provisioned: enter error state
```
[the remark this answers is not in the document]
fmuyassarov: Well not having HostData will break pivoting at least, in a sense that all the hosts will be re-provisioned. So target cluster will be off for some time.

design/baremetal-operator/hostdata_crd.md
fmuyassarov: Agreed. Updated the documents. Thanks.

design/baremetal-operator/hostdata_crd.md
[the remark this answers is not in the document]
andfasano: @fmuyassarov I think you mentioned a fair point. Once published, CRs data can be consumed also by other operators. The decomposition is bringing up a scenario where part of the data could be lost (something not possible before), so the owner should take any action to reconstruct them when they are missing. And if it's not possible to immediately rebuild them (as in the case under discussion, an host already provisioned), the owner should notify somehow such condition.

kashifest (approved):
/lgtm

design/baremetal-operator/hostdata_crd.md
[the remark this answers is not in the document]
fmuyassarov: Hi @andfasano . I will just write here also about the discussion we had on the community meeting with @zaneb today.

Since normally operator is going to add a finalizer on the HostData, it won't be deleted by an accident until BareMetalHost still exists. It can happen only internationally by the user, who deletes the finalizer manually to able to delete the HostData. And if that's the case, then it is user's fault as Zane pointed out. So, if user deleted HostData when it should not be, then it is responsibility of the user to recover the cluster before proceeding for example with the pivoting. Otherwise, I think we don't have any corner case that would result in the HostData lost while BMH is provisioned. Perhaps we can add an event for such a scenario just for the sake of the user notification.

As such, I've updated this part of the document. Please let me know if you have any case in your mind. Thanks.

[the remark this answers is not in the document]
andfasano: The use case I had in mind was when the resource could be deleted unintentionally or due a bug, maybe by another operator (and not necessarily by a human being). Even though the finalizer can greatly reduce such risk, that's a chance theoritically possible thanks to to the decomposition (while before, with a single CR, it was a  "all or nothing" scenario), and ignoring completely such event doesn't appear to me a good choice.
From the previous discussions I realized that in case of an already provisioned host it may be not desired to re-trigger an unwanted inspection, so it seems fully reasonable for me the idea of at least raising an event to properly inform the users (or any other potential automated listener) in such case.

PR #239, opened 2022-01-31
Correct comunity meeting time information

Rozzii:


README.md
```diff
@@ -31,7 +31,7 @@ come talk to us!
 * Subscribe to the [Metal³ Development Mailing List](https://docs.google.com/document/d/1d7jqIgmKHvOdcEmE2v72WDZo9kz7WwhuslDOili25Ls/edit)
   for the project related anouncements, discussions and questions.
 * Come and meet us in our weekly community meetings on every
-  Wednesday at 13:00 UTC on [Zoom](https://zoom.us/j/97255696401?pwd=ZlJMckNFLzdxMDNZN2xvTW5oa2lCZz09)
+  Wednesday at 14:00 UTC on [Zoom](https://zoom.us/j/97255696401?pwd=ZlJMckNFLzdxMDNZN2xvTW5oa2lCZz09)
```
dtantsur, line 34: Should we maybe be honest and say that the time is not bound to UTC, but to central european time?

dtantsur:
/approve

but a question inline

README.md
Rozzii: Well, I am not sure that changing from UTC to EET would give us any benefit, I know the majority of the developers working on the project are living in EET but not all of them and who knows what the future holds for the project in this regard. 

I have checked and the k8s community is using UTC https://www.kubernetes.dev/community/community-groups/ so I would say let's stay aligned with them but I that is just my opinion. 

We could also discuss this on the next community meeting if you'd like to do so.

Rozzii: Although a combination of EET and EEST could be useful as UTC is not following daylight saving procedure so we would need to adjust it all the time, so actually I will change my opinion I support writing something like:

That the meeting  is at EET 16:00   From  2:00 a.m. on the second Sunday of March until 2:00 a.m. on the first Sunday of November and then it is EEST 16:00 .

furkatgofurov7 (commented):
LGTM

README.md
zaneb: I believe the current calendar invitation is denominated in (US) EST.
Every time DST changes in either the US or EU, everyone gets confused and the meeting splits for a week or two, then a new calendar invitation goes out with a different time zone.
We need to pick one time zone (UTC or otherwise) and stick to it, advertise that here, and in the meeting invitation. (This would be a good topic for tomorrow's meeting.)

Issue #220, opened 2022-01-14
User-guide: add doc to briefly describe Node Reuse feature in CAPM3


fmuyassarov:
Add a document briefly explaining the Node Reuse feature in CAPM3.
Describe:
- briefly Node Reuse feature in CAPM3
- why we need it
- how to use it
- what are the caveats 

/kind documentation
/help

PR #246, opened 2022-02-16
📖 User guide: Add nodeReuse feature document

furkatgofurov7:
Fixes: #220

### DIFF
--- a/docs/user-guide/src/capm3/node_reuse.md
+++ b/docs/user-guide/src/capm3/node_reuse.md
@@ -1,0 +1,50 @@
+# Node Reuse
+
+This feature brings a possibility of re-using the same BaremetalHosts (referred to as a host later)
+during deprovisioning and provisioning mainly as a part of the rolling upgrade process in the cluster.
+
+## Importance of scale-in strategy
+
+The logic behind the reusing of the hosts, solely relies on the **scale-in** upgrade strategy utilized by
+Cluster API objects, namely [KubeadmControlPlane](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy) and MachineDeployment.
+During the upgrade process of above resources, the machines owned by KubeadmControlPlane or MachineDeployment are
+removed one-by-one before creating new ones (delete-create method).
+That way, we can fully ensure that, the intended host is reused when the upgrade is kicked in (picked up on the following provisioning for the new machine being created).
+
+**Note:** To achieve the desired *delete first and create after* behavior in above-mentioned Cluster API objects,
+user has to modify:
+
+* MaxSurge field in KubeadmControlPlane and set it to 0 with minimum number of 3 control plane machines replicas
+* MaxSurge and MaxUnavailable fields in MachineDeployment set them to 0 & 1 accordingly
+
+On the contrary, if the scale-out strategy is utilized by CAPI objects during the upgrade, usually create-swap-delete
+method is followed by CAPI objects, where new machine is created first and new host is picked up for that
+machine, breaking the node reuse logic right at the beginning of the upgrade process.
+
+## Workflow
+
+Metal3MachineTemplate (M3MT) Custom Resource is the object responsible for enabling of the node reuse feature.
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: Metal3MachineTemplate
+metadata:
+  name: test1-controlplane
+  namespace: metal3
+spec:
+  nodeReuse: True
+  template:
+    spec:
+      image:
+      ...
+```
+
+There could be two Metal3MachineTemplate objects, one referenced by KubeadmControlPlane for control plane nodes, and the other by MachineDeployment for worker node. Before performing an upgrade, user must set `nodeReuse` field to **true** in the desired Metal3MachineTemplate object where hosts targeted to be reused. If left unchanged, by default, `nodeReuse` field is set to **false** resulting in no host reusing being performed in the workflow. If you would like to know more about the internals of controller logic, please check the original proposal for the feature [here](https://github.com/metal3-io/metal3-docs/blob/main/design/cluster-api-provider-metal3/node_reuse.md)
+
+Once `nodeReuse` field is set to **true**, user has to make sure that scale-in feature is enabled as suggested above, and proceed with updating the desired fields in KubeadmControlPlane or MachineDeployment to start a rolling upgrade.
+
+**Note:** If you are creating a new Metal3MachineTemplate object (for control-plane or worker), rather than using the existing one
+created while provisioning, please make sure to reference it from the corresponding Cluster API object (KubeadmControlPlane or MachineDeployment). Also keep in mind that, already provisioned Metal3Machines were created from the old Metal3MachineTemplate
+and they consume existing hosts, meaning even though `nodeReuse` field is set to **true** in the new Metal3MachineTemplate,
+it would have no effect. To use newly Metal3MachineTemplate in the workflow, user has to reprovision the nodes, which
+should result in using new Metal3MachineTemplate referenced in Cluster API object and Metal3Machine created out of it.
--- a/docs/user-guide/src/SUMMARY.md
+++ b/docs/user-guide/src/SUMMARY.md
@@ -22,7 +22,7 @@
     - [Features]()
         - [Automated cleaning API]()
         - [Remediation](capm3/remediaton.md)
-        - [Node Reuse]()
+        - [Node Reuse](capm3/node_reuse.md)
         - [Pivoting](capm3/pivoting.md)
 - [Ip-address-manager]()
     - [Install Ip-address-manager]()


docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,45 @@
...
+Cluster API Provider Metal3 (CAPM3) Machine controller performs the following:
+
+During deprovisioning:
+
+1. Finds M3MT referenced by Metal3Machine (M3M) and checks if nodeReuse is enabled;
+1. If enabled, it checks Cluster API (CAPI) machine object role (control-plane or worker), and based on that, fetches
+   either KubeadmControlPlane (KCP) or MachineDeployment (MD) name accordingly;
+1. Sets fetched object's name on the corresponding host's labels field.
+
+On the following provisioning:
```
Xenwar, line 18:
```suggestion
During the next provisioning:
```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,45 @@
...
+  **<infrastructure.cluster.x-k8s.io/node-reuse: kcp-object-name>** where after node reuse label, a prefix is
+  appended depending on the owner of the host in question (kcp or md) to avoid mismatch, in cases where KCP and
+  MD object names are the same.
+1. Chooses a host from the filtered list of reused hosts for provisioning.
+
+## Importance of scale-in strategy
+
+The logic behind the reusing of the hosts, solely relies on the scale-in upgrade strategy utilized by
+CAPI objects, i.e KCP or MachineDeployment, where during the upgrade process, the machines owned
+(control-plane or worker) are removed one-by-one before creating a new ones (delete-create method).
```
Xenwar, line 30:
```suggestion
(control-plane or worker) are removed one-by-one before creating new ones (delete-create method).
```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,45 @@
...
+That way, we can fully ensure that, once the upgrade is kicked in, the intended host is reused
+during deprovisioning and the same host is picked up on the following provisioning for the new machine
+being created.
+
+**Note:** To achieve the desired *delete first and create after* like behavior in above-mentioned CAPI objects,
+user has to modify:
+
+* MaxSurge field and set it to 0 during KCP upgrade with minimum of 3 replicas of control plane machines in the
+  cluster, so that control plane machines are removed one-by-one before creating a new ones;
```
Xenwar, line 39:
```suggestion
  cluster, so that control plane machines are removed one-by-one before creating new ones;
```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,45 @@
...
+
+Metal3MachineTemplate (M3MT) CRD is the object to support enabling/disabling node reuse feature. When set to **true**,
+Cluster API Provider Metal3 (CAPM3) Machine controller performs the following:
+
+During deprovisioning:
+
+1. Finds M3MT referenced by Metal3Machine (M3M) and checks if nodeReuse is enabled;
+1. If enabled, it checks Cluster API (CAPI) machine object role (control-plane or worker), and based on that, fetches
+   either KubeadmControlPlane (KCP) or MachineDeployment (MD) name accordingly;
+1. Sets fetched object's name on the corresponding host's labels field.
```
mboukhalfa, line 16: Would be better to explain the label format here: <infrastructure.cluster.x-k8s.io/node-reuse: kcp-object-name>

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,45 @@
...
+
+On the following provisioning:
+
+1. Filters out all hosts with node reuse label set and those that store KCP/MD object name, in a format of i.e:
+  **<infrastructure.cluster.x-k8s.io/node-reuse: kcp-object-name>** where after node reuse label, a prefix is
+  appended depending on the owner of the host in question (kcp or md) to avoid mismatch, in cases where KCP and
+  MD object names are the same.
+1. Chooses a host from the filtered list of reused hosts for provisioning.
+
```
mboukhalfa, line 25: Can we project the steps on an example ?

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,45 @@
+# Node Reuse
+
+This feature brings a possibility of re-using the same BaremetalHosts (referred to as a host later)
+during deprovisioning/provisioning mainly as part of the rolling upgrade process in the cluster.
```
mboukhalfa, line 4: Is this part trying to answer _why we need it?_  just thinking that we might need to emphasize that need for using the same external storage

docs/user-guide/src/capm3/node_reuse.md
furkatgofurov7: I would keep it as is since it is already detailed and more than that can be found in the original proposal of the feature.

docs/user-guide/src/capm3/node_reuse.md
furkatgofurov7: I am not sure if I got the idea, can you elaborate more?

docs/user-guide/src/capm3/node_reuse.md
furkatgofurov7: What is the external storage in this context, do you mean m3 object, i.e m3mt?

fmuyassarov (commented):
Thanks @furkatgofurov7 for the document. This looks good. Can we also explain what is the need for the future at the beginning?  Workflow part seems to be internal to the logic of the controller, I would rather try to avoid how controller works behind the scenes and perhaps add a reference to another document which explains it.

docs/user-guide/src/capm3/node_reuse.md
mboukhalfa: I mean as a necessity that node reuse cover. as we have an external storage attached to a node we prefer reusing the same node and keep the external storage I guess that is one of the main reason to implement nodereuse, isn't it?

furkatgofurov7: Yeah, but that is a motivation of the feature which is described in detail in the original proposal https://github.com/metal3-io/metal3-docs/blob/main/design/cluster-api-provider-metal3/node_reuse.md#motivation.

furkatgofurov7:
> Thanks @furkatgofurov7 for the document. This looks good. Can we also explain what is the need for the future at the beginning? Workflow part seems to be internal to the logic of the controller, I would rather try to avoid how controller works behind the scenes and perhaps add a reference to another document which explains it.

Done.

furkatgofurov7:
@mboukhalfa @fmuyassarov thanks for reviews, take a look again.

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
...
+
+This feature brings a possibility of re-using the same BaremetalHosts (referred to as a host later)
+during deprovisioning and provisioning mainly as part of the rolling upgrade process in the cluster.
+
+## Importance of scale-in strategy
+
+The logic behind the reusing of the hosts, solely relies on the **scale-in** upgrade strategy utilized by
+Cluster API objects, namely [KubeadmControlPlane](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy) and MachineDeployment, where during the upgrade process, the machines owned (control-plane or worker) are removed one-by-one before creating new ones (delete-create method).
+That way, we can fully ensure that, once the upgrade is kicked in, the intended host is reused during deprovisioning and the
+same host is picked up on the following provisioning for the new machine being created.
```
fmuyassarov, line 11:
```suggestion
Cluster API objects, namely [KubeadmControlPlane](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy) and MachineDeployment, where during the upgrade process, the machines owned (control-plane or worker) are removed one-by-one before creating new ones (delete-create method).
That way, we can fully ensure that, once the upgrade is kicked in, the intended host is reused (picked up on the following provisioning for the new machine being created).
```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
...
+spec:
+  nodeReuse: True
+  template:
+    spec:
+      image:
+      ...
+```
+
+There are two Metal3MachineTemplate objects, one referenced by KubeadmControlPlane which is used to create Metal3Machine owned by control plane machine, and the other one referenced by MachineDeployment and used to create Metal3Machine owned by worker machine. Before performing an upgrade, user must set `nodeReuse` field to **true** in the desired Metal3MachineTemplate object where hosts targeted to be reused. If left unchanged, by default, `nodeReuse` field is set to **false** meaning no host reusing performed in the workflow.
+
```
fmuyassarov, line 42:
```suggestion
There could be two Metal3MachineTemplate objects, one referenced by KubeadmControlPlane for control plane nodes, and the other by MachineDeployment for worker node. Before performing an upgrade, user must set `nodeReuse` field to **true** in the desired Metal3MachineTemplate object where hosts targeted to be reused. If left unchanged, by default, `nodeReuse` field is set to **false** meaning no host reusing performed in the workflow.

```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
...
+Once `nodeReuse` field is set to **true**, user has to make sure that scale-in feature is enabled as suggested above, and go on with updating the desired fields in KubeadmControlPlane or MachineDeployment to start a rolling upgrade.
```
fmuyassarov, line 43:
```suggestion
Once `nodeReuse` field is set to **true**, user has to make sure that scale-in feature is enabled as suggested above, and goes on with updating the desired fields in KubeadmControlPlane or MachineDeployment to start a rolling upgrade.
```

fmuyassarov (commented):
Looks good, few more comments left. Thanks
Do you think it is worth of mentioning how user can not set or how can he set nodeReuse on a single Metal3Machine for partial upgrade?

furkatgofurov7:
> Do you think it is worth of mentioning how user can not set or how can he set nodeReuse on a single Metal3Machine for partial upgrade?

This should be already covered I believe, please check https://github.com/metal3-io/metal3-docs/pull/246/files#diff-9308f30303c82fd797e41762f7883864811a32fddde69c8f070b3ec869249f5fR40, let me know if this needs more clarification.

docs/user-guide/src/capm3/node_reuse.md
mboukhalfa: I was thinking if an example that applies those steps on a list of bmh would simplify understanding the node reuse workflow

furkatgofurov7: @mboukhalfa having a controller logic re-explained on this document felt over-burden, so I have completely removed the controller logic from the user guide and provided a reference to it, thus I have removed the points we are discussing here

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
+# Node Reuse
+
+This feature brings a possibility of re-using the same BaremetalHosts (referred to as a host later)
+during deprovisioning and provisioning mainly as part of the rolling upgrade process in the cluster.
```
smoshiur1237, line 4:
```suggestion
during deprovisioning and provisioning as a part of the rolling upgrade process in the cluster.
```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Importance of scale-in strategy
+
+The logic behind the reusing of the hosts, solely relies on the **scale-in** upgrade strategy utilized by
+Cluster API objects, namely [KubeadmControlPlane](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy) and MachineDeployment, where during the upgrade process, the machines owned (control-plane or worker) are removed one-by-one before creating new ones (delete-create method).
```
smoshiur1237, line 9: Can you please break this long sentence? It would be easier for user to understand.

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
...
+That way, we can fully ensure that, once the upgrade is kicked in, the intended host is reused (picked up on the following provisioning for the new machine being created).
```
smoshiur1237, line 10:
```suggestion
In this way, we can ensure the intended host is reused when upgrade is kicked in (picked up on the following provisioning for the new machine being created).
```

docs/user-guide/src/capm3/node_reuse.md
```diff
@@ -0,0 +1,49 @@
...
+
+**Note:** To achieve the desired *delete first and create after* like behavior in above-mentioned Cluster API objects,
```
smoshiur1237, line 12:
```suggestion
**Note:** To achieve the desired *delete first and create after*  behavior in above-mentioned Cluster API objects,
```

smoshiur1237 (changes requested):
Thanks for the writing. Some comment and suggestions are added.

smoshiur1237 (approved):
Thanks for addressing the request. Looks good from my side.
/lgtm

kashifest (approved):

PR #249, opened 2022-02-25
Proposal: Failure Domain (FD) Support for Control-Plane Nodes

Arvinderpal:
Related Issue:  https://github.com/metal3-io/cluster-api-provider-metal3/issues/402

### DIFF
--- a/design/fd-support-kcp.md
+++ b/design/fd-support-kcp.md
@@ -1,0 +1,155 @@
+# Failure Domain (FD) Support for Control-Plane Nodes
+
+## Summary
+
+CAPI supports failure domains (FDs) for control-plane nodes via the KCP
+controller. CAPI will look for the set of FailureDomains in the
+`ProviderCluster.Spec.FailureDomain`, if specified, the field will be copied
+to the `Cluster.Status.FailureDomains`. KCP will choose an FD from this set
+and place its value in `Machine.Spec.FailureDomain`. Currently, KCP tries
+to balance Machines evenly across all FDs.
+
+It is expected that providers will use this chosen FD (in the Machine.Spec)
+in deciding where to place the infrastructure machine. In the case of
+metal3, CAPM3 should associate the Metal3Machine with the corresponding BMH
+in the desired FD.
+
+In public clouds FailureDomains are well defined. For example, in Azure, FDs
+map to availability zones within an azure region. For baremetal, there is no
+such mapping. However, an operator of a baremetal/metal3 cluster may still
+place their hosts across FDs in their data center or region.
+For example, two racks, connected to independent power grids, may
+constitute two FDs. Labels placed on the BMHs, by the operator, can provide a
+means to distinguish between hosts belonging to different FDs.
+
+The proposal is for CAPM3 to choose hosts based on the FD selected by KCP.
+In other words, based on the value in `Machine.Spec.FailureDomain` CAPM3 will
+choose a BMH matching the label corresponding to that FD.
+
+## User Stories
+
+As an operator who has placed their baremetal infrastructure across different
+FDs, I would like CAPM3 to associate control-plane nodes with BMHs from the
+desired FD.
+
+## Goals
+
+* Support selection of BMHs based on the desired FD for the control plane node.
+
+## Non-Goals
+
+* Failure Domains support for worker nodes (i.e. MachineDeployments).
+
+## Proposal
+
+The operator labels the BMH resource with a standard FD label. For
+example, the following standard label could be used on the BMH:
+
+```diff
+  infrastructure.cluster.x-k8s.io/failure-domain=my-fd-1
+```
+
+Today, CAPM3 chooseHost() func associates a Metal3Machine with a specific BMH
+based on labels specified in `Metal3Machine.Spec.HostSelector.MatchLabels`.
+We can expand this capability. The HostSelector field is used to narrow down
+the set of available BMHs that meet the selection criteria. When FDs are
+being utilized (i.e. `Machine.Spec.FailureDomain` != nil and Machine is for a
+control-plane node), we insert the above label into `HostSelector.MatchLabels`.
+This should, in effect, narrow down the set of available BMHs to only those
+in the desired FD. Additional labels in the set could further narrow down the
+set of usable BMHs.
+
+Note that if no FD is specified in the `Machine.Spec.FailureDomain`, then the
+chooseHost() will default to it's current behavior.
+
+### Example Scenario
+
+Consider the following scenario. An operator has FDs: fd-1 and fd-2. These are
+specified in Metal3Cluster:
+
+```yaml
+kind: Metal3Cluster
+spec:
+  failureDomains:
+    fd-1:
+      controlPlane: true
+    fd-2:
+      controlPlane: true
+...
+```
+
+When a control-plane node needs to be created, KCP will choose an FD from this set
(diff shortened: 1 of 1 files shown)


design/fd-support-kcp.md
```diff
@@ -0,0 +1,111 @@
...
+
+* Failure Domains support for worker nodes (i.e. MachineDeployments).
+
+## Proposal
+
+The operator labels the BMH resource with a standard FD label. For
+example, the following standard label could be used on the BMH:
+
+```diff
+  infrastructure.cluster.x-k8s.io/failure-domain=my-fd-1
```
zaneb, line 49: Is this a limitation of the CAPI Failure Domains model, that the failure domain is single-valued?

In the real world failure domains are hierarchical in nature - sometimes you want to spread hosts across multiple datacenters and sometimes just across adjacent racks. (See e.g. @jaypipes [here](http://lists.openstack.org/pipermail/openstack-dev/2014-March/031106.html) and "Partitioning and failure domains" on [this page](https://www.joinfu.com/2018/08/project-mulligans-architecture/).) To handle that in this design you'd need something like a separate controller (HWCC?) to check multiple labels and calculate the failure domain label. So it would be worth considering some kind of selector here, or allowing the user to choose (at the cluster level) the name of the label to match values for.

Arvinderpal: Yes, at this point, CAPI only supports flat FDs. 

Kubernetes has the concept of zones and regions. Regions are composed of zone and thus hierarchical -- [K docs on this topic](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion). A zone is essentially a failure domain within a region. While there is no strict definition of a region, major cloud providers consider regions to be geographically distributed. Additionally, creating a single K cluster that extends across regions is strongly discouraged. 

From a baremetal point of view, we can achieve hierarchy through the use of regions and zones -- for example, a region may represent a DC and a zone a rack (as in your example). In this scenario, a cluster most certainly can extend across "regions".  Unfortunately, this does not align with the cloud provider definition. 

In the near to mid-term, I think we'll only have zones to work with. Longer-term, there is a discussion to be had with the CAPI community in coming up with something that works well for both cloud and baremetal providers. 
By limiting ourselves to flat FDs, we can still mimic some hierarchy -- for example, as described in the [link that you gave](http://lists.openstack.org/pipermail/openstack-dev/2014-March/031106.html), you could make smaller FDs: rack1 -DC1, rack2-DC1, rack1-DC2,  rack2-DC2, etc… Not ideal but would be better than having no support. 

Also, please note that this proposal is only for control-plane nodes, so the initial scope is pretty limited.

zaneb: OK, if the limitation comes from CAPI then there is probably no current need in trying to design anything more complex. Thanks!

design/fd-support-kcp.md
```diff
@@ -0,0 +1,111 @@
...
+In public clouds FailureDomains are well defined,. For example, in Azure, FDs
+map to availability zones within an azure region. For baremetal, there is no
+such mapping. However, an operator of a baremetal/metal3 cluster may still
+place their hosts across FDs in their data center or region.
+For example, two racks, connected to independent power grids, may
+constitue two FDs. Labels placed on the BMHs, by the operator, can provide a
+means to distinguish between hosts belonging to different FDs.
+
+The proposal is for CAPM3 to choose hosts based on the FD selected by KCP.
+In other words, based on the value in `Machine.Spec.FailureDomain` CAPM3 wil
```
lentzi90, line 26:
```suggestion
In other words, based on the value in `Machine.Spec.FailureDomain` CAPM3 will
```

lentzi90 (commented):
LGTM, just a typo below

design/fd-support-kcp.md
```diff
@@ -0,0 +1,111 @@
...
+nodes, and zone labels are mostly applicable for worker nodes, we discuss
+some options here for operators who wish to have these labels applied to
+their control-plane nodes.
+
+### BMH Label Sync
+
+CAPM3 supports a [label sync mechanism](https://github.com/metal3-io/metal3-docs/blob/main/design/sync-labels-bmh-to-node.md)
+between the BMH and its corresponding Kubernetes Node. The sync controller
+could be configured to sync zone labels
+(e.g.`topology.kubernetes.io/zone=my-fd-1`) placed on the BMH to the
```
furkatgofurov7, line 88: would this work out of the box using label sync controller or I am not sure what you meant by could be configured?

design/fd-support-kcp.md
```diff
@@ -0,0 +1,111 @@
...
+
+Today, CAPM3 chooseHost() func associates a Metal3Machine with a specific BMH
+based on labels specified in `Metal3Machine.Spec.HostSelector.MatchLabels`.
+We can expand this capability. The HostSelector field is used to narrow down
+the set of available BMHs that meet the selection criteria. When FDs are
+being utilized (i.e. `Machine.Spec.FailureDomain` != nil and Machine is for a
+control-plane node), we insert the above label into HostSelector.MatchLabels.
+This should, in effect, narrow down the set of available BMHs to only those
+in the desired FD. Additional labels in the set could further narrow down the
+set of usable BMHs.
```
furkatgofurov7, line 60: Once the standard FD label is set in BMH and desired BMH is chosen, do we remove the FD label by the controller or it is the operator's/admin's responsibility? Also, it would be very nice and easy to follow the steps if we could include example CR representations that are used in the workflow in the proposal (e.g. BMH, Machine)

furkatgofurov7 (commented):
@Arvinderpal thanks for the proposal and my apologies for being late on this review. Some questions inline and also ptal at design template we use https://github.com/metal3-io/metal3-docs/blob/main/design/_template.md in metal3 and add some missing paragraphs (Work Items/Test Plan) to the proposal.

design/fd-support-kcp.md
Arvinderpal: @furkatgofurov7 In this section on "FD labels on the Node Obj", I'm mentioning how information about what FD a Node belongs to can be bubbled up to the Node object. I'm doing this for completeness.

The label sync mechanism we have implemented can be configured to sync this zone label. The specifics on how to do this are in the linked proposal.

design/fd-support-kcp.md
Arvinderpal: @furkatgofurov7 The "standard label" should remain on the BMH as long as the host it represents stays within the specified FD. I don't see the point it removing it. 
I'll add snippets of CRs as you asked.

Arvinderpal: Please see "### Example Scenario" section below.

zaneb:
I have no objections, but also no time to do a thorough review.

Arvinderpal:
Can we get consensus on this? 

We have no means today of distributing control-plane nodes across failure-domains. This is major blocker for clusters in production environments that require HA. 

For workers, we have the option to use a MachineDeployment (and therefore a Metal3MachineTemplate) per failure-domain. This is just not possible with KCP. 

Also, the change I'm proposing is quite small. TL;DR Update `chooseHost()` func in CAPM3 to associate control-plane Machines with BMHs based on the value Machine.Spec.FailureDomain. 

@furkatgofurov7 @lentzi90 @zaneb

lentzi90 (approved):

Rozzii:
Based on community meeting discussion this proposal PR has reached Lazy consensus so anyone with approval rights can approve!

PR #250, opened 2022-03-01
🐛Fix all-owners.sh script and update list based on output

furkatgofurov7:
This PR ~~fixes **yq** syntax error while running `all-owners.sh` and~~ introduces  `branch` var to differentiate repository branches, since we have still few repositories with the `master` branch as default, few transitioned to `main` and only _metal3-io.github.io_ is tracked using `source` branch as default. 
ALL-OWNERS file has been updated accordingly as per the script output. 

**Note:** I have gone through all repositories listed in the script and checked the maintainers with the one we have now auto-generated by script in the ALL-OWNERS file and it looks correct. Please let me know if the output is missing someone in the list. Thanks!

### DIFF
--- a/maintainers/ALL-OWNERS
+++ b/maintainers/ALL-OWNERS
@@ -3,22 +3,24 @@
 
 approvers:
 - andfasano
+- ashughorla
 - bfournie
 - derekhiggins
-- digambar15
 - dtantsur
 - elfosardo
-- furkatgofurov7
 - fmuyassarov
+- furkatgofurov7
 - hardys
 - hroyrh
+- imain
 - iurygregory
 - kashifest
 - knowncitizen
-- mhrivnak
 - ogelbukh
+- ravipwaghmare
 - russellb
 - stbenjam
+- xenwar
 - zaneb
 
 emeritus_approvers:
--- a/maintainers/all-owners.sh
+++ b/maintainers/all-owners.sh
@@ -25,8 +25,17 @@ all_owners_raw() {
     else
       filter='.approvers'
     fi
-    curl -s "https://raw.githubusercontent.com/metal3-io/$repo/main/OWNERS" | \
-      yq -y $filter | \
+    if [ "$repo" = "hardware-classification-controller" ] ||  [ "$repo" = "ironic-client" ] \
+      || [ "$repo" = "ironic-hardware-inventory-recorder-image" ] || [ "$repo" = "metal3-dev-env" ] \
+      || [ "$repo" = "metal3-helm-chart" ] || [ "$repo" = "static-ip-manager-image" ]; then
+      branch='master'
+    elif [ "$repo" = "metal3-io.github.io" ]; then
+      branch='source'
+    else
+      branch='main'
+    fi
+    curl -s "https://raw.githubusercontent.com/metal3-io/$repo/$branch/OWNERS" | \
+      yq $filter | \
       grep -v "null" | \
       grep -v "\.\.\."
   done
--- a/maintainers/all-owners.sh
+++ b/maintainers/all-owners.sh
@@ -35,7 +35,7 @@ all_owners_raw() {
       branch='main'
     fi
     curl -s "https://raw.githubusercontent.com/metal3-io/$repo/$branch/OWNERS" | \
-      yq $filter | \
+      yq -y $filter | \
       grep -v "null" | \
       grep -v "\.\.\."
   done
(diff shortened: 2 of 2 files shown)


maintainers/ALL-OWNERS
```diff
@@ -3,22 +3,25 @@
...
 - elfosardo
-- furkatgofurov7
 - fmuyassarov
+- furkatgofurov7
 - hardys
 - hroyrh
+- imain
 - iurygregory
 - kashifest
 - knowncitizen
```
fmuyassarov, line 19: I @knowncitizen still in Metal3? Probably worth of checking and updating here and in metal3-io.github.io repo accordingly

maintainers/ALL-OWNERS
```diff
@@ -3,22 +3,25 @@
 
 approvers:
 - andfasano
+- ashughorla
 - bfournie
 - derekhiggins
-- digambar15
+- dhellmann
```
fmuyassarov, line 9: I think @dhellmann has moved to another project, not sure if this addition is correct. I can see his GitHub handle in some GitHub projects of Metal3.io, but those projects are either forgotten to be updated or archived.

dhellmann: It's OK to leave me out of the list, since I'm not really active with the project these days.

kashifest: Isn't it so that this file is auto-generated using a script? I would rather check if any repo still has @dhellmann in the OWNERS file, rather than manually removing it from here.

dhellmann: That makes sense. A quick check only showed me still in 1 repo: https://github.com/metal3-io/ironic-agent-image/pull/18

furkatgofurov7: > Isn't it so that this file is auto-generated using a script? I

Correct, the list has been updated with the output of the script. I would also support removing people from the corresponding repos if needed and then updating the list in this PR. 



> That makes sense. A quick check only showed me still in 1 repo: [metal3-io/ironic-agent-image#18](https://github.com/metal3-io/ironic-agent-image/pull/18)

Thanks @dhellmann

maintainers/ALL-OWNERS
furkatgofurov7: @fmuyassarov I can see @knowncitizen in metal3-io.github.io repo approvers list, not sure what else I have to check?

furkatgofurov7:
/hold
until https://github.com/metal3-io/ironic-agent-image/pull/18 is merged and I will update this after.

furkatgofurov7:
> until [metal3-io/ironic-agent-image#18](https://github.com/metal3-io/ironic-agent-image/pull/18) is merged and I will update this after.

/hold cancel

maintainers/all-owners.sh
```diff
@@ -25,7 +25,16 @@ all_owners_raw() {
...
-    curl -s "https://raw.githubusercontent.com/metal3-io/$repo/main/OWNERS" | \
+    if [ "$repo" = "hardware-classification-controller" ] ||  [ "$repo" = "ironic-client" ] \
+      || [ "$repo" = "ironic-hardware-inventory-recorder-image" ] || [ "$repo" = "metal3-dev-env" ] \
+      || [ "$repo" = "metal3-helm-chart" ] || [ "$repo" = "static-ip-manager-image" ]; then
+      branch='master'
+    elif [ "$repo" = "metal3-io.github.io" ]; then
+      branch='source'
+    else
+      branch='main'
+    fi
```
andfasano, line 36: Small tip: in order to slightly improve the maintenance (ie avoid re-touching this file when a repo branch is migrated), instead of the hard-coded checks, what about using something like `git ls-remote` to determine if a branch exists or not? Ie: 

```git ls-remote -q --exit-code --heads https://github.com/metal3-io/$repo main``` 

will return `0` if `main` was found, `2` otherwise. Different branches could be tested (`main`, `master`, `source`) to discover the suitable one

furkatgofurov7: @andfasano thanks for the great suggestion, I will take it. Please take a look now.

furkatgofurov7:
Going through the repos under Metal3 org excluding the archived ones once again, I noticed we are missing [mariadb-image repo](https://github.com/metal3-io/mariadb-image/blob/main/OWNERS) in the REPOS list in the script, I will add it to the list if there are no objections.

dtantsur:
> I will add it to the list if there are no objections.

Yeah, it should be there. But its content should now be equivalent to ironic-image.

/approve

furkatgofurov7:
> Yeah, it should be there. But its content should now be equivalent to ironic-image.

@dtantsur PTAL: https://github.com/metal3-io/metal3-docs/pull/253

PR #253, opened 2022-03-04
Include mariadb-image repo to the list of repos in all-owners.sh script

furkatgofurov7:
As per https://github.com/metal3-io/metal3-docs/pull/250#issuecomment-1056964421 adding mariadb-image repo to complete the list of repos which is used to fetch all owners from that list.


fmuyassarov (commented):
/lgtm

PR #259, opened 2022-04-12
User-guide: move BMO images to its specific dir

fmuyassarov:
We should keep project related images on their own dir to ease the maintenance in the future.
/cc @dtantsur @kashifest


furkatgofurov7 (commented):
/lgtm

kashifest (approved):

PR #271, opened 2022-07-08
📖 User-guide: add contributing section with services and contacts

lentzi90:
The idea is to expand this with "ops routines" for all the listed services with periodic tasks that needs to be performed. For example, check the daily CI jobs or patch and reboot the BML jumphosts.

### DIFF
--- a/docs/user-guide/src/contributing/introduction.md
+++ b/docs/user-guide/src/contributing/introduction.md
@@ -1,0 +1,11 @@
+# Contributing to Metal³
+
+This section is dedicated to the various ways you can contribute to Metal³.
+There are many things outside of normal code contributions that you can do.
+Here are some examples:
+
+- Report issues when you find them
+- Write documentation for missing areas
+- Fix documentation that has been outdated
+- Help taking care of the CI system
+- Do presentations or demos about Metal³
--- a/docs/user-guide/src/contributing/services_contacts.md
+++ b/docs/user-guide/src/contributing/services_contacts.md
@@ -1,0 +1,118 @@
+# Ops routines for the Metal³ project
+
+Every project has some services that it depends on for running tests, publishing a web page or building container images for example.
+Metal³ is no exception.
+In this section of the book we try to gather information about these services, who runs them, and how to keep them healthy and secure.
+
+## What services are we responsible for
+
+We need to know which services we use that we are running our selves (i.e. we are responsible for keeping them operational).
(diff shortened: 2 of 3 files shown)


docs/user-guide/src/contributing/services_contacts.md
```diff
@@ -0,0 +1,118 @@
...
+
+We use [Quay](https://quay.io) to store the container images for all projects.
+
+### Bare metal lab hardware contact
+
+Contact: Ian Kumlien (<ian.kumlien@ericsson.com>)
+
+Note this is for hardware issues.
+We are responsible for the software.
+See the [BML wiki](https://wiki.nordix.org/display/CPI/Bare+Metal+Lab).
```
fmuyassarov, line 118: I am not sure if this is needed, because it is needed only for the team that manages the BML but not for everyone

docs/user-guide/src/contributing/services_contacts.md
```diff
@@ -0,0 +1,118 @@
...
+Status page: <https://www.cnstatus.com/>
+
+Cleura (former CityCloud) is the cloud platform we run most of our things on.
+All the Jenkins jobs, Prow and all developer VMs are here.
+They basically donate the resources to us and the normal support email is available if we need to reach them.
+
+### Nordix services
+
+Contact: <discuss@lists.nordix.org>
+
```
fmuyassarov, line 95: Does nordix admins respond to requests made from someone outside of EST? Do we expect any community member to create request there?

lentzi90: Good point. I'm not sure how much we should expect the community to engage but it would be nice to prepare for the possibility that someone outside of EST/Nordix gets involved.

Anyway, the most important part for me is to have this documented somewhere. If that somewhere should be accessible to the community or just EST doesn't matter much to me. I see two main options:

1. [this PR] Make everything accessible to the community and figure out how people outside of EST can get involved. 
2. Make this doc internal to EST (for example on the Nordix wiki).

Let's discuss when people are back from vacation.

docs/user-guide/src/contributing/introduction.md
```diff
@@ -0,0 +1,11 @@
+# Contributing to Metal³
+
+This section is dedicated to the various ways you can contribute to Metal³.
+There are many things outside of normal code contributions that you can do.
+Here are some examples:
+
+- Report issues when you find them
```
dtantsur, line 7: This is so light on details, that I wonder if it makes sense to include this document. Realistically, all these bullet points require a guidance on how to do them.

lentzi90:
Closing in favor of an internal wiki page, since only Nordix members would have access to these anyway.

PR #274, opened 2022-08-01
baremetal boot from volume document

haoziwu:
This design document is for baremetal boot from remote volume,  please refer to [https://github.com/metal3-io/baremetal-operator/pull/1147](url).


Rozzii:
This design document is also related to :  #1146

PR #279, opened 2022-09-02
User guide: Add BMO External inspection

lentzi90:


docs/user-guide/src/bmo/external_inspection.md
```diff
@@ -0,0 +1,35 @@
...
+Why is this needed?
+
+- It allows avoiding an extra reboot for live-images that include their own inspection tooling.
+- It provides an arguably safer alternative to the status annotation in some cases.
+
+Caveats:
+
+- If both `baremetalhost.metal3.io/status` and `inspect.metal3.io/hardwaredetails` are specified on BareMetalHost creation, `inspect.metal3.io/hardwaredetails` will take precedence and overwrite any hardware data specified via `baremetalhost.metal3.io/status`.
+- If the BareMetalHost is in the `Ready` state the controller will not attempt to match profiles based on the annotation.
+  This is because the primary use-case is live-boot images (where no disk image is written to disk), and that profile matching is no longer the preferred interface for specifying root-disk hints.
```
Rozzii, line 35: I am not sure what do you mean by the statement that:
 "profile matching is no longer the preferred interface for specifying root-disk hints."
As far as I understands root device hints are supplied as part of the "spec" and they are independent from inspection and the hardware data visible in the hardware details. The root device hint matching happens inside the IPA so whatever is visible in hardware details  has no effect on what the IPA sees .

lentzi90: I must confess I don't know what it means actually. I just took it from the design doc [here](https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/external-introspection.md#hardware-status-update-proposal)

dtantsur: Profile matching is not the preferred interface for specifying root device hints any more :) Let's probably drop this sentence, it's a bit too specific and potentially outdated.

Rozzii: Yes I agree with dtantsur's, referencing some historical fact (how it worked before) is not relevant in this document and the wider sentence itself  feels strange.  I would drop the last two point in the list  so I would just leave this.
```suggestion
- If both `baremetalhost.metal3.io/status` and `inspect.metal3.io/hardwaredetails` are specified on BareMetalHost creation, `inspect.metal3.io/hardwaredetails` will take precedence and overwrite any hardware data specified via `baremetalhost.metal3.io/status`.
```

lentzi90: How do you mean @Rozzii ? I already dropped the part about profile matching. Should I remove the caveats completely?

docs/user-guide/src/bmo/external_inspection.md
```diff
@@ -0,0 +1,35 @@
...
+Why is this needed?
+
+- It allows avoiding an extra reboot for live-images that include their own inspection tooling.
+- It provides an arguably safer alternative to the status annotation in some cases.
+
+Caveats:
+
+- If both `baremetalhost.metal3.io/status` and `inspect.metal3.io/hardwaredetails` are specified on BareMetalHost creation, `inspect.metal3.io/hardwaredetails` will take precedence and overwrite any hardware data specified via `baremetalhost.metal3.io/status`.
+- If the BareMetalHost is in the `Ready` state the controller will not attempt to match profiles based on the annotation.
+  This is because the primary use-case is live-boot images (where no disk image is written to disk).
```
dtantsur, line 35: This is not the literal reason (which is: we never overwrite an already matched profile). Let's omit the reason completely.

lentzi90: Done

docs/user-guide/src/bmo/external_inspection.md
Rozzii: Yeah it is /lgtm from my part.

Rozzii (approved):

docs/user-guide/src/bmo/external_inspection.md
```diff
@@ -0,0 +1,34 @@
...
+
+Why is this needed?
+
+- It allows avoiding an extra reboot for live-images that include their own inspection tooling.
+- It provides an arguably safer alternative to the status annotation in some cases.
+
+Caveats:
+
+- If both `baremetalhost.metal3.io/status` and `inspect.metal3.io/hardwaredetails` are specified on BareMetalHost creation, `inspect.metal3.io/hardwaredetails` will take precedence and overwrite any hardware data specified via `baremetalhost.metal3.io/status`.
+- If the BareMetalHost is in the `Ready` state the controller will not attempt to match profiles based on the annotation.
```
mboukhalfa, line 34:
```suggestion
- If the BareMetalHost is in the `Available` state the controller will not attempt to match profiles based on the annotation.
```

docs/user-guide/src/bmo/external_inspection.md
```diff
@@ -0,0 +1,34 @@
...
+Here is an example with a BMH that has inspection disabled and is using the external inspection feature to add the HardwareDetails.
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: node-0
+  namespace: metal3
+  annotations:
+    inspect.metal3.io: disabled
```
mboukhalfa, line 19: Probably this annotation can be highlighted in the description I could not get if it is needed with `inspect.metal3.io/hardwaredetails` annotation or not ? because you mentioned that external  annotation is only controlled through `inspect.metal3.io/hardwaredetails`

mboukhalfa (commented):
Some notes here

docs/user-guide/src/bmo/external_inspection.md
lentzi90: Added a bit more about it above now. PTAL

kashifest (approved):

docs/user-guide/src/bmo/external_inspection.md
```diff
@@ -0,0 +1,34 @@
...
+    inspect.metal3.io/hardwaredetails: |
+      {"systemVendor":{"manufacturer":"QEMU", "productName":"Standard PC (Q35 + ICH9, 2009)","serialNumber":""}, "firmware":{"bios":{"date":"","vendor":"","version":""}},"ramMebibytes":4096, "nics":[{"name":"eth0","model":"0x1af4 0x0001","mac":"00:b7:8b:bb:3d:f6", "ip":"172.22.0.64","speedGbps":0,"vlanId":0,"pxe":true}], "storage":[{"name":"/dev/sda","rotational":true,"sizeBytes":53687091200, "vendor":"QEMU", "model":"QEMU HARDDISK","serialNumber":"drive-scsi0-0-0-0", "hctl":"6:0:0:0"}],"cpu":{"arch":"x86_64", "model":"Intel Xeon E3-12xx v2 (IvyBridge)","clockMegahertz":2494.224, "flags":["foo"],"count":4},"hostname":"hwdAnnotation-0"}
+spec:
+  ...
+```
+
+Why is this needed?
```
kashifest, line 26: I will probably highlight this text as a heading of subsection

docs/user-guide/src/bmo/external_inspection.md
```diff
@@ -0,0 +1,34 @@
...
+
+- It allows avoiding an extra reboot for live-images that include their own inspection tooling.
+- It provides an arguably safer alternative to the status annotation in some cases.
+
+Caveats:
```
kashifest, line 31: Same

kashifest (changes requested):

docs/user-guide/src/bmo/external_inspection.md
lentzi90: Hmm I see your point, but we have the same in other places. See e.g. reboot annotation: https://metal3io.netlify.app/bmo/reboot_annotation.html
Should we make them also headings then?

kashifest: IMO yes

lentzi90: Ok, I'll update and try to fix it in all places I can find :slightly_smiling_face:

kashifest: On a second thought lets do it in a separate commit

PR #281, opened 2022-09-07
Add separate OWNERS file for design docs

zaneb:
The approvers list for design docs is intentionally kept quite small, but we would like the freedom to add many more reviewers and approvers for documentation in general. By adding this OWNERS file with the no_parent_owners option set, we disconnect the design docs directory from the rest of the repo.

### DIFF
--- a/design/OWNERS
+++ b/design/OWNERS
@@ -1,0 +1,18 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+ - andfasano
+ - dtantsur
+ - russellb
+ - zaneb
+
+reviewers:
+ - furkatgofurov7
+
+emeritus_approvers:
+ - maelk
+ - hardys
+ - fmuyassarov
+
+options:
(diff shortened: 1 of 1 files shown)


zaneb:
> This is really good approach. So after this we would get rid of the root owners file right?

We should be able to keep the root one and it will apply to everything else in the repo except this directory.

kashifest:
> We should be able to keep the root one and it will apply to everything else in the repo except this.

So who would be in root owners after this file goes in?

zaneb:
This PR doesn't change the root OWNERS, but it does leave us free to add people to it without affecting the `design` directory approvers.

kashifest:
> This PR doesn't change the root OWNERS, but it does leave us free to add people to it without affecting the `design` directory approvers.

+1 . I don’t have right here it seems . But lgtm.

furkatgofurov7 (commented):
Thank you. 

/lgtm

/cc @dtantsur @andfasano

Issue #224, opened 2022-01-14
User-guide: add doc to briefly describe Live ISO feature in BMO


fmuyassarov:
Add a document briefly explaining the Live ISO feature in BMO.
Describe:
- briefly Live ISO feature in BMO
- why we need it
- how to use it
- what are the caveats if any

/kind documentation
/help

PR #284, opened 2022-09-21
User-guide: Add BMO live-iso

furkatgofurov7:
Fixes: https://github.com/metal3-io/metal3-docs/issues/224

### DIFF
--- a/docs/user-guide/src/bmo/live-iso.md
+++ b/docs/user-guide/src/bmo/live-iso.md
@@ -1,0 +1,36 @@
+# Live ISO
+
+The live-iso API in Metal3 allows booting a BareMetalHost with a live ISO image instead of writing an image to
+the local disk using the IPA deploy ramdisk.
+
+Why we need it?
+
+In some circumstances, i.e to reduce boot time for ephemeral workloads, it may be possible to boot an iso
+and not deploy any image to disk (saving the time to write the image and reboot). This API is also useful
+for integration with 3rd party installers distributed as a CD image, for example leveraging the existing
+toolchains like [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/)
+might be desirable.
+
+How to use it?
+
+Here is an example with a BareMetalHost CRD, where iso referenced by the url and `live-iso` set in DiskFormat will be
+live-booted without deploying an image to disk. Additionally, live ISO mode is supported with any
+virtualmedia driver when used as a BMC driver. Also, checksum options are not required in this case, and will be
+ignored if specified:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: live-iso-booted-node
+spec:
+  image:
+    url: http://1.2.3.4/image.iso
+    format: live-iso
+  online: true
+```
+
+**Note**: `rootDeviceHints`, `networkData` and `userData` will not be used
+since the image is not written to disk.
+
+For more details, please see the [design proposal](https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/bmh_live_iso.md).
--- a/docs/user-guide/src/SUMMARY.md
+++ b/docs/user-guide/src/SUMMARY.md
@@ -8,7 +8,7 @@
     - [Features]()
         - [Automated Cleaning](bmo/automated_cleaning.md)
         - [Automatic Secure Boot]()
-        - [Live ISO]()
+        - [Live ISO](bmo/live-iso.md)
         - [Detached annotation](bmo/detached_annotation.md)
         - [Status annotation](bmo/status_annotation.md)
         - [External Inspection]()


docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,40 @@
...
+
+Here is an example with a BaremetalHost CRD, where DiskFormat spec indicates that instead
+of deploying an image to disk, a live-iso image will be booted:
+
+```yaml
+spec:
+image:
+    url: http://1.2.3.4/image.iso
+    format: live-iso
+online: true
```
lentzi90, line 21:
```suggestion
  image:
    url: http://1.2.3.4/image.iso
    format: live-iso
  online: true
```

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,40 @@
...
+
+1. Please make sure that Ironic is configured to use the [ramdisk
+   deploy interface](https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html)
+
+```yaml
+enabled_deploy_interfaces = direct,ramdisk
+```
+
+2. Also, you need to set the provided ISO URL into the
+   [instance_info boot_iso field](https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#virtual-media-ramdisk)
```
lentzi90, line 38: If I understand correctly, this is what BMO will do for you behind the scenes, not something that the user has to do, so I would skip it. Or alternatively reword it so that it is clear that this is something BMO does and you don't have to think about it.

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,40 @@
+# Live-ISO
+
+The live-iso API in Metal3 allows to boot BaremetalHost with live-iso image instead of using IPA deploy ramdisk.
+
+Why we need it?
+
+In some circumstances, i.e to reduce boot time for ephemeral workloads, it may be possible to boot an iso
+and not deploy any image to disk (saving the time to write the image and reboot). Since this alternative
+is distributed as a live-iso, leveraging the existing toolchains like [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/) might be desirable. On top that, Ironic already exposes an [API](https://docs.openstack.org/ironic/latest/admin/ramdisk-boot.html) to support this.
```
lentzi90, line 9: I would skip the last sentence here since it is already obvious that Ironic must support it if BMO does. Or we could say something like "this is implemented using the ramdisk-boot API in Ironic" so the user can read more about it if interested.
```suggestion
is distributed as a live-iso, leveraging the existing toolchains like [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/) might be desirable.
```

furkatgofurov7: I got rid of it, thanks

lentzi90 (approved):
/lgtm

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,32 @@
+# Live-ISO
+
+The live-iso API in Metal3 allows to boot BaremetalHost with live-iso image instead of using IPA deploy ramdisk.
```
dtantsur, line 3: nit: BareMetalHost

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,32 @@
...
+spec:
+  image:
+    url: http://1.2.3.4/image.iso
+    format: live-iso
+  online: true
+```
+
+**Note**  `rootDeviceHints` and `userData` will not be used
+since the image won't be written to disk, and Ironic doesn't
+support passing user-data in addition to the iso attachment at present.
```
dtantsur, line 30: It's not 100% correct, Ironic supports attaching it as a virtual USB stick. The problem is, a lot of hardware does not allow 2 virtual media devices at the same time. I'd probably skip the reasoning. But do mention that network data is also not used.

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,32 @@
...
+
+In some circumstances, i.e to reduce boot time for ephemeral workloads, it may be possible to boot an iso
+and not deploy any image to disk (saving the time to write the image and reboot). Since this alternative
+is distributed as a live-iso, leveraging the existing toolchains like [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/) might be desirable.
+
+How to use it?
+
+Here is an example with a BaremetalHost CRD, where iso referenced by the url and `live-iso` set in DiskFormat will be
+live-booted without deploying an image to disk. Checksum options are not required in this case, and will be ignored if specified:
+
```
dtantsur, line 15: An important bit is missing: the live ISO mode is only supported for virtual media drivers.

docs/user-guide/src/bmo/live-iso.md
dtantsur, line 3: When used in a normal text, it should be "live ISO" or "live CD", not "live-iso". So for example, "with a live ISO image.."

dtantsur, line 3:
```suggestion
The live-iso API in Metal3 allows booting a BareMetalHost with a live ISO image instead of writing an image to the local disk using the IPA deploy ramdisk.
```

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,32 @@
+# Live-ISO
+
+The live-iso API in Metal3 allows to boot BaremetalHost with live-iso image instead of using IPA deploy ramdisk.
+
+Why we need it?
+
+In some circumstances, i.e to reduce boot time for ephemeral workloads, it may be possible to boot an iso
+and not deploy any image to disk (saving the time to write the image and reboot). Since this alternative
+is distributed as a live-iso, leveraging the existing toolchains like [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/) might be desirable.
```
dtantsur, line 9: This sentence reads a bit awkward to me. "this alternative" refers to the live-iso mode, so it cannot be distributed. You probably wanted something like:

This API is also useful for integration with 3rd party installers distributed as a CD image, for example ...

dtantsur (changes requested):

docs/user-guide/src/bmo/live-iso.md
furkatgofurov7: Good point, added now.

docs/user-guide/src/bmo/live-iso.md
furkatgofurov7: Fixed in normal text usage.

docs/user-guide/src/bmo/live-iso.md
furkatgofurov7: Done

docs/user-guide/src/bmo/live-iso.md
furkatgofurov7: I have removed the reasoning part for ironic, however did leave that `rootDeviceHints`, `networkData` and `userData` won't be used due to it image is not written to the disk. Hope that makes sense 🙂

docs/user-guide/src/bmo/live-iso.md
```diff
@@ -0,0 +1,36 @@
...
+and not deploy any image to disk (saving the time to write the image and reboot). This API is also useful
+for integration with 3rd party installers distributed as a CD image, for example leveraging the existing
+toolchains like [fedora-coreos installer](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/)
+might be desirable.
+
+How to use it?
+
+Here is an example with a BaremetalHost CRD, where iso referenced by the url and `live-iso` set in DiskFormat will be
+live-booted without deploying an image to disk. It is important to note that, live ISO mode is only supported when
+redfish-virtualmedia driver is used as BMC driver. Also, checksum options are not required in this case, and will be
```
dtantsur, line 18: Any virtualmedia driver actually.

furkatgofurov7: Thanks, fixed now

furkatgofurov7:
@dtantsur thanks for review, all should be good now, PTAL

kashifest (approved):

smoshiur1237 (approved):
/lgtm

PR #290, opened 2022-10-24
vendor  firmware support

moshe010:
Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/design/baremetal-operator/vendor-firmware-support.md
+++ b/design/baremetal-operator/vendor-firmware-support.md
@@ -1,0 +1,151 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add API to support vendor firmware burn and config
+
+## Status
+
+provisional
+
+## Summary
+
+Add a new interface to allow burn and config vendor firmware.
+
+## Motivation
+
+It desired to burn and config vendor firmware before deploying the OS image.
+With hardware manager framework vendor can extend IPA to burn and configure its firmware.
+Burning and configuring firmware is done in the manual cleaning step. It desired to expose this
+functionality to BMO API.
+
+### Goals
+
+1. Agree on the format of the API and attributes
+2. Agree on Nvidia vendor format
+
+### Non-Goals
+
+1. Agree on others vendor’s format.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+Cloud admin wants to burn and configure vendor firmware before OS deployed on bare-metal host.
+
+## Design Details
+
+### Implementation Details/Notes/Constraints
+
+
+IPA HardwareManagers allows vendors to add firmware burn and config
+in the manual cleaning stage.
+
+The ironic command to trigger the manual cleaning would be in
+the following syntax:
+
+```shell
+baremetal node clean node --clean-steps \
+[{"interface": "deploy", "step": <vendor function>, \
+   "args": {<vendor_arg1>: <vendor_value1>, \
+   <vendor_arg2>: <vendor_value2>, ...}}]
+```
+
+IPA HardwareManagers framework defines the following attributes:
+1. `step` - A function on the vendor hardware manager
+2. `argsinfo` - Arguments that can be required or optional to the step function
+3. `interface` - Should always be the deploy interface
+
+The proposal is to extend the BareMetalHost with VendorFirmware field.
+The VendorFirmware field includes the `step` name and `args` that can be required or optional.
+
+Nvidia hardware manager support `step` - update_nvidia_firmware with `args` - firmware_config and firmware_url
+
+The VendorFirmwareConfig for Nvidia should be the following:
+
+```yaml
+vendor_firmware:
+   update_nvidia_firmware:
+      firmware_config: "http://10.7.12.161/07_2022/config.yaml"
+      firmware_url: "http://10.7.12.161/fw/07_2022"
+```
+
+This yaml will translated to the following ironic command:
+
+``` console
+baremetal node clean node --clean-steps \
+[{"interface": "deploy", "step": "update_nvidia_firmware","args": \
+{"firmware_config": "http://10.7.12.161/07_2022/config.yaml", \
+"firmware_url": "http://10.7.12.161/fw/07_2022"}}]
+```
+
+Please note that VendorFirmware defines vendor clean steps that existing IPA upstream.
+Other vendors can extend IPA to support their firmware as Nvidia did. They may extend the VendorFirmware
+according to their "step" and "args" defined in IPA.
+
+### Risks and Mitigations
+
+None
+
+### Work Items
+
+- Extend the BareMetalHost CRD with the new parameters for VendorFirmware.
+
+- Validation of input values in the YAML parameters
+
+- A function to handle the manual cleaning steps related to Nvidia firmware.
+
+- Unit tests for all the work above
+
+### Dependencies
+
+- Ironic
+- BareMetal-operator
+
+### Test Plan
+
+- Unit tests for the functions
+- Integration testing with actual hardware
+
+### Upgrade / Downgrade Strategy
+
+None
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+Another approch is to define VendorFirmware to contains a string. The string describes
+the manual cleaning steps in json format. For example:
+```yaml
+vendor_firmware:
+   config: '{
+      "interface": "deploy",
+      "step": "update_nvidia_firmware",
+      "args": {
+         "firmware_config": "http://10.7.12.161/07_2022/config.yaml",
+         "firmware_url": "http://10.7.12.161/fw/07_2022"
+   }'
+```
+
+The downside of this approch is that we don't have validate control on the user input.
+
+## References
+
+- [IPA Hardware Managers]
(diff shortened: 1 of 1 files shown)


moshe010:
@dtantsur  can you please review this proposal. We want to enable the Nvidia IPA changes which are done in [1] 

[1] - https://review.opendev.org/c/openstack/ironic-python-agent/+/566544

design/baremetal-operator/vendor-firmware-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+The proposal is to extend the BareMetalHost with VendorFirmware field.
+The VendorFirmware field includes the `step` name and `args` that can be required or optional.
+
+Nvidia hardware manager support `step` - update_nvidia_firmware with `args` - firmware_config and firmware_url
+
+The VendorFirmwareConfig for Nvidia should be the following:
+
+```yaml
+vendor_firmware:
```
adrianchiris, line 73: an alternative is to model clean steps into BMH.  just a thought here...

that is something of this sort:

```
cleanSteps:
  - interface : "deploy"
    step: "update_nvidia_firmware"
    args:
      firmware_config: "http://10.7.12.161/07_2022/config.yaml"
      firmware_url: "http://10.7.12.161/fw/07_2022"
  - interface: "deploy"
    step "another_step"
    args:
        other: "val1"
        yetAnother: "val2"
```


args in this case would be a `map[string] string`

moshe010: I think I cover this on the Alternatives section [1]

[1] -  https://github.com/metal3-io/metal3-docs/pull/290/files/fc5600f7ed3075a649f331ab06887dec07e4c755#diff-9e85ae0576b8227ae061d315a315390d1c6fbfd9bd3cdd7ca27ace7d94f5f69cR128

adrianchiris: yes, its similar, but what i propose above is a bit more structured imo.

zaneb:
The design principles I'd like to see followed here are:

- A _declarative_ API. If we want some particular firmware version the user should specify that and the operator should be able to tell when we have it and when we don't (and thus need to update).
- Preferably vendor-neutral, but failing that in a specifically 'vendor-defined' area (see e.g. the HostFirmwareSettings CRD).
- Not be tied to implementation details in Ironic (e.g. clean_steps)
- Ideally be in a separate CRD and not lumped in to the BMH.

s3rj1k:
> 
>     * Preferably vendor-neutral, but failing that in a specifically 'vendor-defined' area (see e.g. the HostFirmwareSettings CRD).
> 

There is no vendor-neutral way of updating BIOS and BMC Firmware as all of this firmware is actually vendor-specific, even cross same vendor but different platform generation there are differences in update process.

It is true that more newer platforms have a more streamlined approach to firmware upgrade but no vendor follows any RFCs as there are none,

s3rj1k:
Considering all the above I, as a user of BMO, would like to avoid a bloat of code that in the end will not work as there is no possibility to make it generic and even test it on all the platforms that are available on the planet Earth.

Instead this could be just another controller with platform-specific plugin system

moshe010:
> Considering all the above I, as a user of BMO, would like to avoid a bloat of code that in the end will not work as there is no possibility to make it generic and even test it on all the platforms that are available on the planet Earth.
> 
> Instead this could be just another controller with platform-specific plugin system

I am not sure is just another controller because BMO is controlling the Host state. I don't think it a good idea that 2 controller will controlling the host.  We want to run the firmware update in the prepare method [1].  So IMO if we can extend BMO with register vendors plugin and we define grpc api (maybe PrepareDevice)  to trigger them from BMO(  similar concept of device plugin in kubelet [2]. ) BMO operator call PrepareDevice for each registered vendor plugin and the plugin can interact with ironic to preform the manual clean step. 

The pros is that is remove all the vendor code and the vendor can define the api is suitable to set prepare his device.
The down side of it it that we will call clean step per plugin and for BMO which is less efficient  

Thoughts about this approach? 

[1] - https://github.com/metal3-io/baremetal-operator/blob/ea2509bae83102ac9d5d79e6dc6e1352fe86ab78/pkg/provisioner/provisioner.go#L152 
[2] - https://github.com/kubernetes/design-proposals-archive/blob/main/resource-management/device-plugin.md#registration

s3rj1k:
My main concern that this should be opt-in functionality, or at least opt-out.

s3rj1k:
> > 
> 
> Do you mean that this feature can be enabled/disabled by flag?

Disabled by default, for those who need it, they would need to explicitly enable it and maybe do some additional setup, like spinning that ad-hoc GRPC service

Hopefully this would not bring a lot of new dependencies to the code base

moshe010:
> 

right User that enable the feature need to make sure to deploy the plugin(s)  the same node. 
We can use kubelet pluginwatcher pkg [1] to provide use the discover/yregistration//deregistration
if that make sense I will update the PR with this new approach  

[1] - https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/pluginmanager/pluginwatcher/plugin_watcher.go

design/baremetal-operator/vendor-firmware-support.md
dtantsur: Ironic is very explicitly a low-level system, while we have been so far trying to keep BMO API high-level. Exposes clean steps as they are raises a lot of questions. For example: which steps have to run every time and which - only once.

design/baremetal-operator/vendor-firmware-support.md
```diff
@@ -0,0 +1,151 @@
...
+   update_nvidia_firmware:
+      firmware_config: "http://10.7.12.161/07_2022/config.yaml"
+      firmware_url: "http://10.7.12.161/fw/07_2022"
```
dtantsur, line 76: I generally like this design. I think naming could use some improvement, especially considering that we also have OOB firmware updates. What I'd like to see is something like:

```yaml
firmwareUpdates:
    nvidia:
        - url: ...
          config: ...
    system:
        - url:
          config:
```

Here `system` means the future out-of-band updates (to be clear: you don't need to implement them, just to account for them added in the future). `url` is used, again, for compatibility with existing OOB clean steps.

design/baremetal-operator/vendor-firmware-support.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+This yaml will translated to the following ironic command:
+
+``` console
+baremetal node clean node --clean-steps \
+[{"interface": "deploy", "step": "update_nvidia_firmware","args": \
+{"firmware_config": "http://10.7.12.161/07_2022/config.yaml", \
+"firmware_url": "http://10.7.12.161/fw/07_2022"}}]
+```
```
dtantsur, line 86: Does firmware update happen every time the node is provisioned or only once? If once, how is it ensured?

Do you track changes behind the same URL or is it always a new URL?

moshe010: Every time the node is provisioned. In the IPA code we check if the Firmware version/config in the NIC is different from the URL and update if needed. 
we don't track changes behind URL. The idea is that admin will put all his config in folder. If he has a new FW update/config he can put it on new folder (e.g new URL)  and reprovision the node.

moshe010:
> The design principles I'd like to see followed here are:
> 
> * A _declarative_ API. If we want some particular firmware version the user should specify that and the operator should be able to tell when we have it and when we don't (and thus need to update).
> * Preferably vendor-neutral, but failing that in a specifically 'vendor-defined' area (see e.g. the HostFirmwareSettings CRD).
> * Not be tied to implementation details in Ironic (e.g. clean_steps)
> * Ideally be in a separate CRD and not lumped in to the BMH.
The firmware version is complex because different cards (e.g OEM) can have same version but different binary. The idea here with the url is that admin will upload all his firmware binaries for all NIC flavors. He can put all them under version folder. This is also simplify mix NIC configuration of same vendor.

design/baremetal-operator/vendor-firmware-support.md
moshe010: just to clarify we are on the same page when you wrote "out-of-band updates" you mean to update the firmware using iDRAC for example? I know that we have such support with Dell to do out-of-band updates. 
Also not sure I understand "url is used, again, for compatibility with existing OOB clean steps." what are the existing OOO cleans steps?

dtantsur: Yep, that's what I mean: https://docs.openstack.org/ironic/latest/admin/drivers/redfish.html#firmware-update-using-manual-cleaning and https://docs.openstack.org/ironic/latest/admin/drivers/ilo.html#initiating-firmware-update-as-manual-clean-step

design/baremetal-operator/vendor-firmware-support.md
dtantsur: Understood, please add this information to the spec.

We'll need to invent something else for OOB since we cannot check versions there... maybe we think about it once we get there.

dtantsur:
> Disabled by default, for those who need it, they would need to explicitly enable it and maybe do some additional setup, like spinning that ad-hoc GRPC service

We're talking about maybe 100-200 lines of code in BMO, similar to what was added for BIOS. Not sure maintaining another service is worth saving some kilobytes of binary.

If we go down this path, we'll need to invent a way to pass clean steps from the new service to BMO. Which is doable, but is probably going to be a larger BMO change than this.

moshe010:
We decided to split the proposal to 2. 
1. firmware image update support which is captured in https://github.com/metal3-io/metal3-docs/pull/302
2. firmware config support will be posted soon.

PR #297, opened 2022-11-17
Add build arg options section for custom source build of ironic-image

elfosardo:
Source build can be customized using remote source like gerrit refs or commit hashes, or local source in a directory in the container context.


furkatgofurov7 (commented):
/lgtm

PR #302, opened 2023-01-08
firmware update support

moshe010:
Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/design/baremetal-operator/firmware-image-update.md
+++ b/design/baremetal-operator/firmware-image-update.md
@@ -1,0 +1,195 @@
+# Firmware Image Update
+
+## Status
+
+provisional
+
+## Summary
+
+Provide the ability to get and update vendor firmware version.
+
+## Motivation
+
+It desired to update vendor firmware before deploying the OS image of the host.
+
+### Goals
+
+- Agree on the format of the vendor neutral API and attributes
+- Implemented Nvidia firmware update for NICs
+
+### Non-Goals
+
+- Implemented OOB firmware update for ILO and redfish
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+Cloud admin wants to update Nvidia NICS firmware before OS deployed on bare-metal host.
+
+## Design Details
+
+### New API (CRD) Proposal
+
+We propose to introduce a new Custom Resource Definition (CRD) `HostFirmwareImage`. `HostFirmwareImage` will store a list of firmware images to be applied (if required) on the bare-metal host. The `HostFirmwareImage` will be mapped to Baremetal Host object by namespace and name. The spec section will introduce the following attributes:
+
+- url - location to download the firmware image.
+- checksum - checksum of the firmware image. Optional attribute.
+- checksumType - checksum algorithm for the firmware image can be md5, sha256, sha512. Optional attribute.
+- vendor – The hardware vendor which is responsible for that firmware. Current proposal it will be only nvidia and can be extended in future proposals for other vendors.
+- component – The hardware component which this firmware matches. Hardware vendor may have several types of hardware for example Nvidia has NiC’s and GPU. Current proposal will be only `net` and can be extended in future proposals for other vendors.
+- componentFlavor- indicate the component flavor. Optional attribute. For Nvidia NICs it will be the PSID.
+- version - the requested firmware version of the firmware image.
+
+example of the HostFirmwareImage spec section:
+
+```yaml
+spec:
+  firmwareImages:
+    - url: http://firmware_images/fw1.bin
+      checksum: a94e683ea16d9ae44768f0a65942234d
+      checksumType: md5
+      vendor: nvidia
+      component: net
+      componentFlavor: MT_1090120019
+      version: 24.34.1002
+    - url: http://firmware_images/fw2.bin
+      checksum: e07da4de09e9a2ab836bd7e23378f8e45afd5400f0f1cf7da5ecefce45333d5d
+      checksumType: sha1
+      vendor: nvidia
+      component: net
+      componentFlavor: MT_0000000652
+      version: 22.35.1012
+```
+
+The `HostFirmwareImage` status section will introduce the following attributes:
+
+- Firmwares – list of all hardware devices with the following information:
+    - componentID – unique identifier for the hardware. For NICs the unique identifier will be in the following format “`<interface name>_<MAC>”.`
+    - vendor – The hardware vendor which is responsible for that firmware.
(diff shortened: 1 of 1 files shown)


moshe010:
@dtantsur @zaneb  any feedback on the proposal?

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+      component: net
+      componentFlavor: MT_0000000652
+      version: 22.35.1012
+  lastUpdated: 2022-12-27T01:12:17Z
+  conditions: ...
+```
+
+### HostFirmwareSettings Controller Changes
+
+The proposal is to extend the `HostFirmwareSettings` controller which was introduced in the `bulk-set-bios-config` proposal. The controller will be extended to track the `HostFirmwareImage` CRD. The HostFirmwareImage.spec is the desired firmware version and the HostFirmwareImage.status contain the actual firmware. Please note that the correct proposal addresses Nvidia NICs but can be extended to other vendors too as well.
```
zaneb, line 104:
```suggestion
The proposal is to extend the `HostFirmwareSettings` controller which was introduced in the `bulk-set-bios-config` proposal. The controller will be extended to track the `HostFirmwareImage` CRD. The HostFirmwareImage.spec is the desired firmware version and the HostFirmwareImage.status contain the actual firmware. Please note that the current proposal addresses Nvidia NICs but can be extended to other vendors too as well.

```

FWIW, each CRD generally has its own controller, and they all run in the same process. So we wouldn't extend the existing controller, but rather create a new one in the same vein.

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+The NIC firmware version exists today in the data.Extra.Network struct. We will also extend `HardwareData` with the NIC firmware by adding firmware attribute to the details.NIC. BIOS version already exists in `HardwareData` but other devices which do not firmware version is not exposed will required to provide mechanism to retrieve actual firmware and store it in the `HardwareData`.
```
zaneb, line 106: The last sentence does not parse.

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+The reconcile loop of the `HostFirmwareSettings` controller will do the following:
+
+- Validate the `HostFirmwareImage` attributes (e.g., vendor name, component …). We will try to validate at match as we can. If valid set condition to FirmwareImageValid to true.
+- Fetch the corresponding HardwareData and get the Nvidia NIC firmware. The Nvidia firmware contains firmware version and componentFlavor "24.34.1002 (MT_0000000540)" other NIC vendors can extend the parsing according to their NIC firmware string.
```
zaneb, line 111: This requires knowledge of how to get the firmware version for a particular piece of hardware from HardwareDetails to be built in to the controller.
It's really Ironic's job to be providing the abstraction over different kinds of hardware. Ironic has third-party CI set up to run against multiple vendors' hardware. We don't have that kind of community in Metal3, so we try to avoid being in the business of having to implement the hardware abstraction.

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+- Vendor, component or componentID in the spec section do not exist in the bare metal.
+- Actual firmware version exists in the status but no matching component or componentFlavor in the spec.
+
+### Implementation Details/Notes/Constraints
+
+The `HostFirmwareImage` spec section will be translated to ironic manual cleaning steps in the ironic provisioner as following:
+
+```shell
+baremetal node clean control-0 --clean-steps nvidia_nic_firmware_image.json
```
zaneb, line 135: Again, this is a vendor-specific step that would have to be hard-coded into the baremetal-operator.

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+```yaml
+spec:
+  firmwareImages:
+    - url: http://firmware_images/fw1.bin
+      checksum: a94e683ea16d9ae44768f0a65942234d
+      checksumType: md5
+      vendor: nvidia
+      component: net
+      componentFlavor: MT_1090120019
+      version: 24.34.1002
```
zaneb, line 57: It's unfortunate that the user needs to correctly describe the version that corresponds to the URL, but I can't think of a way around that.

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+Provide the ability to get and update vendor firmware version.
+
+## Motivation
+
+It desired to update vendor firmware before deploying the OS image of the host.
+
+### Goals
+
+- Agree on the format of the vendor neutral API and attributes
```
zaneb, line 17: 'Goals' mean the goals of the feature, not the goals of this document.

I would say the goal is to have a consistent API where the vendor-specific parts are well-defined.

design/baremetal-operator/firmware-image-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+provisional
+
+## Summary
+
+Provide the ability to get and update vendor firmware version.
+
+## Motivation
+
+It desired to update vendor firmware before deploying the OS image of the host.
```
zaneb, line 13: FYI, for someone without context this isn't as motivating as you possibly imagine it is :)

zaneb (commented):
Ideally from my perspective, there would be an API in Ironic that returns a list of firmware that can be updated, together with:

* The current version
* The clean_step name to initiate any upgrade
* A list of arguments to the clean_step and their types

As we do for the HostFirmwareSettings, we could pre-populate the Spec with the available fields, and we could detect any changes vs. the status to know if we need to update at our next opportunity.

That's also how the HostFirmwareSettings was done: an API for it was implemented in Ironic first.

dtantsur:
FWIW I'm floating an idea of a more high-level Ironic API. That will mean more work for everyone, of course :)

design/baremetal-operator/firmware-image-update.md
dtantsur: Another note: I'd like us to stop relying on `data.Extra` and move everything we need to the official IPA inventory. The Extra thing increases the size of our ramdisks. With TripleO-RDO ramdisk potentially gone soon, we'll need to use the upstream Ironic ramdisks that don't include extra-hardware support.

PR #313, opened 2023-03-08
Proposal: add Support for User-defined Parameters

zhouhao3:
Add an optional field `extraParametersName` to accept a `Secret` object where to allow user to set vendor specific configuration.

Signed-off-by: Zou Yu <zouy.fnst@fujitsu.com>

### DIFF
--- a/design/baremetal-operator/user-defined-parameters.md
+++ b/design/baremetal-operator/user-defined-parameters.md
@@ -1,0 +1,170 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add Support for User-defined Parameters
+
+## Status
+
+provisional
+
+## Summary
+
+Add an optional field `extraParametersName` to accept a `Secret` object
+where to allow user to set vendor specific configuration.
+
+## Motivation
+
+When deploying BM nodes, some vendor's BMC will require additional parameters
+to adapt to certain situations. For example:
+
+1. When FIPS mode is enabled on the host running the ironic service, the
+   version of SNMP protocol of iRMC driver must be v3. Therefore, additional
+   parameters are needed to specify the version and to provide the corresponding
+   credential information.
+2. With some BMC firmware upgrades, additional configuration may be required.
+   E.g.: For iRMC firmware version >= `iRMC S6 2.00`, we need to additionally
+   provide driver info such as:
+    - **redifsh_address**: iRMC IP address or hostname.
+    - **redfish_username**: iRMC user name with administrative privileges.
+    - **redfish_password**: password of redfish_username.
+    - **redfish_verify_ca**: accepts values those accepted in irmc_verify_ca.
+3. For the production environment, it may be necessary to use a certificate
+   issued by a third-party CA for verification, so additional parameters are
+   required to specify the root certificate, e.g.: `irmc_verify_ca=<cert_path.crt>`.
+
+Considering that the user-defined parameters may contain credential information,
+`Secret` object would be an appropriate carrier for usage.
+
+To sum up, without changing the existing APIs, adding support for extra
+user-defined parameters will bring more freedom to meet the specific
+configuration requirements for some vendors' BM.
+
+### Goals
+
+- Add an optional field `extraParametersName` in BMH to allow specify a
+  `Secret` object.
+- Users can define arbitrary parameters in the `Secret`.
+- The parameters can be passed to ironic's driver_info.
+
+### Non-Goals
+
+- User-defined parameters can override the BMC parameters that set by existing
+  APIs.
+- Verify that user-defined parameters are valid.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a cluster administrator, I would like to build an OpenShift cluster with
+FIPS mode using Fujitsu PRIMERGY servers, but due to [the limitation of iRMC driver](https://github.com/openstack/ironic/blob/21.3.0/ironic/drivers/modules/irmc/common.py#L254-L257),
+I need to set `irmc_snmp_version` to "v3" and provide related credential.
+
+#### Story 2
+
+As a cluster administrator, when deploying Fujitsu PRIMERGY servers with
+firmware version >= `iRMC S6 2.00` using the iRMC driver, I need to provide
+additional parameters due to the following [change](https://docs.openstack.org/ironic/latest/admin/drivers/irmc#node-configuration):
+
+> Fujitsu server equipped with iRMC S6 2.00 or later version of firmware disables IPMI over LAN by default.
+> However user may be able to enable IPMI via BMC settings. To handle this change, irmc hardware type first
+> tries IPMI and, if IPMI operation fails, irmc hardware type uses Redfish API of Fujitsu server to provide
+> Ironic functionalities. So if user deploys Fujitsu server with iRMC S6 2.00 or later, user needs to set
+> Redfish related parameters in driver_info.
+
+Note that so called "Redfish related parameters" above is for iRMC rather than Redfish driver.
+
+## Design Details
+
+### Implementation Details/Notes/Constraints
+
+#### Add New Field
+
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/user-defined-parameters.md
```diff
@@ -0,0 +1,456 @@
...
+    credentialsName: worker-test-bmc-secret
+    extraParametersName: worker-test-bmc-extra-parameters # New field
+```
+
+#### Convert ExtraParametersName to ExtraParameters
+
+Get the specified `Secret` according to `ExtraParametersName`, and then convert
+the data of the `Secret` to `ExtraParameters`.
+
+```diff
```
dtantsur, line 132: Please don't paste actual patches to specifications, it's not the place to review them. We need to agree on the design first.

design/baremetal-operator/user-defined-parameters.md
```diff
@@ -0,0 +1,456 @@
...
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+Due to the limitation of `Secret` object, the value of `ExtraParameter` can only
+be string type.
+
```
dtantsur, line 449: The proposal gives users no way to determine which fields can be used without diving into the Ironic docs.

design/baremetal-operator/user-defined-parameters.md
```diff
@@ -0,0 +1,456 @@
...
+-       if info.host.Status.ErrorType == metal3v1alpha1.RegistrationError || registeredNewCreds {
++       if info.host.Status.ErrorType == metal3v1alpha1.RegistrationError || registeredNewCreds || registeredNewExtraParameters {
+                info.log.Info("clearing previous error message")
+                dirty = clearError(info.host)
+        }
+```
+
+### Risks and Mitigations
+
+None
```
dtantsur, line 418: How will verification work? What if a user supplies an unsupported parameter? What if a user supplies an invalid value?

design/baremetal-operator/user-defined-parameters.md
```diff
@@ -0,0 +1,456 @@
...
+2. With some BMC firmware upgrades, additional configuration may be required.
+   E.g.: For iRMC firmware version >= `iRMC S6 2.00`, we need to additionally
+   provide driver info such as:
+    - **redifsh_address**: iRMC IP address or hostname.
+    - **redfish_username**: iRMC user name with administrative privileges.
+    - **redfish_password**: password of redfish_username.
+    - **redfish_verify_ca**: accepts values those accepted in irmc_verify_ca.
+3. For the production environment, it may be necessary to use a certificate
+   issued by a third-party CA for verification, so additional parameters are
+   required to specify the root certificate, e.g.: `irmc_verify_ca=<cert_path.crt>`.
```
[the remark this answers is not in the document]
zhouhao3: All we want to do is provide a parameter that can be passed to the ironic. As for the subsequent processing of this parameter or whether this parameter is validate, these are not in our proposal. We think these should be handled in `ironic`. I will also add this content to Non-goals.

In this example, we only consider passing the verify_ca path to ironic. As for the mounting and generation of the certificate, it should be handled by the user.

design/baremetal-operator/user-defined-parameters.md
```diff
@@ -0,0 +1,456 @@
...
+When deploying BM nodes, some vendor's BMC will require additional parameters
+to adapt to certain situations. For example:
+
+1. When FIPS mode is enabled on the host running the ironic service, the
+   version of SNMP protocol of iRMC driver must be v3. Therefore, additional
+   parameters are needed to specify the version and to provide the corresponding
+   credential information.
+2. With some BMC firmware upgrades, additional configuration may be required.
+   E.g.: For iRMC firmware version >= `iRMC S6 2.00`, we need to additionally
+   provide driver info such as:
```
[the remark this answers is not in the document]
zhouhao3: No, these are purely extra information for that situation, and it is conceivable that the number of parameters specified by driver_info may increase in future Ironic drivers, not only for iRMC but also for other vendors.

design/baremetal-operator/user-defined-parameters.md
zhouhao3: We can't verify the parameters on the BMO side, so the verification work should be handled by ironic. We'll add that risk here.

design/baremetal-operator/user-defined-parameters.md
zhouhao3: This is indeed a drawback, and we will try to improve it by putting relevant documentation links in the api description. We'll  alos add this to Drawbacks.

design/baremetal-operator/user-defined-parameters.md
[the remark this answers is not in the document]
dtantsur: So, to clarify: Redfish address does not use the iRMC IP, user name and password are different, etc?

Also, I'd prefer we don't try to predict the future and rather concentrate on the problems at hand.

design/baremetal-operator/user-defined-parameters.md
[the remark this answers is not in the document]
dtantsur: > As for the subsequent processing of this parameter or whether this parameter is validate, these are not in our proposal.

This is the problem. BareMetalHost, for better or worse, is designed to be an independent API, not just a pass-thru to Ironic. We need to think in terms of this API. (As an aside, I also regret the way driver_info is designed in Ironic, but that ship has sailed long ago).

> As for the mounting and generation of the certificate, it should be handled by the user.

User in case of upstream Metal3 - sure. If you care about OpenShift, it's not possible for a user to modify the Metal3 pod in a supported way. This is not our limitation, this is how OpenShift works.

design/baremetal-operator/user-defined-parameters.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Motivation
+
+When deploying BM nodes, some vendor's BMC will require additional parameters
+to adapt to certain situations. For example:
+
+1. When FIPS mode is enabled on the host running the ironic service, the
+   version of SNMP protocol of iRMC driver must be v3. Therefore, additional
+   parameters are needed to specify the version and to provide the corresponding
+   credential information.
```
zaneb, line 27: I could potentially be convinced that the passwords for SNMP should be read as separate fields from the existing BMC credentials Secret, if I were convinced that drivers needing to access SNMP (and the passwords being different from the BMC password) were a thing that made sense generally.
The v3 part can easily be inferred by the iRMC BMC plugin based on the presence of the keys.

zaneb (commented):
I'd encourage you to read the design principles documented in https://github.com/metal3-io/metal3-docs/blob/main/design/api-design-principles.md#api-design-principles because this... breaks most of them (especially the first 3).
An unversioned, undocumented catch-all bucket whose meaning depends on the implementation details of Ironic is pretty much the exact opposite of what we are trying to do with Metal³.

design/baremetal-operator/user-defined-parameters.md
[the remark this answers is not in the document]
dtantsur: Yeah, I was about to suggest the same. We already have this pattern for API TLS: define a fixed location where the certificates can be mounted. Then we also need a configuration in ironic.conf (if not already) that is set conditionally on this location existence. Then we update CBO (in case of OpenShift) to mount the secret with the certificates there. No BMO changes required.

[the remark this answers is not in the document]
dtantsur: The thing I was in other places in OpenShift is putting your CA certificate as a string to install-config. That could be taken by the installer and put 1) on the bootstrap VM for control plane installation, 2) in the CBO configuration for the final cluster.

Getting certificate from a server causes a chicken-and-egg problem: how do we verify the certificate of that server? Also not all users will want to establish an additional HTTP server themselves.

design/baremetal-operator/user-defined-parameters.md
dtantsur: If it's just a protocol version though, it can be a parameter in the BMC address: `irmc://ip.address?snmp_version=3`

zhouhao3:
I'm going to close this PR as I've recreated a new [one](https://github.com/openshift/enhancements/pull/1395).

PR #332, opened 2023-05-10
BareMetalHost API v1beta1

dtantsur:


s3rj1k:
If my opinion counts, I like that, removing legacy with API version bump

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Risks and Mitigations
+
+None
+
+### Work Items
+
+- Add `Architecture`
+- Copy type definitions and modify them.
+- Adapt the controller code to work with both API versions.
```
zaneb, line 89: This is not how it works. The controller only deals with the latest version, and you implement a conversion webhook to convert between them. That will have to be part of this work.
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,126 @@
...
+where requested architecture is empty and provided architecture is not.
+
+### Risks and Mitigations
+
+None
+
+### Work Items
+
+- Add `Architecture`
+- Copy type definitions and modify them.
```
zaneb, line 88: We also have to find the 5000 places that `v1alpha1` appears in the controller code and change them. Preferably to something that includes the version in far fewer places (like only in the import lines).

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Both `v1alpha1` and the proposed `v1beta1` resources will be possible
+to handle with the same controller code since the differences are mostly
+cosmetic.
+
+The new architecture handling will make it possible for a
+`PreprovisioningImage` to have an empty architecture requested (currently,
+it defaults to `x86_64`). We expect implementations to use its default
```
zaneb, line 77: Isn't this equivalent to BMO choosing a better default?

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,126 @@
...
+None
+
+### Test Plan
+
+We'll need to test both API versions. Since we already have tests for different
+versions of CAPM3, we may tie them to different versions of BMH.
+
+### Upgrade / Downgrade Strategy
+
+Upgrade to the new API version will involve:
```
zaneb, line 103: The webhook will handle this.

design/baremetal-operator/bmh-v1beta1.md
lentzi90: Yes, +1 to avoid the version. This is what CAPI and many of the providers do also. In CAPM3 we rename the import `infrav1` ([example](https://github.com/metal3-io/cluster-api-provider-metal3/blob/6a9fe1d9e9fbea791094a709d5e79ad99c63b34b/controllers/metal3cluster_controller.go#L27)). Something similar should work well also for BMO. :slightly_smiling_face:

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,126 @@
...
+- Stabilize the BareMetalHost API.
+
+## Proposal
+
+Copy the current API types definition into a new package `v1beta1` with some
+modifications.
+
+## Design Details
+
+The following addition should be done *before* the version is created:
```
lentzi90, line 43: I would suggest actually creating v1alpha2 as a direct copy of the current v1alpha1. Then implement these changes in v1alpha2. When done, make a release. If we are really happy/confident with it, rename v1alpha2 -> v1beta1 before the release. But I think it would be good to actually do a v1alpha2 release first. It can anyway be promoted to v1beta1 without changes if it proves good.

dtantsur: Why? You're essentially asking for double the work. We're definitely not in the alpha stage (for a long time already).

design/baremetal-operator/bmh-v1beta1.md
dtantsur: BMO does not know the default.

design/baremetal-operator/bmh-v1beta1.md
dtantsur: To quote https://kubernetes.io/docs/reference/using-api/#api-versioning about alpha versions:

> The software may contain bugs. Enabling a feature may expose bugs.
> Support for an alpha API may be dropped at any time without notice.
> The API may change in incompatible ways in a later software release without notice.
> The software is recommended for use only in short-lived testing clusters, due to increased risk of bugs and lack of long-term support.

This is obviously false for us. Compare with beta definition:

> The software is well tested. Enabling a feature is considered safe.
> The support for a feature will not be dropped, though the details may change.
> The schema and/or semantics of objects may change in incompatible ways in a subsequent beta or stable API version. When this happens, migration instructions are provided. Adapting to a subsequent beta or stable API version may require editing or re-creating API objects, and may not be straightforward. The migration may require downtime for applications that rely on the feature.

zaneb: > You're essentially asking for double the work.

I think 90% of the work is in getting things set up so that we _can_ add a new version. After that, subsequent new versions should be comparatively easy if we've done it right.

design/baremetal-operator/bmh-v1beta1.md
zaneb: Why is it harder to teach BMO the default (and have an explicit API) than to teach the preprovisioning image controller the default?
The code for both of those exists in the same repo.

dtantsur: At the very least, it will remove the freedom of a preprovisioning image controller to detect the architecture (e.g. the assisted one in OCP can use their InfraEnv resources). In the built-in case, it is the preprovisioning image controller that now has access to DEPLOY_KERNEL/RAMDISK.

design/baremetal-operator/bmh-v1beta1.md
dtantsur: > If we immediately create v1beta1, work on it and then make a release, then any kind of mistakes in the API would require v1beta2 to fix. 

Same applies to alpha, no differences here.

Our software is not alpha. I suspect you're using some definition of "alpha" and "beta" that do not match the kubernetes one that I quoted above. Implicit in what you suggest seems to be an assumption that beta is stable and cannot/should not be changed.

This is not what the kubernetes docs say.

Alpha = do not use in production. Is it what you expect of the new API version (let alone the current one)? At least in OpenShift we've been using BareMetalHost API in production for years already.

Beta = can be used in production but is subject to further changes. This is the state of things the way I see it *already*. I'd probably have more understanding if you suggested that we rename the current v1alpha1 to v1beta1 first, then introduce v1beta1. But that's additional work for both the consumers and myself, and I don't have infinite cycles to juggle with versions just for the sake of it.

lentzi90: I get what you are saying about alpha/beta, but the fact is that the current API is labeled as alpha. This means we have the chance to work on it still under that label. I'm also fine releasing it as v1beta1 in the current state for the reasons you mentioned. That makes sense. Some people seem to think it is a failure if the beta cannot be promoted to v1 without changes. I'm not one of them, but I want to make sure that it is known before we willingly take a path that would require a v1beta2.

Since the plan is to make changes, I think it makes much more sense to be careful about them. I.e. call it alpha since the current version is anyway called alpha. This gives us the chance to iron out any issues before we release it on the world as production ready. It would be very embarrassing to make v1beta1 and realize a day too late that it was completely broken because of some stupid mistake we made, don't you think? Especially considering how good the v1alpha1 is.

dtantsur:
Updated the document with my reasoning against v1alpha2 and added a proposal by Zane to rename online to poweredOn.

design/baremetal-operator/bmh-v1beta1.md
lentzi90: > I don't know if we are going to have a v1beta2. It's up in the air, and this work does not change that.

I mean that if we promote the current v1alpha1 to v1beta1 without changes, then we take the path to also do v1beta2, since we want to implement the changes discussed in this proposal.

> The reality is: we won't know the downsides of any API until people start using it. If we release v1alpha2 now, we'll need to wait months for everyone to switch. While advertising our API as not suitable for production.

Fair enough. I'm open to releasing the current v1alpha1 as v1beta1 (with the caveat about v1beta2 coming up). What I find embarrassing is taking our current production grade API that is labeled as alpha, make changes to it and release as beta, then find out it was broken due to the changes we did. It would be less embarrassing to let it take time and make sure it is tested under the alpha label. But like mentioned before, I'm fine with promoting v1alpha1 as beta directly with the goal of signaling to consumers that this is production ready.

> Then release v1beta1 and wait months more. Then someone comes and says "oh wait, I have a brilliant idea!" :)

The whole idea with this is that two API versions can be identical. If v1alpha2 turns out good, we just release it as v1beta1 when we feel it has been properly tested. No need to then wait more months after that release since it is the exact same as the previous.

To summarize my thoughts

- :heavy_check_mark: Promote v1alpha1 and release it unchanged as v1beta1. Then start working on v1beta2.
- :heavy_check_mark: Create a new v1alpha2 for implementing changes. Can be released as v1alpha2 or v1beta1 depending on how we feel about it.
- :heavy_minus_sign: Create a new v1beta1 for implementing changes. IMO unnecessarily bold move but I can accept it.
- :x: Make changes to v1alpha1. Unacceptable. This is after all the current API used in production. It must stay as it is.

dtantsur: > I mean that if we promote the current v1alpha1 to v1beta1 without changes

I don't think this has been proposed though? Or is it what you would like to do? Making two versions in a short period of time may be confusing.

> find out it was broken due to the changes we did

Could you elaborate what you mean by broken and why it suddenly has become broken (or anyhow more broken than it is)? Most changes here are renaming of fields. The only functional change - the deprecation of MD5 - is essentially forced on us by upstream.

> release it as v1beta1 when we feel it has been properly tested

When is this moment going to happen? What's the definition of "properly tested" going to be? Will anyone even use the new version if they know that v1beta1 is coming soon?

We need to be very careful with assuming that anything we produce will be carefully tested in the near future. We've made this mistake in the Ironic community in the past. We kept producing release candidates that nobody really tested become everyone was waiting for a production-ready release. I'm afraid if we say "v1alpha2 is a candidate for the future v1beta1", the consumers will simply wait for v1beta1 (and then complain if it does not work). Same situation with the proposal of promoting v1alpha1 to v1beta1: why would anyone use that when a v1beta2 is coming soon?

lentzi90: > I don't think this has been proposed though? Or is it what you would like to do?

It is my suggestion for how to solve the "Our software is production ready and has been for long time. We just never advertised it, and I want to fix it as soon as possible, not in several months." issue without compromising quality by rushing the new version.

> Could you elaborate what you mean by broken and why it suddenly has become broken (or anyhow more broken than it is)? Most changes here are renaming of fields. The only functional change - the deprecation of MD5 - is essentially forced on us by upstream.

It would be the first time we do this in BMO, right? I always expect issues then and would prefer to proceed with caution.

> When is this moment going to happen? What's the definition of "properly tested" going to be? Will anyone even use the new version if they know that v1beta1 is coming soon?

Maybe a month? We would definitely use it internally and I hope you too. The main goal for me would be to see that it actually works outside of the CI system and that the release process works so that the artifacts are usable after the release.

> Same situation with the proposal of promoting v1alpha1 to v1beta1: why would anyone use that when a v1beta2 is coming soon?

It is two different things:

1. Show that the project is ready for production (I didn't think this was urgent but you seem to think so)
2. Work on a new release including changes to the API (which should not be rushed in my opinion)

If 1 is not important we can skip it and then there is only the new API to care about. I still suggest labeling it as v1alpha2 in that case because it is the first time we do this with BMO. But if the community wants to do v1beta1 immediately I will accept it. It just seems unnecessarily "bold" with basically nothing to gain. On the other hand if we do v1alpha2 it would require basically no effort to promote it to v1beta1 later and then we avoid (most of) the risk of screwing it up.

dtantsur: > without compromising quality by rushing the new version
> It just seems unnecessarily "bold"
> we avoid (most of) the risk of screwing it up

You keep implying some sort of quality regression stemming from this work without providing any reasoning. We have modified API in the past already, even though in a compatible way. Any change to BMO has a risk of breaking consumers.

If we create v1beta1 now, it probably will work (although the webhook part is the easiest to do wrongly). But how does it prevent us from "compromising quality" when doing v1beta2?

> Maybe a month? We would definitely use it internally and I hope you too. 

A month is not enough for us to switch - we have too many things going on. Knowing another version is coming, I may not even bother with switching because it's not free, and my time is limited.

> On the other hand if we do v1alpha2 it would require basically no effort to promote it to v1beta1 later 

It's not "no effort". We need to update the code, we need to update the consumers, we need to set a sunset date for v1alpha2.

In the end, I can agree to v1alpha1 -> v1alpha2 -> v1beta1 just for the sake of ending this argument, but I still see this as largely pointless. I would like to hear from other community members though, so that I don't have to change this plan again.

lentzi90: Alright let's not drag this out more than necessary. Thank you for the patience :wink: We have discussed it in the team and going for v1beta1 is fine. I'm not understanding fully the need for this but I don't have a strong opinion so to end this thread, go with v1beta1. :slightly_smiling_face: 

There were some additional concerns/comments/ideas though, so expect a few more comments soonish from @kashifest and @Rozzii !

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,140 @@
...
+- Communicate the change to the community.
+
+### Dependencies
+
+None
+
+### Test Plan
+
+We'll need to test both API versions. Since we already have tests for different
+versions of CAPM3, we may tie them to different versions of BMH.
```
kashifest, line 104: I would like to know a bit more concrete idea about how this would be incorporated in the test. Are we thinking about creating new branch. Should this be tied to the issue https://github.com/metal3-io/baremetal-operator/issues/1283 ? How CI would accomodate this greatly depends on how we would want to land this new API. For CAPM3 we have different branches currently for each minor release.

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,140 @@
...
+- Copy type definitions and modify them.
+- Create a conversion webhook to change between versions.
+- Communicate the change to the community.
+
+### Dependencies
+
+None
+
+### Test Plan
+
```
Rozzii, line 102: I was thinking about adding an implementation schedule of checklist.
IMO the implementation process has to already incorporate the branching that we have talked about during MTM.

I would like to have this schedule:
- Implement the aplhav1 related changes e.g. spec.Architecture
- Branch out to alphav1 branch after there are nothing to do to the old api
- Release a patch version of BMO from the alpha v1 branch
- Continue  implementation of v1Beta1 specific features on main branch 
- Branch out v1Beta1
- release v1Beta1 as major 1.0 release of BMO
--- future:
- start planning the scope of v1 in order to have some idea about the future

Rozzii: more about how the branches would be supported here: https://github.com/metal3-io/baremetal-operator/issues/1283

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,140 @@
...
+- Removing `ChecksumType` if its value is `MD5`. For a while, Ironic will
+  still accept MD5 checksums.
+
+### Version Skew Strategy
+
+BMO and CAPM3 version skew may affect the `ChecksumType` handling. Even
+with older CAPM3, users will not be able to use `MD5` as the value, but
+the other values will be available.
+
```
Rozzii, line 121: We have discussed backporting v1beta1 BMO API support to older CAPM3 release branches and we are on the opinion (ATM), that only the latest release branch would get a backport to support v1beta1 of bmo, or there would be a completely new minor release of CAPM3 that could use the new BMO API but old CAPM3's would only work with old BMO and v1alpha1 API. I guess this info can be incorporated in the proposal and this would solve the version skew issues.

I guess if the MD5  related api changes are only done to the v1beta1 BMO api and a user using an old BMO version with an old CAPM3 version with an old Ironic everything would be fine.

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,140 @@
...
+### Upgrade / Downgrade Strategy
+
+The conversion webhook will take care of in-flight CR upgrade:
+
+- Deleting `HardwareProfile`, populating `Architecture` and `RootDeviceHints`
+  if needed.
+
+- Removing `ChecksumType` if its value is `MD5`. For a while, Ironic will
+  still accept MD5 checksums.
```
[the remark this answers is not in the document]
dtantsur: Eventually - yes. We don't have any way around.

design/baremetal-operator/bmh-v1beta1.md
dtantsur: Isn't "Work items" the checklist?

I'm not sure we need to tie supported branches with supported API versions. Not saying it's a bad idea, but in theory we can just introduce a new version and only branch out before removing the old version.

design/baremetal-operator/bmh-v1beta1.md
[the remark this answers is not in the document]
lentzi90: Ok that's fine, we will just need to clearly document that this requires user action.

design/baremetal-operator/bmh-v1beta1.md
Rozzii: Yep the work tickets are fine, I see that my comment was badly written I actually didn't mean follow the API versions with the branches, I would like to follow this metal3-io/baremetal-operator#1283 branching out at the release of every minor and major version.

lentzi90 (approved):

dtantsur:
@zaneb could you take another look and see if this design can get approved?

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,148 @@
...
+### Risks and Mitigations
+
+None
+
+### Work Items
+
+- Refactor the BMO code to avoid referring directly to the current version.
+- Add `Architecture` to the current version.
+- Copy type definitions and modify them.
+- Create a conversion webhook to change between versions.
```
zaneb, line 94: It seems like we're missing a step here, to modify the controller to operate directly on the new API version instead of the old one.

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,148 @@
...
+
+We'll need to test both API versions. Since we already have tests for different
+versions of CAPM3, we will tie them to different versions of BMH. For instance,
+the current version of CAPM3 will switch to the new API version, while all
+stable branches will keep using `v1alpha1`. We'll only need to add one of
+such stable jobs to BMO.
+
+### Upgrade / Downgrade Strategy
+
+The conversion webhook will take care of in-flight CR upgrade:
```
zaneb, line 112: Also reading HardwareDetails from the separate CRD when converting back to v1alpha1.

Rozzii:
Hi @zaneb  @dtantsur 

I would like to introduce `maxSixeGigabye` rootDeviceHint field as part of the v1alpha1 stabiliztaion process,
I have detailed my plan here: https://github.com/metal3-io/baremetal-operator/issues/1308.
Please take a look and let me know about your opinion.

dtantsur:
@zaneb and others: the new revision has changed the approach to the new version by making the transition smoother and ensuring conversion without losses. Please review.

@Rozzii I welcome this, but this can be done without a new version since it's 100% compatible, no?

design/baremetal-operator/bmh-v1beta1.md
dtantsur: I don't believe it's possible, but it's probably also not required if 1) the storage version is v1alpha1, 2) the controller keeps syncing the separate CR and Status.

Rozzii:
> @zaneb and others: the new revision has changed the approach to the new version by making the transition smoother and ensuring conversion without losses. Please review.
> 
> @Rozzii I welcome this, but this can be done without a new version since it's 100% compatible, no?

Yes I believe this would be 100% backward compatible , but I would still modify the API with this I thought all change to the API would cause a version bump.

Rozzii:
I will paste here a link, just to better highlight the parallel discussion about the branching https://github.com/metal3-io/metal3-docs/pull/345
@zaneb @dtantsur @kashifest

zaneb:
For me the question in https://github.com/metal3-io/metal3-docs/pull/332/files#r1245928458 is still not answered by this. The part about the storage version is now explicit, but there's no mention of when the controller starts using the new version.

In fact there's no mention at all of changing the version the controller uses, despite all the preliminary work that has gone into it ("Refactor the BMO code to avoid referring directly to the current version"). I gather that the backward-compatibility changes included here mean we can theoretically continue indefinitely using the older client. So it sounds like we're signing future-us up for a big-bang rewrite of the controller at some unknown time?

dtantsur:
@zaneb updated, PTAL.

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,192 @@
...
+- Refactor the BMO code to avoid referring directly to the current version.
+- Add `Architecture` to the current version.
+- Add `empty` hardware profile.
+- Add `auto` checksum type.
+- Copy type definitions and modify them.
+- Create a conversion webhook to change between versions.
+- Make a new BMO release and communicate the change to the community.
+
+In the future: create a BMO branch and delete `v1alpha1` afterwards.
+This will involve switching the storage format and the controller version
```
zaneb, line 119: Do we have to switch these both at the same time, or could there be a period where the v1alpha1 version still exists in the CRD but the storage format is v1beta1 and the controller uses v1beta1? Obviously any v1alpha1-only features would no longer work after this point, but this might be better than ditching the old version.

In fact, I think this may be essential - in order to migrate existing CRs to the new version we first have to switch the storage format and then rewrite each CR. I think this may be impossible if the storage format changes at the same time that the old version disappears.

zaneb (commented):
OK, I'm happy that this makes explicit how we expect to handle the rest of the transition now.

design/baremetal-operator/bmh-v1beta1.md
dtantsur: Would you like it to be a lasting state (i.e. something operators may potentially be exposed to)? I guess yes, because they'll need to run storage migration (which is something, as far as I understand, that operators run themselves).

zaneb: That's consistent with my understanding, yes.

lentzi90: > could there be a period where the v1alpha1 version still exists in the CRD but the storage format is v1beta1 and the controller uses v1beta1

This is definitely what we should do! The common way to do it is to have a conversion webhook. It can convert everything to the "hub" version (usually the latest, so v1beta1 for us) and it can also convert from the hub version to any other supported version. External tools, humans, etc can use whatever version they want by asking for it as long as it is still supported, but BMO itself would only operate on the hub.

dtantsur: @lentzi90 that's not about the conversion webhook, we already have it planned. The problem is: the new version does not have some fields. Once we switch the controller to it, there is no way to handle these fields any more. This will be a feature loss while `v1alpha1` still exists.

lentzi90: Understood. Then perhaps v1alpha1 needs to be the hub version so the controller can continue working with that?

lentzi90: To add some more context: I think it is required that at least 2 versions are supported at the same time in order to smoothly transition from one to the other. This is how it has been done for cert-manager and Cluster API at least. The thing is, all objects must switch the storage versions and this can only be done when the controller can deal with both versions as far as I know. The steps are like this:

1. BMO with v1alpha1 running. There are BMHs in the cluster obviously stored at v1alpha1.
2. BMO is upgraded to handle v1beta1 *but the existing BMHs are still v1alpha1*. At this point the storage version can be set to v1beta1 and then all BMHs must be "refreshed" in the API to update the stored data.
3. When all objects have been stored at v1beta1 it is safe to drop the v1alpha1 version completely (usually with a new update to the controller and CRDs)

dtantsur: > BMO is upgraded to handle v1beta1 but the existing BMHs are still v1alpha1.

Yep, that's what Zane is talking about. At this stage, v1alpha1 still exists, but features that only exist in that old version no longer work.

> all BMHs must be "refreshed" in the API to update the stored data

Do you know how that works, by the way?

zaneb: >>  all BMHs must be "refreshed" in the API to update the stored data
>
> Do you know how that works, by the way?

The storage version is updated on write. So either you wait for everything to get changed by the user or you build a little program that reads out every CR and writes it back again.

lentzi90: Exactly! I know CAPI and Cert-manager has both implemented this already so we can take notes from that. They did it through their CLIs though so we may need something different.
Ref for CAPI: https://github.com/kubernetes-sigs/cluster-api/issues/6691#issuecomment-1167607669

design/baremetal-operator/bmh-v1beta1.md
```diff
@@ -0,0 +1,204 @@
...
+  that are not part of `v1beta1`
+
+- Create a migration script that reads `v1alpha1` resources and converts them
+  to `v1beta1` ones. This is to ensure that each resource is written at least
+  once with the new storage version.
+
+- Create a new BMO release to get this change to the operators. Instruct them
+  to run the migration script or migrate the objects themselves.
+
+- Delete `v1alpha1` from the supported versions, remove the conversion webhook.
```
zaneb, line 133: I'm not clear on why we need to do this? We could leave that code ~forever without it hurting anything, so why force a rapid change that requires a manual migration within 3-6 months?

dtantsur: @Rozzii @kashifest @lentzi90 what's your opinion on this thread?

lentzi90: The above looks alright to me. I interpret it as just describing how the old version will be phased out (not when).
However, line 135 below changes that a bit. Is the idea to phase out and remove `v1alpha1` in 3-6 months after `v1beta1`? That sounds like a bad idea to me.

I would expect at least a couple of releases in between with both API versions supported. Commonly it has been done so that the old version is first marked as deprecated (but still fully functional) to give a heads up to users. Then in some later release there is a big warning and we stop serving it. At this point, it is still possible to enable it again just through `served: true/false` in the CRD, so ignorant users will get a big scare, but can easily be "saved". Then in the next release the old API version would be completely removed from the code base.

kashifest: A good example of how we plan to remove v1a5 api from CAPM3 is presented here https://github.com/metal3-io/cluster-api-provider-metal3/issues/971. I agree with @dtantsur that we can decide this as a separate issue/discussion.

dtantsur:
Okay, the new revision leaves the actual code removal an open question. @zaneb @kashifest @lentzi90 could you please take another look and lgtm/approve? I think that was the last concern.

PR #339, opened 2023-07-20
Developer environment intro

SafeEHA:
Created developer environment


elfosardo (changes requested):
the overview.md should be renamed to tryit.md and the duplicate dir tryit should be removed

PR #344, opened 2023-08-17
Updates to try-it page

SafeEHA:


dtantsur:
While you're here, could you change the title of the page? "Getting started with Metal3" makes people assume it's for production. Maybe just "Trying Metal3 on a development environment".

/ok-to-test

PR #350, opened 2023-08-30
Attach non-bootable iso to BareMetalHost proposal

hroyrh:
This proposal is for a new feature that allows for attaching a non-bootable iso image to a BMH after it has been provisioned.

### DIFF
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -56,6 +56,13 @@ to attach a non-bootable iso image to hosts.
 We may consider adding support for HTTP credentials and TLS settings in the future
 but it's outside the scope of this proposal.
 
+As part of this proposal, the DataImage attachment will happen after the next reboot
+because many hardwares need a reboot for attaching an image. For example, Dell iDrac
+usually operates in terms of tasks, so once we create a task to attach a virtual
+media, it will wait for the next reboot.
+But we may add such an option in the future, something like - `immediate: true`,
+to immediately attach the DataImage without waiting for reboot, for compatible hardware.
+
 **Note** : As part of this proposal, the dataImage will be attached after the node
 has been Provisioned or ExternallyProvisioned.
 
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -89,12 +89,14 @@ webhook level to validate if a given driver supports the non-bootable ISO featur
 
 As part of the design we are assuming that to attach an image to a BMH object, the user
 will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
-`BareMetalHost` object) and the BMH Controller will add the ownerReference for the corresponding BMH
-to it (example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)). Then, on the next reboot/power on
-of the BMH, the `DataImage` will be attached and the `BareMetalHost` Status will be updated to reflect
-the status of attachment, including which image was attached. In case of detachment of ISO, similar process
-will be followed where the actual detachment will happen on the next reboot/power on and the BMH Status
-will reflect the status of detachment, example : success, number of failures.
+`BareMetalHost` object). The BMH Controller will add the ownerReference for the corresponding BMH
+(example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)) and it will also add a finalizer to the `DataImage` after its attached.
+Then, on the next reboot/power-on of the BMH, the `DataImage` will be attached and the `DataImage`
+`Status` will be updated to reflect the status of the attachment, including which image was attached.
+In case of detachment of the ISO(`DataImage`), similar process will be followed where the actual
+detachment will happen on the next reboot/power-on and the `DataImage` Status will reflect the status
+of detachment, example : success, number of failures. The finalizer will also be removed once the
+detachment succeeds.
 
 Since the attachment of the ISO happens when a BMH reconcile is triggered as a result of BMH
 reboot/power on, so handling the attachment/detachment of dataImage when the host reaches steady state (refer [`actionManageSteadyState` function](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/controllers/metal3.io/baremetalhost_controller.go#L1414) )
@@ -102,8 +104,8 @@ makes sense since we only reach this state when a host has been either provision
 any power state changes (including reboot via rebootAnnotation) are also handled after the host reaches this state
 ( refer [`manageHostPower` function](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L222) )
 
-We need to cache the attached image along with the attachment status in `Status`
-of the BareMetalHost to know if the image has been attached successfully. This
+We need to cache the attached image along with the attachment status in the `Status`
+of the `DataImage` to know if the image has been attached successfully. This
 will also help us determine if a different image was attached previously which
 needs to be detached first before attaching the new dataImage. Or, if the user
 deletes the dataImage CR and the `Status` contains a cached entry, which will
@@ -112,7 +114,8 @@ again trigger detachment of that image.
 In case the image fails to attach, we will retry until either the attachment succeeds
 or the user removes the dataImage CR. In case of failure in attachment, we will
 always do a detachment before the next retry, and record the reason for failure in the
-logs.
+logs. In case of persistent failures, we will retry with increasing delays until the
+attachment/detachment succeeds.
 
 ## Dependencies
 
@@ -122,12 +125,12 @@ block the development of the BMO API.
 
 ## Upgrade / Downgrade Strategy
 
-This feature is adds a Custom Resource(CR) to the cluster and a label to `BareMetalHost`
-which doesn't makes any changes to the BareMetalHost API, so all the existing
-`BareMetalHost` definitions should continue to work as before.
+This feature adds a Custom Resource(CR) to the cluster which doesn't make
+any changes to the `BareMetalHost` `Spec`, so all the existing `BareMetalHost`
+definitions should continue to work as before.
 
 This feature will become available after upgrade, and downgrade is also pretty
-straightforward where the user just need to remove the `DataImage` CRs.
+straightforward where the user just needs to remove the `DataImage` CRs.
 
 Explicit microversion negotiation support has been added to Gophercloud ([pr](https://github.com/gophercloud/gophercloud/pull/2791))
 which can be used in case we don't want to use the master branch for Ironic.
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -1,0 +1,73 @@
+# Add non-bootable iso attach API to BareMetalHost
+
+## Status
+
+Implementable
+
+## Summary
+
+Add a new field in BareMetalHost so that it is possible to attach a generic
+non-bootable iso image via Ironic, after the host has been provisioned.
+
+## Motivation
+
+In certain scenarios it is desirable to attach a non-bootable iso image
+to a system after provisioning has finished, for example :
+
+As a config data disk to do some post provisioning configuration for the
+system.
+
+For troubleshooting of the node.
+
+### Goals
+
+Expose the Ironic API to attach non-bootable iso images via Metal3.
+
+## Proposal
+
+### BMO Proposal
+
+Add a new optional section in the BareMetalHost spec containing details of the
+non-bootable iso image, example :
+
+```yaml
+  spec:
+    dataImage:
+      url: http://1.2.3.4/image.iso
+      username: metal
+      password: password
+      insecure: false
+```
+
+Note that in this case, the iso image will be attached after the node
+has been provisioned or ExternallyProvisioned.
+
+### Ironic Proposal
+
+The [proposal](https://bugs.launchpad.net/ironic/+bug/2033288) for exposing this functionaility in Ironic has been
+accepted by the upstream Ironic community.
+
+## Design Details
+
+Add an optional section `DataImage` to `BareMetalHostSpec`, with sub-fields
+`Url`, `Username`, `Password`, `Insecure`.
+
+## Dependencies
+
+There is a dependency on the implementation(based on the proposal shared
+above) on Ironic side to finish for this feature to work. But this doesn't
+block the development of the BMO API.
+
+## Upgrade / Downgrade Strategy
+
+This feature is added to the BareMetalHost API as an optional field, so all the
+existing BareMetalHost definitions should continue to work as before.
+
+This feature will become available after upgrade, and downgrade is also
+possible as long as the corresponding fields are not used.
+
+## Alternatives
+
+We can also achieve this using the existing `BMH.Spec.Image` field, but this
+attempts to change the boot order of the system and relies on the host to
+fallback to the installed system when booting the image fails.
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -77,7 +77,7 @@ and `Namespace` as the `BareMetalHost` where we are attaching the image.
 To attach an ISO, first the `DataImage` CR object is created and then it is attached to the
 BMH on its next reboot/power on.
 
-The `BareMetalHostStatus` will cache the image details including the attachment
+The `DataImage` `Status` will cache the image details including the attachment
 status, which will inform us if the attachment succeeded and can also be used
 for detachment - either when the attachment fails and we retry or when the user
 explicitly requests detachment by deleting the `DataImage` CRD that corresponds to a BMH.
@@ -95,8 +95,19 @@ Then, on the next reboot/power-on of the BMH, the `DataImage` will be attached a
 `Status` will be updated to reflect the status of the attachment, including which image was attached.
 In case of detachment of the ISO(`DataImage`), similar process will be followed where the actual
 detachment will happen on the next reboot/power-on and the `DataImage` Status will reflect the status
-of detachment, example : success, number of failures. The finalizer will also be removed once the
+of detachment, example : number of failures. The finalizer will also be removed once the
 detachment succeeds.
+Here is an example of what the `DataImage` `Status` might look like:
+
+```yaml
+status:
+  lastReconciliation: "2024-01-01T12:00:00Z"
+  error:
+    count: 0
+    message: ""
+  attachedImage:
+    url: "http://example.com/images/dataimage.iso"
+```
 
 Since the attachment of the ISO happens when a BMH reconcile is triggered as a result of BMH
 reboot/power on, so handling the attachment/detachment of dataImage when the host reaches steady state (refer [`actionManageSteadyState` function](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/controllers/metal3.io/baremetalhost_controller.go#L1414) )
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -7,7 +7,7 @@ Implementable
 ## Summary
 
 Add a new field in BareMetalHost so that it is possible to attach a generic
-non-bootable iso image via Ironic, after the host has been provisioned.
+non-bootable ISO (a CD "ISO 9660" image) image via Ironic, after the host has been provisioned.
 
 ## Motivation
 
@@ -25,7 +25,8 @@ Expose the Ironic API to attach non-bootable iso images via Metal3.
 
 ### Non-Goals
 
-Supporting non-ISO images such as a USB stick.
+Supporting non-ISO images i.e. media types other than CD/DVD, supporting
+drivers other than Redfish and its derived drivers.
 
 ## Proposal
 
@@ -33,6 +34,16 @@ Supporting non-ISO images such as a USB stick.
 
 Add a new optional section in the BareMetalHost spec containing details of the
 non-bootable iso image, example :
+
+#### Current design
+
+```yaml
+  spec:
+    dataImage:
+      url: http://1.2.3.4/image.iso
+```
+
+#### Desired future design
 
 ```yaml
   spec:
@@ -56,8 +67,8 @@ data:
   password: cGFzc3dvcmQ=
 ```
 
-Note that in this case, the iso image will be attached after the node
-has been provisioned or ExternallyProvisioned.
+**Note** : As part of this proposal, the dataImage will be attached after the node
+has been Provisioned or ExternallyProvisioned.
 
 ### Ironic Proposal
 
@@ -74,23 +85,29 @@ The ISO will be attached when the host is in either the `StateProvisioned` or
 the `StateExternallyProvisioned` state before `actionManageSteadyState` is
 called to maintain the State and manage the host power status.
 
-We want to attach the non-bootable ISO without an extra reboot.
+We want to attach the non-bootable ISO without an extra reboot when both
+attaching the ISO and reboot are requested at the same time.
 
 ### Implementation Details
 
-**ToDo** : Explicitly and clearly define the order of actions for the
-below scenarios.
+As part of the design we are assuming that the user will edit the BareMetalHost
+to add the dataImage spec, and might also request reboot via rebootAnnotation
+or power state change via `online` field.
 
-What should be the order of action for attaching the ISO without an extra
-reboot(i.e attaching when the host is poweredoff), when :
+We want to attach the dataImage first, so handling the attachment/detachment of dataImage
+inside the Reconcile function, after reconciling host data and updating hardware details ([code reference](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L152) ), makes sense since
+any power state change (including reboot via rebootAnnotation) will be handled later on via the stateMachine ( [code reference](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L222) ).
 
-**Scenario 1** : `externallyProvisioned: true, online: true, dataImage: <set>`
+We need to cache the attached image in `Status` of the BareMetalHost to know if the
+image has been attached. This will also help us determine if a different image is
+attached previously which needs to be detached first before attaching the new
+dataImage. Or if the user delets the dataImage spec and the `Status` contains a
+cached entry, which will again trigger detachment of that image.
 
-* Proposed workflow : "attach dataImage" -> ManageSteadyState
-
-**Scenario 2** : `image: <set>, online: true, dataImage: <set>`
-
-* Proposed workflow : "mount image" -> provision -> "provisioning finished"-> "detach image" -> "attach dataImage" ->  ManageSteadyState
+In case the image fails to attach, we will retry until either the attachment succeeds
+or the user removes the dataImage spec. In case of failure in attachment, we will
+always do a detachment before the next retry, and record the reason for failure in the
+logs.
 
 ## Dependencies
 
@@ -105,6 +122,9 @@ existing BareMetalHost definitions should continue to work as before.
 
 This feature will become available after upgrade, and downgrade is also
 possible as long as the corresponding fields are not used.
+
+Explicit microversion negotiation support has been added to Gophercloud ([pr](https://github.com/gophercloud/gophercloud/pull/2791))
+which can be used in case we don't want to use the master branch for Ironic.
 
 ## Alternatives
 
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -6,13 +6,14 @@ Implementable
 
 ## Summary
 
-Add a new Custom Resource(CR), `DataImage`, to the cluster which will make it possible,
-using ownerReferences (example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)) and same `Name` and `Namespace` as the `BareMetalHost`, to attach a generic non-bootable ISO (a CD "ISO 9660" image) image via Ironic, after the host has been provisioned.
+Add a new Custom Resource(CR), `DataImage`, to the cluster to attach a
+generic non-bootable ISO (a CD "ISO 9660" image) image to a provisioned
+or externallyProvisioned BMH using Ironic.
 
 ## Motivation
 
 In certain scenarios it is desirable to attach a non-bootable iso image
-to a system after provisioning has finished, for example :
+to a system after provisioning has finished, for example:
 
 A configuration ISO can be used to provide first-boot configuration to a host
 deployed from a standard image in advance. An operator may have a pool of
@@ -25,8 +26,8 @@ Expose the Ironic API to attach non-bootable iso images via Metal3.
 
 ### Non-Goals
 
-Supporting non-ISO images i.e. media types other than CD/DVD, supporting
-drivers other than Redfish and its derived drivers.
+* Supporting non-ISO images i.e. media types other than CD/DVD
+* Supporting drivers other than Redfish and its derived drivers.
 
 ## Proposal
 
@@ -44,8 +45,7 @@ metadata:
   name: metal-worker-0
   namespace: metal-cluster
 spec:
-  dataImage:
-    url: http://1.2.3.4/image.iso
+  url: http://1.2.3.4/image.iso
 ```
 
 We will use ownerReferences in the `DataImage` CR to link it to the corresponding `BareMetalHost` object.
@@ -71,34 +71,35 @@ which in this proposal is Redfish.
 
 ## Design Details
 
-Create a new Custom Resource(CR), `DataImage`, with spec field `Url` and same `Name`
+Create a new Custom Resource(CR), `DataImage`, with spec field `Url` and the same `Name`
 and `Namespace` as the `BareMetalHost` where we are attaching the image.
 
-The ISO will be attached when the `DataImage` CR object is created and the
-corresponding ownerReference is added for the `BareMetalHost` object,  which will trigger a
-reconciliation. In case a reboot is requested at the same time using
-`RebootAnnotation`, the ISO will be attached first.
+To attach an ISO, first the `DataImage` CR object is created and then it is attached to the
+BMH on its next reboot/power on.
 
 The `BareMetalHostStatus` will cache the image details including the attachment
 status, which will inform us if the attachment succeeded and can also be used
 for detachment - either when the attachment fails and we retry or when the user
 explicitly requests detachment by deleting the `DataImage` CRD that corresponds to a BMH.
 
-We will introduce a new flag, `NonBootableISOImage`, at the webhook level to validate if a given driver supports the
-non-bootable ISO feature.
+We will introduce a new flag, `NonBootableISOImage` (under `pkg/hardwareutils/bmc`), at the
+webhook level to validate if a given driver supports the non-bootable ISO feature.
 
 ### Implementation Details
 
 As part of the design we are assuming that to attach an image to a BMH object, the user
 will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
-`BareMetalHost` object) with `OwnerReference` to the BMH object.
-At the same time and independently, the user might also request a reboot via rebootAnnotation
-or power state change via `online` field.
+`BareMetalHost` object) and the BMH Controller will add the ownerReference for the corresponding BMH
+to it (example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)). Then, on the next reboot/power on
+of the BMH, the `DataImage` will be attached and the `BareMetalHost` Status will be updated to reflect
+the status of attachment, including which image was attached. In case of detachment of ISO, similar process
+will be followed where the actual detachment will happen on the next reboot/power on and the BMH Status
+will reflect the status of detachment, example : success, number of failures.
 
-In such scenarios, we want to attach the dataImage first(before the reboot), so handling the
-attachment/detachment of dataImage when the host reaches steady state (refer [`actionManageSteadyState` function](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/controllers/metal3.io/baremetalhost_controller.go#L1414) )
-makes sense since we only reach this state when a host has been either provisioned or externallyProvisioned, and also
-any power state change (including reboot via rebootAnnotation) are also handled after the host reaches this state
+Since the attachment of the ISO happens when a BMH reconcile is triggered as a result of BMH
+reboot/power on, so handling the attachment/detachment of dataImage when the host reaches steady state (refer [`actionManageSteadyState` function](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/controllers/metal3.io/baremetalhost_controller.go#L1414) )
+makes sense since we only reach this state when a host has been either provisioned or externallyProvisioned, and
+any power state changes (including reboot via rebootAnnotation) are also handled after the host reaches this state
 ( refer [`manageHostPower` function](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L222) )
 
 We need to cache the attached image along with the attachment status in `Status`
--- a/design/baremetal-operator/bmh_non-bootable_iso.md
+++ b/design/baremetal-operator/bmh_non-bootable_iso.md
@@ -77,16 +77,17 @@ accepted by the upstream Ironic community.
 
 ## Design Details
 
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,73 @@
...
+
+Add a new optional section in the BareMetalHost spec containing details of the
+non-bootable iso image, example :
+
+```yaml
+  spec:
+    dataImage:
+      url: http://1.2.3.4/image.iso
+      username: metal
+      password: password
```
dtantsur, line 38: username/password have to be inside a secret

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,73 @@
...
+### Goals
+
+Expose the Ironic API to attach non-bootable iso images via Metal3.
+
+## Proposal
+
+### BMO Proposal
+
+Add a new optional section in the BareMetalHost spec containing details of the
+non-bootable iso image, example :
```
dtantsur, line 31: Do we want to support non-ISO images, e.g. a USB stick? If not, please put to the non-goals section (which I don't see here but it's in the template).

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Motivation
+
+In certain scenarios it is desirable to attach a non-bootable iso image
+to a system after provisioning has finished, for example :
+
+As a config data disk to do some post provisioning configuration for the
+system.
+
+For troubleshooting of the node.
```
dtantsur, line 20: This is unclear, please elaborate or remove.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Ironic Proposal
+
+The [proposal](https://bugs.launchpad.net/ironic/+bug/2033288) for exposing this functionaility in Ironic has been
+accepted by the upstream Ironic community.
+
+## Design Details
+
+Add an optional section `DataImage` to `BareMetalHostSpec`, with sub-fields
+`Url`, `Username`, `Password`, `Insecure`.
```
dtantsur, line 53: This information is clear from the previous sections. What I'd like to see here are some details on how the new operation will interact with other BMH states. E.g. when I create a BMH with both `image` and `dataImage` set, at which point is the data image connected?

If I create an externally provisioned host with `online: true` and `dataImage`, we want to make sure we connect the image *before* we try to power on the host.

And so on.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: I was thinking the user might want to run some validations and have some tools to debug, collect logs. Does that make sense ?

dtantsur: The ISO is not bootable, right? So you still need access to the operating system to run any tools from it. If you have it, why do you need an ISO?

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
+# Add non-bootable iso attach API to BareMetalHost
+
+## Status
+
+Implementable
+
+## Summary
+
+Add a new field in BareMetalHost so that it is possible to attach a generic
+non-bootable iso image via Ironic, after the host has been provisioned.
```
dtantsur, line 10: nit: I suspect we all understand what "ISO" means (it's all upper-case, by the way), but maybe we refer to "an ISO 9660 image" or "a CD (ISO 9660) image".

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+powered off hosts with operating system already provisioned. Then, adding
+a host to the cluster boils down to powering it on with a configuration ISO provided.
+
+### Goals
+
+Expose the Ironic API to attach non-bootable iso images via Metal3.
+
+### Non-Goals
+
+Supporting non-ISO images such as a USB stick.
```
dtantsur, line 28: nit: to be strictly correct, "images for media types other than CD/DVD"

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+
```
dtantsur, line 29: Another non-goal for now is to support drivers other than Redfish and the drivers derived from it.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+### BMO Proposal
+
+Add a new optional section in the BareMetalHost spec containing details of the
+non-bootable iso image, example :
+
+```yaml
+  spec:
+    dataImage:
+      url: http://1.2.3.4/image.iso
+      credentialsName: dataimage-secret
```
dtantsur, line 41: The current Ironic patch dropped credentials and disableCertificateVerification because of complexity with their implementation. I suggest we mark this example as the desired future state, but for now only propose the "url" field.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+The ISO will be attached when the host is in either the `StateProvisioned` or
+the `StateExternallyProvisioned` state before `actionManageSteadyState` is
+called to maintain the State and manage the host power status.
+
+We want to attach the non-bootable ISO without an extra reboot.
+
+### Implementation Details
+
+**ToDo** : Explicitly and clearly define the order of actions for the
+below scenarios.
```
dtantsur, line 82: Needs finalizing.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+
+Add an optional section `DataImage` to `BareMetalHostSpec`, with sub-fields
+`Url`, `CredentialsName`, `DisableCertificateVerification` and a secret
+for dataImage with name `CredentialsName`.
+
+The ISO will be attached when the host is in either the `StateProvisioned` or
+the `StateExternallyProvisioned` state before `actionManageSteadyState` is
+called to maintain the State and manage the host power status.
+
+We want to attach the non-bootable ISO without an extra reboot.
```
dtantsur, line 77: What we need to ensure is that if an operator requests a non-bootable ISO at the same time with a reboot, we first attach the ISO then reboot.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+
```
dtantsur, line 78: This proposal does not say anything about two topics:

1) When the image will be detached. To even make detachments possible, we need to cache the attached image in `Status`.

2) How to handle the fact that not all drivers support the attachment API.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,113 @@
...
+block the development of the BMO API.
+
+## Upgrade / Downgrade Strategy
+
+This feature is added to the BareMetalHost API as an optional field, so all the
+existing BareMetalHost definitions should continue to work as before.
+
+This feature will become available after upgrade, and downgrade is also
+possible as long as the corresponding fields are not used.
+
```
dtantsur, line 108: One missing topic here is the one of Ironic API versions. The new API will be introduced in git master soon, but we have a request from the community to not make master versions of Ironic immediately required. I.e. we need a way to support Ironic from the 2023.2 "Bobcat" branch.

We have been working on explicit microversion negotiation support in GopherCloud in https://github.com/gophercloud/gophercloud/pull/2791. We should describe what happens if the required API version is not in Ironic.

design/baremetal-operator/bmh_non-bootable_iso.md
dtantsur: Actually,

3) What happens on a failure?

Rozzii:
I haven't forgot about this, but I had no time to read it yet.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Added that non CD/DVD media types and non Redfish or derived drivers are not supported.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Added current design section, and desired future design.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Added this in implementation details below, should I add it under the design section instead.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,134 @@
...
+#### Current design
+
+```yaml
+  spec:
+    dataImage:
+      url: http://1.2.3.4/image.iso
+```
+
+#### Desired future design
+
```
dtantsur, line 47: Some words here about this design depending on not yet planned Ironic features and is not a blocker for delivering the whole proposal.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,134 @@
...
+```yaml
+  spec:
+    dataImage:
+      url: http://1.2.3.4/image.iso
+      credentialsName: dataimage-secret
+      disableCertificateVerification: false
+```
+
+Secret holding the image url credentials, base64 encoded :
```
dtantsur, line 56: nit: this is redundant, we know how to create secrets.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: dataimage-secret
+type: Opaque
```
dtantsur, line 64: nit: kubernetes.io/basic-auth actually, but that does not matter in practice.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Removed the specific spec details, and replaced with mention of possible future support.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: removed this part

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: removed this part

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,124 @@
...
+The `BareMetalHostStatus` will cache the image details, which will inform
+us if the attachment succeeded and will also be used for detachment, either
+when the attachment fails and we retry or when the user explicitly requests
+detachment by removing the `DataImage` spec from `BareMetalHost` object.
+
+A flag will be used at the webhook level to validate if a given driver supports the
+non-bootable ISO feature. Since the support for attaching an ISO as virtual media is
+equivalent for both `PreprovisioningISOImage` and `NonBootableISOImage`, we plan to
+rename the existing flag for `PreprovisioningISOImage` i.e. `SupportsISOPreprovisioningImage`
+to `SupportVirtualMediaISOImage` and use it for both the cases.
```
Rozzii, line 80: I am not sure about this, why would it be a problem to have separate flags? PreProvisioningISO, VMEDIA and non bootable ISO are all based on VMEDIA but these are very different use cases.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Implementation Details
+
+As part of the design we are assuming that the user will edit the BareMetalHost
+to add the dataImage spec, and might also request reboot via rebootAnnotation
+or power state change via `online` field.
+
+We want to attach the dataImage first, so handling the attachment/detachment of dataImage
```
Rozzii, line 88: Should we block the attachment of a DataImage for provisioned BMHs explicitly?
I think yes, or at least have an annotation or flag or some option to disable this for provisioned BMHs, because this seems to a me a potential source of issues.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,124 @@
...
+inside the Reconcile function, after reconciling host data and updating hardware details ([code reference](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L152) ), makes sense since
+any power state change (including reboot via rebootAnnotation) will be handled later on via the stateMachine ( [code reference](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L222) ).
+
+We need to cache the attached image in `Status` of the BareMetalHost to know if the
+image has been attached. This will also help us determine if a different image is
+attached previously which needs to be detached first before attaching the new
+dataImage. Or if the user delets the dataImage spec and the `Status` contains a
+cached entry, which will again trigger detachment of that image.
+
+In case the image fails to attach, we will retry until either the attachment succeeds
```
Rozzii, line 98: Will this cause the state of the BMH to go to "provisioning error" or would this be a different error state? or would be in general considered a blocking error?

Rozzii (changes requested):
This proposal looks good in general, the feature proposed here would be a great addition IMO.
I have no issues with the general approach proposed here but, I would like to ask for some qualifications related to some implementation details.

Rozzii:
/hold
Putting this on hold just until I get some clarifications for my questions.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: The attachment of detachment of DataImage will happen during reconciliation of the BMH, so in case of a failure it will raise an error in the logs and requeue the request similar to what we do when [fetching the host here](https://github.com/metal3-io/baremetal-operator/blob/63abe2db90306baa805680ec391b055d9fe5be31/controllers/metal3.io/baremetalhost_controller.go#L115).

design/baremetal-operator/bmh_non-bootable_iso.md
dtantsur: Note that when it comes to Ironic, the difference is very slight.

On the other hand, there may be drivers that don't have the non-bootable ISO API implemented. So maybe a different flag indeed.

design/baremetal-operator/bmh_non-bootable_iso.md
dtantsur: Why so? It's actually one of the potential use cases of this API (although not literally the one we're looking at).

Rozzii: Well, until now we didn't really had any functionality that would modify the content of a provisioned BMH, we are able to detach it and re-provision/clean it but no functionality that would modify the block devices/data present on a machine without re-provisioning. 

In many scenarios I agree that on the fly media attachment could be a useful feature but then there could be other scenarios where users would like to make sure that  no block device gets attached to their node not even by accident after it was provisioned. 

A very simple example is why you could disable USB and other I/O ports on a computer and  put a password on the BIOS, there are operating environments where the possibility of attaching any additional I/O device is viewed as a potential attack vector. 

I agree with you that we should allow attaching non bootable iso to the already provisioned machines, but I would really like an option to disable this for people who see this as a potential attack vector.

dtantsur: > Well, until now we didn't really had any functionality that would modify the content of a provisioned BMH

There is definitely interest in changing BIOS  settings and doing firmware upgrades for deployed hosts, btw. Ironic is getting there in terms of features.

> the possibility of attaching any additional I/O device is viewed as a potential attack vector.

If someone has access to your BMH, your security is already breached.

Rozzii: Yes but I don't only mean this for security reasons, for example there could be a user error and someone accidentally updates the BMH resource (e.g. right now BMO sees all BMH resources in all namespaces ) , my main reason would be to have the option producing "finalized" BMHs.

dtantsur: It's a valid concern, but how is it specific to this feature? You can accidentally deprovision a BMH by nuking its image field.

Doesn't detached state solve the issue?

Rozzii: You are totally right that user error could cause deprovisioning, or re-inspection or something, the issue I see here is that switching the block device (blk from now on) to something else or injecting a blk on the machine rom an other user (for example in multy cluster/user environments) could cause the following issues: sharing the data of an other custommer/user, damaging data integrity in general without noticing ( an accidentally attached block device could modify the behavior of the business logic of the end user even).

Until now  our errors  either caused the machine to not upgrade, not provision or the user might have accidentally deprovisioned or "broke" the machine but with this feature we open up a whole host of data integrity related questions.

Detached annotation only partially solves this problem as that would remove the BMH from management which causes loss of functionality (sure you can bring it back but that would add extra steps seems to be a bit heavy handed)  but I am talking about use cases where user would like to manage the machines just as usual, the only thing the user would want is to avoid any threats to data integrity, a machine being accidentally deprovisioned or broken are  "self contained" issues from the perspective of the data on the machine, the data might be lost, but there is no threat of getting someone else's data, or getting data that would effect the workloads and they would do something unwanted, with this feature these scenarios would all become possible.

From my POV an annotation would be fine to block attaching blk and block modifying the attached blk. Ofc I am also open to other ways that we could provide data integrity protection.

TLDR, attaching an incorrect blk to a running BMH might cause way bigger damage than losing that specific machine.

design/baremetal-operator/bmh_non-bootable_iso.md
zaneb: This is all internal from the user's perspective so it's difficult to really care that much what it's called.
Renaming the function will mean that anybody using the hardwareutils module as a library will have to update their code, which is kind of a pain for them.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,124 @@
...
+to `SupportVirtualMediaISOImage` and use it for both the cases.
+
+### Implementation Details
+
+As part of the design we are assuming that the user will edit the BareMetalHost
+to add the dataImage spec, and might also request reboot via rebootAnnotation
+or power state change via `online` field.
+
+We want to attach the dataImage first, so handling the attachment/detachment of dataImage
+inside the Reconcile function, after reconciling host data and updating hardware details ([code reference](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L152) ), makes sense since
```
zaneb, line 89: This doesn't make sense to me :)
Doing it outside of the state machine means it can't take into account any of the other state. e.g. is it valid to do this in the middle of... provisioning? inspecting? I suspect not.

Handling it in `actionManageSteadyState()` would make sense to me.

design/baremetal-operator/bmh_non-bootable_iso.md
zaneb: > Until now our errors either caused the machine to not upgrade, not provision or the user might have accidentally deprovisioned or "broke" the machine but with this feature we open up a whole host of data integrity related questions.

If you have write access to the BMH you can disable cleaning and reprovision with your own image.
For that matter, you could remove this new annotation you're proposing.

I don't think there is any substitute for using RBAC to not give write access to the BMH to users who aren't fully trusted.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,124 @@
...
+any power state change (including reboot via rebootAnnotation) will be handled later on via the stateMachine ( [code reference](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L222) ).
+
+We need to cache the attached image in `Status` of the BareMetalHost to know if the
+image has been attached. This will also help us determine if a different image is
+attached previously which needs to be detached first before attaching the new
+dataImage. Or if the user delets the dataImage spec and the `Status` contains a
+cached entry, which will again trigger detachment of that image.
```
zaneb, line 96: We may need to record the status of the attachment as well, since we can't really add new states to the state machine.

design/baremetal-operator/bmh_non-bootable_iso.md
dtantsur: Yeah. Ironic itself has RBAC that could be used to e.g. allow some users to provision but not attach ISO or change BIOS settings. I guess we don't have anything like that?

Would it help if we introduced a new CR for attaching ISOs? Then an operator could disable everyone's access to such resources?

Rozzii: This CR + RBAC based solution sounds the best so far, I am fine with it ! Thanks!

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,124 @@
...
+
+**Note** : As part of this proposal, the dataImage will be attached after the node
+has been Provisioned or ExternallyProvisioned.
+
+### Ironic Support
+
+The [proposal](https://bugs.launchpad.net/ironic/+bug/2033288) for exposing this functionaility in Ironic has been
+accepted by the upstream Ironic community.
+
+The [Ironic implementation](https://review.opendev.org/c/openstack/ironic/+/894918) based on the above proposalon has been completed.
```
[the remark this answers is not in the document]
hroyrh: Done

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Made changes to design to use CR + Rbac for dataImage

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Now using a separate flag `NonBootableISOImage` for dataImage

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+#### Design spec
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: DataImage
+metadata:
+  name: metal-worker-0
+  namespace: metal-cluster
+spec:
+  dataImage:
```
dtantsur, line 47: `dataImage` here feels redundant, you can put the url right under the spec.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
+# Add non-bootable iso attach API to BareMetalHost
+
+## Status
+
+Implementable
+
+## Summary
+
+Add a new Custom Resource(CR), `DataImage`, to the cluster which will make it possible,
+using ownerReferences (example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)) and same `Name` and `Namespace` as the `BareMetalHost`, to attach a generic non-bootable ISO (a CD "ISO 9660" image) image via Ironic, after the host has been provisioned.
```
dtantsur, line 10: I'd suggest moving ownerReferences and the example to the implementation details: it's too detailed for the summary.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Motivation
+
+In certain scenarios it is desirable to attach a non-bootable iso image
+to a system after provisioning has finished, for example :
```
dtantsur, line 15: nit: redundant space before :

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+powered off hosts with operating system already provisioned. Then, adding
+a host to the cluster boils down to powering it on with a configuration ISO provided.
+
+### Goals
+
+Expose the Ironic API to attach non-bootable iso images via Metal3.
+
+### Non-Goals
+
+Supporting non-ISO images i.e. media types other than CD/DVD, supporting
```
dtantsur, line 28: nit: convert to a list

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+Other than that, we just need the implementation for driver specific support,
+which in this proposal is Redfish.
+
+## Design Details
+
+Create a new Custom Resource(CR), `DataImage`, with spec field `Url` and same `Name`
+and `Namespace` as the `BareMetalHost` where we are attaching the image.
+
+The ISO will be attached when the `DataImage` CR object is created and the
+corresponding ownerReference is added for the `BareMetalHost` object,  which will trigger a
```
dtantsur, line 78: Do you expect a user to add ownerReferences? I don't think it's a common pattern. I think we need the BMH reconciler to generally watch all DataImages.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+reconciliation. In case a reboot is requested at the same time using
+`RebootAnnotation`, the ISO will be attached first.
```
dtantsur, line 80: There is no "at the same time" any more: we have two different resources. Now I'd expect the DataImage to be created first, but only handled by the BMH controller on reboot/power on.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+explicitly requests detachment by deleting the `DataImage` CRD that corresponds to a BMH.
+
+We will introduce a new flag, `NonBootableISOImage`, at the webhook level to validate if a given driver supports the
+non-bootable ISO feature.
+
+### Implementation Details
+
+As part of the design we are assuming that to attach an image to a BMH object, the user
+will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
+`BareMetalHost` object) with `OwnerReference` to the BMH object.
```
dtantsur, line 94: I feel uneasy about a user setting ownerReferences. At the very least, I have never seen it done in Kubernetes.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Makes sense, will update the proposal.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,138 @@
...
+
+The ISO will be attached when the `DataImage` CR object is created and the
+corresponding ownerReference is added for the `BareMetalHost` object,  which will trigger a
+reconciliation. In case a reboot is requested at the same time using
+`RebootAnnotation`, the ISO will be attached first.
+
+The `BareMetalHostStatus` will cache the image details including the attachment
+status, which will inform us if the attachment succeeded and can also be used
+for detachment - either when the attachment fails and we retry or when the user
+explicitly requests detachment by deleting the `DataImage` CRD that corresponds to a BMH.
```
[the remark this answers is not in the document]
zaneb: Either works, but I think we will want the Status thing anyway.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,136 @@
...
+
+To attach an ISO, first the `DataImage` CR object is created and then it is attached to the
+BMH on its next reboot/power on.
+
+The `BareMetalHostStatus` will cache the image details including the attachment
+status, which will inform us if the attachment succeeded and can also be used
+for detachment - either when the attachment fails and we retry or when the user
+explicitly requests detachment by deleting the `DataImage` CRD that corresponds to a BMH.
+
+We will introduce a new flag, `NonBootableISOImage`, at the webhook level to validate if a given driver supports the
```
zaneb, line 85: You mean in pkg/hardwareutils/bmc?

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,136 @@
...
+non-bootable ISO feature.
+
+### Implementation Details
+
+As part of the design we are assuming that to attach an image to a BMH object, the user
+will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
+`BareMetalHost` object). Then, on the next reboot/power on of the BMH, the `DataImage` will
+be attached and the BMH Controller will add the ownerReference for the corresponding BMH
```
zaneb, line 93: We should probably add the ownerReference immediately.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Yes

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Makes sense, will update the proposal.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Updated the proposal

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Right, I have updated the proposal.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Done.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Makes sense, I've updated the proposal.

design/baremetal-operator/bmh_non-bootable_iso.md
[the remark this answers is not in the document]
hroyrh: I have updated the proposal to add a finalizer to dataImage and also cache it in status.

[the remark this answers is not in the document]
dtantsur: I'm not sure what else we would use BMH.Status for. Just to record the fact that a DataImage exists? I assume we can set up the reconciler so that it gets notified when a DataImage is created with the same name as an existing BMH?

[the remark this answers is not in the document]
dtantsur: UPD: yeah, tracking success is probably a strong enough reason. I just don't really want to update BMH status for this work... Maybe it's enough to use DataImage.Status + a finalizer?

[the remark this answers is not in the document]
zaneb: We will certainly need to record in the Status of some CR which image we've attached, whether it worked, and if it didn't how many failures we've had.

Note that the only thing that can be updated atomically with the BMH Status is other things in the BMH Status. So to record this stuff in the DataImage Status instead, we'd need to be completely sure that there's no sequence of e.g. API write errors that could cause us to get stuck in an inconsistent state. Are we?

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,139 @@
...
+webhook level to validate if a given driver supports the non-bootable ISO feature.
+
+### Implementation Details
+
+As part of the design we are assuming that to attach an image to a BMH object, the user
+will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
+`BareMetalHost` object) and the BMH Controller will add the ownerReference for the corresponding BMH
+to it (example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)). Then, on the next reboot/power on
+of the BMH, the `DataImage` will be attached. The BMH Controller will also add a finalizer to
+the `DataImage` once the attachment succeeds. In case of detachment of ISO, similiar process
```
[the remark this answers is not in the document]
hroyrh: Fixed

design/baremetal-operator/bmh_non-bootable_iso.md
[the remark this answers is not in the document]
dtantsur: Forgot to say: I don't believe we need to update anything in the BMH status in a lockstep?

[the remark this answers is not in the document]
hroyrh: I've updated the proposal to use finalizer + dataImage.status and the BMH reconciler can keep track of the dataImage since it will have the same name and namespace.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Design Details
+
+Create a new Custom Resource(CR), `DataImage`, with spec field `Url` and the same `Name`
+and `Namespace` as the `BareMetalHost` where we are attaching the image.
+
+To attach an ISO, first the `DataImage` CR object is created and then it is attached to the
+BMH on its next reboot/power on.
+
+The `BareMetalHostStatus` will cache the image details including the attachment
```
dtantsur, line 80: This part still references the BMH status.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,142 @@
...
+
+As part of the design we are assuming that to attach an image to a BMH object, the user
+will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
+`BareMetalHost` object). The BMH Controller will add the ownerReference for the corresponding BMH
+(example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)) and it will also add a finalizer to the `DataImage` after its attached.
+Then, on the next reboot/power-on of the BMH, the `DataImage` will be attached and the `DataImage`
+`Status` will be updated to reflect the status of the attachment, including which image was attached.
+In case of detachment of the ISO(`DataImage`), similar process will be followed where the actual
+detachment will happen on the next reboot/power-on and the `DataImage` Status will reflect the status
+of detachment, example : success, number of failures. The finalizer will also be removed once the
```
dtantsur, line 98: Success won't be visible: the DataImage object will be deleted on success.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,142 @@
...
+webhook level to validate if a given driver supports the non-bootable ISO feature.
+
+### Implementation Details
+
+As part of the design we are assuming that to attach an image to a BMH object, the user
+will create the corresponding `DataImage` CR (with the same `Name` and `Namespace` as the
+`BareMetalHost` object). The BMH Controller will add the ownerReference for the corresponding BMH
+(example : [HostFirmwareSettings](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/docs/api.md?plain=1#L664)) and it will also add a finalizer to the `DataImage` after its attached.
+Then, on the next reboot/power-on of the BMH, the `DataImage` will be attached and the `DataImage`
+`Status` will be updated to reflect the status of the attachment, including which image was attached.
```
dtantsur, line 95: Could you actually define how the Status is going to look? An example after attachment would also be good.

dtantsur:
@Rozzii this is quite close, could you take another look?

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Fixed

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,154 @@
...
+In case of detachment of the ISO(`DataImage`), similar process will be followed where the actual
+detachment will happen on the next reboot/power-on and the `DataImage` Status will reflect the status
+of detachment, example : number of failures. The finalizer will also be removed once the
+detachment succeeds.
+Here is an example of what the `DataImage` `Status` might look like:
+
+```yaml
+status:
+  state: attached
```
dtantsur, line 104: nit: this is probably redundant: we can derive it from the presence of `attachedImage`

dtantsur:
/hold cancel

I believe all the past concerns have been addressed, and my approval is already in place.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Makes sense, removed it.

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,153 @@
...
+#### Design spec
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: DataImage
+metadata:
+  name: metal-worker-0
+  namespace: metal-cluster
+spec:
+  url: http://1.2.3.4/image.iso
```
carbonin, line 48: Will this URL be used directly with the host or will this image be downloaded and served by ironic similar to how BMH handles images today?

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,153 @@
...
+    message: ""
+  attachedImage:
+    url: "http://example.com/images/dataimage.iso"
+```
+
+Since the attachment of the ISO happens when a BMH reconcile is triggered as a result of BMH
+reboot/power on, so handling the attachment/detachment of dataImage when the host reaches steady state (refer [`actionManageSteadyState` function](https://github.com/metal3-io/baremetal-operator/blob/9ce684e6462a6ab55ff650cb6c11f9ba1ffb395d/controllers/metal3.io/baremetalhost_controller.go#L1414) )
+makes sense since we only reach this state when a host has been either provisioned or externallyProvisioned, and
+any power state changes (including reboot via rebootAnnotation) are also handled after the host reaches this state
+( refer [`manageHostPower` function](https://github.com/metal3-io/baremetal-operator/blob/1bb45eef449c942711b1c0937ecff2b10a326eb3/controllers/metal3.io/baremetalhost_controller.go#L222) )
```
carbonin, line 116: So what is the expected flow?

1. User provisions a node (BMH is `Provisioned`)
2. User creates DataImage
3. User forces a BMH reboot

Or is the DataImage supposed to exist when the BMH moves to the Provisioned or ExternallyProvisioned state?

Alternatively will this work?

1. Create BMH as `ExternallyProvisioned` and `online: false`
2. Create `DataImage`
3. Set BMH `online: true`

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: Are you talking about image registry ? As specified in the proposal, the image should be an ISO image ( i.e. CD "ISO 9660" image), other than that it can be served over network or be downloaded into a registry.

design/baremetal-operator/bmh_non-bootable_iso.md
hroyrh: I think both these workflows should work. Basically, if the BMH has reached the `Provisoned`/`ExternallyProvisioned` state and a valid `DataImage` exists, then on the next reboot the attachment will happen.

design/baremetal-operator/bmh_non-bootable_iso.md
Rozzii: Thanks!

design/baremetal-operator/bmh_non-bootable_iso.md
```diff
@@ -0,0 +1,153 @@
...
+status:
+  lastReconciliation: "2024-01-01T12:00:00Z"
+  error:
+    count: 0
+    message: ""
+  attachedImage:
+    url: "http://example.com/images/dataimage.iso"
+```
+
+Since the attachment of the ISO happens when a BMH reconcile is triggered as a result of BMH
```
Rozzii, line 112: Is there btw some other way to force  attach the ISO without booting/rebooting the machine? Or should we have a way if we don't have one to make the BMH reconcile and attach the medua without reboot? Wouldn't it be better if we would have an annotation, something like "attachIso" to start reconciliation instead of requiring a reboot? It feels a bit excessive to require a full reboot when a non bootable ISO image is attached.

Rozzii (changes requested):
I am okay with almost everything but I have 1 question left.

design/baremetal-operator/bmh_non-bootable_iso.md
dtantsur: It's somewhat risky. A lot of hardware will not attach a device without a reboot. iDRAC tends to operate in terms of tasks: once you create a task to attach a virtual media, this task will be pending until the next reboot. It may interfere with later tasks, e.g. if you try to clean the host.

Maybe we can later extend the DataImage object with something like `immediate: true`? In any case, I'd not include it into the first version, if that's acceptable for you.

design/baremetal-operator/bmh_non-bootable_iso.md
dtantsur: There is an internal cache in Ironic for this process as well IIRC.

design/baremetal-operator/bmh_non-bootable_iso.md
Rozzii: Thank you for the explanation @dtantsur , I am fine with this reasoning but I would like to ask you @hroyrh  to please include this reasoning in the proposal for future reference and please mention that in the future we could implement a method that would allow us to attach block device without restarting, please also include the iDRAC example also.

Issue #258, opened 2022-04-07
User-guide: add diagram to describe the baremetal inspection/provision workflow


djfjeff:
Having a diagram to explain the inspection/provision workflow for Metal3 (similar to https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.2/html-single/bare_metal_provisioning/index#the-bare-metal-provisioning-service) would greatly help understand how it works.

Also, a diagram will help understand what part to troubleshoot when an issue arise in the workflow.

Rozzii:
/triage accepted
/assign @fmuyassarov 
/kind feature

The community has accepted the proposal for a sequence diagram related to nspection/provision workflow.

fmuyassarov:
@djfjeff we will just need to agree with the community on the common tool to draw diagrams, which hopefully will be decided by the next week's community meeting. And then I can submit a patch with the state diagram.

djfjeff:
/remove-lifecycle stale

This is something that would still be useful to have.

djfjeff:
/remove-lifecycle stale

This is something that would still be useful to have. Is there any progress made on this?

djfjeff:
/remove-lifecycle stale

This is something that would still be useful to have. Is there any progress made on this?

Rozzii:
Any additional documentation would be good, I will move this to frozen, no need to always remove the stale label.
/lifecycle frozen

wrkode:
I've created a diagram proposal in https://github.com/metal3-io/metal3-docs/pull/354

PR #354, opened 2023-09-07
proposed Inspection and Provisioning diagram

wrkode:
This PR addresses issue **https://github.com/metal3-io/metal3-docs/issues/258**
Please understand that I'm not (yet) strong with Metal3, but would something like the be technically correct?


wrkode:
@Rozzii could you please review the diagram for technical correctness?

Rozzii:
Thanks for the contribution @wrkode  I will check this soon.

Rozzii:
@wrkode thanks fro creating this nice diagram, a few nits:
- Please highlight that the machine object comes from CAPI and that between the MAchine and the BMH there is a Metal3Machine object.
- You have mentioned  how the inspections is initiated so I would suggest mentioning how the provisioning is initiated (it starts when the image field gets populated)
- There is a typo in the "Start Provisioning" block "etal3" instead of "Metal3"
- If you used draw.io or some other code->image generator please attach the source code also.
cc @lentzi90 @mboukhalfa @kashifest

lentzi90 (commented):
Thanks for working on this!
To keep things simple, I would advice to focus on just the Baremetal Operator level and the BareMetalHost resource for this diagram and not attempt to include Machines or Metal3Machines. Keeping that in mind, I would change the provisioning section so that the first two boxes are merged into one. There I would put something like

> Image information
> The spec.image field is filled with information about the desired provisioned state to start the provisioning process. This can be done by the user directly, or a controller such as CAPM3.

In the `Store Data` box it may be worth mentioning the [HardwareData CRD](https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/hardwaredata_crd.md), I'm not sure.

Rozzii:
> Thanks for working on this! To keep things simple, I would advice to focus on just the Baremetal Operator level and the BareMetalHost resource for this diagram and not attempt to include Machines or Metal3Machines. Keeping that in mind, I would change the provisioning section so that the first two boxes are merged into one. There I would put something like
> 
> > Image information
> > The spec.image field is filled with information about the desired provisioned state to start the provisioning process. This can be done by the user directly, or a controller such as CAPM3.
> 
> In the `Store Data` box it may be worth mentioning the [HardwareData CRD](https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/hardwaredata_crd.md), I'm not sure.

I can get behind this also, then let's skip mentioning Metal3Machine and Machine objects.

wrkode:
I'll have an updated version soon. thanks for the feedback folks.
for this I've use Excalidraw, but I want to use draw-io, so that I can share code.
expect a new version soon.

dtantsur:
The diagram is mostly correct (I don't think inspection depends on the `online` field; there is a typo in metal3), but I'm curious what the purpose of it is. It's not linked in any document, and we probably have documents explaining the similar thing.

wrkode:
> The diagram is mostly correct (I don't think inspection depends on the `online` field; there is a typo in metal3), but I'm curious what the purpose of it is. It's not linked in any document, and we probably have documents explaining the similar thing.

@dtantsur, I was working against https://github.com/metal3-io/metal3-docs/issues/258, but if you and team find this no more relevant, I can stop and close the PR. please do let me know though.

wrkode:
@Rozzii I feel I do need to discuss this in the community/Project call. could you please let me know when this will happen next time?

Rozzii:
@wrkode are you still interested to finish this?

PR #356, opened 2023-09-22
fix dead link with error code 404

Gutopro:
new link provides documentation for security vulnerability of kubletes ability to self-label node, former link resulted in a dead end returning a 404-'page not found' error.
@tuminoid could you please review this?

### DIFF
--- a/design/sync-labels-bmh-to-node.md
+++ b/design/sync-labels-bmh-to-node.md
@@ -240,7 +240,7 @@ this label.
 
 There is also a security vulnerability associated with kubelet's ability to
 self-label the Node object. This has been
-[addressed](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md).
+[addressed](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-isolation-restriction).
 However, as a consequence, BMH labels that fall within the *.kubernetes.io/*
 space cannot be specified via kubelet's `--node-labels` flag.
 


design/sync-labels-bmh-to-node.md
```diff
@@ -240,7 +240,7 @@ this label.
 
 There is also a security vulnerability associated with kubelet's ability to
 self-label the Node object. This has been
-[addressed](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md).
+[addressed](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-isolation-restriction).
```
tuminoid, line 243: Instead of fixing the link, it seems this whole section would need a rewrite. Referenced KEP is long gone, and probably the issue/vulnerability discussed here as well.

Thoughts @zaneb @dtantsur @furkatgofurov7 ?

dtantsur: I'm out of context on this issue, so if you say so.

zaneb: The KEP [moved](https://github.com/kubernetes/enhancements/pull/2208/files#diff-6b82398e6a50098aba701b8ef1e53e653542f91ea335a95e511f8a8c6b18296e) to https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/279-limit-node-access but I don't know anything else about it.

Rozzii: I would propose adding the KEP link that was shown by @zaneb , could you @Gutopro please change to that link?

I am fine with creating an issue to look into the relevance of the mentioned security issue but this PR only aimed to fix the link so if that is fixed I would be fine with merging this.

Rozzii (changes requested):

Rozzii:
/remove-lifecycle stale
There is no activity on this ticket I will move it to frozen.
/lifecycle frozen

PR #366, opened 2023-11-17
Update the maintainers list

kashifest:
as the title says

PR #367, opened 2023-11-17
fix and simplify all-owners script

tuminoid:
Noticed with @kashifest in #366 that the script isn't working right. Fix and simplify the script. 

1. remove emeritus_approvers from the list, they're not active maintainers
2. simplify the primary branch detection
3. add new repo utility-images
4. remove lower casing the github names, some have legit caps there like "Rozzii", added explicit sorting option to keep the sorting the same in Linux and Mac

### DIFF
--- a/maintainers/all-owners.sh
+++ b/maintainers/all-owners.sh
@@ -5,6 +5,7 @@
 
 set -eu
 
+# repos and which branch is the default
 declare -a REPOS=(
     baremetal-operator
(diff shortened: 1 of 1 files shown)


kashifest (commented):
/approve

kashifest:
cc @Rozzii

Rozzii (approved):
/approve

PR #378, opened 2024-01-18
switch markdownlint container to markdownlint-cli2

tuminoid:
Switch markdownlint container to markdownline-cli2. This CLI version supports the enable/disable rules in markdown files and allows us to ignore issues locally, not just globally.

- project-infra PR https://github.com/metal3-io/project-infra/pull/619 needs to merge for markdownlint here to pass


tuminoid:
/hold
Needs https://github.com/metal3-io/project-infra/pull/619 to merge first.

.markdownlint-cli2.yaml
```diff
@@ -0,0 +1,10 @@
+# Reference: https://github.com/DavidAnson/markdownlint-cli2#markdownlint-cli2yaml
+
+config:
+  ul-indent:
+    # Kramdown wanted us to have 3 earlier, tho this CLI recommends 2 or 4
```
lentzi90, line 5: Just noting here that I think 2 is the most common and would thus be least surprising. As we currently use 3 I am still fine continuing with that at least through the transition. After that we could maybe think about switching to 2, this is what I would prefer.

docs/user-guide/src/developer_environment/tryit.md
```diff
@@ -39,7 +38,7 @@ For execution with VMs
 - Setup passwordless sudo access
 
 ```console
-  $ sudo visudo
+  sudo visudo
```
lentzi90, line 41: This is not correct IMO. If the code block is `console` then there should be a prompt. If it is `bash` or `shell` then no prompt.

lentzi90 (commented):
Great job!

docs/user-guide/src/developer_environment/tryit.md
tuminoid: The linter rule is that you can only show prompt if you show output too.
Arguably the original `console` is just wrong, it should be `shell`.

.markdownlint-cli2.yaml
tuminoid: Yes I agree. We have 95% of the lines as 3 now (counting all repos) and probably 80% even in this repo, since it was that way with the old linter. 

We can definitely do 2 later on, to get rid of this last custom thing we have. After that, we basically have no global rules or config, so we're doing the defaults only, which is optimal.

docs/user-guide/src/developer_environment/tryit.md
lentzi90: Makes sense! Can you change it to shell?

tuminoid: Done

lentzi90 (commented):
/approve

tuminoid:
Bumped to 0.9.0 release of the image from 2 days ago.

smoshiur1237 (commented):
/lgtm

Rozzii (approved):
/approve

tuminoid:
/retest
Test new linters.

tuminoid:
/unhold
Good to go.

PR #381, opened 2024-02-15
✨ Added placeholders that resolve to the latest version tag of a release

peppi-lotta:
Added placeholders that will be resolved in the latest version tag of the given project. Version numbers are resolved after the page is rendered with JavaScript. 

Known issue: User will see placeholders if JavaScript is disabled in browser.

### DIFF
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -68,4 +68,12 @@ Before submitting document change, you can run the same mdbook binary to preview
     ```
 
 You should have the user-guide available now at `localhost:3000`.
-Also, the serve command watches the `src` directory for changes and rebuilds the user-guide for every change.
+Also, the serve command watches the `src` directory for changes and rebuilds theuser-guide for every change.
+
+## Dynamic Version Numbers
+
+Some version numbers are automatically generated. This is done with JavaScript
+after the page is rendered.
+{releaselink:repo:<https://api.github.com/repos/metal3-io/cluster-api-provider-metal3>:stable:any}
+Stable variable can be: any, stable or prerelease. If not set a stable release
+tag will be fetched.
--- a/docs/user-guide/release_version.js
+++ b/docs/user-guide/release_version.js
@@ -1,0 +1,61 @@
+async function getMatch(match) {
+    // Check if the match is already in the session storage
+    let version = sessionStorage.getItem(`${match[0]}`);
+    let version_tag;
+
+    if (version) {
+        // If it is, parse it and use it
+        version_tag = JSON.parse(version);
+    } else {
+        const repo = match[1];
+        const stable = match[3];
+
+        version_tag = await getReleaseLink(repo, stable);
+        sessionStorage.setItem(match[0], JSON.stringify(version_tag));
+        console.log("Fetching version tag from GitHub");
+    }
+
+    return version_tag;
+}
+
+// Function to fetch the latest release tag name from a GitHub repository
+async function getReleaseLink(repo, stable="stable") {
+    const response = await fetch(`${repo}/releases`);
+    let releases = await response.json();
+    
+    switch (stable) {
+        case "any":
+            break;
+        case "stable":
+            releases = releases.filter(release => !release.prerelease);
+            break;
+        case "prerelease":
+            releases = releases.filter(release => release.prerelease);
+            break;
+    }
+
+    return releases[0].tag_name;
+    
+}
+
+async function replacePlaceholders() {
+    // Regular expression to match the placeholders
+    // and parse the api url to get the tag name
+    const regex = /\{releaselink:repo:(.*?)(:stable:(.*?))?\}/g;
+    html = document.body.innerHTML;
+    const matches = [...html.matchAll(regex)];
+
+
+    // Loop through the matches and replace the placeholders
+    // with the tag name in the HTML
+    for (const match of matches) {
+        const tagName = await getMatch(match);
+        if (tagName === undefined) {
+            continue;
+        }
+        html = html.replace(match[0], tagName);
+    }
+
+    document.body.innerHTML = html;
+}
+replacePlaceholders();
--- a/docs/user-guide/src/capm3/installation_guide.md
+++ b/docs/user-guide/src/capm3/installation_guide.md
@@ -16,16 +16,16 @@ install them yourself.
 1. Install Cluster API core compoenents i.e., core, bootstrap and control-plane providers. This will also install cert-manager, if it is not already installed.
 
     ```bash
-     clusterctl init --core cluster-api:v1.1.4 --bootstrap kubeadm:v1.1.4 \
-     --control-plane kubeadm:v1.1.4 -v5
+     clusterctl init --core cluster-api:{releaselink:repo:https://api.github.com/repos/kubernetes-sigs/cluster-api} --bootstrap kubeadm:{releaselink:repo:https://api.github.com/repos/kubernetes/kubernetes} \
+     --control-plane kubeadm:{releaselink:repo:https://api.github.com/repos/kubernetes/kubernetes} -v5
     ```
 
 ## With clusterctl
 
-This method is recommended. You can specify the CAPM3 version you want to install by appending a version tag, e.g. `:v1.1.2`. If the version is not specified, the latest version available will be installed.
+This method is recommended. You can specify the CAPM3 version you want to install by appending a version tag, e.g. `{releaselink:repo:https://api.github.com/repos/metal3-io/cluster-api-provider-metal3}`. If the version is not specified, the latest version available will be installed.
 
 ```bash
-clusterctl init --infrastructure metal3:v1.1.2
+clusterctl init --infrastructure metal3:{releaselink:repo:https://api.github.com/repos/metal3-io/cluster-api-provider-metal3}
 ```
 
 ## With kustomize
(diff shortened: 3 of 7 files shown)


docs/user-guide/README.md
```diff
@@ -69,3 +69,7 @@ Before submitting document change, you can run the same mdbook binary to preview
 
 You should have the user-guide available now at `localhost:3000`.
 Also, the serve command watches the `src` directory for changes and rebuilds the user-guide for every change.
+
+## Dynamic Version Numbers
```
tuminoid, line 73: Empty line after headers.

docs/user-guide/src/capm3/installation_guide.md
```diff
@@ -16,16 +16,16 @@ install them yourself.
...
     ```bash
-     clusterctl init --core cluster-api:v1.1.4 --bootstrap kubeadm:v1.1.4 \
-     --control-plane kubeadm:v1.1.4 -v5
+     clusterctl init --core cluster-api:{{releaselink:repo:https://api.github.com/repos/kubernetes-sigs/cluster-api/releases/latest}} --bootstrap kubeadm:{{releaselink:repo:https://api.github.com/repos/kubernetes/kubernetes/releases/latest}} \
+     --control-plane kubeadm:{{releaselink:repo:https://api.github.com/repos/kubernetes/kubernetes/releases/latest}} -v5
     ```
 
 ## With clusterctl
 
 This method is recommended. You can specify the CAPM3 version you want to install by appending a version tag, e.g. `:v1.1.2`. If the version is not specified, the latest version available will be installed.
```
tuminoid, line 25:
```suggestion
This method is recommended. You can specify the CAPM3 version you want to install by appending a version tag, e.g. `:v1.6.0`. If the version is not specified, the latest version available will be installed.
```
While at it, maybe update the hardcoded example as well, or replace this too.

docs/user-guide/src/introduction.md
```diff
@@ -60,3 +60,10 @@ Guide](https://metal3.io/try-it.html)). Nor is it a reference (for that, see
 the [API Reference
 Documentation](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/api.md),
 and of course, the code itself.)
+
+NOTE: In this book some version numbers are dynamically generated. This is done with JavaScript 
+after rendering the page. If You are seing placeholder text like this 
+'{{releaselink:repo:api-url-ro-get-version}}', 
+this is propably due to your browser having JavaScript turned off. This can be fixed by turning JavaScript
```
tuminoid, line 67: Please limit line length to 80, so it doesn't have to be fixed later.

tuminoid (commented):
Some nits.

In general, good improvement, thank you!

mquhuy:
The implementation looks good, thanks @peppi-lotta. 
I have a question with this design choice though: If I understand correctly, the versions will have to be resolved with github api every time someone reads the book, i.e. not just at build phase, right? It means if someone refreshes the browser multiple times in a row, they might go over the API limit.
IMO it would be much better to do it just once at render phase, and let the final html have the versions already rendered. I saw that mdbook has some [Custom backend syntax](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#custom-backend-commands), have you checked if we could use that instead?

docs/user-guide/src/capm3/installation_guide.md
```diff
@@ -16,16 +16,16 @@ install them yourself.
 1. Install Cluster API core compoenents i.e., core, bootstrap and control-plane providers. This will also install cert-manager, if it is not already installed.
 
     ```bash
-     clusterctl init --core cluster-api:v1.1.4 --bootstrap kubeadm:v1.1.4 \
-     --control-plane kubeadm:v1.1.4 -v5
+     clusterctl init --core cluster-api:{{releaselink:repo:https://api.github.com/repos/kubernetes-sigs/cluster-api/releases/latest}} --bootstrap kubeadm:{{releaselink:repo:https://api.github.com/repos/kubernetes/kubernetes/releases/latest}} \
```
adilGhaffarDev, line 19:
```suggestion
     clusterctl init --core cluster-api:{releaselink:repo:https://api.github.com/repos/kubernetes-sigs/cluster-api/releases/latest} --bootstrap kubeadm:{releaselink:repo:https://api.github.com/repos/kubernetes/kubernetes/releases/latest} \
```
Can we use single braces?

docs/user-guide/release_version.js
```diff
@@ -0,0 +1,28 @@
+// Function to fetch the latest release tag name from a GitHub repository
+async function getReleaseLink(repo) {
+    const response = await fetch(repo);
+    const data = await response.json();
+    return data.tag_name;
```
adilGhaffarDev, line 5: This will grab rc and beta releases too. We should get stable releases only.

peppi-lotta:
> The implementation looks good, thanks @peppi-lotta. I have a question with this design choice though: If I understand correctly, the versions will have to be resolved with github api every time someone reads the book, i.e. not just at build phase, right? It means if someone refreshes the browser multiple times in a row, they might go over the API limit. IMO it would be much better to do it just once at render phase, and let the final html have the versions already rendered. I saw that mdbook has some [Custom backend syntax](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#custom-backend-commands), have you checked if we could use that instead?

To my understanding renderers are used when the book is build. The book is build and then deployed to Netlify. This means that after every release someone would have to rebuild and deploy the book because once in Netlify the content is already static. This is why I used the placeholders to work around the static content to fetched and updated the placehoplders when the client makes the request to see the book. 

The comment about the api rate limits is very valid. I'm thinking I could cache the version values as session variables or something so that an api requests would be made for each unique url and if the same url is in another placeholder then the cached value would be used instead of fetching.

mquhuy:
> > The implementation looks good, thanks @peppi-lotta. I have a question with this design choice though: If I understand correctly, the versions will have to be resolved with github api every time someone reads the book, i.e. not just at build phase, right? It means if someone refreshes the browser multiple times in a row, they might go over the API limit. IMO it would be much better to do it just once at render phase, and let the final html have the versions already rendered. I saw that mdbook has some [Custom backend syntax](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#custom-backend-commands), have you checked if we could use that instead?
> 
> To my understanding renderers are used when the book is build. The book is build and then deployed to Netlify. This means that after every release someone would have to rebuild and deploy the book because once in Netlify the content is already static. This is why I used the placeholders to work around the static content to fetched and updated the placehoplders when the client makes the request to see the book.
> 
> The comment about the api rate limits is very valid. I'm thinking I could cache the version values as session variables or something so that an api requests would be made for each unique url and if the same url is in another placeholder then the cached value would be used instead of fetching.

Thanks for the explanation. I like the idea of having placeholders, but I think we shouldn't need to update it all the time. 
I don't like the idea of using DOM manipulation to render this, also because it goes against how mdbook is intended to be used. Mdbook is supposed to be fast and lightweight, due to the static html nature. By doing this (and adding client-side JS in general), we add several ms to almost every load.

Caching is a way to solve the API limit, but I'm still not sure if it can be done on the client side. Anw my JS knowledge is limited so please feel free to educate me with how you intend to implement it.

IMO it's sufficient that we re-render the Netlify app once a week or so to update the versions. This could be very well-automated, for e.g. with [Netlify deploy API](https://docs.netlify.com/api/get-started/#deploy-with-the-api).

docs/user-guide/release_version.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    return releases[0].tag_name;
+    
+}
+
+async function replacePlaceholders() {
+    // Regular expression to match the placeholders
+    // and parse the api url to get the tag name
+    const regex = /\{releaselink:repo:(.*?)(:stable:(.*?))?\}/g;
```
adilGhaffarDev, line 44: Are we hardcoding :stable: here? If yes, can we get "latest" and "stable" from releaselink marker instead of hardcoding here? for example:
`{releaselink:repo:https://api.github.com/repos/kubernetes-sigs/cluster-api@latest}`
`{releaselink:repo:https://api.github.com/repos/kubernetes-sigs/cluster-api@stable}`
This will make this script more usable in the future.

docs/user-guide/src/capm3/installation_guide.md
```diff
@@ -16,16 +16,16 @@ install them yourself.
...
     ```
 
 ## With clusterctl
 
-This method is recommended. You can specify the CAPM3 version you want to install by appending a version tag, e.g. `:v1.1.2`. If the version is not specified, the latest version available will be installed.
+This method is recommended. You can specify the CAPM3 version you want to install by appending a version tag, e.g. `{releaselink:repo:https://api.github.com/repos/metal3-io/cluster-api-provider-metal3}`. If the version is not specified, the latest version available will be installed.
 
 ```bash
-clusterctl init --infrastructure metal3:v1.1.2
+clusterctl init --infrastructure metal3:{releaselink:repo:https://api.github.com/repos/metal3-io/cluster-api-provider-metal3}
```
mboukhalfa, line 28: Do you know if md-book provide a way to default this placeholder ? 
If not can this be an interesting a feature? if so we can open an issue here like https://github.com/rust-lang/mdBook/issues/144

I think this guides are meant also to work offline then our js will not be able to request github api and we should default these somehow not a big deal but we can create a feature request if you think it is feasible

peppi-lotta:
Thank you for comments en reviews! I will close this PR and make another for a more serves side approach.

PR #384, opened 2024-02-23
Quick start for Metal3

lentzi90:
This adds a quick start guide for how to get started with Metal3.

### DIFF
--- a/docs/user-guide/src/quick-start.md
+++ b/docs/user-guide/src/quick-start.md
@@ -1,0 +1,915 @@
+# Quick-start for Metal3
+
+This guide has been tested on Ubuntu server 22.04. It should be seen as an
+example rather than the absolute truth about how to deploy and use Metal3. We
+will cover two environments and two scenarios. The environments are
+
+1. a baremetal lab with actual physical servers and baseboard management
+   controllers (BMCs), and
+1. a virtualized baremetal lab with virtual machines and sushy-tools acting as
+   BMC.
+
+In both of these, we will show how to use Bare Metal Operator and Ironic to
+manage the servers through a Kubernetes API, as well as how to turn the servers
+into Kubernetes clusters managed through Cluster API. These are the two
+scenarios.
+
+In a nut-shell, this is what we will do:
+
+1. [Setup a management cluster](#management-cluster)
+1. [Setup a DHCP server](#dhcp-server)
+1. [Setup a disk image server](#image-server)
+1. [Deploy Ironic](#deploy-ironic)
+1. [Deploy Bare Metal Operator](#deploy-bare-metal-operator)
+1. [Create BareMetalHosts to represent the servers](#create-baremetalhosts)
+1. [(Scenario 1) Provision the BareMetalHosts](#scenario-1-provision-baremetalhosts)
+1. [(Scenario 2) Deploy Cluster API and turn the BareMetalHosts into a Kubernetes cluster](#scenario-2-metal3-and-cluster-api)
+
+## Prerequisites
+
+You will need the following tools installed.
+
+- docker (or podman)
+- kind or minikube (management cluster, not needed if you already have a "real"
+  cluster that you want to use)
+- clusterctl
+- kubectl
+- htpasswd
+- virsh and virt-install for the virtualized setup
+
+## Baremetal lab configuration
+
+The baremetal lab has two servers that we will call bml-01 and bml-02, as well
+as a management computer where we will set up Metal3. The servers are equipped
+with iLO 4 BMCs. These BMCs are connected to an "out of band" network
+(`192.168.1.0/24`) and they have the following IP addresses.
+
+- bml-01: 192.168.1.13
+- bml-02: 192.168.1.14
+
+There is a separate network for the servers (`192.168.0.0/24`). The management
+computer is connected to both of these networks with IP addresses `192.168.1.7`
+and `192.168.0.150` respectively.
+
+Finally, we will need the MAC addresses of the servers to keep track of which is
+which.
+
+- bml-01: 80:c1:6e:7a:e8:10
+- bml-02: 80:c1:6e:7a:5a:a8
+
+## Virtualized configuration
+
+If you do not have the hardware or perhaps just want to test things out without
+committing to a full baremetal lab, you may simulate it with virtual machines.
+In this section we will show how to create a virtual machine and use sushy-tools
+as a baseboard management controller for it.
+
+The configuration is a bit simpler than in the baremetal lab because we don't
+have a separate out of band network here. In the end we will have the BMC
+available as
+
+- bml-vm-01: 192.168.222.1:8000/redfish/v1/Systems/bmh-vm-01
+
+and the MAC address:
+
+- bml-vm-01: 00:60:2f:31:81:01
+
+Start by defining a libvirt network:
+
+```xml
+<network>
+  <name>baremetal</name>
+  <forward mode='nat'>
+    <nat>
+      <port start='1024' end='65535'/>
+    </nat>
+  </forward>
+  <bridge name='metal3'/>
+  <ip address='192.168.222.1' netmask='255.255.255.0'>
+  </ip>
+</network>
+```
+
+Save this as `net.xml`, define it and start it.
+
+```bash
+virsh -c qemu:///system net-define net.xml
+virsh -c qemu:///system net-start baremetal
+```
+
+Next, we will create a virtual machine. Feel free to adjust at as you see fit,
+but make sure to note the MAC address. That will be needed later. You can also
+create more than one if you like.
+
+```bash
+virt-install \
+  --connect qemu:///system \
+  --name bmh-vm-01 \
+  --description "Virtualized BareMetalHost" \
+  --osinfo=ubuntu-lts-latest \
+  --ram=4096 \
+  --vcpus=2 \
+  --disk size=25 \
+  --graphics=none \
+  --console pty \
+  --serial pty \
+  --pxe \
+  --network network=baremetal,mac="00:60:2f:31:81:01" \
+  --noautoconsole
+```
+
+### Sushy-tools - AKA the BMC
+
+Metal3 relies on baseboard management controllers to manage the baremetal
+servers, so we need something similar for our virtual machines. This comes in
+the form of [sushy-tools](https://docs.openstack.org/sushy/latest/).
+
+We need to create configuration file first:
+
+```conf
+# Listen on 192.168.222.1:8000
+SUSHY_EMULATOR_LISTEN_IP = u'192.168.222.1'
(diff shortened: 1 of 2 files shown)


docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+      - containerPort: 6180
+        hostPort: 6180
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+      # Ironic API
+      - containerPort: 6385
+        hostPort: 6385
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+      # Inspector API
```
dtantsur, line 101: Is it possible to already remove inspector from here?

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.14.3/cert-manager.yaml
+```
+
+### DHCP server
+
+The BareMetalHosts must be able to call back to Ironic when going through the
+inspection phase. This means that they must have IP addresses in a network where
+they can reach Ironic. We will set up a DHCP server for this purpose.
+
+Any DHCP server can be used for this. We will here use the ironic container
```
dtantsur, line 130: nit: let's be consistent in "Ironic" vs "ironic" (I prefer the former)

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+In order to do anything useful, we will need a server for hosting disk images
+that can be used to provision the servers.
+
+Create a folder to hold the disk images:
+
+```bash
+mkdir disk-images
+```
+
+Download an image to use for testing:
```
dtantsur, line 165: This is supposed to be a real baremetal user guide, right? Let's avoid cirros then. I'm impressed you actually found some hardware where it works, but on most it probably won't (missing drivers for NICs, RAID controllers, etc).

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+# Listen on this IP (management computer)
+PROVISIONING_IP=192.168.0.150
+# Give out IP addresses in this range
+DHCP_RANGE=192.168.0.100,192.168.0.149
+```
+
+You can now run the DHCP server like this:
+
+```bash
+docker run --name dnsmasq --rm -d --net=host --privileged --user 997:994 \
```
dtantsur, line 149: Bonus for using capabilities instead of --privileged

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+```
+
+#### Authentication configuration
+
+Create authentication configuration for Ironic and Inspector. You will need to
+generate a username and password for each. We will here refer to them as
+`IRONIC_USERNAME`, `IRONIC_PASSWORD`, `INSPECTOR_USERNAME` and
+`INSPECTOR_PASSWORD`.
+
+Create a file `ironic-auth-config` with configuration for how to access Ironic.
```
dtantsur, line 197: This all will be redundant without inspector. Since this guide is still a draft, maybe avoid inspector already? You can set USE_IRONIC_INSPECTOR=false for now.

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+It looks like this for the baremetal lab:
+
+```bash
+# Same port as exposed in kind.yaml
+HTTP_PORT=6180
+# This is the interface inside the container
+PROVISIONING_INTERFACE=eth0
+# URL where the http server is exposed (IP of management computer)
+CACHEURL=http://192.168.0.150
+IRONIC_FAST_TRACK=true
```
dtantsur, line 248: Not needed, please remove (I want to deprecate the explicit variable).

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,642 @@
...
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: bml-01
+spec:
+  online: true
+  bootMACAddress: 80:c1:6e:7a:e8:10
+  bootMode: legacy
```
dtantsur, line 562: Is it really required in your case? We no longer recommend using legacy boot by default. At the very least, please explain this choice so that the users don't blindly copy it on newer hardware.

dtantsur:
I think it's a much better quick start than pointing people to metal3-dev-env, great job! Some comments inline.

docs/user-guide/src/quick-start.md
lentzi90: Thanks for the heads up! I was not aware of this

docs/user-guide/src/quick-start.md
lentzi90: If you can tell me exactly which to use I'll happily change it :sweat_smile:

docs/user-guide/src/quick-start.md
lentzi90: Sadly it is :slightly_frowning_face: 
I actually tried first with UEFI but it was not supported on the hardware. I will add a comment about it though

docs/user-guide/src/quick-start.md
lentzi90: Unfortunately it turned out to be more work removing inspector everywhere then just using it as is. All the manifests still include and expect Inspector to be there in v0.5.0. I will happily update this once we have a new release that doesn't use Inspector though.
(I hope we can get this guide merged in time for KubeCon :crossed_fingers: )

lentzi90:
Alright I think this is ready for review. The virtualized setup is missing but we can add that separately also. Perhaps easier to review it that way also

lentzi90:
I initially wanted to do the same kind of "tabs" that [CAPI has](https://cluster-api.sigs.k8s.io/user/quick-start) for baremetal vs virtualized, but it turned out to be a bit more involved than I expected (including a makefile and a go tool). We can still do it but I suggest we merge this first as is. Then we can figure out the tabs and adapt the guide.

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
+# Quick-start for Metal3
+
+This guide has been tested on Ubuntu server 22.04. It should be seen as an
+example rather than the absolute truth about how to deploy and use Metal3. We
+will cover two environments and two scenarios. The environments are
+
+1. a baremetal lab with actual physical servers and baseboard management
+   controllers (BMCs), and
+2. a virtualized baremetal lab with virtual machines and sushy-tools acting as
```
tuminoid, line 9:
```suggestion
1. a virtualized baremetal lab with virtual machines and sushy-tools acting as
```
Our generator supports auto numbered lists, right?

If so, I'd like to see it utilized for all numbered lists here.

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
...
+6. Create BareMetalHosts to represent the servers
+7. (Scenario 1) Provision the BareMetalHosts
+8. (Scenario 2) Deploy Cluster API and turn the BareMetalHosts into a Kubernetes
+   cluster
+
+## Prerequisites
+
+You will need the following software installed.
+
+- docker (or podman)
```
tuminoid, line 33: Docker as a softwarere is a name, hence it would need to be capitalized. docker as a CLI command would not. So either fix the capitalization of all the tools, or alternatively L31 could read tools instead of software?

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
...
+  --env-file dnsmasq.env --entrypoint /bin/rundnsmasq \
+  quay.io/metal3-io/ironic
+```
+
+### Image server
+
+In order to do anything useful, we will need a server for hosting disk images
+that can be used to provision the servers.
+
+Create a folder to hold the disk images:
```
tuminoid, line 253:
```suggestion
Create a directory to hold the disk images:
```
Folder is GUI terminology.

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
...
+
+To enable basic auth, we need to create secrets containing the keys
+`IRONIC_HTPASSWD` and `INSPECTOR_HTPASSWD` with values generated from the
+credentials using htpasswd. We will do this by creating two files
+`ironic-htpasswd` and `ironic-inspector-htpasswd` with the following content.
+
+`ironic-htpasswd`:
+
+```bash
+IRONIC_HTPASSWD="output of `htpasswd -n -b -B IRONIC_USERNAME IRONIC_PASSWORD`"
```
tuminoid, line 329:
```suggestion
IRONIC_HTPASSWD="<output of `htpasswd -n -b -B IRONIC_USERNAME IRONIC_PASSWORD`>"
```
`<>` commonly used for placeholders/options.

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
...
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: baremetal-operator-system
+# These are the kustomizations we build on. You can download them and change the URLs to relative
+# paths if you do not want to access them over the network.
+# Note that the ref=v0.5.0 specifies the version to use.
+resources:
+  - https://github.com/metal3-io/baremetal-operator/config/namespace?ref=v0.5.0
```
tuminoid, line 424: Have the YAML in k8s formatting, ie. without leading spaces on the lists.

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
...
+
+6. Deploy Baremetal Operator.
+
+   ```bash
+   kubectl apply -k bmo
+   ```
+
+## Create BareMetalHosts
+
+Now that we have Baremetal operator deployed, let's put it to use by creating
```
tuminoid, line 653:
```suggestion
Now that we have Bare Metal Operator deployed, let's put it to use by creating
```

docs/user-guide/src/quick-start.md
```diff
@@ -0,0 +1,916 @@
...
+   BMC.
+
+In both of these, we will show how to use Baremetal operator and Ironic to
+manage the servers through a Kubernetes API, as well as how to turn the servers
+into Kubernetes clusters managed through Cluster API. These are the two
+scenarios.
+
+In a nut-shell, this is what we will do:
+
+1. Setup a management cluster
```
tuminoid, line 19: Add anchor links to these as well.

tuminoid (commented):
Good work, very much needed improvement!

Some nits and suggestions. Most of them repeat, so I only mentioned each one.

lentzi90:
Thanks @tuminoid I think I have addressed all comments now :slightly_smiling_face:

PR #386, opened 2024-03-01
:sparkles:  Go preprocessor to find and replace releasetags

peppi-lotta:
This is a mdbook preprocessor writen in go. The go method uses a [plugin](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/utils/plugin) to find tags and replace them.  [Cluster-api documentation has implemented](https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/tools/mdbook/releaselink/releaselink.go#L49) this plugin and I'm trying to use the plugin in the same way except for minor changes to fit the metal3 use case better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,20 @@ HOST_PORT ?= 3000
 BIN_DIR := hack
 MDBOOK_BIN := $(BIN_DIR)/mdbook
 MDBOOK_RELEASE_URL := https://github.com/rust-lang/mdBook/releases/download/$(MDBOOK_BIN_VERSION)/mdbook-$(MDBOOK_BIN_VERSION)-x86_64-unknown-linux-gnu.tar.gz
+TOOLS_DIR := hack/tools
+TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/bin)
+
+export PATH := $(PATH):$(TOOLS_BIN_DIR)
+
+## ------------------------------------
+## Resolve placeholders as tags
+## ------------------------------------
+RELEASETAGS := $(TOOLS_BIN_DIR)/mdbook-releasetags
+$(RELEASETAGS): $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR); go build -tags=tools -o $(TOOLS_BIN_DIR)/mdbook-releasetags ./releasetags
+
+.PHONY: releasetags
+releasetags: $(RELEASETAGS)
 
 ## ------------------------------------
 ## Documentation tooling for Netlify
@@ -19,7 +33,7 @@ $(MDBOOK_BIN): # Download the binary
 	curl -L $(MDBOOK_RELEASE_URL) | tar xvz -C $(BIN_DIR)
 
 .PHONY: netlify-build
-netlify-build: $(MDBOOK_BIN) # Build the user guide
+netlify-build: $(RELEASETAGS) $(MDBOOK_BIN) # Build the user guide
 	$(MDBOOK_BIN) build $(SOURCE_PATH)
 
 
@@ -28,7 +42,8 @@ netlify-build: $(MDBOOK_BIN) # Build the user guide
 ## ------------------------------------
 
 .PHONY: build
-build: # Build the user guide
+build: $(RELEASETAGS)
+# Build the user guide
 	$(CONTAINER_RUNTIME) run \
 	--rm -it --name metal3 \
 	-v "$$(pwd):/workdir" \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -28,21 +41,21 @@ netlify-build: $(MDBOOK_BIN) # Build the user guide
...
 .PHONY: serve
-serve: # Serve the user-guide on localhost:3000 (by default)
+serve:# Serve the user-guide on localhost:3000 (by default)
 	$(CONTAINER_RUNTIME) run \
 	--rm -it --init --name metal3 \
 	-v "$$(pwd):/workdir" \
 	-p $(HOST_PORT):3000 \
 	$(IMAGE_NAME):$(IMAGE_TAG) \
-	mdbook serve --open $(SOURCE_PATH) -p 3000 -n 0.0.0.0
+	mdbook serve --open $(SOURCE_PATH) -p 3000 -n 0.0.0.0
```
mquhuy, line 58: Why is the trailing space needed?

docs/user-guide/src/SUMMARY.md
```diff
@@ -19,20 +18,18 @@
       - [Inspect annotation](bmo/inspect_annotation.md)
       - [Reboot annotation](bmo/reboot_annotation.md)
 - [Ironic in Metal3](ironic/introduction.md)
-   - [Install Ironic](ironic/ironic_installation.md)
-   - [Ironic python agent](ironic/ironic-python-agent.md)
-   - [Ironic container images](ironic/ironic-container-images.md)
```
mquhuy, line 24: Shouldn't these lines be indented?

.gitignore
```diff
@@ -3,7 +3,7 @@ design/*.toc.*
 
 # Ignore mdbook binary
 # from being committed.
-hack/mdbook
+hack/tools/bin
```
kashifest, line 6: I think it should be `hack/tools/` ?

docs/user-guide/src/SUMMARY.md
```diff
@@ -19,20 +18,18 @@
...
+   - [Features](capm3/features.md)
       - [Remediation](capm3/remediaton.md)
       - [Node Reuse](capm3/node_reuse.md)
       - [Pivoting](capm3/pivoting.md)
       - [Automated cleaning](capm3/automated_cleaning.md)
 - [Ip-address-manager](ipam/introduction.md)
-   - [Install Ip-address-manager](ipam/ipam_installation.md)
-- [Developer environment]()
-   - [Try it](developer_environment/tryit.md)
+- [Install Ip-address-manager](ipam/ipam_installation.md)
```
kashifest, line 32: indentation changes are wrong here

hack/tools/releasetags/releasetags.go
```diff
@@ -0,0 +1,76 @@
...
+	"sigs.k8s.io/kubebuilder/docs/book/utils/plugin"
+)
+
+type ReleaseTag struct{}
+
+// SupportsOutput checks if the given plugin supports the given output format.
+func (ReleaseTag) SupportsOutput(_ string) bool { return true }
+
+// Process modifies the book in the input, which gets returned as the result of the plugin.
+func (l ReleaseTag) Process(input *plugin.Input) error {
```
mquhuy, line 22: I'm wondering if we could get some caching logic here: instead of querying a releasetag, check if it was queried before, if yes, use the cached value, if no, query it, and store the result in the cache.

I haven't checked the doc, so not sure if it's worth the effort. I'd be ok if you prefer to not do it now.

.gitignore
```diff
@@ -3,7 +3,7 @@ design/*.toc.*
 
 # Ignore mdbook binary
```
mquhuy, line 4:
```suggestion
# Ignore all binaries
```

mquhuy (commented):
Thank you Peppi. I haven't run it, but the code looks legit. I just have some small nits in the comments.

adilGhaffarDev (approved):
nice work! LGTM from my side once the nits are fixed.

docs/README.md
```diff
@@ -72,3 +72,12 @@ All the commands below are executed within mdbook container.
     ```bash
     make clean
     ```
+
+## Markers
+
+There is a preprocessor written in Golang that resolves
+```{{#releasetag owner:"metal3-io" repo:"cluster-api-provider-metal3"}}``` type
+markers as the latest version eg. ```:v1.6.1```.  It takes repository name and owner
```
adilGhaffarDev, line 80:
```suggestion
markers as the latest version e.g. ```:v1.6.1```.  It takes repository name and owner
```
I think spellcheck is failing here.

hack/tools/releasetags/releasetags.go
peppi-lotta: I was thinking of a cache system too. But currently there only a few markers in the book. This is a good addition if more markers are added. So I think I won't implement it for now.

adilGhaffarDev (commented):
these changes should fix the spellcheck.

kashifest (commented):
/approve

PR #393, opened 2024-03-11
Fix the Getting Started Guide link in introduction to try it section

NymanRobin:
Currently the Getting Started Guide link in introduction section gives page not found error.

I suspect this is supposed to take the user to the tryit section, so here is the link fix also changes from relaying on url to file link.
Maybe the wording should be changed as this tutorial is in fact part of the user guide? :thinking: 

However if this understanding is wrong please correct me :D

### DIFF
--- a/docs/user-guide/src/introduction.md
+++ b/docs/user-guide/src/introduction.md
@@ -56,7 +56,7 @@ contributions. The community meets at the following venues:
 
 This user guide aims to explain the Metal³ feature set, and provide how-tos for
 using Metal³. It's not a tutorial (for that, see the [Getting Started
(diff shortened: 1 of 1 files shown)


NymanRobin:
Hey!
Sure, I can tackle those as well. I will change the scope of this PR to fix the try-it links

tuminoid:
> Hey! Sure, I can tackle those as well. ~I will change the scope of this PR to fix the try-it links.~
> 
> EDIT: Seems the blog posts were in another repo will open a PR there as well

You're right, they weren't in this one. Thanks!

/lgtm

PR #404, opened 2024-03-22
Typo in authentication config section

jschoone:


tuminoid:
/ok-to-test

We have just removed inspector from the main, so the whole section needs update soon to get rid of those references. But for the time being, it should get fixed.

tuminoid (commented):
/lgtm

/assign @kashifest

PR #410, opened 2024-04-05
Remove extra OWNERS files and propose more reviewers in root OWNERS

kashifest:
Theres no need to have these extra OWNERS files as the `design` directory has its own and it does not inherit parent OWNER file. Removing these reduces confusion. 

Also proposing more reviewers in the root OWNERS and this list is inherited mainly from the deleted OWNERS file.


smoshiur1237 (commented):
LGTM

lentzi90 (approved):
/approve

lentzi90:
I think we will need approval from someone in each of the removed files

kashifest:
> I think we will need approval from someone in each of the removed files

A couple of them have single entry who have left the project a while ago

kashifest:
@digambar15 @ogelbukh if you are still around Metal3 please review this patch and leave an approve label if no objection with this.

Rozzii (approved):
/lgtm

PR #411, opened 2024-04-08
Live updates (servicing) for hosts

dtantsur:
This design proposal introduces the Ironic servicing feature into BMO.
Servicing is a process of modifying already provisioned nodes, e.g.
changing BIOS settings or updating firmware (the 2 features I want to
concentrate on right now).

### DIFF
--- a/design/baremetal-operator/host-live-updates.md
+++ b/design/baremetal-operator/host-live-updates.md
@@ -1,0 +1,194 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Live updates of hosts
+
+## Status
+
+One of: implementable
+
+## Summary
+
+This proposal enables baremetal-operator to conduct certain actions on already
+provisioned BareMetalHosts.
+
+## Motivation
+
+Currently, all preparation actions, such as BIOS settings, hardware RAID
+configuration and newly firmware updates, happen in the `preparing` state
+before a host is provisioned. While it makes perfect sense for RAID, firmware
+updates can normally happen for already provisioned hosts.
+
+We already have a precedent of modifying running hosts: a `DataImage` can
+be attached on reboot. This proposal extends the same to other actions,
+including potential future ones. This new feature will be called *live updates*
+throughout this document.
+
+### Goals
+
+- Firmware updates and BIOS settings changes are possible for already
+  provisioned hosts.
+- Operators can opt out of live updates using RBAC.
+
+### Non-Goals
+
+- Changing hardware RAID configuration.
+- Covering data images in the same proposal (we may add them later).
+- Supporting actions from inside the running operating system.
+- Live updates without rebooting.
+
+## Proposal
+
+A new auxiliary CRD `HostUpdatePolicy` will define policies for different
+actions (initially, firmware updates and firmware settings).
+The defaults will match the current ones. A policy object will be matched to
+a BMH by its name (similarly to other resources). RBAC can be used to prevent
+a certain user from creating/changing policies.
+
+### User Stories
+
+#### Story 1
+
+As a BareMetalHost user, I want to be able to update system firmware or change
+firmware settings without de-provisioning my hosts.
+
+#### Story 2
+
+As a cluster operator, I want to control whether certain users can run live
+updates on BareMetalHosts.
+
+## Design Details
+
+### Custom Resource Definitions
+
+A new auxiliary CRD `HostUpdatePolicy` will be added with the following Go
+definition (shortened):
+
+```go
+type HostUpdatePolicySpec struct {
+    FirmwareSettings HostUpdatePolicyType
+    FirmwareUpdates HostUpdatePolicyType
+}
+```
+
+`HostUpdatePolicyType` is an enumeration with the following values:
+
+- `onPreparing` (the default) - updates will only happen in the `preparing`
+  state (no live updates).
+- `onReboot` - updates will happen on the next reboot.
+
+In all cases, detached hosts will need to be attached first. Hosts not in the
+`provisioned` state will not be affected.
+
+### New Operational Status
+
+A new operational status `servicing` will be added to designate hosts that are
+undergoing a live update. Once the update finishes, the operational status will
+become `OK` or `error`.
+
+### Example workflow
+
+As an operator, I want to update firmware on my Kubernetes worker node. My
+actions:
+
+1. Mark the affected node as non-schedulable, drain the workloads from it.
+2. Find the corresponding `HostFirmwareComponents` resource, edit it to put
+   an HTTP link to the expected firmware image(s).
+3. Edit the `BareMetalHost` resource to add the reboot annotations.
+4. Wait for the host's `operationalStatus` field to first change to
+   `servicing`, then back to `OK`.
+5. Wait for the node's status to become `Ready` again.
+6. Mark the node as schedulable.
+
+### Implementation Details/Notes/Constraints
+
+- The power management code for provisioned hosts will be updated with a logic
+  to check servicing on each power on. If the policy allows servicing, and
+  there is a change in the corresponding resources (`HostFirmwareSetting` etc),
+  the [servicing](https://docs.openstack.org/ironic/latest/admin/servicing)
+  operation will be invoked.
+
+  Servicing is technically very similar to cleaning, but runs on already
+  provisioned hosts. It reboots into the IPA ramdisk (if needed) and runs
+  a sequence of requested steps. As with cleaning, BMO will be responsible for
+  creating a list of steps to execute. Unlike cleaning, the end state of the
+  host is provisioned and powered on.
+
+  In case of firmware updates, Ironic downloads the requested images into the
+  local HTTP server and provides links to them to the machine via Redfish. Then
+  the machine is rebooted.
+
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,141 @@
...
+- `immediate` - updates cause an immediate reboot and get applied.
+
+In all cases, detached hosts will need to be attached first. Hosts not in the
+`provisioned` state will not be affected.
+
+### New Operational Status
+
+A new operational status `updating` will be added to designate hosts that are
+undergoing a live update. Once the update finishes, the operational status will
+become `OK` or `error`.
```
zaneb, line 92: This makes me nervous. Maybe there should be a separate 'Update status' field with its own state machine?

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,141 @@
...
+- Covering data images in the same proposal (we may add them later).
+- Supporting actions from inside the running operating system.
+- Live updates without rebooting.
+
+## Proposal
+
+A new auxiliar CRD `HostUpdatePolicy` will define policies for different
+actions (initially, firmware updates and firmware settings).
+The defaults will match the current ones. A policy object will be optionally
+linked from a BareMetalHost. RBAC can be used to prevent a certain user from
```
zaneb, line 49: linked by name I hope

design/baremetal-operator/host-live-updates.md
dtantsur: What exactly concerns you? I'm afraid that another state machine may turn the BMO code into an even worse spaghetti than it is.

design/baremetal-operator/host-live-updates.md
dtantsur: Yeah, corev1.LocalObjectReference

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,141 @@
...
+creating/changing policies.
+
+### User Stories
+
+#### Story 1
+
+As a BareMetalHost user, I want to be able to update system firmware without
+de-provisioning my hosts.
```
rhjanders, line 57: suggestion: "update system firmware _and/or change BIOS settings_" ?

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,141 @@
...
+#### Story 2
+
+As a cluster operator, I want to control whether certain users can run live
+updates on BareMetalHosts.
+
+## Design Details
+
+### Custom Resource Definitions
+
+A new auxiliar CRD `HostUpdatePolicy` will be added with the following Go
```
rhjanders, line 68: nit: auxiliary ?

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,141 @@
...
+### Non-Goals
+
+- Changing hardware RAID configuration.
+- Covering data images in the same proposal (we may add them later).
+- Supporting actions from inside the running operating system.
+- Live updates without rebooting.
+
+## Proposal
+
+A new auxiliar CRD `HostUpdatePolicy` will define policies for different
```
rhjanders, line 46: nit: auxiliary ?

rhjanders (commented):
Thank you for the proposal Dmitry. LGTM. Very useful feature.

tuminoid (commented):
Can you add proper description to the PR?

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,155 @@
...
+undergoing a live update. Once the update finishes, the operational status will
+become `OK` or `error`.
+
+### Implementation Details/Notes/Constraints
+
+- The power management code for provisioned hosts will be updated with a logic
+  to check servicing on each power on. If the policy allows servicing, and
+  there is a change in the corresponding resources (`HostFirmwareSetting` etc),
+  the Ironic *servicing* operation will be invoked.
+- Draining hosts will not be handled by BMO. It will be entirely up to the user
```
mhrivnak, line 99: It would help to write out the complete workflow of what needs to happen to update nodes in a running cluster. I'm not completely following what the user needs to do, or even who/what is responsible for initiating the reboot.

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,155 @@
...
+A new operational status `servicing` will be added to designate hosts that are
+undergoing a live update. Once the update finishes, the operational status will
+become `OK` or `error`.
+
+### Implementation Details/Notes/Constraints
+
+- The power management code for provisioned hosts will be updated with a logic
+  to check servicing on each power on. If the policy allows servicing, and
+  there is a change in the corresponding resources (`HostFirmwareSetting` etc),
+  the Ironic *servicing* operation will be invoked.
```
mhrivnak, line 98: What does that operation do exactly? Would it stage firmware updates prior to actually turning the system on? Or something else?

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,155 @@
...
+### New Operational Status
+
+A new operational status `servicing` will be added to designate hosts that are
+undergoing a live update. Once the update finishes, the operational status will
+become `OK` or `error`.
+
+### Implementation Details/Notes/Constraints
+
+- The power management code for provisioned hosts will be updated with a logic
+  to check servicing on each power on. If the policy allows servicing, and
```
mhrivnak, line 96: Would this capability to update firmware only be available then when the BMH is used to cycle power, but not when MCO initiates a reboot?

dtantsur: In this design - yes. I'm not sure how we'll detect that reboot is happening otherwise.

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,194 @@
...
+
+- Although the BMO's reboot annotation will be used as a trigger,
+  the actual reboot will not be handled by BMO in this case. All reboots
+  (at least two in most cases) will be orchestrated by Ironic in reality.
+  BMO will simply remove the reboot annotation after triggering servicing.
+
+- Draining hosts will not be handled by BMO. It will be entirely up to the user
+  to ensure that the hosts can be updated.
+
+  In case of a Kubernetes cluster, the workloads will need to be drained and
```
mhrivnak, line 133: Machine Config Operator can cordon, drain, gracefully reboot, and then un-cordon a node. Can we just use that?

dtantsur: Possibly? That would be an openshift-specific discussion though. I also don't see a huge issue in just documenting it for the operators to do.

mhrivnak: I see. So you're going to document the process in metal3, but not automate it? That would enable people to handle the reboot as gracefully as their specific tooling and/or distribution allows.

elfosardo:
can't approve here but it looks good to me

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Design Details
+
+### Custom Resource Definitions
+
+A new auxiliary CRD `HostUpdatePolicy` will be added with the following Go
+definition (shortened):
+
+```go
+type HostUpdatePolicySpec struct {
```
Rozzii, line 72: I don't want to endulge in bikeshedding but `BMHostUpdatePolicySpec` would be a better name imo, "Host Update" might confuse users of big "management clusters" with a lot of CRs  by such a generic CR name.

design/baremetal-operator/host-live-updates.md
```diff
@@ -0,0 +1,194 @@
...
+  provisioned hosts. It reboots into the IPA ramdisk (if needed) and runs
+  a sequence of requested steps. As with cleaning, BMO will be responsible for
+  creating a list of steps to execute. Unlike cleaning, the end state of the
+  host is provisioned and powered on.
+
+  In case of firmware updates, Ironic downloads the requested images into the
+  local HTTP server and provides links to them to the machine via Redfish. Then
+  the machine is rebooted.
+
+- Although the BMO's reboot annotation will be used as a trigger,
```
Rozzii, line 125: So do I understand correctly that depending on the BMC driver and the targeted firmware the physical machines might get rebooted either 1 or 2 time but from the perspective of a user who only sees the BMH CRs it will just look like: reboot annotation->start servicing->remove the annotation -> BMH in servicing state -> servicing ends BMH goes back to provisioned state.

Rozzii (commented):
I am fine with this I have 1 nit and 1 clarifying question.

design/baremetal-operator/host-live-updates.md
dtantsur: This is a good conversation to have, but right now we already have HostFirmwareSettings and HostFirmwareComponents, this is just following the example.

design/baremetal-operator/host-live-updates.md
dtantsur: Correct (except that "servicing" is an operational status, not a state).

design/baremetal-operator/host-live-updates.md
Rozzii: Indeed, I am fine with staying consistent, but in the future IMO we should change the CR names or at least discuss this.

Rozzii (commented):
/approve

Issue #395, opened 2024-03-20
Go through the quick-start and fix/improve things you find


lentzi90:
Read through and try out our new [quick-start guide](https://book.metal3.io/quick-start).
Was something hard to understand? Did you run into any issues? Maybe you found something that could be improved?
Create a PR with a fix or an issue describing what should be changed.

lentzi90:
I think we can close this for now. We had a couple of related PRs merged

PR #414, opened 2024-04-18
Add baremetal-operator image version in kustomization.yaml example

chess-knight:
Error:
```
"msg":"if kind is a CRD, it should be installed before calling Start","kind":"HostFirmwareComponents.metal3.io","error":"no matches for kind \"HostFirmwareComponents\" in version \"metal3.io/v1alpha1\""
```
Related to #395


lentzi90 (commented):
Thanks for this!
I think it highlights an issue with the releases we do in BMO actually. We should pin the image when we do releases. Until we have that, we will need this though.
/approve

Issue #399, opened 2024-03-20
Add cover pages for feature sections


lentzi90:
The [user-guide](https://book.metal3.io/introduction) has two missing "Feature" and one "Developer environment" sections. These sections do not need a lot of information but it would be great to fill in the gaps. Write some kind of generic description for the section and what it contains. Here is an example of what it can look like: https://cluster-api.sigs.k8s.io/security/

lentzi90:
This was fixed in https://github.com/metal3-io/metal3-docs/pull/386

PR #417, opened 2024-04-28
Added a feature section to the user guide. solved issue #399

sanmarg:
Signed-off-by: Sanmarg Paranjpe<sanmarg.paranjpe@gmail.com>
 Changes to be committed:
	new file:   docs/user-guide/src/feature/feature.md


dtantsur (changes requested):
I'm sorry, but this addition does not make much sense to me. What you list are not features, they are components. We have an incomplete feature list in the book, and the list of components is implicit in https://book.metal3.io/introduction (where it actually belongs).

PR #419, opened 2024-05-03
hack/tools: reduce required Go version

dtantsur:
1.22 is very new and is not yet available in most distributions.
It seems that 1.21 works fine.

### DIFF
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module metal3-io/metal3-docs/hack/tools
 
-go 1.22
+go 1.21
 
 require (
 	github.com/blang/semver v3.5.1+incompatible


dtantsur:
See https://github.com/metal3-io/metal3-docs/pull/420

lentzi90 (commented):
/approve

kashifest:
/hold
This perhaps passed the tests since we have a GO_VERSION variable set in the netlify settings and the build used go 1.22.2 anyhow, I will remove it now. I am anyhow making this variable to be set via the config file here #421. Here is the build which passed for this test: 

Screenshot 2024-05-06 at 12 32 16

kashifest:
/hold cancel
I have removed it from the settings, lets do a repush of this branch to see if the build is still happy.

PR #421, opened 2024-05-06
Bump rust to v0.4.37

kashifest:
Rust v0.4.15 is pretty old and we had several versions out since then. Also add GO_VERSION var in the netlify build config. We set it previously in netlify build settings web ui, however setting it explicitly here is more easier and doesn't require special permission for the webui.


kashifest:
/hold
Bumping rust is not solving the build issue

Rozzii (approved):
/approve

lentzi90 (commented):
/approve

PR #425, opened 2024-05-10
Add docker-build target

lentzi90:
This adds a docker-build make target for building the container image. The Dockerfile is also expanded to build the releasetags preprocessor in a build container instead of copying from the host.

The mdbook version is taken from the Makfile as an argument. The rust version is bumped since the old version could not build the newer mdbook version. (The versions in the Dockerfile and Makefile were different.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ design/*.toc.*
 
 # Ignore all binaries
 # from being committed.
-hack/tools/
+hack/tools/bin
 # Ignore mdbook boilerplate
 # from being committed.
 docs/user-guide/book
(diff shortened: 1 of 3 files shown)


lentzi90:
It seems like we have not published a new container image for this in ~2 years. How about we just always build it locally? I don't see much need for publishing this image and if we always build locally there is one less thing to worry about.
The `build` and `serve` make targets already depend on the image so it would be trivial to make them depend on building the image also, instead of pulling it.

peppi-lotta:
Excellent change in my opinion! Works great. 

/lgtm

tuminoid:
It doesn't seem to work unless built with `make docker-build` that feeds the mdbook version. Can we have sensible default for that, so it works if you just `docker build ...` it?

lentzi90:
Sure I can add some default version!

tuminoid (commented):
/lgtm

Works like a charm, thanks!

kashifest (commented):
/approve

PR #428, opened 2024-05-16
User guide: migrate BMC addressing information from BMO

dtantsur:
Adds a new page "Supported hardware" with information loosely based on
the BMO docs with these modifications:
- Split one large table into sections based on the underlying
  technology.
- Copy supported Dell firmware versions from OpenShift docs.
- Write clarifications on Redfish support status.
- Highlight supporting iLO 6.
- Highlight which iLO 5 drivers support hardware RAID.

### DIFF
--- a/docs/user-guide/src/bmo/supported_hardware.md
+++ b/docs/user-guide/src/bmo/supported_hardware.md
@@ -1,0 +1,97 @@
+# Supported hardware
+
+Metal3 supports many vendors and models of enterprise-grade hardware with
+a *BMC* ([Baseboard Management Controller][bmc]) that supports one of the
+remote management protocols described in this document. On top of that, one of
+the two boot methods must be supported:
+
+1. Network boot. Most hardware supports booting a Linux kernel and initramfs
+   via TFTP. Metal3 augments it with [iPXE][ipxe] - a higher level network boot
+   firmware with support for scripting and TCP-based protocols such as HTTP.
+
+   Booting over network relies on DHCP and thus requires a *provisioning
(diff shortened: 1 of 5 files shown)


dtantsur:
@lentzi90 my thinking was: this information is based on the code in baremetal-operator: both the URL format and the list of available drivers is defined there. So it's sort of a property of BareMetalHost API rather than something Metal3-wide (indeed, Ironic supports more drivers and uses a different format).

lentzi90:
That makes sense, but on the other hand it also propagates to CAPM3 since that builds on top of BMO. So CAPM3 won't work with any other hardware either.

docs/user-guide/src/bmo/supported_hardware.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. Network boot. Most hardware supports booting a Linux kernel and initramfs
+   via TFTP. Metal3 augments it with [iPXE][ipxe] - a higher level network boot
+   firmware with support for scripting and TCP-based protocols such as HTTP.
+
+   Booting over network relies on DHCP and thus requires a *provisioning
+   network* for isolated L2 traffic between the Metal3 control plane and the
+   machines.
+
+2. Virtual media boot. Some hardware model support directly booting an ISO 9660
```
tuminoid, line 16:
```suggestion
1. Virtual media boot. Some hardware model support directly booting an ISO 9660
```
Markdown numbered lists are recommended to be a list of `1.` as they will be auto numbered on rendering, removing the need of adjusting them if list length changes.

tuminoid (commented):
/lgtm

Great addition to the user guide, thanks!

PR #433, opened 2024-05-23
fix broken link to a kubelet kep

tuminoid:
Fix broken link to a kubelet KEP.

Supersedes: #356 as it is abandoned.


lentzi90 (commented):
/approve

Issue #446, opened 2024-06-13
Fix dead links at crs_reference.md


tuminoid:
```
| FILE: ./crs_reference.md
| [✖] https://github.com/metal3-io/metal3-dev-env/tree/master/crs
| 
| 3 links checked.
| 
| ERROR: 1 dead links found!
| [✖] https://github.com/metal3-io/metal3-dev-env/tree/master/crs → Status: 404
```

/triage accepted
/help
/good-first-issue

Split from https://github.com/metal3-io/metal3-docs/issues/328

chibuikeIg:
Hi team, can I be assigned this issue?

tuminoid:
Sure!

/assign @chibuikeIg

chibuikeIg:
Thank you! Will get it done shortly.

Rozzii:
/remove-lifecycle stale
Are you still interested in this @chibuikeIg ?

tuminoid:
Took over, made #499 to close this.

PR #447, opened 2024-06-14
fixed broken link reference to CRs

chibuikeIg:
### Which problem is this PR solving?

Closes issue: Fix dead links at crs_reference.md #446
Fixes: #446

### Description of the changes

Changed from https://github.com/metal3-io/metal3-dev-env/tree/master/crs to https://github.com/metal3-io/metal3-dev-env/tree/b5a53ecf9576c6deab81f692733782d4c5e8091a/crs

### How was this change tested?

Ensured the link works both on Browser and when clicked on the Doc.

### DIFF
--- a/crs_reference.md
+++ b/crs_reference.md
@@ -8,7 +8,7 @@ they reference to each other. To see the example CRs check
 ## Environment Variables
 
 The user is required to set the following environment variables before applying
-the [CRs](https://github.com/metal3-io/metal3-dev-env/tree/master/crs)
+the [CRs](https://github.com/metal3-io/metal3-dev-env/tree/b5a53ecf9576c6deab81f692733782d4c5e8091a/crs)
 
 ```console
 CLUSTER_NAME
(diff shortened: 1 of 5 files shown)


tuminoid:
Thanks for the PR! Please fix the DCO (sign-off the commit) and repush.

/ok-to-test
/cc @kashifest 

PTAL if the linking the last version of the CRS is the right link to replace the dead link, or was the content actually migrated elsewhere.
/hold

dtantsur:
> PTAL if the linking the last version of the CRS is the right link to replace the dead link, or was the content actually migrated elsewhere.

I think it was migrated under the `base` subdirectory, and indeed we should link to the most recent version.

kashifest:
I think this whole crs_reference.md file is obsolete, its still pointing to cluster-api-provider-baremetal! and v1a2 and v1a3 of apis, the crs are non existent in dev-env since I believe

tuminoid:
We have two contradicting opinions it seems. It is the only file in the root of the repo and indeed looks obsolete, so insttead of fixing the link we could delete the whole file? @dtantsur @kashifest

PR #453, opened 2024-06-28
User guide: document RAID

dtantsur:

### DIFF
--- a/docs/user-guide/src/bmo/raid.md
+++ b/docs/user-guide/src/bmo/raid.md
@@ -1,0 +1,173 @@
+# RAID setup
+
+[RAID](https://en.wikipedia.org/wiki/RAID) is a technology that allows creating
+volumes with certain properties out of two or more physical disks. Depending on
+the [RAID level](https://en.wikipedia.org/wiki/Standard_RAID_levels), you may
+be able to merge several disks into a larger one or achieve redundancy.
+
+Metal3 supports two RAID implementation:
+
+- *Hardware RAID* is implemented by hardware itself and can be configured
+  through the machine's BMC.
+- *Software RAID* is implemented by the Linux kernel and can be configured
+  using the standard `mdadm` tool.
+
+To create or delete RAID volumes, you need to edit the `spec.raid` field of the
+`BareMetalHost` resource, changing either the `hardwareRAIDVolumes` or the
+`softwareRAIDVolumes` array. If the host is in the `available` state, it will
+be moved to the `preparing` state and the new settings will be applied. After
+some time, the host will move back to `available`, and the resulting changes
+will be reflected in its `status.raid` field.
+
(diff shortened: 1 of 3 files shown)


docs/user-guide/src/bmo/raid.md
```diff
@@ -0,0 +1,169 @@
...
+
+Metal3 supports two RAID implementation:
+
+- *Hardware RAID* is implemented by hardware itself and can be configured
+  through the machine's BMC.
+- *Software RAID* is implemented by the Linux kernel and can be configured
+  using the standard `mdadm` tool.
+
+To create or delete RAID volumes, you need to edit the `spec.raid` field of the
+`BareMetalHost` resource, changing either the `hardwareRAIDVolumes` or the
```
MahnoorAsghar, line 16: Do we support the softwareRAIDVolumes bit? I remember only testing via hardwareRAIDVolumes myself.

dtantsur: We don't support it in OpenShift, but it exists upstream in Metal3.

MahnoorAsghar: Perhaps we can also mention a little early in the document that software RAID support is experimental.

MahnoorAsghar:
LGTM!

docs/user-guide/src/bmo/raid.md
```diff
@@ -0,0 +1,169 @@
+# RAID setup
+
+[RAID](https://en.wikipedia.org/wiki/RAID) is a technology that allows creating
+volumes with certain properties out of 2 or more physical disks. Depending on
```
tuminoid, line 4:
```suggestion
volumes with certain properties out of two or more physical disks. Depending on
```

docs/user-guide/src/bmo/raid.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+The most common RAID levels are `0`, `1`, `5` and `1+0`. Levels `2`, `6`,
+`5+0` and `6+0` are also supported by Metal3 but may not be supported by all
+hardware models. The level dictates the minimum number of physical disks and
+the maximum size of a RAID volume.
+
+**Note:** because of values like `1+0`, RAID level is a string, not a number.
+
+You can use the boolean `rotational` field to filter out physical disks that
```
tuminoid, line 61: Can we elaborate a bit on this. Does setting rotational to true mean the raid is only created from rotational disks, and false it is created from ssds, and absense of the value (null) that any will be used?

tuminoid (commented):
Again, great addition to documentation. Some minor nits.

dtantsur:
@tuminoid done, please check

tuminoid:
> @tuminoid done, please check

/lgtm
Thanks!

Rozzii (commented):
/approve

kashifest (commented):
/approve

PR #457, opened 2024-07-04
User guide: explain root device hints

dtantsur:
Add plenty of links to the new document since it's quite important and
often overlooked. Explain how the deprecated profiles are related to it.
Provide an example of using HardwareData (another often overlooked
feature).

Also sneak in a couple of fixes for invalid relative URLs (sorry!).

### DIFF
--- a/docs/user-guide/src/bmo/root_device_hints.md
+++ b/docs/user-guide/src/bmo/root_device_hints.md
@@ -1,0 +1,118 @@
+# Root Device Hints
+
+Bare-metal machines often have more than one block device, and in many cases
+a user will want to specify, which of them to use as the root device. *Root
+device hints* allow selecting one device or a group of devices to choose from.
+You can provide the hints via the `spec.rootDeviceHints` field on your
+`BareMetalHost`:
+
+```yaml
+spec:
+  # ...
+  rootDeviceHints:
+    wwn: "0x55cd2e415652abcd"
+```
+
+**Hint:** root device hints in Metal3 are closely modeled on the Ironic's [root
+device hints][ironic-hints], but there are important differences in available
+hints and the comparison operators they use.
+
+**Warning:** the default root device depends on the hardware profile as
+explained below. Currently, `/dev/sda` path is used when no hints are
+specified. This value is not going to work for NVMe storage. Furthermore, Linux
+does not guarantee the block device names to be consistent across reboots.
+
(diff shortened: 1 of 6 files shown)


kashifest (commented):
/approve

docs/user-guide/src/bmo/root_device_hints.md
```diff
@@ -0,0 +1,119 @@
+# Root Device Hints
+
+Bare-metal machines often have more than one block device, and in many cases
```
tuminoid, line 3: One day we will unify how we write bare-metal, baremetal or bare metal, but not today :)

docs/user-guide/src/bmo/root_device_hints.md
```diff
@@ -0,0 +1,119 @@
...
+    "rotational": true,
+    "serialNumber": "drive-scsi0-0-0-0",
+    "sizeBytes": 32212254720,
+    "type": "HDD",
+    "vendor": "QEMU"
+  }
+]
+```
+
+Here you can find all necessary fields to populate the root device hints before
```
tuminoid, line 98:
```suggestion
Here you can find all necessary fields to populate the root device hints before
```
Here is typically used to provide a link or following output, but now its after the output, so here is weird to me.  Could this string be placed above the output?

tuminoid (commented):
A nit.

Thanks for contributing many valuable documentation improvements @dtantsur !

docs/user-guide/src/bmo/root_device_hints.md
dtantsur: Let me see

docs/user-guide/src/bmo/root_device_hints.md
dtantsur: I think, and wikipedia seems to be consistent with it in https://en.wikipedia.org/wiki/Bare_machine, we need to use bare-metal as an auxiliary word and "bare metal" as a separate concept.

jayofdoom: FWIW, it appears in OpenStack, we're officially "Bare Metal" for Ironic. I don't think we use that consistently though.

tuminoid: I'll resolve this for now, thanks for taking a look and providing input @dtantsur @jayofdoom

PR #458, opened 2024-07-08
User guide: copy provisioning states information

dtantsur:
This information is copied more or less as it is from the BMO source
code, including the diagram. I'm not exactly fond of how the information
is presented, but it's the best I can do now (and I strongly believe we
need at least something about the states in our documentation).

The only original part is the introduction and the mention of
operationalStatus/errorType.


docs/user-guide/src/bmo/state_machine.md
```diff
@@ -0,0 +1,132 @@
...
+will stay in the Inspecting state until this process is completed.
+
+### Preparing
+
+When setting up RAID or changing firmware settings, the host will be in
+Preparing state.
+
+### Available
+
+A host in the Available state is available to be provisioned.
```
mboukhalfa, line 106:
```suggestion
A host in the Available state is ready to be provisioned.
```

docs/user-guide/src/bmo/state_machine.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Provisioning
+
+While an image is being copied to the host and it is being configured
```
mboukhalfa, line 110:
```suggestion
While an image is being copied to the host and configured
```

docs/user-guide/src/bmo/state_machine.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Provisioning state machine
+
+![BaremetalHost provisioning state transitions](images/provisioning-states.png)
+
+## Provisioning states
+
+### Created
+
+Newly created hosts move immediately to Discovered or Registering. No
```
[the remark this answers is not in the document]
dtantsur: I was wondering myself when copying this. I'll double-check with the code.

docs/user-guide/src/bmo/state_machine.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Registering
+
+The host will stay in the Registering state while the BMC access
+details are being validated.
+
+### Inspecting
+
+After the host is registered, an IPA image will be booted on it
+using a ramdisk. The agent collects information about the available
```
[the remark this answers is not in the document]
dtantsur: It's also not quite perfect: the host does not do anything itself.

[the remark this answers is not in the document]
mboukhalfa: Well, it is up to you. My aim was just to simplify the sentence. Here, I think we are not describing the specific method of how the image will be loaded to the machine. We are simply stating that, in this state, the VM will boot this image. Whether it gets the image from PXE or a CD-ROM, the boot process is ultimately carried out by the machine itself, isn't it? The way you changed it is also not bad.

docs/user-guide/src/bmo/state_machine.md
```diff
@@ -0,0 +1,135 @@
...
+- `error` -- error has occured, see the `status.errorType` and
+  `status.errorMessage` fields for details.
+- `delayed` -- cannot proceed with the provisioning because the maximum number
+  of the hosts in the given state has been reached.
+- `detached` -- the host is detached, no provisioning actions are possible
+  (see [detached annotation](./detached_annotation.md) for details).
+
+## Provisioning state machine
+
+![BaremetalHost provisioning state transitions](images/provisioning-states.png)
```
zaneb, line 64: Is it worth hotlinking directly to the diagram in the repo (i.e. https://raw.githubusercontent.com/metal3-io/baremetal-operator/main/docs/BaremetalHost_ProvisioningState.png) so that we don't have to keep maintaining a copy over here?
It's hard enough to keep it in sync with the code when they live in the same repo.

zaneb (commented):
/approve

Rozzii (commented):
/lgtm

docs/user-guide/src/bmo/state_machine.md
dtantsur: Yeah, I was thinking of dropping the diagram from there and redo this diagram to be more high-level.

PR #459, opened 2024-07-08
Add support for UEFI HTTP boot

elfosardo:
Proposal to add support for UEFI HTTP boot adding a new hardware driver `redfish-uefihttp` that uses the already implemented ironic `redfish-https` boot interface.

### DIFF
--- a/design/baremetal-operator/uefi-http-boot.md
+++ b/design/baremetal-operator/uefi-http-boot.md
@@ -1,0 +1,132 @@
+# Add UEFI HTTP boot hardware driver
+
+## Status
+
+Implementable
+
+## Summary
+
+This design proposes to add a new hardware driver `redfish-uefihttp` to
+support the UEFI HTTP boot interface using the Redfish API.
+
+## Motivation
+
+HTTP boot is available since at least 9 years, it was standardized in UEFI
+version 2.5, and it's considered as a natural valid replacement for
+PXE boot.
+It addresses PXE issues for security, relying on HTTPs and using a
+reliable TCP connection, and scalability and performance, for example with
+the possibility of using HTTP load balancers.
+In general HTTP boot grants a faster, more reliable and more secure way to
+obtain operating system images and EFI executables during a network boot
+process.
+It is also compatible with UEFI Secure Boot, which is quite hard to impossible
+to implement using standard PXE boot method.
+
+### Goals
+
+Support UEFI HTTP boot interface using Redfish API to boot from an ISO images.
+
+### Non-Goals
+
+- HTTP boot is a UEFI only feature, so support for Legacy (BIOS) can't and
+won't be taken into consideration.
+- Any non-Redfish support for HTTP boot.
+- This implementation covers only booting from an ISO image, it won't take
+into consideration booting using kernel plus ramdisk.
+- Despite HTTP boot being a valid replacement and an improvement on PXE
+boot methods, this change does not aim to replace, change, deprecate, or
+remove the current PXE boot methods in use.
+
+## Proposal
+
+Add a new hardware driver called `redfish-uefihttp`, similar to
+`redfish-virtualmedia`, to the Baremetal Operator.
+The new hardware driver supports UEFI HTTP boot using the ironic
+`redfish-https` boot interface.
+
+## Design Details
+
+Using exclusively Redfish API, the accepted value of `Firmware Interface` and
+`RAIDInterface` for the `redfish-uefihttp` hardware driver will be `redfish`.  
+Secure Boot will also be supported.
+
+When the `redfish-uefihttp` hardware driver is used we need to configure the
+`redfish-https` boot interface in ironic, which is currently not enabled
+in ironic-image.
+
+### Implementation Details/Notes/Constraints
+
+It should be enough to add the new `redfish-uefihttp` hardware driver to
+BMO to support the UEFI HTTP boot feature.  
+As for the `redfish-virtualmedia` driver, the new `redfish-uefihttp` relies
+on network boot from an ISO image that is composed by Ironic, but unlike
+Virtual Media boot, it requires a Provisioning Network to be configured.  
+This is because the UEFI HTTP boot is essentially "secured PXE" designed
+to overcome limitations of the traditional PXE boot method, and it requires
+functional DNS, DHCP and HTTP servers to be configured.  
+The ISO image is provided by the HTTP server and the host boots from it
+exactly like it used to boot from a standard ramdisk during PXE booting.
+
+### Risks and Mitigations
+
+The UEFI HTTP boot provides a high level of security for a network boot
+process, relying on secure communication between BMC and HTTP server to
+transfer all the files and images needed to boot the host.
+
+### Work Items
+
+- Add a new hardware driver called `redfish-uefihttp` to BMO.
+- Add `redfish-https` to supported boot interfaces to the ironic-image.
+- Update metal3 documentation with details on the new hardware driver and
+how to use it.
+
+### Dependencies
+
+Ironic implementation of the `redfish-https` boot interface [has been
+already completed](https://review.opendev.org/c/openstack/ironic/+/900964).
+
+### Test Plan
+
+The new hardwre driver can currently be tested only on supported hardware
+as the Redfish support must be enabled and supported by the BMC.  
+Although the support for UEFI HTTP boot has been added to sushy-tools, so
+the current BMO e2e tests set can be expanded to cover this scenario.
+
+Unit tests will be implemented as usual.
+
+### Upgrade / Downgrade Strategy
+
+The new hardware driver will be available after upgrading the version
+of Baremetal Operator.
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+Adding one more hardware driver with a different boot interface complicates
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,102 @@
+# Add UEFI HTTP boot hardware driver
+
+## Status
+
+Implementable
+
+## Summary
+
+This design proposes to add a new hardware driver `redfish-http-boot` to
```
dtantsur, line 9: I"m slightly inclined to omit the word "boot" and use something like `redfish-uefihttp` for consistency with both the existing `redfish-virtualmedia` and what Ironic has.

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Goals
+
+Support UEFI HTTP boot interface using Redfish API.
+
+### Non-Goals
+
+- HTTP boot is a UEFI only feature, so support for Legacy (BIOS) can't and
+won't be taken into consideration.
+- Any non-Redfish support for HTTP boot.
```
dtantsur, line 32: We need to make it clear that the current approach relies on booting ISOs, not a kernel/ramdisk pair.

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,102 @@
...
+
+When the `redfish-http-boot` hardware driver is used we need to configure the
+`redfish-https` boot interface in ironic, which is currently not enabled
+in ironic-image.
+
+### Implementation Details/Notes/Constraints
+
+It should be enough to add the new `redfish-http-boot` hardware driver to
+BMO to support the UEFI HTTP boot feature.
+
```
dtantsur, line 55: Let's maybe provide a short summary of how the new boot interface works (including the ISO aspect) for those less familiar with ironic news?

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,102 @@
...
+### Dependencies
+
+Ironic implementation of the `redfish-https` boot interface [has been
+already completed](https://review.opendev.org/c/openstack/ironic/+/900964).
+
+### Test Plan
+
+The new hardwre driver can currently be tested only on supported hardware
+as the Redfish support must be enabled and supported by the BMC.
```
dtantsur, line 77: Looks like sushy-tools is supported? https://review.opendev.org/c/openstack/sushy-tools/+/923592

If so, we'll need to extend the BMO e2e tests.

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,102 @@
...
+The new hardware driver will be available upgrading the version of Baremetal
+Operator.
+
+### Version Skew Strategy
+
+None
+
+## Drawbacks
+
+None
```
dtantsur, line 92: Complicating the operator's decision is definitely a drawback. We'll need to explain in which cases each boot interfaces will be chosen.

design/baremetal-operator/uefi-http-boot.md
dtantsur: The new boot interface will require provisioning network, correct? Worth mentioning as well.

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Alternatives
+
+The users will keep using other hardware drivers already implemented in
+Baremetal Operator API.
```
dtantsur, line 97: This section should explain why the existing drivers don't cover all cases, for instance:

1. iPXE is incompatible with secure boot and may have issues on certain platforms.
2. Virtual media is a heavy thing to implement on the BMC side and requires providing the BMC with access to the control plane.

design/baremetal-operator/uefi-http-boot.md
elfosardo: that makes sense, I'll change it

design/baremetal-operator/uefi-http-boot.md
elfosardo: yep, I was unsure, thanks for verifying it

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,126 @@
...
+### Goals
+
+Support UEFI HTTP boot interface using Redfish API to boot from an ISO images.
+
+### Non-Goals
+
+- HTTP boot is a UEFI only feature, so support for Legacy (BIOS) can't and
+won't be taken into consideration.
+- Any non-Redfish support for HTTP boot.
+- This implementation covers aonly booting from an ISO image, it won't take
```
dtantsur, line 33: nit: only

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,126 @@
...
+HTTP boot is available since at least 9 years, it was standardized in UEFI
+version 2.5, and it's considered as a natural valid replacement for
+PXE boot.
+It addresses PXE issues for security, relying on HTTPs and using a
+reliable TCP connection, and scalability and performance, for example with
+the possibility of using HTTP load balancers.
+In general HTTP boot grants a faster, more reliable and more secure way to
+obtain operating system images and EFI executables during a network boot
+process.
+
```
dtantsur, line 23: nit: Worth mentioning UEFI secure boot: iPXE is very hard to make compatible with it.

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,126 @@
...
+None
+
+## Drawbacks
+
+Adding one more hardware driver with a different boot interface complicates
+the decision process to decide which boot interface to use, in particular in
+this case between `redfish-virtualmedia` and `redfish-uefihttp` since they both
+rely on network booting from an ISO image.  
+In general, the main difference is that the UEFI HTTP boot supports Secure
+Boot, while virtual media does not, but on the other hand, no configuration
```
Rozzii, line 109: How come no configuration drive functionality? Is it not possible to boot IPA thus not possible to send over the config drive via IPA API? or what do you men by no configuration drive functionality? Could you please clarify this?

Rozzii (commented):
Generally looks fine this feature would be a nice addition, I am a bit unclear on the config drive limitations as per my comment.

design/baremetal-operator/uefi-http-boot.md
dtantsur: I don't think Riccardo meant that, but I agree, it is confusing.

elfosardo: yeah, sorry for the confusion, I had to unconfuse myself too here!
this is not related to IPA at all
what I meant is that no static configuration can be injected at boot time, DHCP is a hard requirement
does that help?

Rozzii: Bare with me I am bit slow to understand this, what do you mean by static configuration? 
Do you mean we can't use the feature with servers that have static IPs for their interfaces? So same way as for iPXE the UEFI HTTP booting also requires the BOOTP protocol thus there must be a DHCP server present at all times to facilitate the boot media discovery?

elfosardo: @Rozzii exactly that
I will change the wording

tuminoid:
> Generally looks fine this feature would be a nice addition, I am a bit unclear on the config drive limitations as per my comment.

Holding for this, feel free to clear when cleared.
/hold

Rozzii (commented):
/lgtm

It looks okay to me, I won't remove the "hold" because there might be further discussion by others.

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,132 @@
...
+The new hardwre driver can currently be tested only on supported hardware
+as the Redfish support must be enabled and supported by the BMC.  
+Although the support for UEFI HTTP boot has been added to sushy-tools, so
+the current BMO e2e tests set can be expanded to cover this scenario.
+
+Unit tests will be implemented as usual.
+
+### Upgrade / Downgrade Strategy
+
+The new hardware driver will be available upgrading the version of Baremetal
```
MahnoorAsghar, line 100: nit: ...will be available after upgrading...

design/baremetal-operator/uefi-http-boot.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Alternatives
+
+The users will keep using other hardware drivers already implemented in
+Baremetal Operator API.  
+The existing interfaces do not cover some important cases and have some
+limitations, such as:
+
+- iPXE does not support Secure Boot.
+- To rely on Virtual Media boot is necessary to grant access to the
```
MahnoorAsghar, line 126: nit: To rely on Virtual Media boot, it is necessary...

elfosardo:
thanks @MahnoorAsghar , I've fixed the nits

MahnoorAsghar:
LGTM :)

elfosardo:
/unhold
a week has passed

Rozzii (commented):
/lgtm

PR #461, opened 2024-07-25
Ironic Standalone Operator

dtantsur:
This design proposal discusses ironic-standalone-operator: provides a
motivation for its creation, describes its goals and outlines the
current design and features, as well as the plans for the nearest
future.

### DIFF
--- a/design/ironic-standalone-operator.md
+++ b/design/ironic-standalone-operator.md
@@ -1,0 +1,375 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Ironic Standalone Operator
+
+## Status
+
+implementable
+
+## Summary
+
+This proposal discussed the [ironic-standalone-operator][ir-op] project that
+was written using inspiration from OpenShift's
+[cluster-baremetal-operator][cbo]. The project is already under the Metal3's
+umbrella, so this document serves to describe its design goals, future plans
+and the rough API shape.
+
+[ir-op]: https://github.com/metal3-io/ironic-standalone-operator
+[cbo]: https://github.com/openshift/cluster-baremetal-operator
+
+## Motivation
+
+Ironic is not trivial to install and operate. Although we provide
+[ironic-deployment scripts][ironic-deployment] as part of BMO, there are still
+a lot of moving parts where things can be wrong. Configuring Ironic through
+environment variables is error-prone and complicates upgrades. The operator
+pattern is standard and ubiquitous in the Kubernetes world to manage complex
+software. Metal3 should use it as well.
+
+[ironic-deployment]: https://github.com/metal3-io/baremetal-operator/tree/main/ironic-deployment
+
+### Goals
+
+- Provide a recommended way to install Ironic and its satellite services for
+  using with Metal3.
+- Make it easy to install and manage Ironic for Metal3 newcomers.
+- Provide a Kubernetes operator that can also be used outside of Metal3.
+- Pave the way for highly-available Ironic installations.
+
+### Non-Goals
+
+Explicitly not planned:
+
+- Tailor the new operator to use cases outside of Metal3 except for the most
+  minor things.
+- Support for versions of ironic-image predating this proposal (e.g. the ones
+  containing ironic-inspector).
+- Deprecate or discourage alternative ways to install Ironic for Metal3, get
+  rid of ironic-image/mariadb-image.
+- Install BMO, IPAM or CAPM3 via the same operator.
+
+May happen in the future but not as part of this proposal:
+
+- Radically change the installed architecture. For example, we have bold plans
+  to look into dropping the host networking requirement.
+- Stabilize the API design.
+
+## Proposal
+
+### User Stories
+
+As an administrator, I want to be able to install Ironic in a way that is
+suitable for Metal3 by installing an operator and creating custom Kubernetes
+resources.
+
+## Design Details
+
+This proposal adds a new repository under the Metal3 umbrella:
+ironic-standalone-operator. It is a Kubernetes operator that exposes a few
+Custom Resources and manages an Ironic installation.
+
+### Naming
+
+The project has undergone a heavy discussion on its naming. The initial name
+was straightforward: ironic-operator. However, it was quickly found to conflict
+with a few existing projects, including a pretty active one developed by Red
+Hat as part of its OpenStack offering.
+
+Another candidate was metal3-ironic-operator. The arguments against it were
+inconsistency with other Metal3 projects (we don't call BMO
+metal3-baremetal-operator even though baremetal-operator is pretty generic) and
+the desire to make the new operator usable outside of pure Metal3.
+
+The argument against using the word "standalone" was that this word is
+overloaded in the OpenStack context and may be unclear to people without this
+context. A poll among contributors showed that the intention of the word is at
+least more or less clear to us, and that it clearly conveys the difference from
+the Red Hat's OpenStack operator.
+
+A few code names were also discussed but ruled out because of a potential user
+confusion and possible trademark issues.
+
+Note that there is no established acronym for ironic-standalone-operator like
+we have for baremetal-operator (BMO) or CAPM3. Using ISO is definitely going to
+be confusing. This document will be referring to ironic-standalone-operator or
+just "the operator" in cases where it does not cause confusion with a human
+operator.
+
+### Implementation Details/Notes/Constraints
+
+This section describes the current state of the project. It is not an attempt
+to fix the details forever. I'm using it to give a reader a clearer idea what
+the operator currently does.
+
+#### Current architecture
+
+The operator has two controllers: for MariaDB and for Ironic plus its auxiliary
+containers.
+
+The MariaDB controller, also referred to as the *database controller* in this
+context, starts a MariaDB instance in a *deployment* using
+[mariadb-image][mariadb-image]. As with Metal3 now, MariaDB is optional: if it
+is not configured, SQLite is used instead.
+
+The Ironic controller starts and manages the following components:
+
+- Ironic itself
+- HTTPD for serving images and iPXE scripts
+- Dnsmasq for DHCP and TFTP
+- Ramdisk logs publisher
+- IPA downloader
+
+All these components are used in the same way as in a traditional Metal3
+installation. Note that the IPA Downloader fate is under discussion: there is
+a strong desire to make it optional and maybe replace with a different method
+of delivering IPA images.
+
+Unlike the current Metal3, the operator requires authentication and will create
+secrets with random credentials when a user does not provide them. We're
+considering to do the same with TLS, but it requires [figuring out CA
+integration][issue4].
+
+[mariadb-image]: https://github.com/metal3-io/mariadb-image/
+[issue4]: https://github.com/metal3-io/ironic-standalone-operator/issues/4
+
+#### HA architecture
+
+The *non-HA* architecture is the architecture that Metal3 uses now. All Ironic
+components are run in a single *deployment*.
+
+The *HA* architecture is a new concept in ironic-standalone-operator. It
+involves running a copy of Ironic and HTTPD per control plane node (so, 3
+copies in most cases). This has two benefits:
+
+1. Ironic can be updated in a rolling fashion without an interruption in the
+   service.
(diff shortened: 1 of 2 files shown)


design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+[cluster-baremetal-operator][cbo]. The project is already under the Metal3's
+umbrella, so this document serves to describe its design goals, future plans
+and the rough API shape.
+
+[ir-op]: https://github.com/metal3-io/ironic-standalone-operator
+[cbo]: https://github.com/openshift/cluster-baremetal-operator
+
+## Motivation
+
+Ironic is not trivial to install and operate. Also we provide
```
elfosardo, line 27: "Even if we provide" ?

dtantsur: Also-vs-Although 🤦🏽 Will fix if I need to update.

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+
+The `Ironic` object is much more complex and should probably be split into more
+custom resources as we polish its internal architecture. Currently, it uses
+nested structures to logically group fields. Here are the most important fields
+(omitting various fine-tuning for brevity):
+
+- `credentialsRef` - a reference to a secret with credentials (generated if
+  missing)
+- `databaseRef` - a reference to an `IronicDatabase` object (if needed)
+- `distributed` - a boolean flag that enables the HA architecture
```
MahnoorAsghar, line 250: Why not just call this field `highlyAvailable` or `highAvailability`

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+- `networking` - a nested structure that defines networking (see below)
+- `nodeSelector` - a selector for nodes to run Ironic on
+- `tlsRef` - a reference to a TLS secret to use for the service
+
+The `networking` sub-structure deserves a separate consideration:
+
+- `apiPort`, `imageServerPort`, `imageServerTLSPort` allow overriding listening
+  ports for the services
+- `bindInterface` - a boolean flag that makes Ironic listen on only the
```
MahnoorAsghar, line 259: It may sound too simplistic, but can we call this provisioningInterfaceOnly? )bindInterface can sound like another kind of interface.)

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+  provisioning interface
+- `dhcp` - another nested structure with DHCP parameters (see below)
+- `externalIP` - IP through which nodes deployed over virtual media access
+  Ironic and HTTPD
+- `interface`, `ipAddress`, `macAddresses` - various ways to specify the
```
MahnoorAsghar, line 264: Similarly, how does provisioningInterface, provisioningIPAddress and provisioningMACAddress sound?

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+### Upgrade / Downgrade Strategy
+
+By default, the operator will be tightly coupled with the version of Ironic
+(and, eventually, IPA) that it installs. A release of the operator will follow
+each release of ironic-image, and they release branches will match. The `main`
+branch will continue following the latest container image.
+
+After each full reconciliation, the operator will store the version of Ironic
+it has just installed in the `status`. In the future, this will allow to apply
+any logic on upgrade. However, we will try to keep ironic-image self-upgrading
```
MahnoorAsghar, line 325: 'However, we will try to keep ironic-image self-upgrading for the sake of users that do not use ironic-standalone-operator.' What do you mean?

design/ironic-standalone-operator.md
dtantsur: That's a good question. I guess `distributed` is more precise technically, while `highAvailability` may be more user-friendly. I'd like to hear more opinions on the topic.

design/ironic-standalone-operator.md
dtantsur: There is a potential for confusion here. We call "provisioning network" a *dedicated* network for the PXE and other provisioning traffic. Ironic can work without it, and the variables work regardless of whether you have a "provisioning network".

Maybe I'm overthinking it? Would like more opinions here as well.

design/ironic-standalone-operator.md
dtantsur: I concur but see below about the word "provisioning".

design/ironic-standalone-operator.md
dtantsur: I mean, if you just use ironic-image, and you upgrade it by replacing the container, we will try to support this case still. Any ideas on how to phrase it better?

design/ironic-standalone-operator.md
dtantsur: If we do decide to change this, I'd rather rename the whole "networking" sub-field into something like "provisioningNetwork" to avoid duplicate prefixes.

MahnoorAsghar: But the rest of the sub-fields seem unrelated to the provisioning network (like dhcp and external ip)?
How would the interface, ipAddress and macAddresses fields work if there is no provisioning interface?
(I am not too familiar with ironic's networking, so just trying to understand the expected behaviour)

design/ironic-standalone-operator.md
MahnoorAsghar: You mean that for example, if someone was running ironic-standalone-operator v1.0 (with ironic-image v1.0), and just replaced their ironic-image with v1.1, ironic-standalone-operator will continue to work?
If that's what you mean, then 'for the sake of users that do not use ironic-standalone-operator.' can be confusing..

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+rather than a clean one (as is the case for SQLite). To accommodate this:
+
+- We'll create a new container entrypoint that will run the *online data
+  migrations* for Ironic while the service is already running (a part of
+  the upgrade process that we've ignored so far).
+- We'll probably need to upgrade the database schema separately from Ironic to
+  avoid running it 3+ times in parallel. Maybe it will take a form of a *Job*.
+- BMO will need to be updated to handle more cases of unexpected provision
+  state. E.g. what needs to be done when the BMH is *inspecting* but the node
+  is found in a completely wrong state like `active` or `clean wait`.
```
hardys, line 345: There is an additional complexity to consider if we're going to support mariadb combined with persistent volumes, when the mariadb version changes you either have to ensure clean shutdown of the container or enable some kind of backup/restore workflow due to the mariadb checks around versioning for the redo log - [see here](https://github.com/suse-edge/charts/pull/135#issuecomment-2236635629) for example.

This can likely be resolved with with pod lifecycle hooks, or perhaps some kind of job pre/post upgrade but it's additional complexity which I don't think we currently consider in the community deployment examples.

design/ironic-standalone-operator.md
dtantsur: DHCP is very-very much related to the provisioning network: it's DHCP on the provisioning network. ExternalIP is not directly related indeed, maybe we'd need to move it out then.

> How would the interface, ipAddress and macAddresses fields work if there is no provisioning interface?

There must be a provisioning interface, but it can be on some existing network rather than an isolated provisioning network. Confusing, I know. This is why I'm pondering our usage of "provisioning" here.

design/ironic-standalone-operator.md
dtantsur: Ah, I see the source of confusion. What I wanted to say was: users of ironic-image without the operator will be able to do it still. I'll work on a different way to phrase it.

design/ironic-standalone-operator.md
dtantsur: Very good point, thank you!

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+provisioning network: for dnsmasq and for httpd/ironic. This is a new
+operational requirement that I'd like to avoid at this stage.
+
+Q: Why does the same dnsmasq limitation affect the HA architecture. A: The HA
+architecture is completely new here, so we can introduce new requirements
+without regression in the operational experience.
+
+Q: Why using *DaemonSets* if *StatefulSets* provide us an easier way to address
+separate Ironic instances? A: While we're relying on host networking, making
+several Ironic instances co-exist on the same Kubernetes node is too complex.
```
Rozzii, line 223: From Kubernetes HA perspective it is also more reliable to have the ironic pods present on separate control plane nodes. In other words clustering all Ironic pods on e.g. 1  controlplane node wouldn't be HA anyways as that node could be lost.

Rozzii (commented):
/approve

Rozzii (commented):
/lgtm
From my perspective it is fine given that you will address the comments.

Rozzii:
/hold
Just holding to avoid someone putting approve on it while my lgtm is active and merging it

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+certificate that is valid for its RPC access IP or hostname. This problem has
+been extensively discussed in the [initial HA proposal][issue3], and here is
+the proposed solution, at least for the MVP case:
+
+The Ironic controller will generate a self-signed CA and pass its public and
+private parts into each Ironic container. Each Ironic container will generate
+its private key certificate and sign the certificate with this CA. The CA will
+be trusted **only** for the RPC purpose, removing the possibility of abuse.
+To reduce the number of code paths, this process will happen unconditionally,
+even when TLS for Ironic itself is not enabled.
```
kashifest, line 205: Just for my understanding what prevents us from enabling TLS unconditionally for ironic itself?

design/ironic-standalone-operator.md
kashifest: I would vote for highAvailability or haEnabled

kashifest (commented):
Thanks a lot @dtantsur few questions/nits inline

kashifest:
I can approve it already and hold can be taken back when the reviews are addressed
/approve

design/ironic-standalone-operator.md
```diff
@@ -0,0 +1,369 @@
...
+
+### User Stories
+
+As an administrator, I want to be able to install Ironic in a way that is
+suitable for Metal3 by installing an operator and creating custom Kubernetes
+resources.
+
+## Design Details
+
+This proposal adds a new project under the Metal3 umbrella:
```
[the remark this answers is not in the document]
dtantsur: Project is overloaded, right? I'll change to repository.

design/ironic-standalone-operator.md
dtantsur: Well, nothing really. We'd need to decide if we want to use a CertificateSigningRequest (and require a manual approval) or just generate a self-signed certificate. Opinions?

design/ironic-standalone-operator.md
dtantsur: Okay, let's go with highAvailability.

design/ironic-standalone-operator.md
Rozzii: I guess the end goal would be to have an integration with cert-manager as in most K8s environments cert-manager is handling the certificates.

design/ironic-standalone-operator.md
Rozzii: I would also say to not use "provisioning", on Bare Metal Life Cycle management level the name "Provisioning Nework" makes sense but if we just talk about an ironic instance not that much. Everything Ironic does with the machin goes through the provisioning network, ofc he BMC might have it's own separate network but ipxe, dhcp, bootp,  http image server, IPA - Ironic communication they all happen in the "provisioning" network but from Ironic perspective it is simply "the Network".

Rozzii:
/lgtm
@dtantsur  feel free to remove the hold if you feel the time is right.

Rozzii (commented):
/lgtm

MahnoorAsghar:
LGTM

design/ironic-standalone-operator.md
dtantsur: My only worry about the manual step (approving the cert request). Is it also considered standard?

I mean, in the situation when the admin has not provided any certificate via the secret.

kashifest:
> @kashifest do you feel like your comments have been (at least mostly) addressed? The TLS one is probably the only outstanding (and we can discuss it later if you wish).

I am fine with this as is. We can continue the TLS discussion as you suggested.
/hold cancel

PR #471, opened 2024-09-17
Add white background to ipa-provisioning.png

peppi-lotta:
Image is illegible when browser is in dark mode. Added a white background to image to fix this. 
Image in question is the last one in this page: https://book.metal3.io/bmo/introduction


tuminoid (commented):
/lgtm

Good catch, it is horrible in dark mode.

Rozzii (commented):
/approve
Thanks, good catch!

PR #475, opened 2024-10-03
Migrate BareMetalHost and HardwareData API docs from BMO to user-guide

SafeEHA:


Rozzii:
IMO this looks good.
/approve

tuminoid (changes requested):
We already have status annotation page, this creates second one. We also have BMH states already.

Also none of these pages are linked in the index, so they're not visible to readers.

Rozzii:
> We already have status annotation page, this creates second one. We also have BMH states already.
> 
> Also none of these pages are linked in the index, so they're not visible to readers.

Ohh you are right, @SafeEHA  as @tuminoid  mentioned  we have these pages:  https://book.metal3.io/bmo/inspect_annotation and https://book.metal3.io/bmo/status_annotation , if there is any extra info in the annotations.md then move it to the inspect and status annotation pages please otherwise you can just delete that from the BMO side.

tuminoid:
> Thanks, @tuminoid, i noticed the status and inspect annotations already existing in the user-guide have slightly different content in comparison to the ones on BMO, the code block is yaml formatted on the user guide currently and json formatted on BMO, the descriptive content is also slightly different... should I go ahead to take out the new annotations file i created since it already exists? I'll go ahead to make API visible on the index

We should merge the documents so the user guide is as comprehensible as possible.

tuminoid:
> One final concern, the annotation files on BMO (which is the files referenced in the issue) have different content with the files in user-guide currently, should i focus on merging the files in the user-guide only? its a bit confusing @tuminoid

@Rozzii can you assist @SafeEHA which portions to keep and which take?

dtantsur:
/hold

I don't think the PR is in the right direction.

First, the remaining API docs were not meant to get migrated from BMO.

Second, while inspect and status annotations have some use case overlaps, they also have completely disjoint use cases. If we do decide to merge the pages, we need to add redirects to old pages so that existing links don't lead into nowhere.

Finally, any missing API features we migrate here should be written from a case-based perspective. What am I trying to achieve? How do I do it? What are possible issues?

docs/user-guide/src/bmo/api.md
```diff
@@ -0,0 +1,117 @@
...
+1. The host `spec.image.url` field must contain a URL for a valid
+   image file that is visible from within the cluster and from the
+   host receiving the image.
+2. The host must have `online` set to `true` so that the operator will
+   keep the host powered on.
+3. The host must have all of the BMC details.
+
+To initiate deprovisioning, clear the image URL from the host spec.
+
+## Unmanaged Hosts
```
dtantsur, line 26: This is covered in the state machine docs (expand if needed)

docs/user-guide/src/bmo/api.md
```diff
@@ -0,0 +1,117 @@
...
+
+**Metal³** introduces the concept of **BareMetalHost** resource, which
+defines a physical host and its properties. See [BareMetalHost
+CR](https://doc.crds.dev/github.com/metal3-io/baremetal-operator/metal3.io/BareMetalHost/v1alpha1)
+or check the source code at `apis/metal3.io/v1alpha1/baremetalhost_types.go`
+for a detailed API description. See the [user
+guide](https://book.metal3.io/bmo/introduction) for information on how to
+manage hosts.
+
+## Triggering Provisioning
```
dtantsur, line 13: This is already covered in more details in the provisioning docs

docs/user-guide/src/bmo/api.md
```diff
@@ -0,0 +1,117 @@
...
+data retrieved from Ironic. HardwareData holds the same name and
+namespace as its corresponding BareMetalHost resource. Currently,
+HardwareData doesn't have *Status* subresource but only the *Spec*.
+
+See [HardwareData
+CR](https://doc.crds.dev/github.com/metal3-io/baremetal-operator/metal3.io/HardwareData/v1alpha1)
+or check the source code at `apis/metal3.io/v1alpha1/hardwaredata_types.go`
+for a detailed API description.
+
+## PreprovisioningImage
```
dtantsur, line 98: This also deserves its own advanced page

Rozzii:
This should be finished
/remove-lifecycle stale

PR #481, opened 2024-10-22
User guide: fix missing checksumType in samples

dtantsur:
I completely forgot that BMO defaults to md5, not auto. These examples
don't work because Ironic is trying to match a SHA256 checksum against
an MD5 one.


lentzi90 (commented):
Good catch!
/lgtm

PR #487, opened 2024-11-27
Add v27.0 supported ironic-image

elfosardo:


dtantsur:
DCO missing.

Also, do we need to remove one version from supported at the same time?

elfosardo:
> DCO missing.
Fixed!

> Also, do we need to remove one version from supported at the same time?
not necessarily, we agreed to have up to 3 supported branches
we will move v25.0 to Tested and v24.0 to EoL once we remove 1.6 release series after 1.9 series is released

tuminoid:
> > DCO missing.
> 
> Fixed!
> 
> > Also, do we need to remove one version from supported at the same time?
> 
> not necessarily, we agreed to have up to 3 supported branches we will move v25.0 to Tested and v24.0 to EoL once we remove 1.6 release series after 1.9 series is released

24.1 is also there, used by 1.7 which is moving from Supported to Tested when 1.9 is released. 24.1 will remain in Tested  as well.

25.0 is kind of weird fall between, as it is not matching to any CAPM3 release, so it is only tested when someone sends a PR there. There is no periodics targeting any Ironic-image release directly.

tuminoid (commented):
/lgtm

We can adjust the CI test matrix, when 1.9 comes out. 27.0 will be paired with 1.9 series then, while it is now paired with 1.8 as 1.9 is not out yet.

Issue #435, opened 2024-05-24
Troubleshooting guide


lentzi90:
We should add a troubleshooting guide with the most common issues and solutions. Since we integrate with Cluster API, it also makes sense to link to [their troubleshooting page](https://cluster-api.sigs.k8s.io/user/troubleshooting), and if possible give some guidance on when an issue is with CAPI and when it is with Metal3.

## Things to include

BMO/Ironic/IPA:

- How to verify that BMO and Ironic are operational to rule out configuration errors
  - Containers should be running (not restarting, this could indicate that BMO cannot connect to Ironic)
  - Ironic should not be "waiting for IP". Check Ironic logs!
- What does inspection errors look like, what could be the cause and solution?
  - BMC credentials could be wrong or missing. These issues should show up in the BareMetalHost status or as events (to be confirmed)
  - The host is not able to communicate back results to Ironic (this will result in a timeout). Access to serial logs is needed to determine the exact issue in these cases.
  - Incompatible configuration. For example, attempt to use virtualmedia or UEFI when not supported. This should show up in the BareMetalHost status or as events (to be confirmed).
- Provisioning errors
  - Image errors (wrong checksum, missing image, image too large to decompress or too large for disk).
    This should show up in the BMH status (to be confirmed)
  - No root device found
    This should show up in the BMH status (to be confirmed)

CAPM3/IPAM:

- No BareMetalHost available/matching. This should show as an event when describing the Metal3Machine (to be confirmed)
- Provider ID missing. This can happen if `noCloudProvider` is set to `false` on the Metal3Cluster when no external cloud provider is used.
- nodeRef missing. Really a CAPI level issue. Could be caused by failure to boot the image or failure to join it to the cluster. Access to the node or serial logs is needed to determine the issue. Especially cloud-final logs are of interest.

PR #490, opened 2024-12-13
Troubleshooting: Verify BMO and Ironic healthy

lentzi90:
Add section on how to verify that BMO and Ironic are healthy as well as
example output for both healthy and unhealthy examples.

I also moved the troubleshooting doc to the top level. My reasoning for this is that users may not know at what level an error is. They need a single place to search instead of first having to figure out if the issue belongs under BMO or CAPM3 or something else.

Part of #435


docs/user-guide/src/troubleshooting.md
```diff
@@ -0,0 +1,72 @@
+# Troubleshooting
+
+## Verify that Ironic and Baremetal Operator are healthy
```
tuminoid, line 3: Consistency: either Baremetal or BareMetal, you have both in this change.

docs/user-guide/src/troubleshooting.md
```diff
@@ -0,0 +1,72 @@
...
+  automatedCleaningMode: disabled
+```
+
+Alternatively, you can wait for 3 cleaning retries to finish. After that, the
+host will be deleted. If you do care about cleaning, you need to figure out why
+it does not finish.
+
+## MAC address conflict on registration
+
+If you force deletion of a host after registration, BareMetal Operator will not
```
tuminoid, line 61: Here is example of BareMetal in paragraph.

tuminoid (commented):
Nit.

tuminoid (commented):
/lgtm

docs/user-guide/src/SUMMARY.md
```diff
@@ -46,4 +45,5 @@
 - [Try it (for developers)](developer_environment/tryit.md)
 - [Version Support](version_support.md)
 - [Project Security Policy](security_policy.md)
+- [Troubleshooting FAQ](troubleshooting.md)
```
dtantsur, line 48: It could be harder to discover the guide here, especially for people who're not interested in IPAM or even CAPM3. Maybe at least move it after IPAM and group the last 3 items into something like "For developers"?

docs/user-guide/src/troubleshooting.md
dtantsur, line 3: Honestly, I liked the FAQ format where each title is how a user would google it. Like, what would prompt an operator to look at this section you're adding?

lentzi90: Sure, I also like that format, but I know also that an unhealthy Ironic or BMO could show itself in probably dozens of ways. It is more of a sanity check and something I would expect to direct users to when they are having issues that are hard to pinpoint. So I think it is important to have this somewhere. If you want, we can split this into troubleshooting and FAQ maybe?

docs/user-guide/src/SUMMARY.md
lentzi90: I moved it up now but didn't add the group for now. The thing is, if we add a group, that has to link to a proper page also or the build fails. So we would need an "intro page" for that group. I'm also a bit hesitant to "hide" the version support and security policy in that way since they may well be of interest to users. Finally, if we do the group, it would be logical to also change the path (e.g. `/security_policy.md` -> `development/security_policy.md`) and I believe we have links to the policy from practically every repo in the organization :grimacing:

tuminoid: Don't hide security policy, it must stay on top level.

docs/user-guide/src/troubleshooting.md
```diff
@@ -0,0 +1,72 @@
...
+baremetal-operator-controller-manager-85b896f688-j27g5   1/1     Running   0             3m35s
+ironic-6bcdcb99f8-6ldlz                                  1/3     Running   1 (24s ago)   3m37s
+```
+
+### Waiting for IP
+
+Make sure to check the logs also since Ironic may be stuck on "waiting for IP".
+
+```bash
+kubectl -n baremetal-operator-system logs deploy/ironic
```
dtantsur, line 40: I suspect this won't work with IrSO... maybe should an example using the just shown pod name `ironic-6bcdcb99f8-6ldlz`?

More importantly, this section lacks any pointers on what to do about the "waiting for IP" problem (check networking in IrSO or wherever we normally have it?)

lentzi90: Updated

tuminoid:
/cc @elfosardo @dtantsur 
Have we agreed how its going going to be setup?

dtantsur:
/approve

I have some doubts about the format, but the new text is a clear improvement, so let's merge it.

PR #493, opened 2024-12-24
User guide: first take at documenting IrSO

dtantsur:

### DIFF
--- a/docs/user-guide/src/irso/install-basics.md
+++ b/docs/user-guide/src/irso/install-basics.md
@@ -1,0 +1,162 @@
+# Installing Ironic
+
+This document covers installing Ironic in different scenarios. You need to
+answer a few questions before you can pick the one that suits you:
+
+- Which physical network interface will be used for provisioning? Without any
+  configuration, Ironic will use the host cluster networking.
+
+- If you use a dedicated network interface, are you going to use the built-in
+  Keepalived service to configure the IP address on the control plane node
+  where the Ironic pod is located? If not, you need to make sure the interface
+  has a usable address on this node.
+
+- Do you want to use network boot (iPXE) during provisioning? DHCP adds more
+  requirements and requires explicit configuration. Without it, only virtual
+  media provisioning is possible (see [supported
+  hardware](../bmo/supported_hardware.md)).
+
+- Are you going to use TLS for the Ironic API? It is not recommended to run
+  without TLS. To enable it, you need to manage the TLS secret. [Cert
+  Manager](https://cert-manager.io/) is the recommended service for it.
+
+## Using Ironic
+
+Regardless of the scenario you choose, you will need to create at least an
+`Ironic` object and wait for it to become ready:
+
+```bash
+NAMESPACE="test-ironic"  # change to match your deployment
+kubectl create -f ironic.yaml
+kubectl wait --for=condition=Ready --timeout="10m" -n "$NAMESPACE" ironic/ironic
+```
+
+If the resource does not become `Ready`, check its status and the status of the
+corresponding `Deployment`.
+
+Once it is ready, get the credentials from the associated secret, e.g. with
+
+```bash
+SECRET=$(kubectl get ironic/ironic -n "$NAMESPACE" --template={{.spec.apiCredentialsName}})
+USERNAME=$(kubectl get secrets/$SECRET -n "$NAMESPACE" --template={{.data.username}} | base64 -d)
+PASSWORD=$(kubectl get secrets/$SECRET -n "$NAMESPACE" --template={{.data.password}} | base64 -d)
+```
+
+Now you can point BMO at the Ironic's service at `ironic.test-ironic.svc`.
+
+## Scenario 1: no network boot, no dedicated networking
+
+In this scenario, Ironic will use whatever networking is used by the cluster.
+No DHCP will be available, bare-metal machines will be provisioned using
+virtual media. Since there is no dedicated network interface, Keepalived is
+also not needed.
+
+It is enough to create the following resource:
+
+```yaml
+apiVersion: ironic.metal3.io/v1alpha1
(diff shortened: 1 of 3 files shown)


docs/user-guide/src/irso/install-basics.md
```diff
@@ -0,0 +1,157 @@
...
+- If you use a dedicated network interface, are you going to use the built-in
+  Keepalived service to manage the IP address on it? If not, you need to make
+  sure the interface has a usable address on each control plane node.
+
+- Do you want to use network boot (iPXE) during provisioning? DHCP adds more
+  requirements and requires explicit configuration. Without it, only virtual
+  media provisioning is possible (see [supported
+  hardware](../bmo/supported_hardware.md)).
+
+- Are you doing to use TLS for the Ironic API? It is not recommended to run
```
MahnoorAsghar, line 18: nit: 'going' instead of 'doing'

docs/user-guide/src/irso/install-basics.md
```diff
@@ -0,0 +1,157 @@
+# Installing Ironic
+
+This document covers installing Ironic in different scenarios. You need to
+answer a few questions before you can pick the one that suits you:
+
+- Which physical network interface will be used for provisioning? Without any
+  configuration, Ironic will use the host cluster networking.
+
+- If you use a dedicated network interface, are you going to use the built-in
```
MahnoorAsghar, line 9: A little bit confused about this: if we have multiple control plane nodes, should we make sure one interface on each control plane node has an IP address (which means one ironic per control plane node using that IP address)? Shouldn't that IP address be the same, or shouldn't we have a proxy?

docs/user-guide/src/irso/introduction.md
```diff
@@ -0,0 +1,72 @@
...
+On every source code change, a new IrSO image is built and published at
+`quay.io/metal3-io/ironic-standalone-operator`. To install it in your cluster,
+you can use the Kustomize templates provided in the source repository:
+
+```console
+git clone https://github.com/metal3-io/ironic-standalone-operator
+cd ironic-standalone-operator
+git checkout -b <DESIRED BRANCH OR main>
+
+make install deploy
```
lentzi90, line 26: In the future we should start publishing the output of `kustomize build` as a release artifact so users can just do a `kubectl apply -f [URL]`. This is fine for now though, no need to change!

docs/user-guide/src/irso/introduction.md
```diff
@@ -0,0 +1,72 @@
...
+See [installing Ironic with IrSO](./install-basics.md) for information on how
+to use these resources.
+
+## How is Ironic installed?
+
+By default, IrSO installs Ironic as a single pod on a **control plane** node.
+This is because Ironic currently requires *host networking*, and thus it's not
+advisable to let it co-exist with tenant workload.
+
+### Installed services
```
lentzi90, line 50: Coming from a kubernetes background, this section can be a bit confusing. The terminology is easy to misunderstand since some of the words have specific meanings in kubernetes context (e.g. service and deployment). That said, I'm not quite sure how I would change it :grimacing: 

Is it so that all of these below are containers in the ironic pod? I would then maybe prefer to talk about them as "ironic sidecars". What do you think?

docs/user-guide/src/irso/install-basics.md
```diff
@@ -0,0 +1,157 @@
...
+  without TLS. To enable it, you need to manage the TLS secret. [Cert
+  Manager](https://cert-manager.io/) is the recommended service for it.
+
+## Using Ironic
+
+Regardless of the scenario you choose, you will need to create at least an
+`Ironic` object and wait for it to become ready:
+
+```bash
+NAMESPACE="scenario-N"  # change to match your deployment
```
lentzi90, line 28: I don't quite get the point of having a different namespace for each scenario. Is it expected that people deploy all 3 together? If not, I would perhaps just leave it out.

lentzi90 (commented):
Very nice!

docs/user-guide/src/irso/introduction.md
MahnoorAsghar: Perhaps 'An Ironic instance (?) always contains these three components:'

tuminoid (commented):
Besides what is already mentioned by others, I don't see any major issues here. It is a solid first stab at documentation, thank you!

docs/user-guide/src/irso/install-basics.md
dtantsur: The role of keepalived is to maintain the same IP address, regardless of which node is hosting Ironic. I'll try to clarify a bit.

docs/user-guide/src/irso/introduction.md
dtantsur: Yeah, I totally agree. Do we already do it for CAPM3?

dtantsur:
Hopefully addresses all the comments. Also added information about versions.

docs/user-guide/src/irso/install-basics.md
MahnoorAsghar: So the bottom line is: Make sure you have a usable IP address on the control plane node hosting ironic. It's possible to use keepalived for this. 
Its clearer now, thank you!

MahnoorAsghar:
LGTM!

docs/user-guide/src/irso/introduction.md
lentzi90: Yes, it is part of how CAPI expects things to work. So it is published as `infrastructure-components.yaml` on every release. That allows `clusterctl init --infrastructure=metal3` to automatically find the latest and install it :slightly_smiling_face:

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

Issue #443, opened 2024-06-13
Fix dead links at foundation-proposal.md


tuminoid:
```
| FILE: ./design/community/foundation-proposal.md
| [✖] https://github.com/cncf/toc/blob/main/process/project_proposals.md
| 
| 8 links checked.
| 
| ERROR: 1 dead links found!
| [✖] https://github.com/cncf/toc/blob/main/process/project_proposals.md → Status: 404
```

/triage accepted
/help
/good-first-issue

Split from https://github.com/metal3-io/metal3-docs/issues/328

aditya7302:
@tuminoid
 ```https://github.com/cncf/sandbox/issues/new?assignees=&labels=New&projects=&template=application.yml&title=%5BSandbox%5D+%3CProject+Name%3E?assignees=&labels=New&projects=&template=application.yml&title=%5BSandbox%5D+%3CProject+Name%3E```
 is this a valid link for the dead  link for this issue ?

tuminoid:
> @tuminoid `https://github.com/cncf/sandbox/issues/new?assignees=&labels=New&projects=&template=application.yml&title=%5BSandbox%5D+%3CProject+Name%3E?assignees=&labels=New&projects=&template=application.yml&title=%5BSandbox%5D+%3CProject+Name%3E` is this a valid link for the dead link for this issue ?

No, https://github.com/cncf/toc/tree/main/process seems to be the right one, as it is v1.6 of the proposal process and the old link pointed to v1.5 of th same.

geekrypter:
Hi @tuminoid, can I work on this issue?

tuminoid:
> Hi @tuminoid, can I work on this issue?

Sure, there has been no further activity in weeks.
/assign @geekrypter

geekrypter:
Hi @tuminoid, sorry, I couldn't find time to work on this issue before. There seems to be no further activity on pull request 477. I am happy to work on this now, so please let me know if I can raise a separate pull request to fix this issue.

tuminoid:
Please go ahead. @geekrypter

PR #494, opened 2024-12-24
Fix dead link in foundation-proposal.md

geekrypter:
What is the change?
This change fixes a dead link in foundation-proposal.md.

Why the change?
Fixes: https://github.com/metal3-io/metal3-docs/issues/443


kashifest (commented):
/lgtm

PR #497, opened 2025-01-17
Modifying notes on checksum to reflect recent updates regarding empty checksumType field

fracappa:
Hi everyone.

As suggested by community members, I would like to propose those changes regarding checksumType notes to reflect modifications introduced in this [PR](https://github.com/metal3-io/baremetal-operator/pull/2174).

The idea, is allowing the empty `checksumType` to behave the same way as the `auto` case.

This could be taken as a proposal for a new version of BMO (it could be 0.10.0).

Glad to receive any possible feedback.

Thanks.

### DIFF
--- a/docs/user-guide/src/bmo/provisioning.md
+++ b/docs/user-guide/src/bmo/provisioning.md
@@ -108,12 +108,9 @@ provisioning](./live-iso.md) that works differently from a normal one.
 
 ### Notes on checksums
 
-Unlike Ironic itself, Metal3 currently assumes the checksum algorithm to be MD5
-when no `checksumType` value is provided. Since more secure algorithms, such as
-SHA256 or SHA512, are popular nowadays, care must be taken to provide the
-correct `checksumType`. The value of `auto` will make Ironic detect the
-checksum type from its length and will become the default in the next version
-of the BareMetalHost API.
+Starting from BMO v0.10.0, leaving `checksumType` empty prompts Ironic to
+automatically detect the checksum type based on its length. In earlier versions,
+this behavior can be achieved by setting `checksumType` to `auto`.
 
 The `checksum` value can be provided either as a URL or as the hash value
 directly. Providing a URL is more convenient in case of public cloud images,


docs/user-guide/src/bmo/provisioning.md
```diff
@@ -108,12 +108,7 @@ provisioning](./live-iso.md) that works differently from a normal one.
 
 ### Notes on checksums
 
-Unlike Ironic itself, Metal3 currently assumes the checksum algorithm to be MD5
-when no `checksumType` value is provided. Since more secure algorithms, such as
-SHA256 or SHA512, are popular nowadays, care must be taken to provide the
-correct `checksumType`. The value of `auto` will make Ironic detect the
-checksum type from its length and will become the default in the next version
-of the BareMetalHost API.
+Setting `checksumType` to `auto` or leaving it empty prompts Ironic to automatically detect the checksum type based on its length.
```
tuminoid, line 111: Please limit line length to 80.

I would also not delete the note on the baremetalhost API change. API change != this change.

tuminoid (commented):
/ok-to-test

docs/user-guide/src/bmo/provisioning.md
```diff
@@ -108,12 +108,9 @@ provisioning](./live-iso.md) that works differently from a normal one.
...
 
-Unlike Ironic itself, Metal3 currently assumes the checksum algorithm to be MD5
-when no `checksumType` value is provided. Since more secure algorithms, such as
-SHA256 or SHA512, are popular nowadays, care must be taken to provide the
-correct `checksumType`. The value of `auto` will make Ironic detect the
-checksum type from its length and will become the default in the next version
-of the BareMetalHost API.
+Setting `checksumType` to `auto` or leaving it empty prompts Ironic to
+automatically detect the checksum type based on its length and will become
+the default in the next version of the BareMetalHost API.
```
dtantsur, line 113: The last part is no longer relevant: the default in the next version will be the same as the default you introduced.

Instead, it may be useful to mention the version of BMO where the empty string works the same as auto.

fracappa: Do we know if it will be BMO v 0.10.0?

tuminoid: Next version of BMO per milestones is v0.10.0. I don't think we're doing anything this release cycle to warrant major bump

tuminoid (commented):
/lgtm

PR #502, opened 2025-01-22
Update command to create userdata

iurygregory:
The current example takes into consideration the file is a JSON, someone reported to us downstream that they couldn't run the command. Doing some tests the person found that we should pass a yaml file.

This PR updates the example in our docs.


tuminoid (commented):
Example data given before the command is also in YAML, makes sense :) 

/lgtm

lentzi90 (commented):
/approve

PR #503, opened 2025-02-06
Update MariaDB release docs

peppi-lotta:
Specify how the tagging is done.

### DIFF
--- a/processes/releasing.md
+++ b/processes/releasing.md
@@ -28,16 +28,38 @@ Release documentation in release branches may be outdated.
 After releasing `v1.x.y` version of CAPM3, create an annotated tag with `capm3-`
 prefix + release version in
 [mariadb-image](https://github.com/metal3-io/mariadb-image) Github repository.
+Origin points to `metal3-io`.
+
+For minor releases:
+
+```bash
+git tag -s -a capm3-v1.x.y -m capm3-v1.x.y
+git push origin capm3-v1.x.y
+```
+
+For patch releases:
 
 ```bash
 git tag -s -a capm3-v1.x.y -m capm3-v1.x.y [optional sha, or tag^{}]
 git push origin capm3-v1.x.y
 ```
 
+The last part should point to a previous patch. For example for tag `v.1.x.2`:
+
+```bash
+git tag -s -a capm3-v1.x.2 -m capm3-v1.x.2 capm3-v1.x.1^{}
+```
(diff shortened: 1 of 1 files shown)


smoshiur1237 (commented):
/approve

processes/releasing.md
```diff
@@ -28,17 +28,39 @@ Release documentation in release branches may be outdated.
...
 ```bash
 git tag -s -a capm3-v1.x.y -m capm3-v1.x.y [optional sha, or tag^{}]
 git push origin capm3-v1.x.y
 ```
 
+The last part should point to a previous patch. For example for tag `v.1.x.2`:
+
+```bash
+git tag -s -a capm3-v1.x.2 -m capm3-v1.x.2 capm3-v1.x.1^{}
+```
```
lentzi90, line 51: I don't understand this. Why should it point to a previous patch? What does it do?

lentzi90: Wouldn't this actually tag the same commit as the previous tag? The last positional argument should be the commit that should be tagged. This should definitely NOT be the same as the previous tag in general.
/hold

tuminoid: We don't test mariadb main versus release branches, and we don't have release branches in mariadb. We thus cannot be sure if the latest would work (say, we introduce a security feature or env variable).  This is why we've always retagged the same tag the branch was originally tested with. When doing new minor, we tag the head of main, as main is always tested with main, so they're guaranteed to work at the time. 

See L44, and check previous tags to see this pattern. This PR is just making the command more obvious I guess.

tuminoid: Also reference: We have backlog issue of making mariadb release process better (maybe more standalone, similar to what were discussion for keepalived), or maybe even retiring our mariadb image in favor of just configuring an upstream mariadb (as you've suggested).

peppi-lotta: Yes, the release should be made better at some point. This is just a minor update mean while. I just wanted to make the docs more unambiguous and add things that I've had to ask about.  

Can we unhold this? I could add the reasoning given in these comment to the docs if that seems necessary.

lentzi90: It would be good to have an explanation for why this is done in the doc, but since it is at least correct I'm fine unhold.
/hold cancel

lentzi90 (commented):
/approve

PR #507, opened 2025-02-26
Update IrSO matrix for ironic-image 28.0

elfosardo:


docs/user-guide/src/irso/introduction.md
```diff
@@ -79,7 +79,7 @@ values for each branch and release of the operator:
 
 | Operator version | Ironic version(s) | Default version |
 | ---------------- | ----------------- | --------------- |
-| latest (main)    | 27.0              | latest          |
+| latest (main)    | 28.0, 27.0        | latest          |
```
tuminoid, line 82: We do have two IRSO releases, should we be documenting those?

elfosardo: I've updated it to include the released versions

docs/user-guide/src/irso/introduction.md
```diff
@@ -77,9 +77,11 @@ A major and minor version can be supplied to the `Ironic` resource to request
 a specific branch of ironic-image (and thus Ironic). Here are supported version
 values for each branch and release of the operator:
 
-| Operator version | Ironic version(s) | Default version |
-| ---------------- | ----------------- | --------------- |
-| latest (main)    | 27.0              | latest          |
+| Operator version | Ironic version(s)  | Default version |
+| ---------------- | -----------------  | --------------- |
+| latest (main)    | latest, 28.0, 27.0 | latest          |
+| 0.2.0            | 28.0, 27.0         | latest          |
```
tuminoid, line 83: It is a bit weird that default is latest, but it is not listed as supported :)

dtantsur: Mmm, yeah. Since latest is always supported, we should either add it everywhere or take it out from all versions.

elfosardo: probably adding it everywhere as it's the default in all versions so far

tuminoid (commented):
/lgtm

PR #508, opened 2025-03-11
Remove mention of IronicDatabase from the IrSO docs

dtantsur:
We're going to deprecate it and mariadb-image in favour of 3rd party
operators, such as mariadb-operator. See the bug for more details:
https://github.com/metal3-io/ironic-standalone-operator/issues/142


tuminoid (commented):
/lgtm

PR #509, opened 2025-03-13
Refresh the BMH v1beta1 design based on recent developments

dtantsur:
- BMO has changed the default ChecksumType to auto without waiting for
  v1beta1, so now remove the explicit MD5 type instead.

- Remove support for VDI and VMDK disk formats. They were disabled in
  the Ironic image validation fixes, and nobody has complained so far.
  These formats have never been tested by the Ironic or Metal3
  communities.

- Introduce an explicit InspectionMode similar to AutomatedCleaningMode.
  It will replace setting inspect.metal3.io to "disabled" with a proper
  API and pave the way to eventually add a Redfish-based option.

- Replace UserData, MetaData and NetworkData with UserDataName,
  MetaDataName and NetworkDataName. It has been discovered that allowing
  secrets from a different namespace poses a security issue, so the
  Namespace field is not functional any more.

### DIFF
--- a/design/baremetal-operator/bmh-v1beta1.md
+++ b/design/baremetal-operator/bmh-v1beta1.md
@@ -66,16 +66,32 @@ The following addition should be done **before** the version is created:
 - Ensure that a `HardwareData` CR exists for any host that has
   `HardwareDetails` in the status.
 
+- Add a `Spec.InspectionMode` to replace the `disabled` value of the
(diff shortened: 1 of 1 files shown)


elfosardo:
/lgtm
always forget I'm not an approver here '^__^

tuminoid:
LGTM.

/cc @lentzi90 @Rozzii @kashifest

lentzi90 (commented):
/approve

PR #521, opened 2025-05-16
Give upstream as a argument to workflow call

peppi-lotta:
Give upstream as a argument to workflow call


tuminoid (commented):
/lgtm
/hold
https://github.com/metal3-io/project-infra/pull/1015 needs to be merged first.
/cc @lentzi90

PR #526, opened 2025-05-26
User guide: retitle BMO features to explain what they do

dtantsur:
Currently, many titles specify how the feature is called instead of what
it does. Operators won't be looking for a certain annotation, they'll
try to find a feature.

I'm keeping URLs the same to avoid breaking any bookmarks.


tuminoid (commented):
/lgtm

Good improvement. Should we rename the files to match? The titles make a lot of sense as filenames too IMO, would improve SEO etc.

tuminoid:
> > Should we rename the files to match? The titles make a lot of sense as filenames too IMO, would improve SEO etc.
> 
> I'm worried about breaking existing bookmarks. I myself already have a few pages at hand.

Yup, there is that downside of course. We can see from link checker runs that CAPI etc break their pages once in a while, so maybe we can do some rearrangements at some point.

lentzi90 (commented):
/approve

PR #547, opened 2025-06-06
add workflow to auto-reapprove workflows

tuminoid:
We now have a link checker workflow in this repo, and it needs constant reapproving, when new contributors make PRs and revise commits. Add the same auto-reapprover workflow as in other repos.


lentzi90 (commented):
/approve

PR #551, opened 2025-06-19
Update support matrix and examples for IrSO

elfosardo:


lentzi90 (commented):
/approve

docs/user-guide/src/irso/introduction.md
```diff
@@ -74,12 +74,13 @@ A major and minor version can be supplied to the `Ironic` resource to request
...
-| Operator version | Ironic version(s)        | Default version |
-| ---------------- | ------------------------ | --------------- |
-| latest (main)    | latest, 29.0, 28.0, 27.0 | latest          |
-| 0.3.0            | latest, 29.0, 28.0, 27.0 | latest          |
-| 0.2.0            | latest, 28.0, 27.0       | latest          |
-| 0.1.0            | latest, 27.0             | latest          |
+| Operator version | Ironic version(s)              | Default version |
+| ---------------- | ------------------------------ | --------------- |
+| latest (main)    | latest, 30.0, 29.0, 28.0, 27.0 | latest          |
+| 0.4.0            | 30.0, 29.0, 28.0, 27.0         | 30.0            |
```
dtantsur, line 80: 0.4.0 still supports latest, it's just not used by default any more

(maybe remove latest from all numbered releases for simplicity?)

tuminoid: I second the latter

elfosardo: if we remove latest from old numbered releases, then we'll have to change the default (in code I mean, not just here)
I'm actually wondering if we should just not support any numbered releases below 0.4.0 at this point, and remove them from the scheme
they're all superseded by 0.4.0 anyway

Rozzii: We could have it like this "up to latest" in the Ironic version column, and if at some point we would land a breaking change in the ironic-image then we can just change it to e.g. "up to v40.0" instead of listing all the versions.

dtantsur: Yeah, maybe we remove `latest` from everywhere except for `main`. And then add a note that all versions support `latest` but it's not recommended to use it with any versions or IrSO except for `latest`.

docs/user-guide/src/irso/introduction.md
```diff
@@ -74,12 +74,13 @@ A major and minor version can be supplied to the `Ironic` resource to request
...
+| 0.3.0            | latest, 29.0, 28.0, 27.0       | latest          |
+| 0.2.0            | latest, 28.0, 27.0             | latest          |
+| 0.1.0            | latest, 27.0                   | latest          |
 
 **NOTE:** the special version value `latest` always installs the latest
 available version of ironic-image and Ironic.
```
dtantsur: Like, update this note with "This version value is supported by all releases of IrSO but only works reliably in the `latest` release."

PR #552, opened 2025-06-24
Propose  BMH specific preprovisioningExtraKernelParams spec field

Rozzii:
This commit:

- Introduces the proposal for BMH support for node specific extra kernel command line arguments for pre-provisioning images.
- Proposal also contains provisions on how preprovisioningImage CRs are handled
  together with the new BMH spec.

The motivation behind the commit can be found in the proposal.

### DIFF
--- a/design/baremetal-operator/node-specific-kernel-args.md
+++ b/design/baremetal-operator/node-specific-kernel-args.md
@@ -1,0 +1,240 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Preprovisioning kernel argument support
+
+## Status
+
+implemented
+
+## Summary
+
+This document describes the BareMetalHost spec attribute for
+configuring preprovisioning image kernel arguments with support for
+PreprovisioningImage integration and format-specific parameter handling.
+
+## Motivation
+
+There are situations where a user would need node specific kernel arguments
+for the preprovisioning image. The reason could be the use of a mixed set of
+hardware, troubleshooting provisioning or boot issues, selectively configuring
+custom logic e.g. enabling IPA plugins or other services embedded in the
+preprovisioning image.
+
+### Goals
+
+1. Implement a string type optional spec field named `preprovisioningExtraKernelParams`
+1. Extend the generic `Provisioner` interface to support the new optional field
+1. Implement support for the `preprovisioningExtraKernelParams` in the Ironic
+   provisioner
+1. Combine the values of the `preprovisioningExtraKernelParams` and the
+   `preprovisioningImage`'s `status.extraKernelParams` field when a
+   `preprovisioningImage` is referenced by the BMH and the `status.extraKernelParams`
+   is not empty
+1. Handle format-specific parameter behavior for different PreprovisioningImage
+   formats
+1. Support kernel parameter updates across all lifecycle management operations
+
+### Non-Goals
+
+1. Implement validation for the new optional field
+1. Provide a default value for the new spec field
+1. Implement conditional logic to overwrite default kernel parameters on
+   per node basis without using the %default% mechanism
+
+## Proposal
+
+This document describes the BareMetalHost spec attribute for
+configuring preprovisioning image kernel arguments with intelligent
+parameter combination logic based on PreprovisioningImage formats.
+
+The proposal involves a new spec field modification to the BMO API
+that is fully backward compatible as the new `preprovisioningExtraKernelParams`
+field is optional.
+
+### Implementation Details/Notes/Constraints
+
+The `preprovisioningExtraKernelParams` field is located directly under the
+`spec` section as a string field. The feature provides intelligent parameter
+combination based on PreprovisioningImage format and ownership:
+
+#### Basic Usage
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: bm0
+spec:
+  online: true
+  bmc:
+    address: "idrac://192.168.122.1:6230/"
+    credentialsName: bm0-bmc-secret
+  preprovisioningExtraKernelParams: "console=ttyS0,115200 intel_iommu=on"
+  # ...
+  bootMACAddress: "52:54:00:b7:b2:6f"
+```
+
+#### Parameter Combination Logic
+
+The feature supports four distinct scenarios with 3 outcomes based on
+PreprovisioningImage format and ownership:
+
+1. **No PreprovisioningImage or Provisioner without PPI support**
+   - Uses only BMH `preprovisioningExtraKernelParams`
+   - Formatted with `%default%` prefix by Ironic provisioner
+
+1. **ISO format PreprovisioningImage**
+   - **BMH parameters are ignored** (kernel params already embedded in ISO)
+   - Warning logged when BMH params are specified but ignored
+   - Only PPI kernel parameters are used
+
+1. **InitRD format PreprovisioningImage**
+   - **BMH and PPI parameters are combined**
+   - Final result: `BMH_params + " " + PPI_params`
+   - Both external and BMO-managed PPIs supported
+
+1. **BMO-managed PreprovisioningImage (default provider)**
+   - Uses only BMH `preprovisioningExtraKernelParams`
+   - PPI parameters from default provider are ignored
+   - Maintains consistency with non-PPI deployments
+
+#### Ownership Detection
+
+The feature distinguishes between:
+
+- **External PreprovisioningImages**: Created/managed by external controllers
+- **BMO-managed PreprovisioningImages**: Created by BMO's default image provider
+
+This distinction enables different parameter handling strategies and prepares
+for future multi-tenancy support.
+
+#### Ironic Integration
+
+The Ironic provisioner:
+
+- Always prepends `%default%` to kernel parameters for consistency
+- Updates BMC driver_info with `kernel_append_params` during registration
+- Supports conditional driver_info updates when parameters change
+- Maintains compatibility across all BMC protocols (IPMI, Redfish, iDRAC)
+
+#### Lifecycle Operation Support
+
+Kernel parameters are updated during all relevant lifecycle operations:
+
+- **Registration**: Initial parameter setup
+- **Preparing**: Parameters refreshed before cleaning
+- **Provisioning**: Parameters applied for deployment
+- **Servicing**: Parameters updated for maintenance operations
+
+This enables dynamic kernel parameter changes for debugging and operational
+needs.
+
+### Format-Specific Behavior Matrix
+
+| PreprovisioningImage Format | BMH Params | PPI Params | Final Result | Notes |
+|----------------------------|-------------|------------|--------------|-------|
+| No PPI / No PPI Support    | Used        | N/A        | `%default% BMH_params` | Standard behavior |
+| ISO (External)             | Ignored     | Used       | `PPI_params` | Warning logged if BMH params specified |
+| ISO (BMO-managed)          | Used        | Ignored    | `%default% BMH_params` | Consistent with no-PPI behavior |
+| InitRD (External)          | Used        | Used       | `%default% BMH_params PPI_params` | Parameters combined |
+| InitRD (BMO-managed)       | Used        | Ignored    | `%default% BMH_params` | Consistent with no-PPI behavior |
+| Unknown Format             | Used        | Ignored    | `%default% BMH_params` | Fallback behavior |
+
+### Risks and Mitigations
+
+Providing faulty kernel arguments might result in boot issues or unexpected
+behavior during the lifecycle of the preprovisioning image. The feature
+provides several mitigations:
+
+- Graceful fallback to BMH parameters when PPI processing fails
+- Warning logs for format-specific parameter behavior
+- Maintains %default% prefix for Ironic consistency
+- No validation prevents breaking existing workflows
+
+## Design Details
+
+The feature extends the BMO architecture with:
+
+1. **Enhanced BareMetalHost Controller**
+   - `retrievePreprovisioningExtraKernelParamsSpec()` for intelligent parameter
+     combination
+   - Ownership detection logic for external vs BMO-managed PreprovisioningImages
+   - Format-specific parameter handling with switch-based logic
+
+1. **PreprovisioningImage Controller Integration**
+   - `findOwnerBMH()` for controller owner reference resolution
+   - BMH kernel parameter injection for initrd format images
+   - Groundwork for future multi-tenancy support
+
+1. **Provisioner Interface Extensions**
+   - Extended `ManagementAccessData` with `PreprovisioningExtraKernelParams`
+     field
+   - Backward compatibility with existing provisioner implementations
+   - Graceful degradation for provisioners without PPI support
+
+1. **Ironic Provisioner Enhancements**
+   - Updated BMC driver implementations for kernel parameter support
+   - Conditional driver_info updates based on parameter changes
+   - `fmtPreprovExtraKernParams()` helper for consistent parameter formatting
+
+### Work Items
+
+- String type optional spec field named `preprovisioningExtraKernelParams`
+- Extended generic `Provisioner` interface to support the new optional field
+- Support for `preprovisioningExtraKernelParams` in the Ironic driver
+- Extended data types used in different operational states to hold
+  `preprovisioningExtraKernelParams`
+- Comprehensive unit tests covering all parameter combination scenarios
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+### Goals
+
+1. Implement a string array type optional spec field named
+   `preProvisionKernelArgs`
+2. Extend the generic `Provisioner` interface to support the new optional field
+3. Implement support for the `preProvisionKernelArgs` in the Ironic provisioner
+
+### Non-Goals
+
+1. To implement any sort value check for the new optional spec
```
lentzi90: I guess *some* kind of validation may be done but I take it this means that we are not going to check the content, only that it is a string array?
```suggestion
1. To implement any sort validation for the new optional field
```

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+2. Providing a default value for the new spec field
+
+## Proposal
+
+This document proposes a new BareMetalHost spec attribute intended for
+configuring the pre-provisioning image's (usually IPA) kernel arguments.
+
+The proposal involves the introduction of a new spec field thus a modification
+to the BMO API. The API change is intended to be backward compatible as the
+new `preProvisionKernelArgs` field intended to be optional.
```
lentzi90:
```suggestion
new `preProvisionKernelArgs` field is optional.
```

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+The `preProvisionKernelArgs` list would be received by the `Provisioner`
+interface as is. The handling of the list would be up to the particular
+implementations of the `Provisioner` interface. In case of the `Ironic`
+provisioner the elements from the `preProvisionKernelArgs` would be
+concatenated into a single string such that the string would start with the
+first element (lowest index) from the list and would end with the last
+(highest index) element. The elements between the first and last element would
+be concatenated in an ascending order based on their index number in the array.
+
+Taken the BMH snippet from above the kernel argument list generated by
```
lentzi90:
```suggestion
Given the BMH snippet from above, the kernel argument list generated by
```

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+
+The user has to be aware that the kernel arguments supplied here will fully or
+partially overwrite the default arguments depending on how the particular
+`Provisioner` is handling the argument list. In case of Ironic provisioner
+the `preProvisionKernelArgs` will completely replace the default argument
+list used by Ironic.
+
+## Design Details
+
+Most of the design is very straightforward an easy to combine with the
```
lentzi90:
```suggestion
Most of the design is very straightforward and easy to combine with the
```

Rozzii:
Something that might be interesting to
/cc @dtantsur @elfosardo

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+### Work Items
+
+- Implement a string array type optional spec field named
+  `preProvisionKernelArgs`
+- Extend the generic `Provisioner` interface to support the new optional field
+- Implement support for the `preProvisionKernelArgs` in the Ironic driver
+- Implement unit tests
+- Add a check to the existing BMO e2e workflow that logs in to an active IPA
+  via ssh or serial console and checks whether the custom kernel args were
+  propagated properly or not.
```
lentzi90: :+1: 
I have recently learned also that it is possible to use the kernel command line to [configure the network](https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt). Maybe we could use that to verify also, by configuring a static IP outside of DHCP range? If IPA becomes reachable at that IP we can consider the test passed.

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Dependencies
+
+- Ironic in case of the `Ironic Provisioner`
+
+### Test Plan
+
+- Unit tests for the functions
+
```
lentzi90:
```suggestion
```

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+...
+
+```
+
+The example contains a snippet of a common IPA kernel argument list.
+The `preProvisionKernelArgs` list would be received by the `Provisioner`
+interface as is. The handling of the list would be up to the particular
+implementations of the `Provisioner` interface. In case of the `Ironic`
+provisioner the elements from the `preProvisionKernelArgs` would be
+concatenated into a single string such that the string would start with the
```
lentzi90: They are concatenated with spaces in between, right? I think that may be worth to point out explicitly here also.

lentzi90 (commented):
Nice! I have just a few nits/typos and one suggestion to clarify how the strings are concatenated (with spaces?)

lentzi90 (commented):
/lgtm

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+There are situations where a user would need node specific kernel arguments
+for the pre-provisioning image. The reason could be the use of a mixed set of
+hardware, troubleshooting provisioning or boot issues, selectively configuring
+custom logic e.g. enabling IPA plugins or other services embedded in the
+pre-provisioning image.
+
+### Goals
+
+1. Implement a string array type optional spec field named
+   `preProvisionKernelArgs`
```
dtantsur: Please keep spelling and word choice consistent: `preprovisioningKernelParams`

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+2. Extend the generic `Provisioner` interface to support the new optional field
+3. Implement support for the `preProvisionKernelArgs` in the Ironic provisioner
+
```
dtantsur, line 41: You also need to extend the PreprovisioningImage object to accept these params.

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+ name: bm0
+spec:
+ online: true
+ bmc:
+   address: idrac://192.168.122.1:6230/
+   credentialsName: bm0-bmc-secret
+ preProvisionKernelArgs:
```
dtantsur: I'd slightly prefer a string. The problem is: what if any item contains a space? Are you going to quote it? How?

Since Ironic accepts a string, and PPI returns a string, maybe use a string here.

design/baremetal-operator/node-specific-kernel-args.md
dtantsur: Actually, it's called `ExtraKernelParams` in PPI. There is a reason for that: these options are appended to the defaults in ironic-image. I'd prefer to keep it this way for the new feature, otherwise it won't be usable for anyone with ironic-image-level customizations (like we have to provide something for CoreOS/Ignition).

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,166 @@
...
+### Risks and Mitigations
+
+Providing a faulty list kernel arguments might result in boot issues
+or unexpected behavior during the life cycle of the pre-provisioning image.
+
+The user has to be aware that the kernel arguments supplied here will fully or
+partially overwrite the default arguments depending on how the particular
+`Provisioner` is handling the argument list. In case of Ironic provisioner
+the `preProvisionKernelArgs` will completely replace the default argument
+list used by Ironic.
```
dtantsur: See above, I'd prefer to avoid a full override. It will require a user to guess Ironic options that must stay present.

design/baremetal-operator/node-specific-kernel-args.md
dtantsur: Ref: https://github.com/metal3-io/baremetal-operator/pull/1109

design/baremetal-operator/node-specific-kernel-args.md
Rozzii: We can keep them separate, I didn't intend to make this new spec to be applied for the "PreprovisioningImage" I thought that already has a spec for this purpose.

design/baremetal-operator/node-specific-kernel-args.md
Rozzii: I have changed it to a string.

design/baremetal-operator/node-specific-kernel-args.md
Rozzii: I have dropped the full override approach for this proposal.

Rozzii:
WIP implementation: https://github.com/metal3-io/baremetal-operator/pull/2576 at the moment without tests, but functionally this already worked .

EDIT: status and the preprovImage CR support is missing ATM

design/baremetal-operator/node-specific-kernel-args.md
Rozzii: I have changed the name it in the proposal (might have missed a few places), it makes sense as I have dropped the functionality to overwrite the defaults in this proposal.

When it comes to the future I think there is still value in allowing the user to overwrite the defaults especially in debug scenarios, I think that proposal could be a useful followup proposal.

design/baremetal-operator/node-specific-kernel-args.md
dtantsur: That will create a wild inconsistency that will cause users file bugs for me with "`preprovisioningExtraKernelParams` does not work" :( Because they'll expect OpenShift's version of Metal3 to work the same way as upstream, and without PPI support it won't.

dtantsur: Btw it's fair if you ask me to follow-up with PPI support myself if you don't have cycles for that. I just want it to be in the design.

Rozzii: I have already looked into it, my original impression was that the user is already able to set the Kernel Args via the preprovisioningImage CR's spec but now I understand that it is not possible as on that CR the kernel args are only part of the status and not the spec. I have also checked https://github.com/openshift/image-customization-controller so I am more familiar with the use-case of the preprovisioningImage CR.

Based on what I know now It makes sense to include preprovisioningImage in the proposal.

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,201 @@
...
+1. Implement a string type optional spec field named `preprovisioningExtraKernelParams`
+2. Extend the generic `Provisioner` interface to support the new optional field
+3. Implement support for the `preprovisioningExtraKernelParams` in the Ironic provisioner
+4. Combine the values of the `preprovisioningExtraKernelParams` and the
+   `preprovisioningImage`'s `status.extraKernelParams` field if a
+   `preprovisioningImage` is refferences by the BMH and teh `status.extraKernelParams`
+   is not empty
+5. Introducing a `preprovisioningExtraKernelParams` related status field that
+   would display the last applied `preprovisioningExtraKernelParams` value.
+   Let's call this field `status.lastAppliedPreprovisioningExtraKernelParams`
```
fmuyassarov: Perhaps something shorter would be easier but I don't have strong opinion on this
Example, `status.lastAppliedKernelParams`

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Goals
+
+1. Implement a string type optional spec field named `preprovisioningExtraKernelParams`
+2. Extend the generic `Provisioner` interface to support the new optional field
+3. Implement support for the `preprovisioningExtraKernelParams` in the Ironic provisioner
+4. Combine the values of the `preprovisioningExtraKernelParams` and the
+   `preprovisioningImage`'s `status.extraKernelParams` field if a
+   `preprovisioningImage` is refferences by the BMH and teh `status.extraKernelParams`
+   is not empty
```
fmuyassarov: Is it a good idea to construct a value of a field from two different sources? I imagine things might go wrong, for example how do we handle a case when prepovisioningImage object referenced by BMH doesn't exist for some reason?

Does this mean that a node might have different kernel parameters compared to what is asked via `.spec.preprovisioningExtraKernelParams`, in case  preprovisioningImage `status.extraKernelParams` has something non-empty? Should not we have a single source of truth for defining the parameters and I guess the status sub-resource was never meant for defining the parameters but more for summarizing the current state right? Perhaps whoever writes into `status.extraKernelParams` should only rely on BMH spec and not something else? Or is this a transitional state that we eventually target to move to a single source?

fmuyassarov: @dtantsur thoughts ?

design/baremetal-operator/node-specific-kernel-args.md
Rozzii: I have been thinking about this, and our offline discussions and now I am on the opinion that I should separate this status topic to a different proposal/discussion, because right now we don't really provide status for applying provisioning or preprvisioning data to Ironic API so we might need a more comprehensive approach to give feedback about what, user/network/meta data and what preprovisioning data was applied. We might also need a status flag to show if the previously mentioned data references/values have changed since the last LCM operation (inspect,provision,deprovision,clean, service).

If I would leave the status topic out from the proposal I think it would be still aligned with our current feedback mechanism (or the lack of it) and then we can improve our feedback mechanism with the followup proposal.

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,201 @@
...
+pre-provisioning image.
+
+### Goals
+
+1. Implement a string type optional spec field named `preprovisioningExtraKernelParams`
+2. Extend the generic `Provisioner` interface to support the new optional field
+3. Implement support for the `preprovisioningExtraKernelParams` in the Ironic provisioner
+4. Combine the values of the `preprovisioningExtraKernelParams` and the
+   `preprovisioningImage`'s `status.extraKernelParams` field if a
+   `preprovisioningImage` is refferences by the BMH and teh `status.extraKernelParams`
```
lentzi90:
```suggestion
   `preprovisioningImage` is referenced by the BMH and the `status.extraKernelParams`
```

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,201 @@
...
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+ name: bm0
+spec:
+ online: true
+ bmc:
+   address: idrac://192.168.122.1:6230/
+   credentialsName: bm0-bmc-secret
+ preprovisioningExtraKernelParams: "--timeout 60000 systemd.journald.forward_to_console=yes ipa-debug=1"
```
lentzi90:
```suggestion
  preprovisioningExtraKernelParams: "--timeout 60000 systemd.journald.forward_to_console=yes ipa-debug=1"
```

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,201 @@
...
+configuring the pre-provisioning image's (usually IPA) kernel arguments.
+
+The proposal involves the introduction of a new spec field thus a modification
+to the BMO API. The API change is intended to be backward compatible as the
+new `preprovisioningExtraKernelParams` field is optional.
+
+### Implementation Details/Notes/Constraints
+
+The `preprovisioningExtraKernelParams` field would be located directly under `spec`
+section. The format of the new field would be a list/array of individual
```
lentzi90: This has been changes elsewhere to be a string

```suggestion
section. The format of the new field would be a string.
The new field would be optional and it would be considered
```

lentzi90 (commented):
/lgtm

design/baremetal-operator/node-specific-kernel-args.md
```diff
@@ -0,0 +1,240 @@
...
+provides several mitigations:
+
+- Graceful fallback to BMH parameters when PPI processing fails
+- Warning logs for format-specific parameter behavior
+- Maintains %default% prefix for Ironic consistency
+- No validation prevents breaking existing workflows
+
+## Design Details
+
+The feature extends the BMO architecture with:
```
dtantsur, line 161: nit: this is too detailed for a design proposal IMO

dtantsur:
/approve

We have a discussion on the exact implementation, but the design is solid.

Issue #561, opened 2025-08-08
Server cannot fetch the ironic python agent


t-h2o:
# Server cannot fetch the ironic python agent

## Context

From the [quickstart](https://book.metal3.io/quick-start).

While the boot process.
The bare metal server try to fetch from the IP of the kind container.

```
http://172.19.0.2:6180/uuid/deploy_kernel....
connection timed out (https://ipxe.org/4c0a6092)
```

## Fix

Adding this environment variable in `ironic/ironic_bmo.env` \
will edit this [line](https://github.com/metal3-io/ironic-image/blob/37b859947e7e506e28b76c93ae7a95639f9bc8dc/ironic-config/ironic.conf.j2#L28) of the ironic configuration.

```
IRONIC_IP=192.168.0.150
```

t-h2o:
Should we close this issue, since it has been resolved by this  [pull request](#562)?

Rozzii:
> Should we close this issue, since it has been resolved by this [pull request](#562)?

At least for me the link is not working, could you @t-h2o check please?

t-h2o:
> At least for me the link is not working, could you [@t-h2o](https://github.com/t-h2o) check please?

✔️  https://github.com/metal3-io/metal3-docs/pull/562

Rozzii:
> > At least for me the link is not working, could you [@t-h2o](https://github.com/t-h2o) check please?
> 
> ✔️ [#562](https://github.com/metal3-io/metal3-docs/pull/562)

/close

PR #562, opened 2025-08-08
fix(ironic/config): add ironic ip variable

t-h2o:
# Add ironic ip variable

See https://github.com/metal3-io/metal3-docs/issues/561

### DIFF
--- a/docs/user-guide/src/quick-start.md
+++ b/docs/user-guide/src/quick-start.md
@@ -353,6 +353,7 @@ HTTP_PORT=6180
 PROVISIONING_INTERFACE=eth0
 # URL where the http server is exposed (IP of management computer)
 CACHEURL=http://192.168.0.150
+IRONIC_IP=192.168.0.150
 IRONIC_KERNEL_PARAMS=console=ttyS0
 # IP where the BMCs can access Ironic to get the virtualmedia boot image.
 # This is the IP of the management computer in the out of band network.


t-h2o:
> I don't think this will work. Have you tried it?

Yes I tried and that works.

> So if we set IRONIC_IP like this, ironic will wait for it to get bound inside the container and that will never happen.

While my quickstart debugging session. I solved my issue by setting the first environment variable with "IP" of ironic container without thinking the other possible impacts.

After changing this value, the server received the good IP address (not the container IP but the management computer).

I didn't dig into the ironic/container details to understand the impact.

lentzi90:
Oh, sorry I mixed it up with PROVISIONING_IP. This may work actually!
I'm curious if using IRONIC_IP is a better way to go than IRONIC_BASE_URL and IRONIC_HTTP_URL that we used [in BMO](https://github.com/metal3-io/baremetal-operator/commit/1306fe42ed784032699c614368d0288a4499b237) to work around this. What do you think @dtantsur ?

Rozzii:
> Oh, sorry I mixed it up with PROVISIONING_IP. This may work actually! I'm curious if using IRONIC_IP is a better way to go than IRONIC_BASE_URL and IRONIC_HTTP_URL that we used [in BMO](https://github.com/metal3-io/baremetal-operator/commit/1306fe42ed784032699c614368d0288a4499b237) to work around this. What do you think @dtantsur ?

We are already following a bad practice that has become very visible during IPv6 related work, we should not rely on IPs we should utilize domain names for all things we can but that would add an extra layer of complexity.

Anyhow I don't think we should use IRONIC_IP as the "go to" or recommended way to configure things.

lentzi90 (commented):
Thanks for the info @Rozzii !
I think I will try to "simplify" the BMO e2e in this way. Even though it may not be the perfect solution, it is at least fewer extra variables.
/approve

PR #513, opened 2025-04-11
remove the unused multilingual field from book.toml

szabgab:
See https://github.com/szabgab/public-mdbooks/issues/10

PR #565, opened 2025-08-11
remove unused multilingual option from book.toml

tuminoid:
Remove unused multilingual option as it has been removed from mdbook.

See https://github.com/szabgab/mdbooks.code-maven.com/issues/10

Supersedes stale #513.


smoshiur1237 (commented):
/lgtm

PR #579, opened 2025-10-07
User guide: remove mentions of Inspector from quick start

dtantsur:
This is just a quick fix, we need to switch the entire installation
process to IrSO.

Similarly, auth-config is not needed (for either services) since the
Inspector removal.


lentzi90 (commented):
I was going to say we need to bump the versions also, but I see you have another PR for that, so
/approve
I am planning to run through this guide soon and try to create an automated test for it so it will be easier to keep up to date.

PR #582, opened 2025-10-13
Bump versions

lentzi90:
We have been stuck on older versions due to #584.
The builder has been changed to use Noble now so we can bump the mdbook version. This PR also bumps the go version used to build the plugins for mdbook.

The newer version of mdbook is stricter with the summary table of content. We cannot have two links pointing to the same page, like we do for BMO basic and advanced features. To fix this I simply split it into two pages.
Maybe not the nicest solution. I am open to suggestions for how to do it instead. I tried using anchor links first but mdbook interpreted those as file names instead (e.g. `features.md#advanced`).

I have rebased this PR on top of #583 (that fixes the build container) and then bumped the build container here to noble instead of focal.

Fixes: #584


lentzi90:
Ok that was not so straight forward...
/hold

lentzi90:
The issue is that the netlify builder we use is very old. We should bump to one that is Noble based probably.
I'll see what I can do about that.

lentzi90:
I am going to close this. We need to bump the netlify builder before we can do this anyway. In the mean time, https://github.com/metal3-io/metal3-docs/pull/583 fixes the build container for local development.

lentzi90:
Created an issue to track: https://github.com/metal3-io/metal3-docs/issues/584
Closing

lentzi90:
@fmuyassarov was able to update the netlify builder to Noble. Let's see how this works now

lentzi90:
This PR now includes the commit from https://github.com/metal3-io/metal3-docs/pull/583 also.

lentzi90:
Let's leave this on hold until #583 is merged. It has all the changes to the dockerfile so they are neatly separated. After it is merged we can bump the versions here without affecting anything else.

lentzi90:
#583 is merged and I have rebased this.
/hold cancel

PR #583, opened 2025-10-14
Use netlify build container

lentzi90:
This makes the build container much closer to the real netlify environment. We use their official build container and install mdbook the same way we do when publishing the site.

- Use official netlify build container
- Simplify variables (remove MDBOOK_VERSION)
- Download mdbook binary directly in container instead of building, same as netlify
- Set user/group ID for container mounts to avoid issues with file permissions

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,5 +1,5 @@
 ARG BUILD_IMAGE=docker.io/golang:1.21.9@sha256:7d0dcbe5807b1ad7272a598fbf9d7af15b5e2bed4fd6c4c2b5b3684df0b317dd
-FROM $BUILD_IMAGE as builder
+FROM $BUILD_IMAGE AS builder
 WORKDIR /workspace
 
 # Run this with docker build --build_arg $(go env GOPROXY) to override the goproxy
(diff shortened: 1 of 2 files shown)


docs/Dockerfile
```diff
@@ -22,10 +22,10 @@ ARG ARCH=amd64
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} \
   go build -tags=tools -a -ldflags '-extldflags "-static"' \
   -o mdbook-releasetags ./releasetags
+ARG MDBOOK_RELEASE_URL="https://github.com/rust-lang/mdBook/releases/download/v0.4.37/mdbook-v0.4.37-x86_64-unknown-linux-gnu.tar.gz"
```
lentzi90, line 25: This is annoying... Since they include the version in the file name itself, we cannot just point this URL at the latest release. It is set form the Makefile anyway but we should try to keep this URL in sync with what we have in the Makefile.

fmuyassarov: I guess there is no process that directly executes the docker build on this Dockerfile right? In other words, we always build via Makefile targets, if that's correct, should we instead pass the `MDBOOK_RELEASE_URL` variable  from the Makefile as an argument to the Dockefile?

lentzi90: Yes we do this, I just added the URL here as "default" value, since otherwise docker will show warnings that the ARG does not have a proper default value

docs/Dockerfile
```diff
@@ -22,10 +22,10 @@ ARG ARCH=amd64
...
+RUN curl -L "${MDBOOK_RELEASE_URL}" \
+  | tar xvz -C /workspace
 
-FROM rust:1.78.0-slim@sha256:517c6272b328bc51c87e099ef4adfbc7ab4558af2d757e8d423c7c3f1cbbf9d5
-ARG MDBOOK_VERSION="0.4.37"
-RUN cargo install mdbook --vers ${MDBOOK_VERSION}
-RUN cp /usr/local/cargo/bin/mdbook /usr/bin/mdbook
-COPY --from=builder /workspace/mdbook-releasetags /usr/bin/mdbook-releasetags
+FROM netlify/build:focal
```
fmuyassarov, line 29:
```suggestion
FROM netlify/build:noble
```
since we just switched to it?

lentzi90: Yeah, sorry for the confusion. I pushed this PR before we change it.
I have included the changes here in https://github.com/metal3-io/metal3-docs/pull/582 and bumped the version there. I'm not sure what makes more sense. Maybe I should bump to noble here and then we bump the mdbook in https://github.com/metal3-io/metal3-docs/pull/582?
Or I just squash everything in that PR and we close this?

fmuyassarov: No strong opinion, but what if we do all the version related changes in #582 and leave everything else as is here? I think the only thing you would need to move out from here to #582 is the go version change.

lentzi90: Yeah that sounds reasonable!

lentzi90:
Ok I think this is good to go in. All version bumps can then be done in https://github.com/metal3-io/metal3-docs/pull/582 that I will rebase when this is merged.

PR #586, opened 2025-10-15
Propose support for Ironic Standalone Networking

alegacy:
This adds a proposed design to add support for the new Ironic Networking service which provides the ability to control Top-of-Rack (TOR) switches without requiring any other OpenStack services (e.g., Neutron).

The Ironic Networking service is still under development.  It is working its way thru the review process which you can check on here:
https://review.opendev.org/q/(status:open+or+is:wip)+topic:feature/standalone-networking

### DIFF
--- a/design/ironic-standalone-networking-service.md
+++ b/design/ironic-standalone-networking-service.md
@@ -1,0 +1,776 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add support for Ironic Networking service
+
+## Status
+
+provisional
+
+## Summary
+
+This design proposes adding support for the Ironic Networking service to
+Metal³, enabling automatic configuration of Top-of-Rack (ToR) switch ports
+during the lifecycle of bare metal nodes. The Ironic Networking service
+provides a pluggable framework for managing switch port configurations such as
+VLAN assignments, port modes (access/trunk/hybrid), and LAG settings.
+This integration will allow Metal³ users to define network configurations
+declaratively through Kubernetes CRDs, which will be automatically applied to
+physical switch ports as nodes are provisioned and deprovisioned.  Switch ports
+will be identified using LLDP information collected during initial inspection
+of nodes during enrollment.
+
+## Motivation
+
+Currently, Metal³ users must manually configure ToR switch ports before
+provisioning bare metal hosts, which creates operational overhead and
+opportunities for misconfiguration. The Ironic Networking service provides a
+standardized way to automate switch port configuration, but Metal³ lacks
+integration with this service. By adding support for Ironic Networking, Metal³
+can provide a fully automated, declarative workflow for both server and network
+provisioning.
+
+### Goals
+
+- Enable optional support for Ironic Networking service in Metal³ deployments
+- Allow users to declaratively define switch configurations through Kubernetes
+  CRDs
+- Automatically configure switch ports during node provisioning and
+  deprovisioning
+- Maintain backward compatibility for deployments not using Ironic Networking
+
+### Non-Goals
+
+- Supporting all possible switch vendor-specific features beyond the base
+  capabilities provided by networking-generic-switch
+- Implementing custom switch drivers (users will use existing drivers from
+  networking-generic-switch)
+- Managing switch infrastructure itself (firmware updates, switch provisioning,
+  etc.)
+- Supporting dynamic network reconfiguration after node deployment (initial
+  scope is provisioning-time configuration only)
+
+## Proposal
+
+This proposal introduces these main components:
+
+1. **New CRDs**: `BareMetalSwitch` and `HostNetworkAttachment` to define
+   switch configurations and network interface configurations respectively
+1. **Enhanced Ironic CRD**: Add service enablement flag and service
+   network configuration to the `ironic.metal3.io` CRD
+1. **Enhanced BareMetalHost CRD**: Add `networkInterfaces` field to reference
+   network configurations for each network interface
+1. **Ironic Networking Pod**: Deploy the Ironic Networking service
+   as a separate Pod alongside the existing main Ironic Pod.
+1. **Share Secret**: To decouple IrSO from the management of switch config
+   files and SSH keys, two Secrets will be shared between BMO and IrSO to
+   hold the contents of those files to be passed opaquely to the
+   ironic-networking service.
+
+### User Stories
+
+#### Story 1
+
+As a Metal³ operator, I want to define my ToR switch configurations once in
+Kubernetes CRDs so that Metal³ can automatically manage switch port
+configurations without manual intervention.
+
+The operator creates `BareMetalSwitch` resources for each ToR switch:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalSwitch
+metadata:
+  name: rack1-tor1
+  namespace: default-ironic-namespace
+spec:
+  driver: generic-switch
+  deviceType: dell_os10
+  address: 192.168.1.10
+  credentials:
+    type: "password"
+    secretName: switch-credentials
+  macAddress: 3e:c5:c0:cb:fa:f6
+```
+
+BMO reconciles the contents of all `BareMetalSwitch` instances and builds a
+config file which is stored in a Secret that is mounted to the
+ironic-networking Pod by IrSO opaquely. Similarly, any SSH key credentials are
+aggregated into a Secret that is mounted to the ironic-networking Pod by
+IrSO.  The ironic-networking container is responsible for reading the
+config files and reacting to any changes therein.
+
+The credentials Secret is expected to contain keys for the `username`,
+`password`, and (optionally) `enable-secret`.  It may also contain an SSH
+private key in which case the Secret type should be `kubernetes.io/ssh-auth`
+and the field name should be `ssh-privatekey`.
+
+#### Story 2
+
+As a Metal³ user, I want to define network configurations for my bare metal
+hosts declaratively so that switch ports are automatically configured when nodes
+are provisioned.
+
+The user creates a `HostNetworkAttachment` defining the desired network
+configuration for an access mode port for the primary network interface:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: HostNetworkAttachment
+metadata:
+  name: machine-network-vlan-100
+  namespace: some-cluster-namespace
+spec:
+  mode: access
+  nativeVLAN: 100
+```
+
+The user optionally creates other `HostNetworkAttachment` instances
+defining the desired network configuration for trunk ports for the secondary
+network interface(s):
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: HostNetworkAttachment
+metadata:
+  name: data-network-vlans-200-202
+  namespace: some-cluster-namespace
+spec:
+  mode: trunk
+  allowedVLANs: [200, 201, 202]
+```
+
+Then references them in the `BareMetalHost`:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-1
+  namespace: some-host-namespace
+spec:
+  # ... other fields ...
+  networkInterfaces:
+  - macAddress: "aa:bb:cc:dd:ee:01"
+    networkAttachment:
+      namespace: some-cluster-namespace
+      name: machine-network-vlan-100
+    switchport:
+      switchId: "00:11:22:33:44:55"
+      portId: "ethernet1/2/3"
+  - macAddress: "aa:bb:cc:dd:ee:02"
+    networkAttachment:
+      namespace: some-cluster-namespace
+      name: data-network-vlans-200-202
+    switchport:
+      switchId: "00:11:22:33:44:55"
+      portId: "ethernet4/5/6"
+```
+
+When the host is provisioned, Metal³ automatically configures the corresponding
+switch port for the primary interface to access mode on VLAN 100, and the
+secondary interface to trunk mode with tagged VLANs 200-202.
+
+Ideally, `HostNetworkAttachment` instances should only be referenced by
+`BareMetalHost` instances after initial inspection so that NIC information is
+present in the `HardwareData` object and available for validation of the
+network interface info by a validation webhook. If inspection has not yet run
+the validation webhook _could_ defer validation until the end of the initial
+inspection phase.
+
+#### Story 3
+
+As a Metal³ operator, I want to configure a service network VLAN that allows
+bare metal hosts to communicate with Ironic during inspection and provisioning,
+separate from their final tenant network configuration.
+
+The operator configures the service network in the Ironic resource:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: Ironic
+metadata:
+  name: default
+  namespace: default-ironic-namespace
+spec:
+  networkingService:
+    enabled: true
+    networkDriver: ironic-networking
+    providerNetworks:
+    - type: idle
+      mode: access
+      nativeVLAN: 10
+    - type: inspection
+      mode: access
+      nativeVLAN: 10
+```
+
+Ironic will ensure that switch ports are always reset to this VLAN
+configuration whenever a node is deprovisioned or whenever the final networking
+configuration of a port (the tenant network) must be cleared.  This
+ensures that it is always possible to delete the BMH and re-add it without any
+manual intervention required on the switch.
+
+## Design Details
+
+### New CRDs
+
+#### BareMetalSwitch
+
+The `BareMetalSwitch` CRD represents a physical ToR switch managed by Ironic
+Networking:
+
+```go
+// SwitchCredentialType defines the type of credential used for switch authentication.
+type SwitchCredentialType string
+
+const (
+    // SwitchCredentialTypePassword indicates password-based authentication
+    SwitchCredentialTypePassword SwitchCredentialType = "password"
+    // SwitchCredentialTypePublicKey indicates SSH public key-based authentication
+    SwitchCredentialTypePublicKey SwitchCredentialType = "publicKey"
+)
+
+// SwitchCredentials defines the credentials used to access the switch.
+type SwitchCredentials struct {
+    // Type specifies the type of credential used for authentication.
+    // +kubebuilder:validation:Enum=password;publicKey
+    Type SwitchCredentialType `json:"type"`
+
+    // The name of the secret containing the switch credentials.
+    // For password authentication, requires keys "username" and "password".
+    // For SSH key authentication, requires keys "username" and "ssh-privatekey".
+    SecretName string `json:"secretName"`
+}
+
+// BareMetalSwitchSpec defines the desired state of BareMetalSwitch.
+type BareMetalSwitchSpec struct {
+    // Important: Run "make generate manifests" to regenerate code
+    // after modifying this file
+
+    // Address is the network address of the TOR switch (IP address or hostname).
+    // +kubebuilder:validation:Required
+    Address string `json:"address"`
+
+    // MACAddress is the MAC address of the switch management interface.
+    // +kubebuilder:validation:Pattern=`[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}`
+    MACAddress string `json:"macAddress"`
+
+    // Driver specifies the type of driver to use for switch management.
+    // Examples: "generic-switch"
+    // +kubebuilder:validation:Required
+    Driver string `json:"driver"`
+
+    // DeviceType specifies the specific device type or model of the switch.
+    // This must be a value supported by the driver selected by `Driver`.
+    // Examples: "netmiko_dell_os10", "netmiko_cisco_nxos"
+    // This helps with driver-specific functionality.
+    DeviceType string `json:"deviceType"`
+
+    // Credentials defines how to authenticate with the switch.
+    // +kubebuilder:validation:Required
+    Credentials SwitchCredentials `json:"credentials"`
+
+    // Port specifies the management port to connect to (e.g., SSH port 22, HTTPS port 443).
+    // If not specified, the driver will use its default port.
+    // +kubebuilder:validation:Minimum=1
+    // +kubebuilder:validation:Maximum=65535
+    // +optional
+    Port *int32 `json:"port,omitempty"`
+
+    // DisableCertificateVerification disables verification of server 
+    // certificates when using HTTPS to connect to the switch. This is 
+    // required when the server certificate is self-signed, but is insecure 
+    // because it allows a man-in-the-middle to intercept the connection.
+    // +optional 
+    DisableCertificateVerification *bool `json:"disableCertificateVerification,omitempty"`
+}
+```
+
+#### HostNetworkAttachment
+
+The `HostNetworkAttachment` CRD defines how a switch port should be
+configured:
+
+```go
+// SwitchportMode defines the switchport mode for network interfaces.
+// +kubebuilder:validation:Enum=access;trunk;hybrid
+type SwitchportMode string
+
+const (
+    // SwitchportModeAccess sets the interface to access mode (single VLAN)
+    SwitchportModeAccess SwitchportMode = "access"
+    // SwitchportModeTrunk sets the interface to trunk mode (multiple VLANs)
+    SwitchportModeTrunk SwitchportMode = "trunk"
+    // SwitchportModeHybrid sets the interface to hybrid mode (access + trunk)
+    SwitchportModeHybrid SwitchportMode = "hybrid"
+)
+
+// HostNetworkAttachmentSpec defines the desired switchport configuration.
+type HostNetworkAttachmentSpec struct {
+    // Mode defines the switchport mode (access, trunk, or hybrid)
+    // +kubebuilder:validation:Enum=access;trunk;hybrid
+    // +kubebuilder:default=access
+    Mode SwitchportMode `json:"mode,omitempty"`
+
+    // NativeVLAN is the untagged VLAN ID for the port
+    // +kubebuilder:validation:Minimum=1
+    // +kubebuilder:validation:Maximum=4094
+    NativeVLAN int32 `json:"nativeVLAN,omitempty"`
+
+    // AllowedVLANs is a list of VLAN IDs allowed on trunk/hybrid ports
+    // Only valid for trunk and hybrid modes
+    // +optional
+    AllowedVLANs []int32 `json:"allowedVLANs,omitempty"`
+}
+```
+
+### Enhanced Ironic CRD
+
+Add the following fields to the `IronicSpec`:
+
+```go
+type ProviderNetworkConfig struct {
+    // Type specifies which provider network this configuration applies to.
+    // +kubebuilder:validation:Enum=idle;inspection;cleaning;rescuing;servicing;provisioning
+    Type string `json:"type"`
+
+    // Mode specifies the switch port mode for a provider network
+    Mode SwitchPortMode `json:"mode"`
+
+    // NativeVLAN specifies the native VLAN ID
+    NativeVLAN int `json:"nativeVLAN"`
+
+    // AllowedVLANs specifies the list of allowed VLANs for trunk/hybrid modes
+    AllowedVLANs []int `json:"allowedVLANs,omitempty"`
+}
+
+type NetworkingService struct {
+    // Enabled enables the Ironic Networking Service integration
+    Enabled  bool `json:"enabled,omitempty"`
+
+    // NetworkDriver sets the default network interface in the Ironic
+    // configuration file.  This defaults to "ironic-networking".
+    // +optional
+    // +kubebuilder:default=ironic-networking
+    NetworkDriver string `json:"networkDriver,omitempty"`
+
+    // RPCPort is the internal RPC port used for Ironic Networking
+    // Only change this if the default value causes a conflict on your deployment
+    // +optional
+    RPCPort *int `json:"rpcPort,omitempty"`
+
+    // ProviderNetworks defines the provider network configurations for Ironic
+    ProviderNetworks []ProviderNetworkConfig `json:"providerNetworks,omitempty"`
+
+    // SwitchDrivers sets the supported switch drivers for the service.
+    SwitchDrivers []string `json:"switchDrivers,omitempty"`
+
+    // SwitchConfigSecretName references a Secret that contains the set of
+    // switch config entries to be used by the Ironic Networking service
+    SwitchConfigSecretName string `json:"switchConfigSecretName"`
+
+    // SwitchCredentialsSecretName references a Secret that contains the set of
+    // SSH private key files that are referenced from the configs contained
+    // within the SwitchConfigSecretName.
+    SwitchCredentialsSecretName string `json:"switchCredentialsSecretName"`
+}
+
+type IronicSpec struct {
+    // ... existing fields ...
+
+    // NetworkingService provides configuration attributes for the Networking
+    // Service if it is required.
+    // +optional
+    NetworkingService *NetworkingService `json:"networkingService,omitempty"`
+}
+```
+
+### Enhanced BareMetalHost CRD
+
+Add the following field to the `BareMetalHostSpec`:
+
+```go
+type BareMetalHostSpec struct {
+    // ... existing fields ...
+
+    // NetworkInterfaces defines the network configuration for each NIC
+    NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
+}
+
+// NetworkInterface defines the network configuration for a specific interface.
+type NetworkInterface struct {
+    // Name of the network interface (e.g., "eth0", "ens1f0")
+    // This must match the name of a NIC in status.hardware.nics.
+    // If MACAddress is set then this field is ignored.
+    // +optional
+    Name string `json:"name,omitempty"`
+
+    // MAC address of the network interface.  This must match the MAC address
+    // of the network interface in status.hardware.nics.
+    // +optional
+    // +kubebuilder:validation:Pattern=`[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}`
+    MACAddress string `json:"macAddress,omitempty"`
+
+    // NetworkAttachment references the HostNetworkAttachment for this interface
+    NetworkAttachment *NetworkAttachmentRef `json:"networkAttachment"`
+}
+
+type NetworkAttachmentRef struct {
+    // Name of the HostNetworkAttachment
+    Name string `json:"name"`
+
+    // Namespace of the HostNetworkAttachment (defaults to BMH namespace)
+    Namespace string `json:"namespace,omitempty"`
+}
+```
+
+### Ironic Configuration
+
+When `Ironic.NetworkingService.Enabled` is true, Metal³ will configure
+Ironic with the following sections:
+
+```ini
+...
+
+[ironic_networking_json_rpc]
+host_ip = <IRONIC_NETWORKING_SERVICE_IP>
+port = 6190
+auth_strategy = http_basic
+auth_type = http_basic
+http_basic_auth_user_file = /etc/ironic/htpasswd
+username = admin
+password = <GENERATED_PASSWORD>
+
+[ironic_networking]
+rpc_transport = json-rpc
+idle_network = <SERVICE_NETWORK_MODE>/<SERVICE_NETWORK_NATIVE_VLAN>
+provisioning_network = <SERVICE_NETWORK_MODE>/<SERVICE_NETWORK_NATIVE_VLAN>
+```
+
+For nodes with network configurations defined, the network interface driver
+should be set to `ironic-networking`.
+
+### Ironic Networking Service Configuration
+
+Metal³ will generate a configuration file for the Ironic Networking service:
+
+```ini
+[DEFAULT]
+debug = <DEBUG_SETTING>
+rpc_transport = json-rpc
+auth_strategy = http_basic
+http_basic_auth_user_file = /etc/ironic/htpasswd
+
+[service_catalog]
+auth_type = http_basic
+username = admin
+password = <GENERATED_PASSWORD>
+endpoint_override = https://<IRONIC_API_INTERNAL_IP>:6385
+cafile = /etc/ssl/certs/ca-bundle.crt
+
+[ironic_networking_json_rpc]
+host_ip = <IRONIC_NETWORKING_SERVICE_IP>
+port = 6190
+auth_strategy = http_basic
+auth_type = http_basic
+http_basic_auth_user_file = /etc/ironic/htpasswd
+username = admin
+password = <GENERATED_PASSWORD>
+
+[ironic_networking]
+switch_config_file = /etc/ironic/networking/switch-configs.conf
+driver_config_dir = /etc/ironic/drivers
+enabled_switch_drivers = generic-switch
+serialize_switch_operations = True
+allowed_vlans =
+```
+
+### Switch Configuration File
+
+For each `BareMetalSwitch` resource, Metal³ will generate an entry in the
+switch configuration file:
+
+```ini
+[switch:<SWITCH_NAME>]
+driver_type = <DRIVER_TYPE>
+device_type = <DEVICE_TYPE>
+address = <ADDRESS>
+username = <USERNAME>
+password = <PASSWORD>
+enable_secret = <ENABLE_SECRET>
+mac_address = <MAC_ADDRESS>
+```
+
+This file will be stored in a Secret and mounted to the Ironic Networking
+container. When any `BareMetalSwitch` is created, updated, or deleted, the
+Secret will be regenerated and the Ironic Networking service will be restarted.
+
+### Ironic Port Configuration
+
+When a `BareMetalHost` has `networkInterfaces` defined, Metal³ will update the
+corresponding Ironic Port's `extra` field with a `switchport` dictionary
+conforming to the schema:
+
+```json
+{
+  "extra": {
+    "switchport": {
+      "mode": "access",
+      "native_vlan": 100,
+      "allowed_vlans": [200, 300, 400]
+    }
+  }
+}
+```
+
+This information represents one half of the details required to configure the
+switch port.  The identity of the switch and the attached port are obtained
+from LLDP and stored in the port's `local_link_connection` field, which is
+populated during initial inspection.
+
+### Implementation Details/Notes/Constraints
+
+1. **Service Network Setup**: Before creating `BareMetalHost` resources, users
+   must ensure that server switch ports are manually configured to the service
+   VLAN to allow initial communication with Ironic. This enables the initial
+   inspection to complete which provides LLDP information for the Node. This
+   is a one-time setup requirement. (This is known as the "idle" VLAN in
+   Ironic terminology.)
+
+1. **Validation Timing**: Network interface validation can only occur after
+   inspection completes, as network interface information is not available
+   before that point. A condition in the `BareMetalHost` status field will be
+   updated after inspection to indicate if Network interfaces are valid.
+
+1. **HostNetworkAttachment Immutability**: To prevent accidental
+   misconfiguration, users cannot modify or delete a `HostNetworkAttachment`
+   while it is referenced by any `BareMetalHost`. The validation webhook will
+   reject such operations.
+
+1. **Switch Configuration Changes**: When `BareMetalSwitch` resources are
+   modified, the entire switch configuration Secret is regenerated and the
+   Ironic Networking service is restarted. This may cause brief service
+   interruptions.
+
+1. **Credential Management**: Switch credentials are stored in Kubernetes
+   Secrets and referenced by `BareMetalSwitch` resources. The Metal³ controller
+   will read these Secrets to populate the switch configuration file.
+
+1. **Container Deployment**: The Ironic Networking service will run as a
+   separate container in the metal3 pod, alongside the existing metal3-ironic
+   container. It will share the `/etc/ironic` volume for configuration files
+   and the htpasswd file.
+
+1. **Network Driver Selection**: The default network interface in Ironic's
+   configuration file will be set to `ironic-networking` unless overridden by
+   the `Ironic.Spec.NetworkingService.NetworkDriver` attribute.  Optionally,
+   BMO could be extended to allow overriding this on a per-node attribute
+   controlled separately by an environment variable or other configuration
+   data specific to BMO.
+
+1. **Provider Networks Scope**: The provider networks configuration defines
+   which VLAN details should be used by Ironic for each of the provider
+   networks it manages (e.g., idle network, inspection network, cleaning
+   network, etc...).
+
+1. **Generic Switch Device Types**: The `deviceType` field in `BareMetalSwitch`
+   must be one of the device types supported by networking-generic-switch.
+   Valid values include, but are not limited to, `netmiko_cisco_ios`,
+   `netmiko_dell_force10`, `juniper_junos`. The full list can be found in the
+    [setup.cfg](https://github.com/openstack/networking-generic-switch/blob/master/setup.cfg#L35)
+   of the networking-generic-switch project.
+
+### Risks and Mitigations
+
+**Risk**: If Metal³'s instantiation of Ironic is not configured to use
+persistent storage for the Ironic database, and the Ironic Pod is restarted
+then all node and port information is lost, including the LLDP information
+and switch port configuration derived from the BMH configuration.
+
+**Mitigation**: The BMH reconciler can be modified to restore the last known
+set of Port objects including their `port.extra.switchport` and
+`port.local_link_connection` attributes using information from the
+`HardwareData` object.
+
+**Risk**: Switch configuration errors could render hosts unreachable or prevent
+provisioning.
+
+**Mitigation**: Implement thorough validation of `BareMetalSwitch` and
+`HostNetworkAttachment` resources. Add status conditions to report
+configuration errors. Ensure service network is properly configured before
+attempting to provision hosts.
+
+**Risk**: Credential exposure in switch configuration files.
+
+**Mitigation**: Store the switch configuration file in a Kubernetes Secret with
+appropriate RBAC restrictions. Use Secret references for credentials in the
+`BareMetalSwitch` CRD rather than embedding them directly.
+
+**Risk**: Ironic Networking service failures could impact provisioning
+workflows.
+
+**Mitigation**: Make the feature optional via the
+`Ironic.NetworkingService.Enabled` flag. Implement health checks for the
+Ironic Networking service. Add detailed status reporting and error messages.
+
+### Work Items
+
+1. Implement ironic-image configuration updates for networking support
+1. Extend the `Ironic` CRD with `NetworkingService` information
+1. Implement Ironic Networking service Pod deployment
+1. Define and implement the `BareMetalSwitch` CRD and controller
+1. Implement switch configuration and SSH key file Secret generation
+   in the BareMetalSwitch controller
+1. Define and implement the `HostNetworkAttachment` CRD and controller
+1. Extend the `BareMetalHost` CRD with `networkInterfaces` field
+1. Implement Ironic Port `extra.switchport` population in the BMH controller
+1. Implement validation logic for network interface configurations
+1. Add status reporting for network configuration state
+1. Implement re-population of Ironic Node/Port information if database is
+   not persistent
+1. Update Metal³ documentation with Ironic Networking setup instructions
+1. Create example manifests for common switch configurations
+
+### Dependencies
+
+- **Ironic Networking service**: Requires the Ironic Networking service
+  container image to be available. This is part of the Ironic image; therefore,
+  the only requirement is to use a version of Ironic that supports the new
+  service.  Version `TBD` supports the required service.
+
+- **networking-generic-switch**: The generic-switch driver depends on the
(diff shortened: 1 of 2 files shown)


design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+- Managing switch infrastructure itself (firmware updates, switch provisioning,
+  etc.)
+- Supporting dynamic network reconfiguration after node deployment (initial
+  scope is provisioning-time configuration only)
+
+## Proposal
+
+This proposal introduces four main components:
+
+1. **New CRDs**: `BareMetalSwitch` and `BareMetalNetworkAttachment` to define
```
dtantsur: Somewhat nitpicking here, but I'd prefer `HostNetworkAttachment`. There is no written policy AFAIK, but my gut feeling is to use `BareMetal...` for physical entities and `Host...` for things that a related to BMH (see HostUpdatePolicy, HostFirmwareSettings, etc).

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+   switch configurations and network interface configurations respectively
+2. **Enhanced Provision CRD**: Add `enableIronicNetworking` flag and service
```
dtantsur: Provisioning CR is OpenShift-ism, it does not exist in upstream Metal3.

This functionality has to be brought to https://github.com/metal3-io/ironic-standalone-operator. Possibly as a separate CR from the existing Ironic? I don't remember how tied the new service is to Ironic itself. We can discuss the exact design once I return from the OpenInfra Summit, if you wish.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+  deviceType: dell_os10
+  address: 192.168.1.10
+  credentials:
+    type: "password"
+    secretName: switch-credentials
+  macAddress: 3e:c5:c0:cb:fa:f6
+```
+
+Metal³ automatically generates the Ironic Networking switch configuration file
+and restarts the service when switches are added or modified.
```
dtantsur: Here is a problem: BMO does not manage Ironic services, it's done by a different operator (IrSO update, CBO in OpenShift).

If you want to go down this path, it probably becomes a resource in IrSO and is managed on the same level as Ironic itself. Which is pretty unfortunate, but maybe it's the way.

Maybe, at least as a stopgap measure, IrSO could mount a predefined Secret that the BareMetalSwitch will later modify as BareMetalSwitch resources are created or edited?  This is by no means an elegant solution as well, of course.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+The credentials Secret is expected to contain keys for the `username`,
+`password`, and (optionally) `enable-secret`.
+
+#### Story 2
+
+As a Metal³ user, I want to define network configurations for my bare metal
+hosts declaratively so that switch ports are automatically configured when nodes
+are provisioned.
+
+The user creates a `BareMetalNetworkAttachment` defining the desired network
```
dtantsur: See above: I'd rather call this `HostNetworkAttachment` for consistency with a few existing BMH-related resources.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+configuration for an access mode port for the primary network interface:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalNetworkAttachment
+metadata:
+  name: machine-network-vlan100
+spec:
+  mode: access
+  nativeVlan: 100
```
dtantsur: nit: I'd call it `nativeVLAN`, but I won't insist

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-1
+spec:
+  # ... other fields ...
+  networkInterfaces:
+  - macAddress: "aa:bb:cc:dd:ee:01"
```
dtantsur, line 158: Is there any value in making each record here a separate CRD (like HostNetworkInterface)? This is not a suggestion, I'm genuinely asking.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+When the host is provisioned, Metal³ automatically configures the corresponding
+switch port for the primary interface to access mode on VLAN 100, and the
+secondary interface to trunk mode with tagged VLANs 200-202.
+
+#### Story 3
+
+As a Metal³ operator, I want to configure a service network VLAN that allows
+bare metal hosts to communicate with Ironic during inspection and provisioning,
+separate from their final tenant network configuration.
```
dtantsur, line 189: This section has to be changed to not reference the OpenShift-specific resource.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+    // +kubebuilder:validation:Maximum=65535
+    // +optional
+    Port *int32 `json:"port,omitempty"`
+
+    // SkipTLSVerify disables verification of server certificates when using
+    // HTTPS to connect to the switch. This is required when the server
+    // certificate is self-signed, but is insecure because it allows a
+    // man-in-the-middle to intercept the connection.
+    // +optional
+    SkipTLSVerify *bool `json:"skipTLSVerify,omitempty"`
```
dtantsur: We already have prior art in form of `disableCertificateVerification`, I suggest using the same name

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+}
+
+type BareMetalSwitchStatus struct {
+    // Conditions track the state of the switch configuration
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
```
dtantsur: Which conditions are we going to expose?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+type BareMetalNetworkAttachmentSpec struct {
+    // Mode defines the switchport mode (access, trunk, or hybrid)
+    // +kubebuilder:validation:Enum=access;trunk;hybrid
+    // +kubebuilder:default=access
+    Mode SwitchportMode `json:"mode,omitempty"`
+
+    // NativeVLAN is the untagged VLAN ID for the port
+    // +kubebuilder:validation:Minimum=1
+    // +kubebuilder:validation:Maximum=4094
+    NativeVLAN int32 `json:"nativeVlan,omitempty"`
```
dtantsur: nit: Either the Go variable should be NativeVlan, or the JSON one should be nativeVLAN

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+    // +optional
+    AllowedVLANs []int32 `json:"allowedVlans,omitempty"`
+}
+
+type BareMetalNetworkAttachmentStatus struct {
+    // Conditions track the state of the network attachment
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+    // AttachedHosts lists the BareMetalHosts currently using this attachment
+    AttachedHosts []string `json:"attachedHosts,omitempty"`
```
dtantsur: Do we have strong reasons not to use `OwnerReferences`? You want to avoid garbage collecting?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+    // NetworkInterfaces defines the network configuration for each NIC
+    NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
+}
+
+// NetworkInterface defines the network configuration for a specific interface.
+type NetworkInterface struct {
+    // Name of the network interface (e.g., "eth0", "ens1f0")
+    // This must match the name of a NIC in status.hardware.nics.
+    // If MACAddress is set then this field is ignored.
```
dtantsur, line 409: Let's make sure we validate assumptions like this in the webhook

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+    // +optional
+    NetworkAttachment *NetworkAttachmentRef `json:"networkAttachment,omitempty"`
+}
+
+type NetworkAttachmentReference struct {
+    // Name of the BareMetalNetworkAttachment
+    Name string `json:"name"`
+
+    // Namespace of the BareMetalNetworkAttachment (defaults to BMH namespace)
+    Namespace string `json:"namespace,omitempty"`
```
dtantsur, line 428: This may have security implications. We already had to fix a CVE where referencing a secret from another namespace could be used for privilege escalation. A HostNetworkAttachment should not consider anything sensitive.. but my gut feeling is to avoid cross-namespace referencing.

Do you have a strong case for it?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+Add the following to `BareMetalHostStatus`:
+
+```go
+type BareMetalHostStatus struct {
+    // ... existing fields ...
+
+    // NetworkConfigurationReady indicates whether network configurations are
+    // valid
+    NetworkConfigurationReady bool `json:"networkConfigurationReady,omitempty"`
```
dtantsur: Let's use a condition for this instead? At least that seems standard in Kubernetes.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+When `enableIronicNetworking` is true, Metal³ will configure Ironic with the
+following sections:
+
+```ini
+...
+
+[networking_json_rpc]
+host_ip = <IRONIC_NETWORKING_SERVICE_IP>
+port = 8090
```
dtantsur: Let's avoid ports in 80* area, we already got a nasty conflict with 8089.

A lot depends on whether we actually need to run the networking service in the same pod. If we do, it will run with host networking, so we need to be mindful about ports. So far, we tend to use 618X ports for internal purposes (6189 was the one we picked for the normal RPC).

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+address = <ADDRESS>
+username = <USERNAME>
+password = <PASSWORD>
+enable_secret = <ENABLE_SECRET>
+mac_address = <MAC_ADDRESS>
+```
+
+This file will be stored in a Secret and mounted to the Ironic Networking
+container. When any `BareMetalSwitch` is created, updated, or deleted, the
+Secret will be regenerated and the Ironic Networking service will be restarted.
```
dtantsur, line 511: Continuing from above: this is a bit of a problematic point since the Ironic pod(s) are not managed by baremetal-operator, rather by ironic-standalone-operator (upstream) or CBO (openshift).

If we can make ironic-networking read these files on fly, maybe we can change the mounted secret in runtime without restarting anything? I'm open to other ideas too.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+switch port.  The identity of the switch and the attached port are obtained
+from LLDP and stored in the port's `local_link_connection` field, which is
+populated during initial inspection.
+
+### Implementation Details/Notes/Constraints
+
+1. **Service Network Setup**: Before creating `BareMetalHost` resources, users
+   must ensure that server switch ports are manually configured to the service
+   VLAN to allow initial communication with Ironic. This is a one-time setup
+   requirement. (This is known as the "idle" VLAN in Ironic terminology.)
```
dtantsur: Let us clarify that this is because of the LLDP discovery described in the previous section. This is an important operation decision.

I wonder now if we need some what for a user to provide these values instead...

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+2. **Validation Timing**: Network interface validation can only occur after
+   inspection completes, as network interface information is not available
+   before that point. The `NetworkConfigurationReady` status field will be
+   updated after inspection.
```
dtantsur: Let's be explicit that we depend on (in-band) inspection for all this work: some people disable inspection.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+3. **BareMetalNetworkAttachment Immutability**: To prevent accidental
+   misconfiguration, users cannot modify or delete a `BareMetalNetworkAttachment`
+   while it is referenced by any `BareMetalHost`. The controller will reject
```
dtantsur: nit: s/controller/validating webhook/

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+   such operations.
+
+4. **Switch Configuration Changes**: When `BareMetalSwitch` resources are
+   modified, the entire switch configuration Secret is regenerated and the
+   Ironic Networking service is restarted. This may cause brief service
+   interruptions.
```
dtantsur, line 558: See above

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+5. **Credential Management**: Switch credentials are stored in Kubernetes
+   Secrets and referenced by `BareMetalSwitch` resources. The Metal³ controller
+   will read these Secrets to populate the switch configuration file.
+
+6. **Container Deployment**: The Ironic Networking service will run as a
+   separate container in the metal3 pod, alongside the existing metal3-ironic
+   container. It will share the `/etc/ironic` volume for configuration files
+   and the htpasswd file.
```
dtantsur, line 567: These files are generated on fly in Metal3, so I really wonder if you need to share the pod. Not sharing the pod will make at least one thing much easier: you won't be forced into host networking.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+7. **Network Driver Selection**: The Baremetal Operator will automatically set
+   the Ironic node's network interface driver to `standalone-experimental`
+   when:
+   - `enableIronicNetworking` is true in the Provision resource, AND
+   - The `BareMetalHost` has `networkInterfaces` defined
+
+8. **Service Network Scope**: The service network configuration applies to a
+   subset of the Ironic provider networks. Specifically, it configures only
+   the idle and inspection networks. All other provider networks (provisioning,
+   cleaning, rescuing, servicing) will default to the tenant VLAN configured on
```
dtantsur: This seems wrong: service network applies to all these. Most people will want the same network for inspection, cleaning and such, while having separate network(s) for the tenant workload.

nit: Metal3 does not support rescuing.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+   each port. This means only two VLANs are needed: the service VLAN for
+   Ironic operations, and the tenant VLAN(s) configured per-port.
+
+9. **Generic Switch Device Types**: The `deviceType` field in `BareMetalSwitch`
+   must be one of the device types supported by networking-generic-switch.
```
dtantsur, line 582: nit: let's link to n-g-s docs for context

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+   Valid values include, but are not limited to, `cisco_ios`, `dell_force10`,
+   `juniper_junos`.  The full list can be found in the setup.cfg file of the
+   networking-generic-switch project.
+
+### Risks and Mitigations
+
+**Risk**: Metal³'s instantiation of Ironic currently does not include
+persistent storage for the Ironic database.  If the Ironic Pod is restarted
```
dtantsur: OpenShift does not, upstream Metal3 can work with a real database.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+then all node and port information is lost, including the LLDP information
+and switch port configuration derived from the BMH configuration.
```
dtantsur, line 593: This is a significant issue. Very significant: every OpenShift update involves rebuilding the database.

Now I wonder if we should persist this information somewhat on the Kubernetes side, e.g. in HardwareData objects. Or really, as I mused above, create a field for that inside networkInterfaces and either accept input from a human or populate it after inspection.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+**Mitigation**: The BMH reconciler can be modified to restore the last known
+set of Port objects including their `port.extra.switchport` and
+`port.local_link_connection` attributes.
```
dtantsur: Yep, this. Then we need to cache local_link_connection.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,734 @@
...
+
+**Risk**: Switch configuration errors could render hosts unreachable or prevent
+provisioning.
+
+**Mitigation**: Implement thorough validation of `BareMetalSwitch` and
+`BareMetalNetworkAttachment` resources. Add status conditions to report
+configuration errors. Ensure service network is properly configured before
+attempting to provision hosts.
```
dtantsur, line 606: We need to make sure that if cleaning is disabled, we don't try to access switches or validating anything on BMH deletion. So that even broken hosts can be cleanly deleted.

design/ironic-standalone-networking-service.md
alegacy: Makes sense.  Will update.

design/ironic-standalone-networking-service.md
alegacy: Sorry, the fqdn of `provisionings.metal3.io` lead me to believe it was a metal3 thing.  I'll figure out how to fit this into Ironic Standalone Operator instead.

design/ironic-standalone-networking-service.md
alegacy: Oh ya, that's interesting.   In my current prototype, I have CBO generate an empty Secret and mount it to the Ironic Networking Pod, CBO then passes that Secret name over to BMO and BMO updates the contents whenever any of the switch CRDs are updated, and finally the Ironic Networking Pod notices that the Secret has changed (via the file mount) and then restarts itself (note: I used to have code to dynamically reload the drivers, but took that out because of a review comment).   

The switch is a standalone CRD (i.e., it is not coupled to the BMH or NetworkAttachment in any way) so I guess it could become a resource of IrSO directly and BMO/CBO would know nothing about it.    I think that's probably the simplest, but I'm ok either way.  

I'll plan on making it an IrSO resource unless you think otherwise.

Thanks for pointing this out.

design/ironic-standalone-networking-service.md
alegacy: Will do.

design/ironic-standalone-networking-service.md
alegacy: Ya, it's never clear how to do these... In Go, it's clear that it should be `NativeVLAN`, but not so clear in other places... I also prefer `nativeVLAN` for the yaml.

design/ironic-standalone-networking-service.md
alegacy: I thought about that, but I don't think there is any value in that.  In fact, I think it unnecessarily drives up the complexity.  The information is tightly coupled to the BMH and there is no chance it could be reused by reference anywhere else... so I think this is the best place for it.

design/ironic-standalone-networking-service.md
alegacy: Ok, didn't realize `Provisioning` was not a metal3 thing... will look at what's available in IrSO.

design/ironic-standalone-networking-service.md
alegacy: Ok.  Will change.

design/ironic-standalone-networking-service.md
alegacy: I won't lie... I think I put this in out of habit and don't really have a clear idea here.   Was probably thinking we could possibly try accessing the switch to test the credentials, but might be overkill in a first implementation.

design/ironic-standalone-networking-service.md
alegacy: I'll change the JSON to `nativeVLAN`

design/ironic-standalone-networking-service.md
alegacy: Actually, I meant to remove this and do the reference checking from the BMH side instead.

design/ironic-standalone-networking-service.md
alegacy: Good point.

design/ironic-standalone-networking-service.md
alegacy: Ok, maybe we need to discuss this a bit then.... the reason I did it this way is that my assumption is that BMHs may be organized in groups and put into namespaces according to those groups (whatever they may be), but then someone may end up picking up BMHs from different groups to build a cluster and that cluster would have a set of NetworkAttachments that should be applied to each of those BMHs. ... the attachments and BMH wouldn't necessarily all be in the same namespace.

design/ironic-standalone-networking-service.md
alegacy: Good point.

design/ironic-standalone-networking-service.md
alegacy: Ok, no problem.

design/ironic-standalone-networking-service.md
alegacy: I had code in my Ironic review to monitor the file and dynamically load the drivers, but was asked to take it out for the initial pass and maybe add it back in later if needed.   That would be my preference as I don't like restarting services to reload stuff, but, as the outsider, don't want to break the existing model that already exists.

design/ironic-standalone-networking-service.md
alegacy: A user providing the LLDP information is highly problematic.  That will definitely lead to user errors and incorrect switch ports being updated and therefore unintentional outages!   

Will add a further clarification about LLDP/inspection.

design/ironic-standalone-networking-service.md
alegacy: Ok, will add.

design/ironic-standalone-networking-service.md
alegacy: Will change.

design/ironic-standalone-networking-service.md
alegacy: Understood.

design/ironic-standalone-networking-service.md
alegacy: Was just trying to follow the existing pattern.  Of course, if it is better/easier/simpler to  just do a separate Pod then I'm open to that.

design/ironic-standalone-networking-service.md
alegacy: Will add.

design/ironic-standalone-networking-service.md
alegacy: ack

design/ironic-standalone-networking-service.md
alegacy: Ya, I think it is best in the `HardwareData` object.

design/ironic-standalone-networking-service.md
alegacy: Agreed... LLDP info should go into `HardwareData` after inspection.

design/ironic-standalone-networking-service.md
alegacy: I _think_ that's ok... with my latest revision in Ironic setting the port's VLAN back to the "idle" VLAN should happen naturally without any extra steps so when transitioning into a state where delete would be accepted the switch port should already be ok.

design/ironic-standalone-networking-service.md
alegacy: My thinking was this...  In the OpenShift case at least, all operations are always done on the final VLAN (i.e., the "tenant" VLAN) since ToR configuration is done manually ahead of time.  Therefore, it seemed reasonable to assume that everything except for "inspection" could be done on the tenant VLAN... inspection being the exception here since we need hosts to sit on the "idle" VLAN prior to being deployed so that initial inspection always works.

Today, I have coded the Ironic standalone-experimental driver such that any provider networks that doesn't have a specific VLAN configured in the config file we fall back to the tenant VLAN (i.e., the VLAN in the `port.extra.switchport` attribute).  Therefore if we only set the idle_network, and inspection_network then all other networks will default to the tenant network and things should work as they do today (except for inspection which would happen on the new "idle" network).

Recall that demo I did a few weeks ago... you pointed out that the final tenant VLAN seemed to get configured too early in the provisioning phase and that it should not happen until the final deploy step.   Well, I tried to address that by setting all of the provider networks to the same as the idle network.  That didn't work because in OpenShift the assisted installer image is first booted before the final image is deployed.  That image ran a DHCP on all interfaces and noticed that all interfaces got an IP address on the same network and complained and refused to finish the install.   So I reverted back to this approach of setting inspection=idle,  and then all of the rest to the tenant VLAN and that worked.  

Perhaps another discussion is in order here.

alegacy:
@dtantsur I've addressed the minor stuff... will need to discuss some of the larger items next week.  Thank you for your review!

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,739 @@
...
+    Name string `json:"name"`
+
+    // Namespace of the HostNetworkAttachment (defaults to BMH namespace)
+    Namespace string `json:"namespace,omitempty"`
+}
+```
+
+### Ironic Configuration
+
+When `Ironic.NetworkingService.Enabled` is true, Metal³ will configure
```
diconico07, line 434: This would only be valid for a deployment using IrSO, there should also be a clear way to deploy without IrSO I believe.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,739 @@
...
+
+6. **Container Deployment**: The Ironic Networking service will run as a
+   separate container in the metal3 pod, alongside the existing metal3-ironic
+   container. It will share the `/etc/ironic` volume for configuration files
+   and the htpasswd file.
+
+7. **Network Driver Selection**: The Baremetal Operator will automatically set
+   the Ironic node's network interface driver to `standalone-experimental`
+   when:
+   - `enabled` is true in the `Ironic.NetworkingService.Enabled` resource
```
diconico07: The Ironic resource is not managed by the BMO, and BMO shouldn't rely on its presence, so will require a configuration option for BMO to enable that behavior rather than using the Ironic resource state here.

kashifest:
Overall I don't have any objection any idea. Rather I think its a welcome addition. I would need time to a deep dive on the technical aspect of how the proposed functionality would land in Metal3.

design/ironic-standalone-networking-service.md
dtantsur: These assumptions are very-very different from the assumptions Ironic has about networking.

The strong expectation is that this feature is used to establish a security boundary between the provisioning network (where insecure PXE traffic happens and where IPA is running) and tenant networks (with also potentially insecure workflows), as well as between tenant networks. You're excluding the first part, which the Ironic community sees as pretty important.

In fact, there is nothing special about inspection network from the Ironic perspective. With fast track (which we need to make work eventually, remember), inspection flows into cleaning, which flows into provisioning. The security properties of all 3 networks is the same, and so is the desire to not mix the PXE/IPA/whatever traffic with the customer's workloads.

Assisted installer is, yes, a different topic. We need to make it work on the "service" networks, especially since inspection can be both enabled and disabled, and fast track basically means that the assisted agent may end up running on the inspection network.

design/ironic-standalone-networking-service.md
dtantsur: Our existing patterns are sometimes problematic and reflect the past desire to do things quickly :)

design/ironic-standalone-networking-service.md
alegacy: Ok, I'll schedule a call with you to level set on some of these concepts.

design/ironic-standalone-networking-service.md
alegacy: Yes, sorry, I meant to reword this to indicate that BMO should accept an environment variable (e.g., IRONIC_NETWORKING_ENABLED) to control this behaviour -- defaulted to false of course.  I'll revise.

design/ironic-standalone-networking-service.md
alegacy: My apologies.  I understood IrSO to be the primary deployment method.   Could you point me to some documentation that describes the existing non-IrSO deployments?

alegacy:
> It is an overall nice feature coming, I have some concern about how it's tightly tied into IrSO by design here.

Thanks @diconico07 ... I'll review the user-guide and amend the document with information about non-IrSO scenarios.

design/ironic-standalone-networking-service.md
dtantsur: We're moving towards IrSO being the primary way of deployment. There is nothing wrong with other people using other means (in fact, there have been requests to bring the Suse's charts upstream), but Allain is right in defaulting to describing the IrSO approach.

diconico07: True, I missed the change in the docs to phase out other ways of deployments (last time I looked at that part all method were still equally listed).

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,739 @@
...
+
+As a Metal³ operator, I want to define my ToR switch configurations once in
+Kubernetes CRDs so that Metal³ can automatically manage switch port
+configurations without manual intervention.
+
+The operator creates `IronicSwitch` resources for each ToR switch:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: IronicSwitch
```
dtantsur: In the spirit of moving the switch handling (back) to BMO, this one should probably be called BareMetalSwitch (like BareMetalHost).

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,739 @@
...
+metadata:
+  name: rack1-tor1
+  namespace: default-ironic-namespace
+spec:
+  driverType: generic-switch
+  deviceType: dell_os10
```
dtantsur, line 92: It may be a ridiculous request.. but when watching your videos, I keep getting confused about driverType vs deviceType, they look the same to me. Maybe s/driverType/driver/ ? or is it just me?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,739 @@
...
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: Ironic
+metadata:
+  name: default
+  namespace: default-ironic-namespace
+spec:
+  networkingService:
+    enabled: true
+    networkDriver: standalone-experimental
```
dtantsur: I'm a bit confused about the meaning of this field. Isn't it the Node's network interface? Why is it a part of the networking service configuration?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,739 @@
...
+    Type SwitchCredentialType `json:"type"`
+
+    // The name of the secret containing the switch credentials.
+    // For password authentication, requires key "password".
+    // For SSH key authentication, requires keys "username", "privateKey", and optionally "publicKey".
+    SecretName string `json:"secretName"`
+}
+
+// IronicSwitchSpec defines the desired state of IronicSwitch.
+type IronicSwitchSpec struct {
```
dtantsur: See above about naming and deviceType vs driverType

design/ironic-standalone-networking-service.md
dtantsur: To recap our recent discussion. We don't actually need Ironic to monitor this file, it's enough that ironic-networking supports SIGHUP to restart its worker. We can use the existing pattern in ironic-image to reload services on secret changes.

Then we can decouple switch configuration from IrSO. IrSO will only pre-create a secret and mount it. BMH will get the secret's name from its configuration and write the configuration into it as BareMetalSwitch resources are changed. As an aside, it's probably easier to support several files rather than just one: then BMO can modify only one key of the secret each time instead of rewriting the whole secret. But it's not a blocker.

The benefits are twofold. On one hand, we don't introduce a hard dependency on IrSO for the whole feature. On the other - by logically grouping BMH and BMS in the same operator, we make it easier to develop this feature in the future when we hopefully get proper API for switches. It's also less surprising (hopefully) for future contributors, even if the approach itself does feel slightly hacky.

design/ironic-standalone-networking-service.md
dtantsur: Yeah, I agree.

Although BMO does support IrSO optionally, so when the IrSO integration is enabled, we can indeed rely on this field.

design/ironic-standalone-networking-service.md
dtantsur: Could you summarize what we decided back then?

design/ironic-standalone-networking-service.md
alegacy: Ok, will update the proposal with this new approach going forward.  Thanks.

design/ironic-standalone-networking-service.md
alegacy: Ok, will switch to having ironic-networking into its own Pod.

design/ironic-standalone-networking-service.md
alegacy: Ya, I agree... maybe `driver:` and `deviceType:` would be more appropriate.

design/ironic-standalone-networking-service.md
alegacy: This is meant to control which network interface to set as the default network interface in Ironic's config file (passed down to ironic-image via an env variable).

design/ironic-standalone-networking-service.md
alegacy: Agreed

design/ironic-standalone-networking-service.md
dtantsur: Ah, cool, then let's make the intention clear. Maybe something like `useByDefault: true`?

But BMO should have the final say, I guess.

design/ironic-standalone-networking-service.md
alegacy: My recollection is that the ultimate goal is that the tenant network should only be applied as the very last step before the final o/s image is booted which implies that we should use this 'service' network in all provider networks ideally.  

But we know that there are issues with how some of our OpenShift usecases operate in this respect so some care will be needed here in the short-term.  Perhaps a compromise would be to expose all of the provider networks here and let the user decide which ones to map to which VLAN that way we could better manage our OpenShift concerns by explicitly setting the network patterns that work there.

design/ironic-standalone-networking-service.md
alegacy: how about i rename it to `defaultNetworkInterface`? ...and possibly add `IRONIC_NETWORK_INTERFACE`  to BMO?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,772 @@
...
+    // +kubebuilder:validation:Maximum=65535
+    // +optional
+    Port *int32 `json:"port,omitempty"`
+
+    // DisableCertificateVerification disables verification of server 
+    // certificates when using HTTPS to connect to the switch. This is 
+    // required when the server certificate is self-signed, but is insecure 
+    // because it allows a man-in-the-middle to intercept the connection.
+    // +optional 
+    DisableCertificateVerification *bool `json:"disableCertificateVerification,omitempty"`
```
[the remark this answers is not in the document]
alegacy: `DisableCertificateVerification` is useful for lab deployments, but not recommended for production use cases.

Rozzii:
I was on vacation for 3 weeks so I was not actively reviewing, back now.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,771 @@
...
+
+This design proposes adding support for the Ironic Networking service to
+Metal³, enabling automatic configuration of Top-of-Rack (ToR) switch ports
+during the lifecycle of bare metal nodes. The Ironic Networking service
+provides a pluggable framework for managing switch port configurations such as
+VLAN assignments, port modes (access/trunk/hybrid), and LAG settings.
+This integration will allow Metal³ users to define network configurations
+declaratively through Kubernetes CRDs, which will be automatically applied to
+physical switch ports as nodes are provisioned and deprovisioned.  Switch ports
+will be identified using LLDP information collected during initial inspection
```
Rozzii, line 24: In the document there is no mention of providing a "manual" option to select the "switch interfaces" on the machine under management.

I would like you to please include a workflow in your design that allows the user to set designate the switch interfaces "manually" to make the feature viable in environments where LLDP is not available.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,771 @@
...
+      name: machine-network-vlan-100
+  - macAddress: "aa:bb:cc:dd:ee:02"
+    networkAttachment:
+      namespace: some-cluster-namespace
+      name: data-network-vlans-200-202
+```
+
+When the host is provisioned, Metal³ automatically configures the corresponding
+switch port for the primary interface to access mode on VLAN 100, and the
+secondary interface to trunk mode with tagged VLANs 200-202.
```
Rozzii, line 176: My question relates to the practicalities of the interface configuration steps on the specific machine under management. 

When does the VLAN "attachment" happen? In the current code a BMH becomes "provisioned" after Ironic finishes with all "deployment steps" and initiates a reboot. When the machine boots into the "node image" or "operational OS" or "disk image" the BMH is already provisioned and it has rebooted and Systemd will start the cloud-init services and among other things run the cloud init network configuration service. Is there any possibility that the the existing network management workflow collides with the Ironic networking? 

Currently we  utilize the cloud-init openstack network-data file that is part of the Ironic node's config drive. Does the proposed ironic networking service also utilizes the same network-data file on the config drive or does it have some other mechanism to configure the interfaces?

design/ironic-standalone-networking-service.md
alegacy: Did you already have something in mind for where such information would be added to the BMH configuration?  I could add it as an optional attribute in the `networkInterfaces` list of the BMH under an attribute like `switchport`.  e.g.:

```yaml
kind: BareMetalHost
metadata:
  name: worker-1
  namespace: some-host-namespace
spec:
  # ... other fields ...
  networkInterfaces:
  - macAddress: "aa:bb:cc:dd:ee:01"
    networkAttachment:
      namespace: some-cluster-namespace
      name: machine-network-vlan-100
    switchport:
      switchId: "00:11:22:33:44:55"
      portId: "ethernet1/2/3"
  - macAddress: "aa:bb:cc:dd:ee:02"
    networkAttachment:
      namespace: some-cluster-namespace
      name: data-network-vlans-200-202
    switchport:
      switchId: "00:11:22:33:44:55"
      portId: "ethernet4/5/6"
```

design/ironic-standalone-networking-service.md
alegacy: > When does the VLAN "attachment" happen?

The switchport configuration would happen just before the reboot into the final image.  Therefore, the switchport should be configured properly and whatever interface configuration was passed to the image via cloud-init should work properly.

> Does the proposed ironic networking service also utilizes the same network-data file on the config drive or does it have some other mechanism to configure the interfaces?

ironic-networking only manages the switchport configuration on the switch.  If you are already passing network data into the node's config drive that will remain the same.   The ironic-networking driver can synthesize a network-data file if one was not provided but won't override it if it already exists.

@dtantsur keep me honest here.

design/ironic-standalone-networking-service.md
Rozzii: Your example above looks acceptable to me on the first glance. I would like to ask @s3rj1k to please take a look here as he is using and implementing quite advanced "network management" so he might notices edge cases that I don't.

design/ironic-standalone-networking-service.md
Rozzii: So if I understand correctly, let's say if I want one of the interfaces on the target machine be part of a specific tagged VLAN, the proposed feature will only configure the switch, the user will be required to provide a network-data file via the existing workflow to BMH with the machine network interface configuration. 

If my understanding is correct then all is good, my question has been answered.

Rozzii (commented):
I have a feeling that this feature will go through a few changes in the future especially when multi tenancy features will also be implemented. In any case as an initial proposal I am quite satisfied, after my two ongoing discussions are resolved, I will lgtm this proposal.

design/ironic-standalone-networking-service.md
alegacy: yes, that's the case.  ...and if the user doesn't provide that network-data file then the Ironic network interface _should_ be able to synthesize it automatically.

alegacy:
> I have a feeling that this feature will go through a few changes in the future especially when multi tenancy features will also be implemented. In any case as an initial proposal I am quite satisfied, after my two ongoing discussions are resolved, I will lgtm this proposal.

@Rozzii Thanks.  Could you maybe elaborate a little on which multi tenancy features you envision in this context?

dtantsur:
@alegacy Adam might have in mind future integration with HostClaims API, which decouple users and admins, in a sense: https://github.com/metal3-io/metal3-docs/blob/main/design/hostclaim-multitenancy.md.

Once we get there, we'll need to figure out what information is admin-only, what users can see and what users should be able to modify. Ideally, this proposal should be designed in a way that will not make it too complex in the future.

design/ironic-standalone-networking-service.md
dtantsur: Yeah, network data is software configuration, here we're talking about hardware (switches) configuration.

design/ironic-standalone-networking-service.md
s3rj1k: I wonder if we actually want this in BMH object as opposed to having a dedicated object and setting Claim/OwnerReference ?

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,771 @@
...
+    NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
+}
+
+// NetworkInterface defines the network configuration for a specific interface.
+type NetworkInterface struct {
+    // Name of the network interface (e.g., "eth0", "ens1f0")
+    // This must match the name of a NIC in status.hardware.nics.
+    // If MACAddress is set then this field is ignored.
+    // +optional
+    Name string `json:"name"`
```
s3rj1k: What if I want to match based on MAC only?

Possible case if unstable NIC naming

alegacy: If `macAddress` is set it takes precedence so that should satisfy your use case.

design/ironic-standalone-networking-service.md
alegacy: I assume you mean breaking out `networkInterfaces` as a separate object.  What do you see as the main advantage of that?

design/ironic-standalone-networking-service.md
s3rj1k: makes sense, we should in that case set Name with `omitempty` and do validation in admission I suppose

design/ironic-standalone-networking-service.md
s3rj1k: > I assume you mean breaking out networkInterfaces as a separate object.

Yes

Just thinking here, we already have a flow in BMH where we link `cloud-init` network data as external secret and already have `hardwareData` as external object, it kind of feels doing this is in the same spirit.

Do mind I don't have strong opinion here as `hardwareData` migration over the years kind of still in the process and not finished.

alegacy: Ok, I'll let others chime in here as well.  Since the `networkInterfaces` list already references the `HostNetworkAttachment` object, breaking it out would create a 2nd level of dereference which would increase complexity.  That pattern may be different enough from the 2 you mentioned.   Something to consider.

design/ironic-standalone-networking-service.md
alegacy: updated

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+  networkInterfaces:
+  - macAddress: "aa:bb:cc:dd:ee:01"
+    networkAttachment:
+      namespace: some-cluster-namespace
+      name: machine-network-vlan-100
+    switchport:
+      switchId: "00:11:22:33:44:55"
+      portId: "ethernet1/2/3"
+  - macAddress: "aa:bb:cc:dd:ee:02"
+    networkAttachment:
```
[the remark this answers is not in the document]
alegacy: resolved.  `networkAttachmentRef` -> `networkAttachment`

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+    Name string `json:"name,omitempty"`
+
+    // MAC address of the network interface.  This must match the MAC address
+    // of the network interface in status.hardware.nics.
+    // +optional
+    // +kubebuilder:validation:Pattern=`[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}`
+    MACAddress string `json:"macAddress,omitempty"`
+
+    // NetworkAttachment references the HostNetworkAttachment for this interface
+    NetworkAttachment *NetworkAttachmentRef `json:"networkAttachmentRef"`
```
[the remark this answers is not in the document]
alegacy: It is not optional.  `networkAttachmentRef` -> `networkAttachment`

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+4. Extend the `BareMetalHost` CRD with `networkInterfaces` field
+5. Implement switch configuration ConfigMap and SSH key file Secret generation
+   in the BareMetalSwitch controller
+6. Implement Ironic Networking service Pod deployment
+7. Implement ironic-image configuration updates for networking support
+8. Implement Ironic Port `extra.switchport` population in the BMH controller
+9. Implement validation logic for network interface configurations
+10. Add status reporting for network configuration state
+11. Implement re-population of Ironic Node/Port information if database is
+    is not persistent
```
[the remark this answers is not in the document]
alegacy: fixed

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+**Upgrade**:
+
+- The feature is opt-in via `Ironic.NetworkingService.Enabled` flag, so
+  existing deployments are unaffected
+
+- Users can enable the feature by:
+  1. Updating to a Metal³ version with Ironic Networking support
+  2. Creating `BareMetalSwitch` resources for their switches
+  3. Setting `Ironic.NetworkingService.Enabled: true` in the Ironic resource
+  4. Defining `Ironic.NetworkingService.ServiceNetwork` configuration
```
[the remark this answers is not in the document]
alegacy: fixed

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+
+1. **New CRDs**: `BareMetalSwitch` and `HostNetworkAttachment` to define
+   switch configurations and network interface configurations respectively
+2. **Enhanced Ironic CRD**: Add service enablement flag and service
+   network configuration to the `ironic.metal3.io` CRD
+3. **Enhanced BareMetalHost CRD**: Add `networkInterfaces` field to reference
+   network configurations for each network interface
+4. **Ironic Networking Pod**: Deploy the Ironic Networking service
+   as a separate Pod alongside the existing main Ironic Pod.
+5. **Share ConfigMap/Secret**: To decouple IrSO from the management of
```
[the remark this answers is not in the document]
alegacy: fixed

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+  deviceType: dell_os10
+  address: 192.168.1.10
+  credentials:
+    type: "password"
+    secretName: switch-credentials
+  macAddress: 3e:c5:c0:cb:fa:f6
+```
+
+BMO reconciles the contents of all `BareMetalSwitch` instance and builds a
+config file which is stores into a ConfigMap that is mounted to the
```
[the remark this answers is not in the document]
alegacy: fixed

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+address = <ADDRESS>
+username = <USERNAME>
+password = <PASSWORD>
+enable_secret = <ENABLE_SECRET>
+mac_address = <MAC_ADDRESS>
+```
+
+This file will be stored in a Secret and mounted to the Ironic Networking
+container. When any `BareMetalSwitch` is created, updated, or deleted, the
+Secret will be regenerated and the Ironic Networking service will be restarted.
```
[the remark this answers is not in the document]
alegacy: Fixed.  Both should have said `Secret`.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+
+BMO reconciles the contents of all `BareMetalSwitch` instance and builds a
+config file which is stores into a ConfigMap that is mounted to the
+ironic-networking Pod by IrSO opaquely. Similarly, any SSH key credentials are
+aggregated into a Secret that is mounted to the ironic-networking Pod by
+IrSO.  The ironic-networking container is responsible for reading the
+config files and reacting to any changes therein.
+
+The credentials Secret is expected to contain keys for the `username`,
+`password`, and (optionally) `enable-secret`.  It may also contain an SSH
```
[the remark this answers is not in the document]
alegacy: Fixed.  Both should have said `Secret`

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+    Mode PortMode `json:"mode"`
+
+    // NativeVLAN specifies the native VLAN ID
+    NativeVLAN int `json:"nativeVLAN"`
+
+    // AllowedVLANs specifies the list of allowed VLANs for trunk/hybrid modes
+    AllowedVLANs []int `json:"allowedVLANs,omitempty"`
+}
+
+type NetworkingService {
```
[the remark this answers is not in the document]
alegacy: fixed

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+Add the following fields to the `IronicSpec`:
+
+```go
+type ProviderNetworkConfig struct {
+    // Type specifies which provider network this configuration applies to.
+    // +kubebuilder:validation:Enum=idle;inspection;cleaning;rescuing;servicing;provisioning
+    Type string `json:"type"`
+
+    // Mode specifies the switch port mode for a provider network
+    Mode PortMode `json:"mode"`
```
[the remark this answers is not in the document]
alegacy: fixed

alegacy:
Hi @Rozzii ... have you had a minute to re-review this?

Rozzii:
/lgtm
From my side it is a accepted!

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+- Managing switch infrastructure itself (firmware updates, switch provisioning,
+  etc.)
+- Supporting dynamic network reconfiguration after node deployment (initial
+  scope is provisioning-time configuration only)
+
+## Proposal
+
+This proposal introduces these main components:
+
+1. **New CRDs**: `BareMetalSwitch` and `HostNetworkAttachment` to define
```
tuminoid, line 61: We are now enforcing 1.1.1 style numbering.

alegacy: done

tuminoid:
/unhold
Thank you, and sorry for inconvenience.

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+```
+
+When the host is provisioned, Metal³ automatically configures the corresponding
+switch port for the primary interface to access mode on VLAN 100, and the
+secondary interface to trunk mode with tagged VLANs 200-202.
+
+Ideally, `HostNetworkAttachment` instances should only be referenced by
+`BareMetalHost` instances after initial inspection so that NIC information is
+present in the `HardwareData` object and available for validation of the
+network interface info by a validation webhook. If inspection has not yet run
```
dtantsur, line 181: Side note: I doubt you can make a webhook access a linked resource anyway

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+    // NetworkDriver sets the default network interface in the Ironic
+    // configuration file.  This defaults to "ironic-networking".
+    // +optional
+    // +kubebuilder:default=ironic-networking
+    NetworkDriver string `json:"networkDriver,omitempty"`
+
+    // RPCPort is the internal RPC port used for Ironic Networking
+    // Only change this if the default value causes a conflict on your deployment
+    // +optional
+    RPCPort *int `json:"rpcPort,omitempty"`
```
dtantsur, line 365: nit: we'll need to see if we put the port here or with the other ports in `Networking` (I suspect the latter)

design/ironic-standalone-networking-service.md
```diff
@@ -0,0 +1,776 @@
...
+
+**Risk**: If Metal³'s instantiation of Ironic is not configured to use
+persistent storage for the Ironic database, and the Ironic Pod is restarted
+then all node and port information is lost, including the LLDP information
+and switch port configuration derived from the BMH configuration.
+
+**Mitigation**: The BMH reconciler can be modified to restore the last known
+set of Port objects including their `port.extra.switchport` and
+`port.local_link_connection` attributes using information from the
+`HardwareData` object.
```
dtantsur, line 598: I'd like a bit more details on this one in a follow-up tracker (a bug or an update to this spec). I guess we could cache this information in `HardwareData.Status` (which is currently empty).

PR #585, opened 2025-10-14
Quick-start update

lentzi90:
The guide has been broken by upstream changes, and is a bit outdated
anyway. This is a quite substantial refresh of it that also introduces IrSO.

- Bump k8s to v1.34.1
- Deploy Ironic using IrSO
- Switch to UEFI boot mode

Fixes #567

PR #587, opened 2025-10-16
Add embed tool for embedding files in markdown

lentzi90:
This adds a plugin to mdbook for embedding files from any repo on github.
My plan is to use this when refreshing the quick-start guide. To see how it works, check this WIP PR: https://github.com/metal3-io/metal3-docs/pull/585

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,14 +12,15 @@ MDBOOK_BIN := $(TOOLS_BIN_DIR)/mdbook
 export PATH := $(PATH):$(TOOLS_BIN_DIR)
 
 ## ------------------------------------
-## Resolve placeholders as tags
+## mdbook plugins
 ## ------------------------------------
 RELEASETAGS := $(TOOLS_BIN_DIR)/mdbook-releasetags
 $(RELEASETAGS): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go build -tags=tools -o $(TOOLS_BIN_DIR)/mdbook-releasetags ./releasetags
 
-.PHONY: releasetags
-releasetags: $(RELEASETAGS)
+MDBOOK_EMBED := $(TOOLS_BIN_DIR)/mdbook-embed
+$(MDBOOK_EMBED): $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR); go build -tags=tools -o $(TOOLS_BIN_DIR)/mdbook-embed sigs.k8s.io/cluster-api/hack/tools/mdbook/embed
 
 ## ------------------------------------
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -12,14 +12,15 @@ MDBOOK_BIN := $(TOOLS_BIN_DIR)/mdbook
...
 ## ------------------------------------
-## Resolve placeholders as tags
+## mdbook plugins
 ## ------------------------------------
 RELEASETAGS := $(TOOLS_BIN_DIR)/mdbook-releasetags
 $(RELEASETAGS): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go build -tags=tools -o $(TOOLS_BIN_DIR)/mdbook-releasetags ./releasetags
 
-.PHONY: releasetags
-releasetags: $(RELEASETAGS)
```
lentzi90, line 22: I removed the phony for releasetags since it seemed unnecessary to me. We were not using it anywhere so the only thing it could be used for is to manually build the binary releasetags.

lentzi90:
/cc @peppi-lotta
This should be useful when creating tests for the quick-start. We can embed whatever manifests we use for the tests directly in the guide. That way we always keep them in sync :slightly_smiling_face:

hack/tools/tools.go
tuminoid, line 1: Hmm I thought go 1.24 did not need these anymore?  https://go.dev/doc/go1.24  

Could we try that approach here?

lentzi90: Good point, I did not think about that!

lentzi90:
/hold
I'll try to use go tools instead as @tuminoid suggested

lentzi90:
I got rid of the `tools.go` file by using the `tool` directive in the go.mod. The rest is still the same. If we only used the `embed` tool directly, we could run it with `go tool embed`, but since we rely on netlify for actually building the site, I thought it was nicer to build it with `go build` the same way we do for the internal `releasetags` tool.
/hold cancel

tuminoid:
All the tests are pending since yesterday...  I'll try re-trigger by editing description.

tuminoid:
> All the tests are pending since yesterday... I'll try re-trigger by editing description.

Nope... maybe repush, @lentzi90 ?

lentzi90:
I'm not sure if they matter though? If you click on them everything is already done and green. Github is just not aware of it (after the AWS outage yesterday). But I can push to make sure

tuminoid:
> I'm not sure if they matter though? If you click on them everything is already done and green. Github is just not aware of it (after the AWS outage yesterday). But I can push to make sure

Ah ok, I thought they never completed.
/lgtm

PR #594, opened 2025-11-05
add AGENTS.md

tuminoid:
Add AGENTS.md to allow AI assistants to be more effective, especially the Code Review Bot.

Note: do not create .github/copilot-instructions.md as it will prevent current version of Copilot CLI from reading AGENTS.md at all.

CLAUDE.md is added as symlink as Cladue STILL does not support AGENTS.md. It will be removed when Claude does. All the other CLIs do support it already.

### DIFF
--- a/.cspell-config.json
+++ b/.cspell-config.json
@@ -215,6 +215,7 @@
     "sdk",
     "service.d",
     "sh",
+    "shellcheck",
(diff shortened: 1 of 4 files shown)


lentzi90 (commented):
/lgtm

PR #595, opened 2025-11-13
Troubleshooting FAQ: CDC devices exposed by some BMCs

dtantsur:

### DIFF
--- a/docs/user-guide/src/troubleshooting.md
+++ b/docs/user-guide/src/troubleshooting.md
@@ -159,6 +159,28 @@ You have options:
 - Update your BareMetalHost definition with the `bootMode` value that matches
   what your hardware supports.
 
+### Duplicate network interfaces
+
+If inspection succeeds on a subset of the hosts and fails for the rest, and you
+see the following message in the Ironic logs:
+
+```text
+DuplicateNodeOnLookup: Multiple nodes with port addresses [...] were found
+```
+
+there may be at least two different reasons for that:
+
+1. One host has `bootMacAddress` from a wrong host.
+1. Hosts have virtual network devices with a hardcoded MAC address (the same
+for all of them).
+
+The latter is a common case with virtual devices provided by BMCs. Such devices
+are represented to the operating system similarly to USB ethernet dongles. They
+use drivers like `cdc_ether` or `cdc_ncm`, and names like "HPE Virtual NIC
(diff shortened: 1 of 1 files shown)


docs/user-guide/src/troubleshooting.md
```diff
@@ -159,6 +159,28 @@ You have options:
...
+see the following message in the Ironic logs:
+
+```text
+DuplicateNodeOnLookup: Multiple nodes with port addresses [...] were found
+```
+
+there may be at least two different reasons for that:
+
+1. One host has `bootMacAddress` from a wrong host.
+2. Hosts have virtual network devices with a hardcoded MAC address (the same
```
tuminoid:
```suggestion
1. Hosts have virtual network devices with a hardcoded MAC address (the same
```

tuminoid (commented):
Nit, otherwise LGTM.

docs/user-guide/src/troubleshooting.md
dtantsur: I know that markdown will work this way too, but isn't it nicer when markdown can be read without rendering them?

tuminoid: "1. 1." style formatting allows you just changing the lines you need, instead of re-numbering the whole list if you add something in middle, and it gets rendered nicely. Con is that if you read it raw, you need to render in your head or render in your editor for numbers.

Either way is fine, but I think majority of docs is 1.1. style, hence the suggestion.

tuminoid (commented):
/lgtm

docs/user-guide/src/troubleshooting.md
```diff
@@ -159,6 +159,28 @@ You have options:
...
+If inspection succeeds on a subset of the hosts and fails for the rest, and you
+see the following message in the Ironic logs:
+
+```text
+DuplicateNodeOnLookup: Multiple nodes with port addresses [...] were found
+```
+
+there may be at least two different reasons for that:
+
+1. One host has `bootMacAddress` from a wrong host.
```
kashifest, line 173: This line is a bit difficult to understand for me. Do you mean one host has  a wrong `bootMacAddress` which is incorrect and matching with another host?

kashifest: ping @dtantsur

tuminoid:
@dtantsur should we reopen and finish?

PR #596, opened 2025-11-14
Add tip to speed up release process to releasing doc

Sunnatillo:
This PR adds tip to speed up release process to releasing doc. 

Fixing issues identified with hack/verify-release.sh script would really speed up release process.

### DIFF
--- a/processes/releasing.md
+++ b/processes/releasing.md
@@ -8,6 +8,12 @@ the post-release actions required to add new release branches to the CI.
 High-level release process for a major/minor release. For patch releases, only
 release the relevant versions.
 
+**TIP**: To speed up the release process, run the `hack/verify-release.sh`
+script on all release branches of all repos (and on the main branch for
+pre-releases). The script shows if any modules need updating or if there are
+CVE fixes required. Open PRs to fix these issues in parallel to make the
+release process faster.
+
 **NOTE**: [Ironic-image](https://github.com/metal3-io/ironic-image) (via
 upstream Ironic) and
 [IrSO](https://github.com/metal3-io/ironic-standalone-operator) (linked to


Sunnatillo:
> This is already part of each repositorys `docs/releasing.md`, not sure if it needs to be duplicated here?

`docs/releasing.md` is for repo specific. I followed that. For each repo I had to open and wait few PRs until release proceeds. The idea here to check each release branch in each repo at the beginning of release process.

tuminoid:
> > This is already part of each repositorys `docs/releasing.md`, not sure if it needs to be duplicated here?
> 
> `docs/releasing.md` is for repo specific. I followed that. For each repo I had to open and wait few PRs until release proceeds. The idea here to check each release branch in each repo at the beginning of release process.

I don't see the benefit as its a pre-check for before cutting anything, same as this doc,, but if you feel this benefits someone, then lets merge it.

processes/releasing.md
```diff
@@ -8,6 +8,12 @@ the post-release actions required to add new release branches to the CI.
 High-level release process for a major/minor release. For patch releases, only
 release the relevant versions.
 
+**TIP**: To speed up the release process, run the `hack/verify-release.sh`
```
tuminoid, line 11: The more you edit this, the more I dislike it. The point of `verify-release.sh` is to make sure something wasn't missed. Its not supposed to be "part" of release process to bump the CVEs etc. All those should've happened as part of normal development process, and the script is just safeguard for verification of that work.

Sunnatillo:
> > > This is already part of each repositorys `docs/releasing.md`, not sure if it needs to be duplicated here?
> > 
> > 
> > `docs/releasing.md` is for repo specific. I followed that. For each repo I had to open and wait few PRs until release proceeds. The idea here to check each release branch in each repo at the beginning of release process.
> 
> I don't see the benefit as its a pre-check for before cutting anything, same as this doc,, but if you feel this benefits someone, then lets merge it.

I have updated slightly to clarify my point about speeding up the process.

processes/releasing.md
Sunnatillo: This paragraph is giving the same idea as here:
https://github.com/metal3-io/baremetal-operator/blob/230a562f7e3b09299f60fc4c9c684b0210707149/docs/releasing.md?plain=1#L18

PR #599, opened 2025-11-25
v1beta1: plan to remove BareMetalHost.Spec.Firmware

dtantsur:
This field has only been supported by the iLO and iRMC drivers, and both
are now deprecated. Because of vast vendor differences, it's impractical
to implement the field for Redfish, so it will be removed in favour of
HostFirmwareSettings.


MahnoorAsghar:
lgtm

PR #571, opened 2025-09-11
Add `fromPoolAnnotation` field for IPPool reference via annotation

s3rj1k:
Adds `fromPoolAnnotation` field to enable IPPool reference through annotations, providing an alternative method for IP address pool assignment in `networkData`.

PR #625, opened 2025-12-03
Add docs for FromPoolAnnotation based IPPools

s3rj1k:
related: https://github.com/metal3-io/metal3-docs/pull/571/files#diff-0dbc75ee9b90ce043e2c9ad2deaf45a8e3624b7611f6970b1c848d3ffff8a10c

### DIFF
--- a/docs/user-guide/src/capm3/annotation_based_ippool.md
+++ b/docs/user-guide/src/capm3/annotation_based_ippool.md
@@ -1,0 +1,98 @@
+# Annotation-based IPPool Reference
+
+The `fromPoolAnnotation` field of a `Metal3DataTemplate` enables dynamic
+IPPool name resolution from annotations on BareMetalHost, Machine, or
+Metal3Machine objects. This allows flexible IPPool assignment
+based on deployment-time decisions rather than static
+template configuration.
+
+## Use Cases
+
+The `fromPoolAnnotation` field can be used in both IPv4 and IPv6 network
+configurations:
+
+- Network configuration to specify which IPPool to use for IP address allocation
+- Gateway configuration to specify which IPPool to use for gateway IP resolution
+- Route gateway configuration to specify which IPPool to use for route gateway
+  IP resolution
+
+## Field Specification
+
+The `fromPoolAnnotation` field contains:
+
+- **object**: The object type to read the annotation from (`baremetalhost`,
+  `machine`, `metal3machine`)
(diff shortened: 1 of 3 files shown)


s3rj1k:
@kashifest can you please proof-read this?

docs/user-guide/src/capm3/annotation_based_ippool.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Field Specification
+
+The `fromPoolAnnotation` field contains:
+
+- **object**: The object type to read the annotation from (`baremetalhost`,
+  `machine`, `metal3machine`)
+- **annotation**: The annotation key containing the IPPool name
+
+> **Note:** When `fromPoolAnnotation` is set, `fromIPPool` and `fromPoolRef`
```
kashifest, line 27: Shall we mention briefly also that when not set , whats the workflow so that users are aware and can choose which way to go?

s3rj1k: It might openup a full documentation story of other ways to use fromAnnotation based features, I don't really want to do this in single PR, like there are no extended documentation for other fromAnnotation features

kashifest: no no, my idea is not to write all those in detail, but something like if this is not used this and this default workflow will be the way to go

s3rj1k: Something like this?
https://github.com/k0rdent-oot/metal3-docs/commit/ff9b9e579274275916e68c7f2ee77d15c7af088f

kashifest: Yes thanks, I am happy with this.

kashifest:
@lentzi90 @adilGhaffarDev can we take this doc update in?

docs/user-guide/src/capm3/annotation_based_ippool.md
```diff
@@ -0,0 +1,97 @@
+# Annotation-based IPPool Reference
+
+The `fromPoolAnnotation` field enables dynamic IPPool name resolution from
```
lentzi90: nit: could you clarify here that we are talking about Metal3DataTemplates? For example:

```suggestion
The `fromPoolAnnotation` field of a `Metal3DataTemplate` enables dynamic IPPool name resolution from
```

PR #635, opened 2025-12-08
:seedling: Add ironic-image v33

elfosardo:
Set ironic-image v29 as Tested.
Set ironic-image image v27, v28 and v30 as EOL.

### DIFF
--- a/docs/user-guide/src/version_support.md
+++ b/docs/user-guide/src/version_support.md
@@ -75,19 +75,20 @@ maintained as supported releases.
 
 Following table summarizes Ironic-image release/test process:
 
-| Minor release | Status    | Ironic Branch      |
-| ------------- | --------- | ------------------ |
-| v32.0         | Supported | stable/2025.2      |
-| v31.0         | Supported | bugfix/31.0        |
-| v30.0         | Supported | bugfix/30.0        |
-| v29.0         | Supported | stable/2025.1      |
-| v28.0         | Supported | bugfix/28.0        |
-| v27.0         | Supported | bugfix/27.0        |
-| v26.0         | EOL       | bugfix/26.0        |
-| v25.0         | EOL       | bugfix/25.0 (EOL)  |
-| v24.1         | EOL       | stable/2024.1      |
-| v24.0         | EOL       | bugfix/24.0 (EOL)  |
-| v23.1         | EOL       | bugfix/23.1 (EOL)  |
+| Minor release | Status    | Ironic Branch       |
+| ------------- | --------- | ------------------- |
+| v33.0         | Supported | bugfix/33.0         |
+| v32.0         | Supported | stable/2025.2       |
+| v31.0         | Supported | bugfix/31.0         |
+| v30.0         | EOL       | bugfix/30.0         |
+| v29.0         | Tested    | stable/2025.1       |
+| v28.0         | EOL       | bugfix/28.0 (EOL)   |
+| v27.0         | Tested    | bugfix/27.0 (EOL)   |
+| v26.0         | EOL       | bugfix/26.0 (EOL)   |
+| v25.0         | EOL       | bugfix/25.0 (EOL)   |
+| v24.1         | EOL       | stable/2024.1 (EOL) |
+| v24.0         | EOL       | bugfix/24.0 (EOL)   |
+| v23.1         | EOL       | bugfix/23.1 (EOL)   |
 
 ## Image tags
 


docs/user-guide/src/version_support.md
```diff
@@ -77,12 +77,13 @@ Following table summarizes Ironic-image release/test process:
...
 | Minor release | Status    | Ironic Branch      |
 | ------------- | --------- | ------------------ |
+| v33.0         | Supported | bugfix/33.0        |
 | v32.0         | Supported | stable/2025.2      |
 | v31.0         | Supported | bugfix/31.0        |
 | v30.0         | Supported | bugfix/30.0        |
 | v29.0         | Supported | stable/2025.1      |
 | v28.0         | Supported | bugfix/28.0        |
-| v27.0         | Supported | bugfix/27.0        |
+| v27.0         | EOL       | bugfix/27.0        |
```
tuminoid: 27 is still "tested" in our CI, until 1.12 is cut.

elfosardo: @tuminoid I think we should set this to EOL, or introduce a different word ("tested" ?)
the ironic bugfix branch is long gone, and the latest update to ironic-image release-27 is 4 months old
it doesn't really look "supported" to me :)

tuminoid: Exactly, we use word "Tested" for this for other components. CI exists, but they're not actively maintained.

elfosardo: /facepalm right....
I changed other versions too, although I'm not sure we can really call them Tested
I would probably just put v28 and v30 as EOL or introduce an "Untested" word for those cases

docs/user-guide/src/version_support.md
```diff
@@ -77,12 +77,13 @@ Following table summarizes Ironic-image release/test process:
 
 | Minor release | Status    | Ironic Branch      |
 | ------------- | --------- | ------------------ |
+| v33.0         | Supported | bugfix/33.0        |
 | v32.0         | Supported | stable/2025.2      |
 | v31.0         | Supported | bugfix/31.0        |
-| v30.0         | Supported | bugfix/30.0        |
+| v30.0         | Tested    | bugfix/30.0        |
```
tuminoid: But now I'm confused, we're adding a Dependabot config for v28 and v30 you just moved to Tested. Dependabot = new PRs = meaning they'd be actively maintained?

tuminoid: Also to add confusion, like you mentioned is that we're not even testing some of these in metal3, because they're not tied to CAPM3 release and we don't have periodics for ironic-image... :)

elfosardo: the dependabot change precedes this change
if we can decide on how to name those branches, then maybe we can change the dependabot config as a consequence

tuminoid: So, looking at the same file in the end for testing matrix, we can see that we're not testing any even releases, so v28, v30 and v32 are not tested by CAPM3. v33 will be paired with 1.12. v27, v29, v31 are tested.

Those can be demoted to EOL, unless they're maintained by upstream, in which case we can leave them Supported and add Dependabot on them.

So this takes us to:
- v27 tested
- v28 eol
- v29 supported
- v30 eol
- v31 supported
- v32 supported
- v33 supported

and after 1.12 is cut, we can trim down to:
- v27 eol
- v28 eol
- v29 tested
- v30 eol
- v31, v32, v33 supported

Did I do mental math correctly? :)

tuminoid:
/hold
Just in case to avoid mid-pondering merging.

docs/user-guide/src/version_support.md
elfosardo: ok, that works
in ironic upstream we have support only for 31 and 33 at the moment

tuminoid: Were not using v32 either, so EOL that then too?

elfosardo: IrSO uses it

dtantsur: We keep the last 3 branches actively supported, it's 31, 32 and 33 right now. It's way too early to EOL 32.

docs/user-guide/src/version_support.md
```diff
@@ -77,12 +77,13 @@ Following table summarizes Ironic-image release/test process:
 
 | Minor release | Status    | Ironic Branch      |
 | ------------- | --------- | ------------------ |
+| v33.0         | Supported | bugfix/33.0        |
 | v32.0         | Supported | stable/2025.2      |
 | v31.0         | Supported | bugfix/31.0        |
-| v30.0         | Supported | bugfix/30.0        |
+| v30.0         | EOL       | bugfix/30.0        |
 | v29.0         | Supported | stable/2025.1      |
```
dtantsur: If 30 is EOL, 29 should at most be tested, otherwise it's a strange gap in the coverage. The policy just above literally talks about "2-3 branches", which I tend to generously interpret as "3 branches" :)

elfosardo:
@dtantsur @tuminoid  I think I'm going to change this so we take into account 1.12 already, my understanding is that it's going to be cut soon, so in the end we'll have
v27 eol
v28 eol
v29 tested
v30 eol
v31, v32, v33 supported

docs/user-guide/src/version_support.md
```diff
@@ -75,19 +75,20 @@ maintained as supported releases.
...
-| v23.1         | EOL       | bugfix/23.1 (EOL)  |
+| Minor release | Status    | Ironic Branch       |
+| ------------- | --------- | ------------------- |
+| v33.0         | Supported | bugfix/33.0         |
+| v32.0         | Supported | stable/2025.2       |
+| v31.0         | Supported | bugfix/31.0         |
+| v30.0         | EOL       | bugfix/30.0         |
+| v29.0         | Tested    | stable/2025.1       |
+| v28.0         | EOL       | bugfix/28.0 (EOL)   |
+| v27.0         | EOL       | bugfix/27.0 (EOL)   |
```
tuminoid: v27 needs to be tested until 1.12 is out.

dtantsur: Ugh, I hope we can eventually stop relying on ancient versions.

On a related note, maybe we can invent a better word than "Tested"? Tested to me implies more commitment than just "works in one specific scenario". I don't have any alternatives in mind though...

tuminoid: I just commented on the IRSO version bump PR that we should sit down and agree what to do with the IRSO/Ironic vs rest of Metal3 couplings. Which parts we can ignore and which are real. If we don't have any real dependencies, we could in tests just bump IRSO for old release branches too (when those branches actually are using IRSO, cough, and not old manifests).

tuminoid (commented):
/lgtm

PR #647, opened 2026-01-08
:sparkles: Add GHaction to test quick-start guide

peppi-lotta:
Add a GitHub action for testing the quickstart guide.


.github/workflows/quick-start-test.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Install libvirt
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libvirt-daemon-system qemu-kvm virt-manager libvirt-dev
+
+    - name: Run Quick Start Test
+      # We need a new shell to pick up the new group. That is why we do the sudo -s -u $USER ...
+      # Remove the pre-installed go version. We install the exact version we need.
+      run: |
+        sudo usermod -a -G libvirt $USER
```
tuminoid:
```suggestion
        sudo usermod -a -G libvirt "${USER}"
```

.github/workflows/quick-start-test.yml
```diff
@@ -0,0 +1,39 @@
...
+        sudo -s -u $USER --preserve-env bash ${{ github.workspace }}/hack/quick-start-test/quick-start-test.sh
```
tuminoid:
```suggestion
        sudo -s -u "${USER}" --preserve-env bash "${{ github.workspace }}/hack/quick-start-test/quick-start-test.sh"
```

.gitignore
```diff
@@ -17,3 +17,8 @@ docs/presentations/plugin
 # Development containers (https://containers.dev/)
 .devcontainer
 .claude
+
+# Quick start test temp files
```
tuminoid: Put these into `hack/quick-start/.gitignore`

docs/user-guide/src/quick-start.md
```diff
@@ -40,9 +40,16 @@ You will need the following tools installed.
   cluster that you want to use)
 - clusterctl
 - kubectl
-- htpasswd
 - virsh and virt-install for the virtualized setup
 
+There are multiple files you will create when following this quick start guide.
+Set the ```QUICK_START_BASE``` environment variable to the base where you are
```
tuminoid:
```suggestion
Set the `QUICK_START_BASE` environment variable to the base where you are
```

hack/quick-start/cleanup-clusters.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# If a cluster was created using Cluster API, delete that first
+kubectl delete cluster my-cluster
+
+# Delete all BareMetalHosts with `kubectl delete bmh <name>`. This ensures that
+# the servers are cleaned and powered off.
+kubectl delete bmh --all
+
+# Delete the management cluster
+kind delete cluster
```
tuminoid: missing EOL

hack/quick-start/ensure/ensure_clusterctl.sh
```diff
@@ -0,0 +1,40 @@
...
+set -eux
+
+USR_LOCAL_BIN="/usr/local/bin"
+MINIMUM_CLUSTERCTL_VERSION=v1.12.1
+
+# Ensure the clusterctl tool exists and is a viable version, or installs it
+verify_clusterctl_version()
+{
+    # If clusterctl is not available on the path, get it
+    if ! [ -x "$(command -v clusterctl)" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "$(command -v clusterctl)" ]]; then
```

hack/quick-start/ensure/ensure_go.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -eux
+
+MINIMUM_GO_VERSION=go1.24.11
+
+# Ensure the go tool exists and is a viable version, or installs it
+verify_go_version()
+{
+    # If go is not available on the path, get it
+    if ! [ -x "$(command -v go)" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "$(command -v go)" ]]; then
```

hack/quick-start/ensure/ensure_kubectl.sh
```diff
@@ -0,0 +1,50 @@
...
+set -eux
+
+USR_LOCAL_BIN="/usr/local/bin"
+MINIMUM_KUBECTL_VERSION=v1.34.1
+
+# Ensure the kubectl tool exists and is a viable version, or installs it
+verify_kubectl_version()
+{
+    # If kubectl is not available on the path, get it
+    if ! [ -x "$(command -v kubectl)" ]; then
```
tuminoid:
```suggestion
    if ! [[ -x "$(command -v kubectl)" ]]; then
```

hack/quick-start/ironic/kustomization.yaml
```diff
@@ -4,3 +4,4 @@ namespace: baremetal-operator-system
 resources:
 - ironic.yaml
 - certificate.yaml
+
```
tuminoid: extra empty line

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,126 @@
...
+    REQUIRED_FILES=(
+        "jammy-server-cloudimg-amd64.img"
+        "CENTOS_10_NODE_IMAGE_K8S_v1.34.1.qcow2"
+        "CENTOS_10_NODE_IMAGE_K8S_v1.34.1.raw"
+        "ipa-centos9-master.tar.gz"
+    )
+
+    missing_files=0
+    for file in "${REQUIRED_FILES[@]}"; do
+        if [ ! -f "${DISK_IMAGE_DIR}/${file}" ]; then
```
tuminoid:
```suggestion
        if [[ ! -f "${DISK_IMAGE_DIR}/${file}" ]]; then
```

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,126 @@
...
+            missing_files=1
+            break
+        fi
+    done
+
+    if [ "$missing_files" -eq 1 ]; then
```
tuminoid:
```suggestion
    if [[ "${missing_files}" -eq 1 ]]; then
```

dtantsur:
This change creates yet another place where we have the virtual environment scripts that already exist in some form in metal3-dev-env, BMO, CAPM3 and potentially other places. We really need to use one, but at the very least let's not add scripts to *docs*. I suspect the most reasonable step is to reuse CAPM3's scripts since the quick start covers CAPI integration.

peppi-lotta:
> This change creates yet another place where we have the virtual environment scripts that already exist in some form in metal3-dev-env, BMO, CAPM3 and potentially other places. We really need to use one, but at the very least let's not add scripts to _docs_. I suspect the most reasonable step is to reuse CAPM3's scripts since the quick start covers CAPI integration.

I do agree with you about virtual environments being in many places and we definitely need work towards having one general virtual environment that is used everywhere. However, this change is testing word for word what quick start guide instructs the user to do. Over half of the scripts that are in this PR are directly embedded to the quick start guide.

Quick start is purely a docs thing and that is why I feel like docs repo is the best place to have the action that verifies if the quick start guide is up to date. I'm open to discuss more about where these scripts could live but at the moment there is no other virtual environment that is an exact match to what the quick start guide is instructing to do and it's necessary to have this test somewhere to ensure quick start is actually kept up to date and working.

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,127 @@
...
+    "${QUICK_START_BASE}/cleanup-clusters.sh"
+    docker stop image-server
+    "${QUICK_START_BASE}/cleanup-virtlab.sh"
+}
+
+ensure_env
+setup
+create_bmhs
+scenario_2
+cleanup
```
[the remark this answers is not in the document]
peppi-lotta: I don't think this comment is valid. Everything should be called conditionally because we are expecting everything to work and if anything fails it's and indicator that quick start has some command that is not working as it's supposed to. We don't want the clean up to fail anything.

.github/workflows/test-quick-start-guide.yml
```diff
@@ -0,0 +1,39 @@
+name: E2E Test
+
+on:
+  workflow_call:
+    inputs:
+      runner:
+        type: string
+        default: "oracle-vm-4cpu-16gb-x86-64"
```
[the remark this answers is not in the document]
peppi-lotta: @tuminoid can we make the oracle-vm-4cpu-16gb-x86-64 available in this repo as well or should this test use ubuntu-latest. I copied this runner from BMO e2e tests as this whole tests as similar to BMO.

hack/quick-start/irso/kustomization.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: ironic-standalone-operator-system
+
+resources:
+- https://github.com/metal3-io/ironic-standalone-operator/releases/latest/download/install.yaml
+
+configMapGenerator:
```
adilGhaffarDev: no need to generate configmap and patch it in the deployment, now install.yaml already has configmap named `ironic-standalone-operator-ironic-standalone-operator-config`, you can just patch that config map with `IPA_BASEURI` .

hack/quick-start/ensure/ensure_go.sh
lentzi90, line 1: Do we really need go? I was hoping not, and from a quick skim through it doesn't look like we are using it. 🙂

peppi-lotta: We don't need go and I removed that now. I had some brain short circuit and thought kind requires go and that's why I had it here.

peppi-lotta:
I have edited the workflow files. Now the test would be run once a day as periodic and on PRs when quick start guide is edited.

I opened a discussion about this in the last community meeting. There was a consensus that this test is needed and there was no strong objection to this location. I did the workflow file changes based on the conversation had in the community meeting.

@dtantsur is it ok to get this merged?

.github/workflows/test-quick-start-guide.yml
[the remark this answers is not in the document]
lentzi90: It would be good to try this workflow on a standard runner first. I think we could get it to work with `ubuntu-latest`. Could you put a temporary commit that makes the workflow run on this PR?

hack/quick-start/bmh-01.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: metal3.io/v1alpha1
 kind: BareMetalHost
 metadata:
-  name: bml-vm-01
+  name: bmh-vm-01
```
lentzi90, line 4: nit: I would keep the name as it was. It is referenced in the guide. The `kind` already tells us it is a BMH, the name tells us then that this is vm number 1 that is part of our BML 🙂 
At least that was my thinking

hack/quick-start/capm3-vars.sh
```diff
@@ -1,10 +1,14 @@
...
 # Baremetal lab image variables
 # export IMAGE_URL="http://192.168.0.150/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.qcow2"
 # export IMAGE_CHECKSUM="http://192.168.0.150/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.qcow2.sha256sum"
 # export IMAGE_FORMAT="qcow2"
 # Virtualized setup variables
 export IMAGE_URL="http://192.168.222.1/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.raw"
-export IMAGE_CHECKSUM="http://192.168.222.1/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.raw.sha256sum"
+# Calculate checksum from the actual raw image file
+IMAGE_CHECKSUM=$(sha256sum "${QUICK_START_BASE}/disk-images/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.raw" | awk '{print $1}')
+export IMAGE_CHECKSUM
```
lentzi90: Revert this please. We calculate the sum and store it in the file when we download the image already. This here is duplicating the work.

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,124 @@
...
+# DHCP and image servers, and deploying Ironic and baremetal operators.
+#------------------------------------------------------------------------------
+set -eux
+
+# Default QUICK_START_BASE to the absolute path of this script's directory if not already set.
+export QUICK_START_BASE=${QUICK_START_BASE:="$(dirname -- "$(readlink -f "${BASH_SOURCE[0]}")")"}
+
+ensure_env() {
+    sudo apt-get update
+    sudo apt-get install -y libvirt-daemon-system qemu-kvm virt-manager libvirt-dev
```
lentzi90: This is duplicated from the workflow. I would suggest removing it from here to keep the script more portable. Then it will be theoretically possible to run this on centos or other distros also, as long as the dependencies are installed first.

```suggestion
```

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,124 @@
...
+    if ! kubectl wait --for=condition=Ready --timeout=600s machine --all; then
+        echo "ERROR: Machine failed to reach 'Ready' state within timeout."
+        exit 1
+    fi
+}
+
+setup_disk_images_dir() {
+    DISK_IMAGE_DIR="${QUICK_START_BASE}/disk-images"
+    REQUIRED_FILES=(
+        "jammy-server-cloudimg-amd64.img"
```
lentzi90, line 87: Note to self: we should optimize the test to avoid downloading images we do not need. If we just test scenario 2, don't download all of them.
Let's keep it like this for this PR though.

hack/quick-start/quick-start-test.sh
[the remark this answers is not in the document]
lentzi90: I would say the bigger risk here is that if there is any error we just exit. No cleanup and no log collection. I think it is ok for now, but it is something we will need eventually. It may be better to handle it outside this script though.

lentzi90 (commented):
Finally, I have reviewed it completely. There is a lot of good here! I basically only have a few nits below.
One this I would want before merging though, is to test it properly. Could you add a commit where the workflow runs on your branch and also where it embeds the files from your branch? Then we can see the proper rendering and also verify that the test runs. Once we see this, I am happy to approve and then we can remove the test commit again.

hack/quick-start/capm3-vars.sh
```diff
@@ -1,10 +1,14 @@
+#!/usr/bin/env bash
+
 # Baremetal lab image variables
 # export IMAGE_URL="http://192.168.0.150/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.qcow2"
 # export IMAGE_CHECKSUM="http://192.168.0.150/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.qcow2.sha256sum"
 # export IMAGE_FORMAT="qcow2"
+# Export the checksum from the disk image file (extract just the hash part, not the filename)
+IMAGE_CHECKSUM=$(awk '{print $1}' "${QUICK_START_BASE}/disk-images/CENTOS_10_NODE_IMAGE_K8S_v1.34.1.raw.sha256sum")
+export IMAGE_CHECKSUM
```
lentzi90, line 9: This shouldn't be needed. Is there some issue if the filename is included? It should be standard procedure to have the checksum in this format, including the filename.

docs/user-guide/examples/bmh-01-provision.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: metal3.io/v1alpha1
 kind: BareMetalHost
 metadata:
-  name: bml-vm-01
+  name: bmh-vm-01
```
lentzi90: nit: I would keep the name as it was. It is referenced in the guide. The kind already tells us it is a BMH, the name tells us then that this is vm number 1 that is part of our BML 🙂
At least that was my thinking

hack/quick-start/setup-image-server-dir.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-mkdir disk-images
+mkdir "${QUICK_START_BASE}/disk-images"
 
-pushd disk-images || exit
+pushd "${QUICK_START_BASE}/disk-images" || exit
 wget https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-amd64.img
```
lentzi90: nit: We should add `--quiet` to all these wget commands now that they will run in CI. As it is now we have ~50000 lines of wget output in the logs 🙈

docs/user-guide/examples/bmh-01-provision.yaml
peppi-lotta: fixed

hack/quick-start/setup-image-server-dir.sh
peppi-lotta: fixed

hack/quick-start/capm3-vars.sh
peppi-lotta: This was the only way I got this to work. If I included the file name there was a failure.

.github/workflows/test-quick-start-guide.yml
```diff
@@ -0,0 +1,39 @@
...
+permissions: {}
+
+jobs:
+  test:
+    name: Quick Start Test
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: ${{ inputs.timeout-minutes }}
+
+    steps:
+    - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
```
tuminoid: v6.0.2 is latest

.github/workflows/test-quick-start-guide.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        ref: ${{ inputs.ref }}
```
tuminoid, line 29:
```suggestion
      with:
        ref: ${{ inputs.ref }}
        persist-credentials: false
```

.github/workflows/test-quick-start-guide-periodic.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+permissions: {}
+
+jobs:
+  periodic-e2e-test:
+    if: github.repository == 'metal3-io/metal3-docs'
+    uses: ./.github/workflows/test-quick-start-guide.yml
```
tuminoid:
```suggestion
    uses: ./.github/workflows/test-quick-start-guide.yml # zizmor: ignore[unpinned-uses] 
```

.github/workflows/test-quick-start-guide-periodic.yml
```diff
@@ -0,0 +1,17 @@
...
+    with:
+      runner: "ubuntu-latest"
+      timeout-minutes: 90
+    permissions:
+      contents: read
```
tuminoid: Move this up to under if, so its more logical to read.

.github/workflows/test-quick-start-guide.yml
```diff
@@ -0,0 +1,39 @@
...
+      timeout-minutes:
+        type: number
+        default: 90
+
+permissions: {}
+
+jobs:
+  test:
+    name: Quick Start Test
+    runs-on: ${{ inputs.runner }}
```
tuminoid, line 21:
```suggestion
    runs-on: ${{ inputs.runner }}
    permissions:
      contents: read
```
Its a public repo, so its not explicitly needed, but since we do checkout, it is nice to have avoid implicit assumptions.

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,121 @@
...
+
+#------------------------------------------------------------------------------
+# This script sets up a quick start test environment for Metal3 by
+# configuring a virtual lab, bootstrapping a Kind cluster, setting up
+# DHCP and image servers, and deploying Ironic and baremetal operators.
+#------------------------------------------------------------------------------
+set -eux
+
+# Default QUICK_START_BASE to the absolute path of this script's directory if not already set.
+export QUICK_START_BASE=${QUICK_START_BASE:="$(dirname -- "$(readlink -f "${BASH_SOURCE[0]}")")"}
```
tuminoid, line 11: `:=` ?

hack/quick-start/quick-start-test.sh
```diff
@@ -0,0 +1,127 @@
...
+        "${QUICK_START_BASE}/setup-image-server-dir.sh"
+    else
+        echo "All required disk images are present."
+    fi
+}
+
+cleanup() {
+    echo "Cleaning up the quick start test environment..."
+    "${QUICK_START_BASE}/cleanup-clusters.sh"
+    docker stop image-server
```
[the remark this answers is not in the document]
tuminoid: Hmm, this comment is valid. Cleanup should disable failures `set +e`  or `|| true`, otherwise the cleanup might stop if any cleanup command fails.

hack/quick-start/setup-image-server-dir.sh
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
```
tuminoid, line 2: Why not `set -eu` ?

hack/quick-start/setup-virtual-lab.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
```
tuminoid, line 2: Why not `set -eu` ?

hack/quick-start/start-image-server.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+docker run --name image-server --rm -d -p 80:8080 \
```
tuminoid, line 2:
```suggestion
#!/usr/bin/env bash

set -eu

docker run --name image-server --rm -d -p 80:8080 \
```

.github/workflows/test-quick-start-guide.yml
peppi-lotta: fixed

.github/workflows/test-quick-start-guide.yml
peppi-lotta: fixed

.github/workflows/test-quick-start-guide-periodic.yml
peppi-lotta: fixed

.github/workflows/test-quick-start-guide-periodic.yml
peppi-lotta: fixed

.github/workflows/test-quick-start-guide.yml
peppi-lotta: fixed

hack/quick-start/quick-start-test.sh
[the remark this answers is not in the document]
peppi-lotta: This cleanup step runs the cleanup commands that we specify in the quick-start. We don't need this cleanups step for the run  per se because the runner instance gets removed after the run and be don't run the same test on the same runner multiple times.

Here I think it's valid that the fail run fails at this clean up step because it might mean we need to update the quick-start-guide somehow if a command we are telling the users to run is failing for some reason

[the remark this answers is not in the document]
tuminoid: OK, that's fair.

hack/quick-start/quick-start-test.sh
peppi-lotta: I've tried to build this so that this would be easy to use to set up a local environment as well. I've done testing where this `quick-start-test.sh` is outside of the "quick starts root" and that is why I have left `:=`. I'm okay with removing it because it is not relevant for github test runs more for if a user or developer want to use the script.

hack/quick-start/setup-image-server-dir.sh
peppi-lotta: I'm happy to add these. I've copied this file as they currently are in [docs/user-guide/examples](https://github.com/metal3-io/metal3-docs/tree/main/docs/user-guide/examples)  with minimal as possible changes.

hack/quick-start/setup-virtual-lab.sh
peppi-lotta: I'm happy to add these. I've copied this file as they currently are in [docs/user-guide/examples](https://github.com/metal3-io/metal3-docs/tree/main/docs/user-guide/examples)  with minimal as possible changes.

hack/quick-start/start-image-server.sh
peppi-lotta: done

kashifest (commented):
/lgtm

PR #650, opened 2026-02-04
try fix reapprover workflow

tuminoid:
The reapprover workflow does not seem to trigger most of the times. One probable cause is race condition of pull request workflows being created and this one running. This one might run before the workflows are created, so it doesn't approve them as it doesn't see them.

Also, remove open/ready_for_review triggers as those are not relevant for this workflow. Add labeled instead, so adding /ok-to-test should trigger this and approve the workflows.


tuminoid:
I will replicate this to other repos, if it solves the race condition as expected.

PR #651, opened 2026-02-05
add glossary to the book

tuminoid:
Add glossary.md with all the relevant abbreviations and key terms to help reader understand the terms used in Metal3 and in baremetal environments. 

Does not include "common knowledge" abbreviations, like IP, API, L2 etc.


tuminoid:
> /approe

Typo, try again 
/cc @Rozzii

dtantsur:
/approve

This is very useful, thanks! It could be even more useful if you added links to explanations: e.g. IrSO could lead to its user guide page, iPXE or Bifrost - to respective web sites, etc. Maybe even wikipedia for some concepts like ISO.

tuminoid:
> /approve
> 
> This is very useful, thanks! It could be even more useful if you added links to explanations: e.g. IrSO could lead to its user guide page, iPXE or Bifrost - to respective web sites, etc. Maybe even wikipedia for some concepts like ISO.

I considered it, and decided against it (for now), as it might end up being high maintenance due number of links. Maybe in follow-up.

PR #659, opened 2026-02-23
📖 improve failure domain example scenario in fd-support-kcp

shibaPuppy:
Improves the example scenario in the failure domain design doc to use concrete rack names instead of abstract identifiers.   

also serves as a design document proposing how CAPM3 should support failure domains with Kubeadm Control Plane (KCP).


tuminoid (commented):
Address the linter failures.

IMO it looks a bit too verbose, but not expert on this.
/cc @Sunnatillo @Rozzii 
PTAL

shibaPuppy:
I also split the DataTemplate selection proposal into a separate file `fd-datatemplate-selection.md` to keep `fd-support-kcp.md` focused on the original BMH selection design.

shibaPuppy:
@Sunnatillo  @Rozzii, could you take a look when you get a chance? I've addressed the linter failures.

Rozzii:
> Address the linter failures.
> 
> IMO it looks a bit too verbose, but not expert on this. /cc @Sunnatillo @Rozzii PTAL

Indeed the proposal contains a lot of CR examples that are indirectly related to the proposal with realistic example values, I don't mind it though.

Rozzii:
PTAL /cc @adilGhaffarDev @kashifest 

The technical part is OK for me, I am fine with the verbosity of the proposal but if you folks think it is too much then I am OK with that too, I can see how that can be a bit disorienting for people.

adilGhaffarDev:
looks fine to me.
thanks for the proposal.

design/fd-datatemplate-selection.md
```diff
@@ -0,0 +1,416 @@
...
+      └── spec.template.spec.dataTemplate → 1 Metal3DataTemplate
+                                              └── networkData (single config)
+```
+
+This creates two problems:
+
+1. **No per-node subnet differentiation**: All CP nodes receive the same
+   subnet configuration, preventing true network-level HA across different
+   L2/L3 segments.
+2. **Static VLAN tagging**: `NetworkDataLinkVlan` in `Metal3DataTemplate` is
```
Sunnatillo:
```suggestion
1. **Static VLAN tagging**: `NetworkDataLinkVlan` in `Metal3DataTemplate` is
```

This should be one to pass markdown lint. Markdown will itself set correct number.

design/fd-datatemplate-selection.md
```diff
@@ -0,0 +1,416 @@
...
+```
+
+### Result — Metal3Machines with per-FD DataTemplates
+
+KCP automatically creates 3 Machine objects (distributed across the 3 failure
+domains), and each Machine triggers the creation of a corresponding
+Metal3Machine. The CAPM3 controller then:
+
+1. Syncs `Machine.Spec.FailureDomain` → `Metal3Machine.Spec.FailureDomain`
+2. Looks up `failureDomainDataTemplates` in the Metal3MachineTemplate
```
Sunnatillo: same here

design/fd-datatemplate-selection.md
```diff
@@ -0,0 +1,416 @@
...
+3. Overrides `Metal3Machine.Spec.DataTemplate` with the FD-specific reference
```
Sunnatillo: and here

shibaPuppy: @Sunnatillo 
replaced the numbered list with dashes to avoid the ordering lint issue.

lentzi90:
Hmm... ok I will try to retitle to shake the workflows.
/retitle 📖 improve failure domain example scenario in fd-support-kcp

PR #663, opened 2026-02-26
Adds detailed providerID workflow in CAPM3

kashifest:
Adds documentation around providerID workflow in CAPM3

### DIFF
--- a/docs/user-guide/src/SUMMARY.md
+++ b/docs/user-guide/src/SUMMARY.md
@@ -52,6 +52,7 @@
       - [ClusterClass](capm3/clusterclass.md)
       - [Failure Domain](capm3/failure_domain.md)
       - [Annotation-based IPPool](capm3/annotation_based_ippool.md)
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
Looks good to me!
There are some assumptions that kubeadm must be used here. This is not true, but we have similar assumptions throughout the docs already so I am ok with it. Just something to keep in mind. It is possible to do all of this with for example k3s and it would be good to start acknowledging these alternatives in our docs.
/approve

Edit: Just to clarify, I think it is **good** that we use kubeadm as *example*, but we should not say that this is the **only** way.

kashifest:
> Edit: Just to clarify, I think it is **good** that we use kubeadm as _example_, but we should not say that this is the **only** way.
 
 Good point, we can but when we don't test those its hard to claim what works and what doesn't. Perhaps the wording should be `for example here's how to configure it via kubeadm`. As you said, we are predominantly kubeadm in our docs until we test other alternatives.

PR #667, opened 2026-03-04
suppress accepted zizmor findings

tuminoid:
Add inline zizmor ignore comments for accepted findings: unpinned project-infra main refs and intentional pull_request_target trigger.


kashifest (commented):
/lgtm

PR #669, opened 2026-03-05
add zizmor scanner

tuminoid:
Zizmor scans workflow files to keep them hardened. It runs on push to produce a report in security tab, and it runs on PRs and blocks them if they is any errors.


tuminoid:
/hold
This is expected to fail to a dependabot-cooldown warning.

tuminoid:
> /hold This is expected to fail to a dependabot-cooldown warning.

/unhold
This is now passing, dependabot cooldown is set to 7 days for github actions and npm too.

PR #675, opened 2026-03-24
update self-assessment

tuminoid:
Mark it complete, add IRSO to actors, add OSPS baseline information, and Mistral Compute as case-study.


smoshiur1237 (commented):
/lgtm

PR #679, opened 2026-04-08
design: add IPPool random allocation strategy

shibaPuppy:
proposing an opt-in AllocationStrategy field with sequential (default) and random values.

### DIFF
--- a/design/ip-address-manager/ippool-random-allocation-strategy.md
+++ b/design/ip-address-manager/ippool-random-allocation-strategy.md
@@ -1,0 +1,135 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# IPPool Random Allocation Strategy
+
+## Status
+
+implementable
+
+## Summary
+
+Introduce a configurable `AllocationStrategy` field on `IPPoolSpec` so that
+operators can choose between `sequential` (the existing behavior, default)
+and `random` allocation when issuing IP addresses from an `IPPool`.
+
+## Motivation
+
+Today the Metal3 IPAM controller always allocates the first available
+address in a pool. While simple and deterministic, this has the following
+drawbacks:
+
+- Whenever workloads are recreated, the same low-numbered IPs are
+  repeatedly reused, which can collide with systems that key on a specific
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/ok-to-test
/cc @lentzi90 @Sunnatillo @peppi-lotta @kashifest

kashifest (commented):
While I do understand the need for randomized IP allocation, and I am largely ok with the change, but my first question would be does it really solve the issue with residual state in external systems such as ARP caches, firewall state, and monitoring history? If the same randomized IP is allocated in quick succession, we would still hit it no? Anyways, as said I am fine with this stratergy, we need to just make sure the code paths which assume sequential IPs remain the same for backwards compatibility which I guess this proposal promises. I will let it be open for some time before approving to let others also review and comment.

shibaPuppy:
thanks for the review, @kashifest!

you raise a valid point — random allocation doesn't eliminate the possibility of reusing a recently freed IP, but it significantly reduces the probability. with a pool of N addresses, sequential reuse is nearly guaranteed, whereas random allocation drops that to ~1/N.

beyond residual state, there's also a security motivation: predictable IP allocation makes it easier for attackers to anticipate which IP a newly provisioned node will receive. This opens the door to pre-positioned ARP spoofing, IP squatting, or MITM attacks during the provisioning window — when nodes are most vulnerable. Randomization raises the bar for these attacks by removing the predictability. It also reduces information leakage about infrastructure scaling patterns through IP observation.

that said, for stronger residual-state guarantees specifically, a follow-up proposal could introduce a cooldown mechanism (e.g., excluding recently released IPs for a configurable duration). I'd prefer to keep that out of scope here to keep the proposal focused.

regarding backwards compatibility — since AllocationStrategy defaults to sequential, existing users and code paths are completely unaffected. The random strategy is strictly opt-in.

lentzi90 (commented):
/lgtm

Sunnatillo (commented):
LGTM

Thank you @shibaPuppy  for opening a proposal and explaining your use case. 

I agree with your point reducing the probability of using recently freed IP address. I think it is not necessary to implement logic to completely eliminates allocating freed ip addresses

tuminoid (commented):
/approve

shibaPuppy:
@Sunnatillo @kashifest 
could you take another look when you get a chance? Thanks!

kashifest (commented):
/approve

PR #683, opened 2026-04-23
docs: Add Fast Track documentation

MaxRink:
Update the Metal3 book with CAPM3 Fast Track documentation.

Changes:
- Add Fast Track mode behavior matrix and configuration.
- Clarify CAPM3 Fast Track vs Ironic Fast Track.
- Document DisablePowerOff and automated cleaning interactions, including CAPM3 PR #3106 behavior.
- Add monitoring and troubleshooting guidance.

IP reuse was split to #700.

Related: metal3-io/cluster-api-provider-metal3#3111

### DIFF
--- a/docs/user-guide/src/capm3/fast_track.md
+++ b/docs/user-guide/src/capm3/fast_track.md
@@ -1,0 +1,151 @@
+# Fast Track Mode
+
+## Overview
+
+Fast Track mode is an optimization feature in CAPM3 that allows BareMetalHosts
+to skip the deprovisioning power-off cycle during machine deletion when certain
+conditions are met. This can significantly speed up cluster upgrades and node
+replacements by keeping hosts powered on and ready for immediate re-use.
+
+## How It Works
+
+When a Metal3Machine is deleted, CAPM3 normally instructs the BareMetalHost to
+power off and deprovision. With Fast Track mode enabled, CAPM3 can keep the host
+powered on if automated cleaning is configured to only clean metadata.
+
+The behavior is controlled by three factors:
+
+1. **BareMetalHost.Spec.DisablePowerOff**: If `true`, the host always stays
+   online (highest priority)
+1. **CAPM3_FAST_TRACK environment variable**: Set to `true` or `false`
+1. **BareMetalHost.Spec.AutomatedCleaningMode**: Set to `disabled` or `metadata`
+
+### Behavior Matrix
+
+| DisablePowerOff | CAPM3_FAST_TRACK | AutomatedCleaningMode | BMH Online Status        |
+| --------------- | ---------------- | --------------------- | ------------------------ |
+| true            | any              | any                   | **On** (DisablePowerOff) |
+| false           | false            | disabled              | Off                      |
+| false           | true             | disabled              | Off                      |
+| false           | false            | metadata              | Off                      |
+| false           | true             | metadata              | **On** (Fast Track)      |
+
+Since PR #3106 (`Allow disable_power_off together with autoclean`, merged on
+2026-03-05), `DisablePowerOff=true` can be combined with
+`AutomatedCleaningMode=metadata`. In that case, `DisablePowerOff` still takes
+priority and keeps the host online.
+
+The host remains online when:
+
+- `DisablePowerOff` is `true` (takes priority over all other settings), or
+- `AutomatedCleaningMode` is set to `metadata` (not `disabled`) AND
+  `CAPM3_FAST_TRACK` is set to `true`
+
+When both conditions are met, the BareMetalHost remains powered on after the
+Metal3Machine is deleted, allowing it to be quickly re-claimed by a new
+Metal3Machine without waiting for a full power cycle.
+
+## When to Use Fast Track
+
+Fast Track is beneficial in scenarios where:
+
+- **Rolling upgrades**: Nodes are being replaced one at a time, and speed is
+  important
+- **Controlled environments**: You trust that metadata cleaning is sufficient
+  (no need for full disk wipes)
+- **Development/testing**: Quick iteration cycles where full deprovisioning
+  adds unnecessary delay
+
+Fast Track should be avoided when:
+
+- **Security is paramount**: Full disk wiping is required between tenants or
+  workloads
+- **Disk cleaning is needed**: The `disabled` AutomatedCleaningMode is used
+  intentionally
+- **Troubleshooting**: You need hosts to fully deprovision to reset state
+
+## Configuration
+
+### Enabling Fast Track
+
+Fast Track is configured via the `CAPM3_FAST_TRACK` environment variable on the
+CAPM3 controller. It defaults to `false`.
+
+#### Via clusterctl
+
+Add to your clusterctl configuration file:
+
+```yaml
+variables:
+  CAPM3_FAST_TRACK: "true"
+```
+
+#### Via Environment Variable
+
+If deploying the controller directly, set the environment variable:
+
+```bash
+export CAPM3_FAST_TRACK=true
+```
+
+### Configuring AutomatedCleaningMode
+
+For Fast Track to work, your BareMetalHost resources must have
+`AutomatedCleaningMode` set to `metadata`. This can also be combined with
+`DisablePowerOff=true`; in that case `DisablePowerOff` remains the deciding
+factor for keeping the host online. Refer to the
+[Baremetal Operator Documentation](https://book.metal3.io/bmo/introduction)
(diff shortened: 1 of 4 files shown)


docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
+# Fast Track Mode
+
+## Overview
+
+Fast Track mode is an optimization feature in CAPM3 that allows BareMetalHosts
```
Rozzii: I would recommend mentioning here as a disclaimer that this fast track should not be confused with the Ironic fast track option and that the CAPM3 fast track and the Ironic fast track is commonly used together in Metal3 but they are different features and enabling one does not enable the other.

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+to skip the deprovisioning power-off cycle during machine deletion when certain
+conditions are met. This can significantly speed up cluster upgrades and node
+replacements by keeping hosts powered on and ready for immediate re-use.
+
+## How It Works
+
+When a Metal3Machine is deleted, CAPM3 normally instructs the BareMetalHost to
+power off and deprovision. With Fast Track mode enabled, CAPM3 can keep the host
+powered on if automated cleaning is configured to only clean metadata.
```
Rozzii: Please add that the reason behind this is the following: When the automated cleaning is not disabled then the physical machine will boot into the provisioning agent to do the cleaning, in order to speed up the re-use of the physical machine fast track keeps the machine running with the agent even after the cleaning has concluded to save 1 reboot  that would happen when the machine is re-provisioned.

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+## When to Use Fast Track
+
+Fast Track is beneficial in scenarios where:
+
+- **Rolling upgrades**: Nodes are being replaced one at a time, and speed is
+  important
+- **Controlled environments**: You trust that metadata cleaning is sufficient
+  (no need for full disk wipes)
+- **Development/testing**: Quick iteration cycles where full deprovisioning
+  adds unnecessary delay
```
Rozzii, line 67: In all use cases where the BMH is expected to be re-used eventually and power consumption is not a factor.

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+
+Fast Track should be avoided when:
+
+- **Security is paramount**: Full disk wiping is required between tenants or
```
Rozzii: full disk wiping is irrelevant here, BMO will be able to do full disk wiping (zeroing) as part of automated cleaning, Agent usually runs in an autonomous mode in production (or at least it should) so no one can log in, and the file system headers and partition table info is removed and usually the disk is encrypted anyhow so fast track + automated cleaning is safer than no automated cleaning at all because in that scenario the disk is fully intact and as long as the host is not re-uesd nothing will be cleaned away.

TLDR, I would not emphasize the security aspect, IMO it is not relevant here and the reasoning is also incorrect in this case.

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+  workloads
+- **Disk cleaning is disabled intentionally**: `AutomatedCleaningMode=disabled`
+  skips cleaning and therefore cannot provide metadata cleaning for Fast Track
+- **Troubleshooting**: You need hosts to fully deprovision to reset state
```
Rozzii:
```suggestion
- **Troubleshooting**:  physical maintenance is needed e.g. replacement of hardware that requires the machine to be powered off.
```

This reset state reasoning doesn't really mean anything to me (or in general IMO). Let's have specific examples.

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+If deploying the controller directly, set the environment variable:
+
+```bash
+export CAPM3_FAST_TRACK=true
+```
+
+### Configuring AutomatedCleaningMode
+
+For Fast Track to work, your BareMetalHost resources must have
+`AutomatedCleaningMode` set to `metadata`. In CAPM3 workflows, configure this
```
Rozzii: Will most likely change in the future, but anything other than disabled should work, just FYI. Not 100% sure it should be mentioned but it could be.

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+   the BMH's `Online` field:
+   - **Fast Track active**: Host stays online (`Online: true`)
+   - **Fast Track inactive**: Host is powered off (`Online: false`)
+1. CAPM3 waits for the host to reach an available state before fully releasing
+   it
+1. The ConsumerRef and other association data are cleared
+
+## Monitoring
+
+When Fast Track is active, you'll see log messages like:
```
Rozzii:
```suggestion
When Fast Track is active, see log messages will be produced by capm3 like:
```

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,152 @@
...
+
+```text
+Set host Online field based on DisablePowerOff, AutomatedCleaningMode, and Capm3FastTrack host=node-0 automatedCleaningMode=metadata hostSpecOnline=true
+```
+
+When Fast Track keeps a host online, the BareMetalHost will transition directly
+from `Provisioned` to `Available` without going through `Deprovisioning` with a
```
Rozzii: I feels this sentence is a bit confusing, reader could get an impression that the deprovisioning state is skipped even though deprovisioning will still happen. Please try to phrase this a bit more clearly.

docs/user-guide/src/capm3/ip_reuse.md
```diff
@@ -0,0 +1,183 @@
+# IP Reuse (BMH Name-Based Preallocation)
```
Rozzii: Could you put this to a separate PR please? I think we can get them merged faster if they are separated.  These are two very distinct topics, two separate commits. I don't want to block both of them just because one of them might have an issue.

Rozzii (commented):
Thank you for working on the docs, it is almost there!
I have a few comments but nothing major.

Rozzii (commented):
/approve
Thank you!

Rozzii (commented):
/lgtm

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,173 @@
...
+an already-running agent ramdisk across Ironic inspection, cleaning, and
+deployment. CAPM3 Fast Track controls whether CAPM3 leaves the BareMetalHost
+`Online` during Metal3Machine deletion. These features are commonly used
+together in Metal3, but enabling one does not enable the other.
+
+## How It Works
+
+When a Metal3Machine is deleted, CAPM3 normally lets the BareMetalHost power off
+as part of deprovisioning. With Fast Track enabled, CAPM3 can keep the host
+powered on when automated cleaning is not disabled.
```
lentzi90: nit: double negation can be hard to understand. This could be simplified as

```suggestion
powered on when automated cleaning is enabled.
```

On further thought I guess this wording was chosen based on the actual field value (`disabled`)? Maybe we could clarify it as

```suggestion
powered on when automated cleaning is set to anything other than `disabled`.
```

docs/user-guide/src/capm3/fast_track.md
```diff
@@ -0,0 +1,173 @@
...
+| DisablePowerOff | CAPM3_FAST_TRACK | AutomatedCleaningMode         | BMH Online Status        |
+| --------------- | ---------------- | ----------------------------- | ------------------------ |
+| true            | any              | any                           | **On** (DisablePowerOff) |
+| false           | false            | disabled                      | Off                      |
+| false           | true             | disabled                      | Off                      |
+| false           | false            | not disabled (metadata today) | Off                      |
+| false           | true             | not disabled (metadata today) | **On** (Fast Track)      |
+
+Since PR #3106 (`Allow disable_power_off together with autoclean`, merged on
+2026-03-05), `DisablePowerOff=true` can be combined with
```
lentzi90: We should convert this to release tags. In what release is/will this be included? Users should not need to know about individual PRs.

lentzi90 (commented):
/approve

Rozzii (commented):
/lgtm

PR #687, opened 2026-05-04
Design proposal for fast (out-of-band) inspection

dtantsur:
Mostly generated by Claude Code with manual editing to reduce verbosity.

### DIFF
--- a/design/baremetal-operator/out-of-band-inspection.md
+++ b/design/baremetal-operator/out-of-band-inspection.md
@@ -1,0 +1,303 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Add fast (out-of-band) inspection mode
+
+## Status
+
+provisional
+
+## Summary
+
+Ironic supports two types of hardware inspection: in-band (agent-based)
+and out-of-band (Redfish-based). Metal3 currently only uses in-band
+inspection, which requires booting a ramdisk with the Ironic Python
+Agent (IPA) and takes several minutes to complete. Out-of-band
+inspection uses the Redfish BMC API to collect hardware inventory
+directly, completing in seconds without booting anything.
+
+This proposal adds a new `InspectionMode` value that enables
+out-of-band inspection in BareMetalHost. The feature requires a
+Redfish-compatible BMC and provides a faster, lighter alternative to
+agent-based inspection.
+
+## Motivation
+
+In-band inspection is thorough but slow: the host must boot a
+ramdisk, run the Ironic Python Agent, collect data, and report back.
+This takes minutes per host and requires a functioning provisioning
+network. Out-of-band inspection avoids all of this by querying the
+BMC directly via Redfish, returning results in seconds.
+
+Beyond speed, out-of-band inspection is useful for preparation
+workflows that need hardware information before booting anything.
+For example, an operator might need to discover hardware
+configuration or provide data for templates used in site-specific
+PreprovisioningImages.
+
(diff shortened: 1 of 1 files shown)


design/baremetal-operator/out-of-band-inspection.md
```diff
@@ -0,0 +1,299 @@
...
+
+- `InspectionModeFast`: use the BMC-native inspect interface (e.g.
+  `redfish`). If the BMC's `AccessDetails.InspectInterface()` returns
+  an empty string (i.e. the BMC does not support out-of-band inspection),
+  the provisioner should return an error.
+- Any other value: use `agent` inspect interface (current behavior).
+
+**NOTE:** it seems logical to use the special interface `no-inspect` for hosts
+with inspection disabled. However, this interface is not current enabled in
+ironic-image, so we'll stay with using `agent` by default for now.
```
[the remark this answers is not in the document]
dtantsur: This is the case now, I'm not intending to change it.

design/baremetal-operator/out-of-band-inspection.md
```diff
@@ -0,0 +1,299 @@
...
+updated to include `fast`:
+
+```go
+// +kubebuilder:validation:Enum=disabled;agent;fast
+InspectionMode InspectionMode `json:"inspectionMode,omitempty"`
+```
+
+The existing `inspect.metal3.io` annotation will continue to work for
+triggering re-inspection. The annotation does not select an inspection
+mode; the mode is always taken from the `inspectionMode` spec field.
```
tuminoid, line 113: Q: when the inspect.metal3.io annotation is set on a host with `inspectionMode: fast`, re-inspection will use fast mode, correct? The BMO 3138 says "it won't be possible to set this mode through the old  inspect annotation", which could be read that as "the annotation can't select the mode" rather than "fast mode can't be triggered via annotation."

tuminoid (commented):
Besides the typo pointed out by Copilot already, one clarifying quesiton.

design/baremetal-operator/out-of-band-inspection.md
```diff
@@ -0,0 +1,299 @@
...
+
+A key advantage of `fast` mode is that no preprovisioning image
+(ramdisk) is needed for inspection. The provisioner must not require
+or wait for a `PreprovisioningImage` when operating in fast mode.
+
+### Risks and Mitigations
+
+**Risk:** Users set `fast` mode on IPMI-only hosts that don't
+support Redfish-based inspection.
+**Mitigation:** The provisioner will return a clear error when fast
```
Rozzii: This we could do with admission web hooks to make it simpler, meaning we can even block setting fast on ipmi . This way we don't need to bother with status fields.

Rozzii (commented):
I support the proposal, this feature would be useful, I have one comment.

design/baremetal-operator/out-of-band-inspection.md
dtantsur: > when the inspect.metal3.io annotation is set on a host with inspectionMode: fast, re-inspection will use fast mode, correct? 

Correct. I'll try to clarify it.

dtantsur: I've updated metal3-io/baremetal-operator#3138, do you think this text needs an update? I think it's pretty explicit about my intentions.

lentzi90 (commented):
/approve

design/baremetal-operator/out-of-band-inspection.md
tuminoid: 3138 is fine, and since I did intepret this right as well, I think its fine. It is not as straightforward as it could be since I had to ask I guess, but lets ship this.

tuminoid (commented):
/lgtm

PR #691, opened 2026-05-27
design: ConsumerSettings field for BareMetalHost

pierrecregut:
The consumerSetting fields aggregates a set of settings of the BareMetalHost whose lifespan should not exceed the binding of the BareMetalHost to a given consumer (made explicit through the use of ConsumerRef).

It is very useful in multi-tenant case (HostClaims) to maintain defaults set by the infrastructure manager.

### DIFF
--- a/design/baremetal-operator/consumer-settings.md
+++ b/design/baremetal-operator/consumer-settings.md
@@ -1,0 +1,183 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# Consumer Settings for BareMetalHost
+
+## Status
+
+provisional
+
+## Summary
+
+This design introduces a new field for BareMetalHost named consumerSettings.
+Its lifespan cannot exceed the lifespan of the consumer it is associated to
+(characterized by the consumerRef field).
+
+consumerSettings contain various fields that were already present in the
+BareMetalHost resource or some of its companion resources (HostFirmwareSettings
+and HostFirmwareComponents). When they are defined, their value takes
+precedence over the equivalent field in the ``spec`` part of the resource.
+
+## Motivation
+
+With the advent of HostClaims resource and the desire to share BareMetalHosts
+between several tenants, it is clear that most BareMetalHost settings must
+be cleared between bindings with different consumer resources.
+
+On the other hand, it is desirable to define defaults values for most of those
+fields. In some cases they contain instance specific values that the consumer
+cannot guess before being bound. Those values are set by the infrastructure
+manager.
+
+### Goals
+
+The goal of this new field is to provide a place for tenants to define their
+settings without overwriting the defaults.
+
+### Non-Goals
+
+This design does not specify how and when specific settings are taken into
+account by Ironic.
+
+Ironic may delay the application of some settings (for example the cleaning
+mode). As the consumer may unbind at any time, this may create some security
+issues.
+
+## Proposal
+
+We introduce a ``consumerSettings`` field. It is an optional object containing
+several fields. The field can be defined only if a consumer binds the host
+(``consumerRef`` field is not nil).
+
+The ``consumerSettings`` field contains at least the following fields:
+
+* userData
+* networkData
+* metaData
+* automatedCleaningMode
+
+All those fields are necessary to implement the HostClaim proposal
+while preserving the existence of infrastructure provided defaults.
+
+The ``consumerSettings`` field may then be extended with the
+following fields:
+
+* raid,
+* hostFirmwareSettings (to override the settings field of the
+  HostFirmwareSettings resource associated to the BareMetalHost),
+* hostFirmwareUpdates (to override the updates field of HostFirmwareComponent
+  resource associated to the BareMetalHost).
+
+## Design Details
+
+### Implementation Details/Notes/Constraints
+
+The constraint between the values of ``consumerRef`` and ``consumerSettings``
+will be implemented in the bareMetalHost webhook. It is very easy to
+make a CEL constraint but the webhook already exists.
+
+Most of the changes are located in the BareMetalHost controller. Before setting
+a field, the controller must check if an override exists and use the later.
+
+The HostClaim controller will use the ``consumerSettings`` field and will never
+touch the now protected fields in the bound BareMetalHost.
+
+The Metal3Machine controller should be modified to use the consumerSettings
+and get a cleaner semantics without corner cases especially for ``xxxData``
+fields.
+
+### Risks and Mitigations
+
+Some field like ``automatedCleaningMode`` do not strictly specify the state of
+the server (It specifies how the server will be cleaned for the *next*
+deprovisioning). When we unbind a server, we must make sure that the disk are
+cleaned before the next user even if the previous user disabled
+cleaning. The existence of the field alone is not enough if we do not know
+that an OS was provisioned and nobody cleaned the disk since.
+
+### Work Items
+
+The handling of the data fields will be done first, followed by the cleaning
+mode. Implementation of other features can be done later.
+
+### Test Plan
+
+For each field it is important to check the transitions from infrastructure
+default to user provided settings and back.
+
+### Upgrade / Downgrade Strategy
+
+The introduction of the new fields should not create any issue. Old controllers
+(unmodified Metal3Machine controllers) will ignore the fields and continue to
+modify the default fields.
+
+The only risk is for controller that *steal* BareMetalHost by resetting the
+``consumerRef`` field. If ``consumerSettings`` field is defined, the update
+will fail.
+
+## Drawbacks
+
+None except the complexity linked to the number of fields.
+
+## Alternatives
+
+### HostClaims are directly handled by BareMetalHost controller
+
+If the BareMetalHost controller is aware of HostClaims:
+
+* it can access a bound HostClaim through the ConsumerRef,
+* check that the binding is legitimate by checking the ``status.baremetalhost``
+  field of the HostClaim,
+* directly access the secrets associated to the HostClaim or any other field
+  in the HostClaim (eg. automatedCleaningMode) and give them priority over
+  the values stored in its own specification.
+
+The main advantages are:
+
+* it avoids copies of secrets.
+* because secrets are not copied in the namespace of the BareMetalHost, they
+  may stay hidden from administrators in charge of the servers.
+
+The main drawbacks/risks are:
+
+* it is HostClaim specific and will not benefit other BareMetalHost consumers
+  (for example Metal3Machine).
+* the expected configuration of the BareMetalHost is less readable for a human:
+  he must access the HostClaim and check its settings.
+* this is a huge change in the logic of the BareMetalHost controller that requires
+  to mediate most access to the BareMetalHost specification through an abstract
+  structure that can give priority to a HostClaim if it exists.
+* there are more risk of reconciliation loops as BareMetalHost and HostClaim
+  need to watch each other resource.
+
+### Infrastructure manager disallow customizing some settings
+
+If the secret pointed by some settings (typically NetworkData) do not respect
+convention of the HostClaim controller (specific label, name convention or both),
+it is considered as set by the infrastructure manager and cannot be overridden
+by the HostClaim controller.
+It is then an error for the end-user to try to set it and the field value can
+only be nil in the HostClaim.
+
+This is clearly less flexible than the proposal but does not require additional
+fields. If this alternative proposal is followed, we could question the way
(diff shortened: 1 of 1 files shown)


pierrecregut:
The main reasons to do a design document are :
* accept that it is needed even if it means that BareMetalHost has a (even more) complex schema.
* agree on the name of the field.
* choose between using the consumerSettings field to let the end user always set all fields or disallow it if the infra manager already set something for a field (typically network). 

Stems from discussions with @dtantsur on https://github.com/metal3-io/baremetal-operator/pull/3152

dtantsur:
I understand that you're not going to be excited about this idea, but I think we need to at least discuss it.

Maybe it's the right time we split the deployment request into a separate CRD that, unlike HostClaims, is tightly coupled to BMH and its controller. Think about this:

```yaml
kind: HostDeployRequest
metadata:
  name: host-0     # matches the BMH name
  namespace: infra # always the same namespace
spec:
  automatedCleaningMode: ...
  image: ...
  customDeploy: ...
  networkInterfaces: ...
  userData: ...
```

Unlike HostClaims, the information is not copied over to the BMH. Instead, the BMH controller uses it directly from HDR resource. The existence of this resource is an request to provision a host, and its deletion is a request to deprovision.

The benefits are:

1. CR creation/deletion are more intuitive than editing fields
2. Less clutter in the BMH, especially as we add fields like networkInterfaces
3. It's finally clear which fields are a part of the inventory and which are instance properties.
4. No issues with stale user data any more. We can even fix the long-standing API quirk that the instance is not rebuilt on userData changes.

I'd even go as far as deprecating instance-related fields on BMH itself.

Thoughts?

/cc @Rozzii @lentzi90 @zaneb

pierrecregut:
If it is named as the BMH (same namespace/name) it is clearly not going to solve the problem of infrastructure defaults which is the problem here. It can also be defined as a HostClaim CRD without the Choose part. If you think that this CR is a replacement for HostClaims, I don't think so : a good part of the power of HostClaims when the hostclaim is not created on the same k8s as the cluster resources is that the choice part is delegated to the cluster where the BMH are. So we are still going to copy secrets in the same way.

dtantsur:
> If you think that this CR is a replacement for HostClaims, I don't think so

Me neither. I'm thinking purely in terms of BareMetalHost resource and its existing complexity, including the difficulty to define which fields belong to instance and which do not.

But now that I've realized that HostClaims can be empty (not deployed), I guess my suggestion does not quite work.

dtantsur:
Per my comment on the meeting, we should discuss the alternative of making the BMH controller aware of HostClaims.

So far, we have assumed that the BMH controller does not know anything about consumers. This is a nice way to split responsibilities, and it makes it easier to reason about the whole system. But I feel like we're hitting the limitations of this approach right now.

If the BMH controller understands HostClaims, we don't need to duplicate any configuration from the claim, the controller will read it. We probably still need to copy over image/customDeploy for sanity reasons, but we won't need new fields for automated cleaning or whatever happens in the future.

design/baremetal-operator/consumer-settings.md
```diff
@@ -0,0 +1,154 @@
...
+while preserving the existence of infrastructure provided defaults.
+
+The ``consumerSettings`` field may then be extended with the
+following fields:
+
+* raid,
+* hostFirmwareSettings (to override the settings field of the
+  HostFirmwareSettings resource associated to the BareMetalHost),
+* hostFirmwareUpdates (to override the updates field of HostFirmwareComponent
+  resource associated to the BareMetalHost).
```
dtantsur, line 73: nit: I'd not commit to this, chances are high we end up with something like HostClaimFirmwareSettings..

design/baremetal-operator/consumer-settings.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Proposal
+
+We introduce a ``consumerSettings`` field. It is an optional object containing
+several fields. The field can be defined only if a consumer binds the host
+(``consumerRef`` field is not nil).
+
+The ``consumerSettings`` field contains at least the following fields:
+
+* userData
```
dtantsur, line 58: I'm curious why you want userData here, it's by definition something a user provides, and an admin cannot have good defaults for it.

pierrecregut: I agree that it is a corner case. But you can always imagine a fixed userData that is only parameterized by user provided metaData. You get something closer to Software/Node as a service than baremetal as a service.... From the previous comments, you probably have guessed that I am not a big fan of infrastructure provided defaults but I do not want to rule out extreme use cases.

pierrecregut:
> BMH controller aware of HostClaims

Yes that could be a solution. HostClaim controller would do the association and record it at both ends as we already do (consumerRef and the BareMetalHost reference we have in status). If the BMH checks both ends we are pretty sure that nobody can bind to a hostClaim and its secrets outside the choose logic of HostClaims (today there was no requirement for the BMH to check the consumerRef was legitimate because of the copy of secrets). So we should not have security issues like there was with cross namespace secret references. This can have interesting consequences (ie simplifications) for the very imperative annotations like reboot or even detach.

pierrecregut:
From an implementation point of view BMH controller would watch HostClaims with a HostClaimToBaremetalHost handler that creates the request from Status.BareMetalHost. I am still a little wary of creating a watch loop because the current implementation has a watch from HostClaims over BMH but I guess we can drop it if we accept that the BMH controller and the HostClaim controller can both update HostClaim conditions (and may be some other status fields - to be checked).

design/baremetal-operator/consumer-settings.md
dtantsur: The problem is: userData and networkData are specific to the image, which is something a user provides. If you take a CoreOS image, providing a cloud-init userData or openstack networkData won't give you anything.

pierrecregut: I would tend to agree with you. That is why Kanod or Sylva never use default data. Specific BMH info is written in their metadata and collected by user supplied templates. But in practice the choice of vetted image may be a singleton or at least have so similar characteristics that they can use the same initialization data.

dtantsur:
> I am still a little wary of creating a watch loop because the current implementation has a watch from HostClaims over BMH but I guess we can drop it if we accept that the BMH controller and the HostClaim controller can both update HostClaim conditions

There may be a middle ground there. If we drop the idea of admin-provided userData/networkData/metaData and assume that corresponding BMH fields are fully owned by a HostClaim, then the association and secret copying logic can fully state in the HostClaim controller.

The BMH controller will only (at least for now) care about automatedCleaningMode from the HostClaim. That does not require updating conditions, to the best of my knowledge. Furthermore, it probably doesn't require watching HostClaim events at all since automatedCleaningMode is only used at a few points in the provisioning workflow. It does not have to be constantly up-to-date in Ironic, only before any deprovisioning (or on enrollment, but it is irrelevant to HostClaims).

pierrecregut:
It seems there is a consensus against consumerSettings for userdata/metadata/networkdata so there is probably no reason to keep this design as is. But I have pushed a new version with the alternative proposed by @dmitry of using the HostClaim as the source of settings directly in the BMH. Dmitry, you suggested to keep the copy logic for secrets and other settings, but in fact, there may be good reason to avoid the copy of secrets and where this new design would provide new security guarantees. With HostClaim initial proposal we have hidden BMC credentials from end-user. But if we avoid copying secrets we could also hide initialization data from infrastructure managers who only have access to the namespace of BareMetalHost. Clearly with access to Redfish and console, they will always have a lot of power on the configuration of the servers and so its content, but the added isolation given by namespace separation would be nice.

I would really like to avoid huge changes on the bareMetalHost controller. The original Hostclaim controller implementation worked with standard BMO release. But if done well most of the changes can be concentrated in host_config_data.go especially if the hostclaim controller set the name in the secret reference hold by the bmh to a forbidden value like -hostclaim- (this would also directly help the end user !).

design/baremetal-operator/consumer-settings.md
```diff
@@ -0,0 +1,183 @@
...
+None except the complexity linked to the number of fields.
+
+## Alternatives
+
+### HostClaims are directly handled by BareMetalHost controller
+
+If the BareMetalHost controller is aware of HostClaims:
+
+* it can access a bound HostClaim through the ConsumerRef,
+* check that the binding is legitimate by checking the ``status.baremetalhost``
```
zaneb, line 133: This is a non-starter because it is not enforced by RBAC.

design/baremetal-operator/consumer-settings.md
zaneb: I agree with Dmitry here. userData is clearly a consumer thing. metaData and preprovisioningNetworkData are clearly an admin thing.

networkData is weird. In OpenStack it is set by the cloud and not by the consumer, so clearly an admin thing. But in Metal³ we historically haven't had different roles for admin/consumer. So there's an argument that it should be overridable by the consumer, and there have been suggestions in the past that there are people relying on this.

In https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit?tab=t.0#heading=h.pi5obn43an36 I put networkData on the admin side, but the comments mention that we might need it on both sides.

Since then we added preprovisioningNetworkData, which is now the default for networkData if it is not supplied. So you could even argue that on the admin side there is sufficient flexibility already and this _only_ needs to be on the consumer side. (I wouldn't make that argument myself though, because admins may have their own reasons for not wanting to expose the preprovisioningNetworkData to consumers.)

zaneb (commented):
> Maybe it's the right time we split the deployment request into a separate CRD that, unlike HostClaims, is tightly coupled to BMH and its controller. ... Unlike HostClaims, the information is not copied over to the BMH.

I think this would be [a mistake](https://groups.google.com/g/metal3-dev/c/qRZnxILVTZk/m/ANf0NP9tAgAJ). Copying the data into the existing BMH structure allows us to keep compatibility with existing clients and also the existing controller implementation. As far as we know, this may rely on all kinds of guarantees about atomic updates. Changing both at the same time would be extremely high-risk. IMHO we should focus on creating the API we want incrementally, by building new controllers that copy data into the BMH, first. Only then could we consider dropping the provisioning functionality from the BMH itself and reimplementing the controller to fetch data from other objects (but I would bet that we won't find much value in doing so).

That said, I don't love the proposal here either. It's confusing to add 2 copies of the fields to the (already bloated) BMH. The override here is the wrong way round IMHO. There should be a separate CR for the default values that get copied into the BMH when the HostClaim does _not_ override them. Potentially this could be added to HostDeployPolicy.

Doing that would bring it pretty close API-wise to what I called `BareMetalAllocation` in the [original API decomposition proposal](https://docs.google.com/document/d/1FL8oA0_WNcPdiC-0zrToinl8aol-NPQUjWe-OkcvTqA/edit?tab=t.0#heading=h.pi5obn43an36). Implementation-wise, though, my plan was that the settings would bubble up one resource at a time to avoid races. So the specs of HostClaim and HostDeployPolicy would be combined and written to the status of HostDeployPolicy, and then the status of HostDeployPolicy would be copied to the Spec of the BMH.

As an aside, I am a big believer of the idea that people who have concrete use cases that they are building real-world implementations of should come up with the final design. But TBH, having done the hard work of figuring out all the stuff that needs to be done to make a multi-tenant cloud, it feels like a missed opportunity when a new design runs into issues immediately after implementation that were already considered and designed for originally, and the result is another entirely new design without reference to existing work. And it's also kinda exhausting.

design/baremetal-operator/consumer-settings.md
dtantsur: It's only a sanity check though, the real security check is via consumerRef. If you can change it, you have access to the BMH anyway.

design/baremetal-operator/consumer-settings.md
dtantsur: > I wouldn't make that argument myself though, because admins may have their own reasons for not wanting to expose the preprovisioningNetworkData to consumers

This is a good point. When combined with the ToR switch work, preprovisioningNetworkData may be quite different from networkData and an operator may not want it to be exposed to a user.

lentzi90 (commented):
> The override here is the wrong way round IMHO. There should be a separate CR for the default values that get copied into the BMH when the HostClaim does not override them. Potentially this could be added to HostDeployPolicy.

I agree with this and would suggest investigating what it would mean to do it through the HostDeployPolicy. It would make sense at least in my mental model of this API that admins manage the settings in this way, just as they do with other things.

I also want to stress one thing particularly: this proposal feels like trying too much at once. We don't have a minimal implementation of HostClaims working yet and we already try to extend with new things. Please focus on the minimum we need to demonstrate the functionality. IMO we could just go with the principal that HostClaims control the `{user|network|meta}Data` and the rest is controlled by admins to begin with. When we have that working we can start thinking about features to make it more flexible.

PR #692, opened 2026-05-27
Design proposal for firmware inventories

dtantsur:
Generated-By: Claude Code (commercial version)

### DIFF
--- a/design/baremetal-operator/firmware-inventory.md
+++ b/design/baremetal-operator/firmware-inventory.md
@@ -1,0 +1,499 @@
+<!--
+ This work is licensed under a Creative Commons Attribution 3.0
+ Unported License.
+
+ http://creativecommons.org/licenses/by/3.0/legalcode
+-->
+
+# FirmwareInventory: Centralized Firmware Image Registry
+
+## Status
+
+provisional
+
+## Summary
+
+We introduce a new Custom Resource, FirmwareInventory, that serves as a
+centralized registry mapping firmware versions to their download URLs for
+a given hardware component. A FirmwareInventory resource targets a
+specific component type (such as `bmc`, `bios`, or `nic`) and can
+optionally select applicable hosts based on their hardware properties.
+
+## Motivation
+
+Today, requesting a firmware update through HostFirmwareComponents
+requires the user to specify both the component name and the full
+download URL for each update. This has several drawbacks:
+
+- **No version tracking.** There is no structured way to record which
+  firmware versions are available for a given component or hardware
+  class. Administrators must maintain this information externally.
+
+- **No hardware-aware scoping.** A firmware image that applies only to
+  Dell BMCs cannot currently be distinguished from one that applies to
+  HPE BMCs at the API level. The administrator must rely on conventions
+  or external tooling to prevent misapplication.
+
+- **Future multi-tenancy issue.** Eventually, we'll need to support firmware
+  updates via HostClaims. Infrastructure administrators may decide to prevent
+  arbitrary images from being applied to their machines, limiting users to
+  a curated list of compatible ones.
+
+### Goals
+
+- Provide a Kubernetes-native resource for declaring available firmware
+  versions and their download locations per component type.
+
+- Allow HostFirmwareComponents to request updates by version alone,
+  with the controller resolving the URL from a matching
+  FirmwareInventory.
+
+### Non-Goals
+
+- Defining how firmware images are built, signed, or distributed.
+  FirmwareInventory only records the mapping from version to URL.
+
+- Automatic firmware update orchestration across fleets. Rollout
+  strategies are the responsibility of external tooling, not
+  FirmwareInventory itself.
+
+- Multi-tenant firmware update policy. Integration with HostClaims,
+  and HostDeployPolicy is out of scope for this proposal and will be addressed
+  separately.
+
+- All NICs will be grouped under the same generic `nic` component, there won't
+  be supported for specific devices (`nic:<ID>`). Given that Redfish `Targets`
+  property is not consistently implemented across vendors, Ironic updates all
+  suitable NICs anyway.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Infrastructure Admin Publishes Firmware Catalog
+
+As an infrastructure administrator, I manage a fleet of Dell and HPE
+servers. I want to declare the available BMC and BIOS firmware versions
+for each vendor in a structured way, so that my team can request
+updates by version without managing URLs.
+
+I create two FirmwareInventory resources in the infrastructure
+namespace:
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: FirmwareInventory
+metadata:
+  name: dell-bmc
+  namespace: infra
+spec:
+  component: bmc
+  hostSelector:
+    matchHardwareData:
+      systemVendor.manufacturer: 'Dell'
+  versions:
+  - version: 1.2.3
+    url: https://firmware.example.com/dell/bmc-1.2.3.img
+  - version: 2.0.0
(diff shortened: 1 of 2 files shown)


pierrecregut:
Firmware inventories are a very nice addition ! But I am not sure that the hostSelector approach is really needed or at least always the right choice. We are speaking of a very fixed relation. Given a brand and a generation of server you should always target a given inventory. If the vendor has the bad idea of changing some naming you relied on for the selector, you should not be impacted. So may be we could also have an inventory field in the HostFirmwareComponent for people who do not like to rely on magic or when their hardware lie about their name.

pierrecregut:
It may be hard to support and even harder to schedule so that there is not too much pressure on the updaters, but a notion of latest for the version in the HostFirmwareComponent could be nice. On the other hand a kind of renovate for HFC and inventories could be a much better idea.

dtantsur:
> So may be we could also have an inventory field in the HostFirmwareComponent for people who do not like to rely on magic or when their hardware lie about their name.

We could, but keep in mind:
1. HFCs are usually created automatically the controller using information from Ironic
2. When we start linking this proposal with HostClaims, users won't be able to choose an inventory

pierrecregut:
HostClaims should only expose versions. I was thinking of an annotation on the BMH defining the HostInventory but we can use a label selector in that case.

pierrecregut:
hostSelector work as conjunction, for the disjunction we use the fact that we can have multiple hostSelectors. 
When it is used with labels you can ensure that you do not need so many disjunctions (that's also why it works well with  but when you match on hardwareData you probably want to say that this inventory is for a Dell (ok single value) with model  matching R[0-9][4-6]0. With the way it is structured today you may end up replicating and so maintaining different maps version -> url.

design/baremetal-operator/firmware-inventory.md
```diff
@@ -0,0 +1,480 @@
...
+## Status
+
+provisional
+
+## Summary
+
+We introduce a new Custom Resource, FirmwareInventory, that serves as a
+centralized registry mapping firmware versions to their download URLs for
+a given hardware component. A FirmwareInventory resource targets a
+specific component type (such as `bmc`, `bios`, or `nic`) and can
```
karampok, line 19: for NICs: how  this proposal would work?
 because same vendor "dell" might have unique bmc/bios firmware inventory url map, but a wide range of nics. Like dell server A has intel 8100, and dell server has mlx cards.

pierrecregut: May be the proposal should distinguish component types (componentType) from component names (component as used today). This is only relevant for nics. A component name for a network card is the concatenanation of its component type  "nic:" and a vendor dependent name that identifies a card. But it is almost unusable from what I have seen with 3 vendors:

* "nic:slot-x" on  a Lenovo where x identifies a pci slot,
*  "nic:DC07A000" for the first embedded HPE card whatever this card is and third party cards are not seen at least on Gen10...,
*  the only sensible output is on Dell (nic:NIC.Embedded.x or nic:NIC.Slot.x where x is an index starting at 1). 

If I understand Ironic source code, this has strictly no use for the update itself: only the type (bmc, bios or nic) is used to decide a reboot policy required after/during the update task and it would make the definition of inventory almost impossible to manage: the slot may move, there may be multiple cards on a server and without real support of targets (the target field in simpleUpdate action) at Ironic level (I think it only applies to target identifying a ComputerSystem, neither device or firmware instance are supported @dtantsur do you agree ?), all cards with the firmware are updated. In fact, you can probably update a storage controller firmware if you pretend it is a nic but it will not be displayed in the results.

This is also why I would prefer a way to state explicitly which Inventory to use:

* either directly in the update element,
* or separately as proposed in an earlier comment but then I need a sensible name which could be the full component  name as found by Redfish. 

But in the FirmwareInventory I would only specify the component type (bmc, bios, nic), not names.

karampok: thanks for the answer.

design/baremetal-operator/firmware-inventory.md
```diff
@@ -0,0 +1,480 @@
...
+  updates:
+  - component: bmc
+    # Existing field: direct URL (still supported for backwards
+    # compatibility)
+    url: <string>
+    # New field: version to resolve via FirmwareInventory
+    version: <string>
+```
+
+When `version` is set and `url` is empty, the HFC controller resolves
```
tuminoid, line 248: Hmm, user normally owns spec and controller status. Here we write-back to spec?

design/baremetal-operator/firmware-inventory.md
```diff
@@ -0,0 +1,480 @@
...
+  # Required. At least one version entry.
+  versions:
+  - version: <semver or vendor string>
+    url: <https:// URL>
+```
+
+**Field details:**
+
+- `component`: Follows the same validation rules as HostFirmwareComponents.
+  Currently must be `bmc`, `bios`, or `nic` (which also covers prefixed
```
tuminoid, line 224: FirmwareInventory with component: bmc matches HFC entries by exact match proabably,  but for NICs, does component: nic:eth0 in an inventory match only HFC entries for that exact NIC? Or should the inventory support a category-level nic that resolves for all nic:* entries? Its only component type with dynamic identifiers right now.

design/baremetal-operator/firmware-inventory.md
```diff
@@ -0,0 +1,480 @@
...
+status:
+  updates:
+  - component: bmc
+    url: https://firmware.example.com/dell/bmc-2.0.0.img
+  inventory:
+    name: dell-bmc
+    namespace: infra
+```
+
+If no matching inventory is found or the requested version does not
```
tuminoid, line 163: Should it be made explicit that Valid must be False (or unset) while resolution is pending (version set, URL empty, no resolution attempt yet). Otherwise there's a race: user sets version, HFC reconciler hasn't run yet, ChangeDetected=True + Valid=True , and the BMH controller attempts to consume an entry 
with an empty URL during Preparing.

design/baremetal-operator/firmware-inventory.md
lentzi90: True, this is a bit messy. We have some objects that are completely controller owned, like IPAddresses, but here we get a mix. I wonder if we could turn this into a completely controller managed object that the user controls through higher abstractions? Say a `FirmwareUpdatePolicy` that drives `HostFirmwareComponents` across multiple hosts?

design/baremetal-operator/firmware-inventory.md
dtantsur: > without real support of targets (the target field in simpleUpdate action) at Ironic level (I think it only applies to target identifying a ComputerSystem, neither device or firmware instance are supported @dtantsur do you agree ?), all cards with the firmware are updated

Correct. Targets are inconsistently implemented so far, and we're working with DMTF to sort the standard's side first.

> But in the FirmwareInventory I would only specify the component type (bmc, bios, nic), not names.

Yep, I agree.

design/baremetal-operator/firmware-inventory.md
dtantsur: > does component: nic:eth0 in an inventory

I don't quite get the question, the line you're commenting on does not allow `nic:eth0` as an acceptable value.

design/baremetal-operator/firmware-inventory.md
dtantsur: I'm not super fond of writing to spec, but I think it's the least confusing UX I can come up with.

And I don't think we're ready for a complete redo of firmware updates :)

dtantsur: Given that the URL is only needed once, maybe we don't need to store it at all?

dtantsur: I went with a new status field, please check.

PR #701, opened 2026-06-18
🌱 Remove mariadb-image references from ironic-standalone-operator

migiFi:
Remove references to archived mariadb-image repo

Direct users to MariaDB Operator.

### DIFF
--- a/design/ironic-standalone-operator.md
+++ b/design/ironic-standalone-operator.md
@@ -49,8 +49,7 @@ Explicitly not planned:
   minor things.
 - Support for versions of ironic-image predating this proposal (e.g. the ones
   containing ironic-inspector).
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
I may lack some context here so please correct me if I'm wrong.
In my eyes this is not needed. It is a design document describing what we wanted to implement at some point in time. It may be valuable to update the design as major changes happen, but I am not sure it is needed for this. There was already a note added below in https://github.com/metal3-io/metal3-docs/commit/b9f17f550559780bfdf039718ddea95d55aa40b8 to clarify the fate of `IronicDatabase`.

Rozzii (commented):
/cancel
Not needed, but thanks for the contribution @migiFi

PR #704, opened 2026-06-23
:seedling: Fix PR Quick start action permissions

peppi-lotta:
I noticed PR job is not working because of miss match in permissions. Update PR action's permission  to match the workflow that is being called from it.


lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #708, opened 2026-07-02
Add small section explaining block move annotations

UncivilizedGhost:
Explaianion of annotations that have been added to CAPM3 for pivoting


docs/user-guide/src/capm3/pivoting.md
kashifest: I think we should update this section as well and tell that its either owner reference chain or clusterctl label

kashifest (commented):
/approve

lentzi90 (commented):
/lgtm

PR #709, opened 2026-07-03
Design: new API for imperative operations

dtantsur:


design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+- Further imperative operations (credential rotation, secure boot
+  configuration) would require even more annotations.
+
+### Goals
+
+- Provide a single CRD for all imperative actions on provisioned
+  hosts (reboot, rebuild, service).
+- Surface progress and errors through standard Kubernetes conditions.
+- Support cancellation of running operations where possible.
+- Deprecate the plain `reboot.metal3.io` annotation.
```
MahnoorAsghar, line 52: Deprecate, but not replace? (L58)

iurygregory: L58 looks correct to me.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+- Triggering re-inspection through HostOperation (inspection applies
+  to `available` hosts, not `provisioned`).
+- Extracting the servicing or rebooting logic into a separate controller.
+- Rename the `onReboot` value of HostUpdatePolicy to something more suitable.
+
+## Proposal
+
+A new CRD `HostOperation` allows users to request an imperative action
+on a provisioned BareMetalHost. Each resource represents exactly one
+action. The spec is immutable after creation (enforced by a webhook),
```
iurygregory, line 75: I do like the idea that is immutable after creation!
So in case the user wants to change this in the future for any reason, the approach is to delete the `HostOperation` CRD ?

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+  startedAt: "..."
+  finishedAt: "..."
+```
+
+Key rules:
+
+- The `host` and `operation` fields are immutable after creation.
+- Exactly one operation must be specified; the webhook rejects zero
+  or multiple.
+- The `cancelled` field can only transition from false to true, never
```
iurygregory, line 147: In case the user creates the CR with `cancelled` set to true, do we notify them somehow? Or we don't create the CR in case the spec contains `cancelled: true`? 
(Maybe I'm overthinking ....)

design/baremetal-operator/host-operation.md
dtantsur: We may differ in terminology here. For me, a deprecation is an intention to replace. Replacing would mean that the old thing stopped working.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+The current mechanism for triggering actions on provisioned hosts
+relies on the `reboot.metal3.io` annotation and on field mutations in
+BareMetalHost. This has several shortcomings:
+
+- Rebuilding an instance with new user/network/meta data requires
+  removing the `image` field, waiting for deprovisioning to start, and
+  re-adding it — an error-prone sequence.
+- There is no way to rebuild without going through cleaning.
+- Servicing (firmware updates, BIOS changes) is coupled to the reboot
+  annotation, forcing a reboot even when one is unnecessary (e.g. BMC
```
jacob-anders, line 36: big thumbs up to this improvement

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+  firmware updates).
+- A HostUpdatePolicy that disallows servicing provides no feedback to
+  the user — changes are silently deferred.
+- Cancelling servicing requires modifying HFS/HFC in non-obvious ways.
```
jacob-anders, line 40: I very much support this, current mechanism has proven problematic (prone to race conditions) and difficult to debug

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+  name: ostest-worker-0-abcdef
+  namespace: openshift-machine-api
+spec:
+  host: ostest-worker-0
+  operation:
+    reboot:
+      mode: soft|hard
+      force: false
+    service:
+      mustReboot: false
```
jacob-anders: would it be a good idea to consider exposing a parameter here which would allow the operator to select single-reboot flow in multi-component updates? 
The idea behind it would be to determine whether we send allow_grouping_reboots=True to Ironic
(working name from https://review.opendev.org/c/openstack/ironic-specs/+/992208/14/specs/approved/multi-component-firmware-updates.rst#111)
It feels like this is separate from mustReboot - or am I mistaken?

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+### Service operation
+
+Triggers Ironic servicing for firmware updates and BIOS settings
+changes. Differences from the current annotation-driven flow:
+
+- **Reboot control**: unless `mustReboot` is true, the decision to
+  reboot is left to Ironic. Some components (e.g. BMC firmware) can
+  be updated without rebooting the host. When `mustReboot` is true,
+  servicing starts with a power-off, matching the current behavior.
+  Note that it will cause a reboot even if no changes are pending.
```
jacob-anders, line 313: further to my earlier comment - let's think what's the best place to put in controls requesting single-reboot flow in Ironic, if desired.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,428 @@
...
+  `cancelled: true`, and deletion blocks on a finalizer until
+  rebuilding completes. This mirrors a current BMO limitation where
+  the only way to abort a failing deployment is to clear the
+  `image`/`customDeploy` fields.
+- **Service**: follows the existing servicing abort flow — Ironic
+  aborts at the next safe step, then the host is booted back. The
+  `Progressing` condition shows reason `Cancelling` during this
+  period. Cancelling a service operation does not revert the
+  underlying HFS/HFC changes — they remain pending and may be
+  applied by a future service operation.
```
jacob-anders, line 360: not so much a comment on this enhancement but more of a mental note for futher discussion: in an event where a firmware update Task/Job is running on the BMC but reboot hasn't been initiated, would we want to be able to abort this and prevent the reboot?

In some cases, if the user realised a mistake has been made (e.g. attempting to flash wrong firmware image) this may be the last call to **EDIT: prevent triggering** ~~trigger~~ a reboot from which the machine may never recover. Most of this would be on the Ironic side, I am happy to investigate this. it would be good to discuss where we want to draw the line in such cases.

jacob-anders (commented):
Thank you for this enhancement, Dmitry. My initial comments, mostly servicing-centric.

design/baremetal-operator/host-operation.md
dtantsur: Can we cancel the task without consequences? I guess that's the key question.

jacob-anders: Unfortunately the answer is "it depends". If memory serves: on Dell, we can, because the firmware updates are staged and Lifecycle Controller will do the actual flashing after reboot. So if we deleted the Jobs, updates should just not happen (to be lab-validated for 100% answer).

However. IIRC, on HP, flash happens in real time and reboot just reloads the firmware. So once the Task starts, damage is already done. But - if reboot is stopped, the operator can at least  try to re-flash the correct firmware I'd hope. Hard to validate without breaking any hardware but in general this is the escape hatch I was thinking of.

design/baremetal-operator/host-operation.md
dtantsur: > It feels like this is separate from mustReboot

I concur. It feels like a property of the host (in which case it goes to HostUpdatePolicy) or of the updates themselves (in which case it goes to HostFirmwareComponents).

design/baremetal-operator/host-operation.md
dtantsur: Yeah, may be worth denying this in the webhook.

design/baremetal-operator/host-operation.md
dtantsur: Yeah, wait for it to finish or cancel it.

design/baremetal-operator/host-operation.md
jacob-anders: > > It feels like this is separate from mustReboot
> 
> I concur. It feels like a property of the host (in which case it goes to HostUpdatePolicy) or of the updates themselves (in which case it goes to HostFirmwareComponents).

I like linking it to the updates themselves. Reason - here is example update flow
1) BMC, BIOS, NIC. Should work in a single reboot.
2) (future - we don't support CPLD/TPM yet) BMC, BIOS, CPLD.  On Dells I tested CPLD needs a reboot separate from other components. So this one time the operator may not want the single-reboot-flow.

Since we don't have a set plan for CPLD/TPM update this is not a show-stopper but it may be wise to futureproof it. In any case we can disregard this in the scope of the Imperative Opearations spec.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+trigger servicing. Users who want to apply pending firmware or BIOS
+changes during a reboot should use the service operation.
+
+The plain `reboot.metal3.io` annotation will be deprecated. While it remains
+supported, adding it when no operation is running will cause the BMH controller
+to create a HostOperation of type `service` with `mustReboot: true`
+automatically. If any operation is already running, the annotation is dropped
+silently.
+
+The coordinated (suffixed) form `reboot.metal3.io/<key>` is not
```
pierrecregut: Keeping annotations and operations at the same time seems to create rather complex interactions. Is coordinated/phased reboot different from a 'halt' operation ? To keep spec immutable, you would need to either cancel or delete it to leave the halt state.  One of the main benefit is that ordering by creation date applies to all operations (halt included).

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+### Custom Resource Definition
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: HostOperation
+metadata:
+  name: ostest-worker-0-abcdef
+  namespace: openshift-machine-api
+spec:
+  host: ostest-worker-0
```
pierrecregut: Even if it is a non goal, what kind of API will be used for hostclaims/m3machines ?Functionally we want a CR whose spec is pushed as a HostOperation on the bound BMH and whose status is synced with this HostOperation. Either we create a new CRD HostClaimOperation with exactly the same fields or we use one single type and specify the kind of the target. In that case, we may prefer:
```yaml
spec:
   host:
     name: otest-worker-0
     kind: BareMetalHost
```

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+### Conflict resolution
+
+Multiple HostOperations may exist for the same host simultaneously.
+The BareMetalHost controller resolves conflicts by sorting pending
+operations by creation timestamp and recording the oldest one in
+`currentOperation`. The operation controller does not proceed (and
+does not update `startedAt`) until the operation's name appears in
+the corresponding host's `currentOperation`.
+
+Failed operations are considered finished: the BMH controller clears
```
elfosardo: isn't this a bit dangerous for a sequence  like service -> reboot ?
we should probably at least document that we can't assume success

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+Key rules:
+
+- The `host` and `operation` fields are immutable after creation.
+- Exactly one operation must be specified; the webhook rejects zero
+  or multiple.
+- Each operation has optional parameters that are specific to operation itself,
+  not to the host it's running on and not to the applicable HostUpdatePolicy.
+- The `cancelled` field can only transition from false to true, never
+  back.
+- `ttlSecondsAfterFinished` controls automatic garbage collection
```
Rozzii, line 149: It might make sense to always keep the last operation or the N number of last operations for transparency purposes rather than cleaning up. I would rather go with an object retention limit that the user could use to specify how many of them  to keep (of the latest ones) and I would also introduce a label or annotation that could  be placed on the object to retain it even if it would exceed the retention limit.
The object retention limit could range from zero to max int value, zero would mean retain all, default would be 1 because IMO it is useful to know  the last host op status, even if only 1 was retained that still could be deleted manually without an error. I would use "finishedAt" for  indexing based on sort-able timestamp as I wrote above.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+status:
+  conditions:
+  - type: Succeeded
+    status: "False"
+    reason: InProgress
+  - type: Progressing
+    status: "True"
+    reason: WaitPowerOff
+  startedAt: "..."
+  finishedAt: "..."
```
Rozzii, line 135: Make the `startedAt` and the `finisedAt` use ISO 8601 time format so these can be used for inexing/sorting and would help CAPM3 avoid expensive time fomat conversions and time measurement and also could be used with my alternative approach that would replace `ttlSecondsAfterFinished `as explained below.

design/baremetal-operator/host-operation.md
Rozzii: I don't like this approach, this could mask serious failures and/or could further degrade the state of the physical host "one bad action followed by a worse". IMO a HostOp failure should be either always a hard failure and stop the execution of the next ops or  at  least it should be configurable on either BMO or BMH level, so users can decide to have "hard failures or soft failures" when the HostOp queue is executed.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+  provisioner calls and updates the HostOperation's conditions to
+  reflect progress. This replaces the current logic that checks for
+  the reboot annotation and calls servicing inline.
+
+- For the rebuild operation with `cleaning: false`, the Ironic
+  provisioner will use the `rebuild` provision target (similar to
+  `nodes.TargetRebuild`). This endpoint is already available in Ironic
+  but not currently used by BMO.
+
+- The `ttlSecondsAfterFinished` field will be handled by the
```
Rozzii: As wrote above I don't like the approach of measuring seconds and removing all the ops objects.

Rozzii (commented):
I believe we need this HostOperation CR but I have a few comments , i don't agree with everything described in this design.

design/baremetal-operator/host-operation.md
dtantsur: The main issue with that is that different operations may have different retention policies. Like, I probably don't care about keeping reboots, but I might prefer to keep all servicing requests. How to achieve that with your suggestion?

Note that the TTL approach is copied from Jobs, so it has prior art in Kubernetes. Are you aware of any Kubernetes controllers that keep N last resources?

design/baremetal-operator/host-operation.md
dtantsur: Then we need another state on HostOperation that says "acknowledged, move on", right?

design/baremetal-operator/host-operation.md
dtantsur: That's implied here

design/baremetal-operator/host-operation.md
dtantsur: Coordinated reboot waits for a 3rd party to remove the annotation. I'm not sure how to express this in HostOperations. Any suggestions?

design/baremetal-operator/host-operation.md
dtantsur: I guess we could have

```yaml
spec:
  optional: false/true
```

Which both defines whether the operation blocks next operations and can be used to unblock next operations (by setting it to `true` after the fact).

design/baremetal-operator/host-operation.md
MahnoorAsghar: The user can also sequence one after another by creating multiple HostOperations...

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+| `WaitPowerOn` | Waiting for the host to power on |
+| `Servicing` | Ironic servicing is running |
+| `Rebuilding` | Re-provisioning is running |
+| `Cancelling` | Cancellation is in progress |
+
+**NOTE:** the list of reasons may be expanded further.
+
+### New BareMetalHostStatus field
+
+A new `currentOperation` field in `BareMetalHostStatus` will hold the
```
MahnoorAsghar: Nit-like suggestion, lets call it currentHostOperation to make it very clear it relates to this CR

design/baremetal-operator/host-operation.md
MahnoorAsghar: I'm not sure I understand this fully, but...wait for the 3rd party to remove the HostOperation CR? Is that not possible? (I do kind of agree that keeping both annotation and CR will be confusing)

MahnoorAsghar: Actually now that I've re-read https://book.metal3.io/bmo/reboot_annotation.html, I'm not sure the co-ordinated annotation maps very well to the CR format. i.e. if all users can see all the coordinated reboot annotations in one place in the BMH, that is easier than...looking at all the HostOperation CRs that indicate coordinated reboots for one BMH.

design/baremetal-operator/host-operation.md
MahnoorAsghar: Perhaps later we can add a field alongside mustReboot; a boolean called singleReboot?

design/baremetal-operator/host-operation.md
MahnoorAsghar: 'blocking' might be a better name instead of 'optional', perhaps

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+  controller should set owner references on HostOperation resources
+  so that they are garbage-collected when the host is deleted.
+
+- **Backward compatibility**: consumers that rely on the reboot
+  annotation need a migration path. Mitigation: the plain annotation
+  continues to work (auto-creating a HostOperation) during a
+  deprecation period.
+
+- **Rebuild cancellation**: the inability to cancel rebuilds may
+  frustrate users. Mitigation: document the limitation clearly and
```
MahnoorAsghar, line 423: Question: if it is possible to do, why not make it part of the scope? Is it because it will cause the scope to become too large for this implementation?

design/baremetal-operator/host-operation.md
Rozzii: For me  the spec.blocking: false/true approach , ofc in this case the value would need to be set to false to "unblock" the operation queue manually.

dtantsur: If we have `blocking`, it will default to `false` organically, is it what we want?

design/baremetal-operator/host-operation.md
lentzi90: Not sure about core Kubernetes, but cert-manager has something like this. [Certificates](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.Certificate) have a `revisionHistoryLimit`: "The maximum number of CertificateRequest revisions that are maintained in the Certificate’s history."
Just noting for reference. I don' have a strong opinion on which way we go here.

design/baremetal-operator/host-operation.md
lentzi90: If I understand @pierrecregut correctly, he is saying that we could have a HostOperation with `operation: halt`. This powers off the BMH and keeps it powered off. To power on, you either cancel the operation or delete the HostOperation.

With this I think it would be possible to deprecate also the coordinated reboot. The initiator creates the HostOperation. Each 3rd party entity adds finalizers to it. Once they are done, they mark the HostOperation for deletion and removes their own finalizer. When all finalizers are gone, the HostOperation goes away and the BMH is powered on.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,470 @@
...
+  operation succeeds with reason `NoChanges`.
+
+### Conflict resolution
+
+Multiple HostOperations may exist for the same host simultaneously.
+The BareMetalHost controller resolves conflicts by sorting pending
+operations by creation timestamp and recording the oldest one in
+`currentOperation`. The operation controller does not proceed (and
+does not update `startedAt`) until the operation's name appears in
+the corresponding host's `currentOperation`.
```
lentzi90: Here I think we have room for improvement (perhaps as a separate extension). I find it hard to see a situation when I would want to execute HostOperations blindly based on creation timestamp. If some sequence is needed, I want to do it in a controlled way.

Perhaps we could take some ideas from the [system-upgrade-controller](https://github.com/rancher/system-upgrade-controller)? We could have a higher level `HostOperationPlan` that specifies the HostOperations as templates along with host selectors, sequence/dependency chain, concurrency and such. A controller could then create the HostOperations, following the plan, and pause/abort as needed. This would also allow us to do HostOperations over multiple BMHs in parallel.

lentzi90 (commented):
Thank you for this proposal! I think it is a clear improvement on the current state.
I have no blockers, but a couple of things to raise for discussion.

This does not improve (or worsen) the situation regarding RBAC for operations. In K8s this is currently an area that is causing some pain around exec and port-forward privileges that are impossible to control in a fine grained way.
The RBAC for us here will be that the user either can create HostOperations or not. There is no way to say if they should be allowed to reboot but not do a rebuild. With policies we may be able to forbid certain operations completely, but we won't be able to say that Alice should be allowed to rebuild while Bob should not.

As mentioned, we do not change the status quo with this proposal. I just want to raise the topic in case some of you see a reason to include this in the design. The alternative with more control would be unique custom resources for each operation I guess? For example, HostReboot, HostRebuild, etc. Honestly, I am not sold on that.

Secondly, I want to point out a potential future extension. What about doing multiple HostOperations in parallel? And controlling a sequence of HostOperations? See my comment below for some more. I don't think it needs to be part of this though. It can be a future thing if we want it.

dtantsur:
> What about doing multiple HostOperations in parallel? 

It's difficult-to-impossible with Ironic. What's the use case you have in mind?

design/baremetal-operator/host-operation.md
dtantsur: You may be overestimating the need for running several operations in parallel, to be honest. My design has to account for this situation, but I really wonder who and why would do it.

One version of this proposal had operations as a list, not as a singular item. I can revert to it if other folks don't object, but it becomes much more complicated when it comes to error reporting and handling (same for a Plan resource).

lentzi90:
> > What about doing multiple HostOperations in parallel?
> 
> It's difficult-to-impossible with Ironic. What's the use case you have in mind?

I was thinking rolling out firmware changes or rebuilding multiple BareMetalHosts. So not parallel as in multiple operations on one host, but parallel as in multiple hosts simultaneously. I'd imagine that would be quite handy if you have dozens of hosts. Even better if we can control it, e.g. do 2 at a time, stop if there are errors.

design/baremetal-operator/host-operation.md
dtantsur: I'll put something like HostOperationPlan in the future ideas. I don't think landing this design precludes having a plan in the future.

dtantsur:
> So not parallel as in multiple operations on one host, but parallel as in multiple hosts simultaneously

Ah. I speculated a bit about a similar resource in https://gist.github.com/dtantsur/2ff8ab698bcf634e3580b3a75636e981#proposal-reboots-and-updates, but I don't think we can pull all these ideas into one proposal.

design/baremetal-operator/host-operation.md
lentzi90: Sounds good! I think keeping the current HostOperation (no list) is preferable. It will be easier to extend by adding higher level objects than starting from a list with all the complications it brings.

lentzi90:
> I don't think we can pull all these ideas into one proposal.

Agreed! It is better to do it separately. Still wanted to mention it :slightly_smiling_face:

design/baremetal-operator/host-operation.md
Rozzii: Core Kubernetes has it also for terminated pod removal.
But we have discussed during Community meeting, if the time based solution can be done in a performance optimized way then I am fine with that too. Meaning after the OP is finished, the OP CR won't be reconciled until the TTL expires , as I understood from @dtantsur 's explanation, it is possible to implement like that.

design/baremetal-operator/host-operation.md
Rozzii: I am fine with **not** blocking by default.

design/baremetal-operator/host-operation.md
Rozzii: Has been discussed further, see https://github.com/metal3-io/metal3-docs/pull/709#discussion_r3731934102

lentzi90 (commented):
I don't see a reason to block this so putting lgtm.
If someone has concerns about the RBAC topic I raised (i.e do you see a need to control what type of operations people can make using RBAC), please shout!
/lgtm

Rozzii:
> I don't see a reason to block this so putting lgtm. If someone has concerns about the RBAC topic I raised (i.e do you see a need to control what type of operations people can make using RBAC), please shout! /lgtm

I am not concerned per say , but a lot of things have been discussed so I think it would be a good idea to clarify the sections that have been discussed. I mean clarify in the text so that  we won't have any doubt over what and why has been done later.

dtantsur:
I'll provide a revision once I have some time, hopefully early this week.

design/baremetal-operator/host-operation.md
dtantsur: Yeah, simply too much work already.

design/baremetal-operator/host-operation.md
dtantsur: I think we can assume a separate `hostClaim` field for simplicity.

design/baremetal-operator/host-operation.md
pierrecregut: @lentzi90 summary is even an improvement on what I had in mind. It is better to have one operation and several finalizers rather than several operations (and would solve @MahnoorAsghar concern).

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,516 @@
...
+- Support cancellation of running operations where possible.
+- Deprecate the plain `reboot.metal3.io` annotation.
+- Enable servicing without requiring a host reboot when the hardware
+  does not need one.
+
+### Non-Goals
+
+- Replacing the coordinated (suffixed) `reboot.metal3.io/<key>`
+  annotation. It requires external synchronization that is outside
+  the scope of this proposal.
```
lentzi90: Should we remove this from the non-goals? The halt operation is described as a replacement for this below.

design/baremetal-operator/host-operation.md
```diff
@@ -0,0 +1,516 @@
...
+
+- Running fast inspection as an operation to update HardwareData for a
+  provisioned instance.
+
+### Risks and Mitigations
+
+- **Orphaned operations**: if a BareMetalHost is deleted while
+  operations are pending, they become orphaned. Mitigation: the BMH
+  controller should set owner references on HostOperation resources
+  so that they are garbage-collected when the host is deleted.
```
lentzi90: I disagree. The controller should only put owner references on resources it creates. To me this garbage collection would be surprising. If I create the HostOperation, I expect to handle deletion of it also.

Could we instead consider using the TTL for this situation also? I'm thinking that a HostOperation without BMH should be considered immediately finished (failed) and can then be cleaned up with the usual TTL rules.

dtantsur: Good point.

PR #712, opened 2026-07-10
Design: support for OCI images for IPA downloader

diconico07:
This design proposal aims to support the distribution of the IPA image as an OCI image. This is based on the downstream work at SUSE for this and our implementation is here presented and linked as PoC.

I disclose the use of AI to write the initial proposal from the PoC, and while I tried to keep this proposal independent from the PoC (as I believe it can be largely improved) I kept it detailed to ensure the way it works and the process are well discussed here.


design/oci-ipa-downloader.md
```diff
@@ -0,0 +1,329 @@
...
+  downloader assumes a compliant single-layer OCI rootfs is provided.
+
+## Proposal
+
+We propose upgrading the `get-resource.sh` entrypoint of the
+`ironic-ipa-downloader` image to detect when `IPA_BASEURI` starts with the
+`oci://` scheme. When detected, the script will leverage an OCI client (such as
+`oras` or equivalent libraries) to fetch and parse the OCI manifest or index.
+
+The OCI image will typically pack the IPA root filesystem inside a single
```
dtantsur, line 93: I'm not super thrilled about having a tarball as an OCI artifact. Both DIB and upstream location distribute separate kernel and initramfs files, and I'd rather keep them as two artifacts that we handle. Doing anything else will require a complex procedure for users that want to use this feature.

dtantsur: We could have a single manifest index that references both kernel and initramfs distinguished by content type (or any other metadata). Whatever we invent, it should be easy to reproduce by any users that has the `oras` command line and the kernel/initramfs pair.

design/oci-ipa-downloader.md
```diff
@@ -0,0 +1,329 @@
...
+     pipeline specifically for wrapping files in a container image, rather
+     than natively supporting standard OCI rootfs images which are more aligned
+     with generic OS-container artifacts.
+1. **Distribute kernel and ramdisk as separate raw OCI artifacts**:
+   We could package the kernel and ramdisk directly as individual raw OCI
+   artifacts (rather than standard container/docker images containing a full
+   root filesystem). This approach is very clean and could eventually be
+   integrated directly into Ironic itself.
+   - *Why rejected for now*: Publishing and managing raw OCI artifacts
+     requires specialized tooling and registries that support them. Standard
```
dtantsur, line 307: `oras` and anything including quay.io and docker's registry.

design/oci-ipa-downloader.md
```diff
@@ -0,0 +1,329 @@
...
+     container/docker images containing a single rootfs layer are much more
+     widely supported by existing CI/CD pipelines, registries, and developer
+     toolchains. This proposal specifically aims to support environments where
+     publishing specialized raw OCI artifacts is not easily doable, making
+     standard container images containing the rootfs the most practical and
```
dtantsur, line 312: There is a load-bearing "standard" here :) For me, standard is two files: kernel and initramfs. This is what DIB outputs and this is what I can download from tarballs.opendev.org. Which standard are you referring to here?

design/oci-ipa-downloader.md
```diff
@@ -0,0 +1,329 @@
...
+   - *Risk*: Converting a large tar layer into a `newc` cpio archive and
+     compressing it via `zstd` is a CPU-intensive operation.
+   - *Mitigation*: This operation occurs exactly once per new image version
+     pulled. The strong digest-based caching layer ensures that on container
+     restarts, no extraction or compression occurs if the image digest is
+     unchanged.
+
+1. **Multi-layer OCI Images**:
+   - *Risk*: The extraction logic assumes a single-layer rootfs layout
+     (`.layers | length == 1`). Standard OCI images could contain multiple
```
dtantsur, line 211: I don't quite understand why you need to care about this, especially if you distribute a single tarball. ORAS just downloads whatever you point it at.

design/oci-ipa-downloader.md
diconico07: Well, here I build the IPA image using a Dockerfile that is (I simplify to keep it readable):
```dockerfile
FROM registry.suse.com/bci/bci-micro:16.0 as target
FROM registry.suse.com/bci/bci-base:16.0 as build
COPY --from=target / /target

RUN zypper --non-interactive --gpg-auto-import-keys --installroot /target install --no-recommends -- \
      kernel-default \
      kernel-firmware-all \
     openstack-ironic-python-agent

FROM scratch
COPY --from=build /target /
```

And this produces a single layer OCI image with the kernel in its usual position and the rest of the filesystem being what needs to be in the initramfs.
I know this is not the standard way of building IPA images, but it's the most convenient in my use-case.

design/oci-ipa-downloader.md
diconico07: "standard container image", means OCI image that you'd be able to `podman run` or `docker run`.

design/oci-ipa-downloader.md
dtantsur: Hmm, but cannot you extract the `openstack-ironic-python-agent` RPM without using Docker?

design/oci-ipa-downloader.md
dtantsur: Sure, but let's be more specific here. So far, most users haven't used a process like that to build IPA.

design/oci-ipa-downloader.md
```diff
@@ -0,0 +1,329 @@
...
+Currently, `ironic-ipa-downloader` retrieves IPA images exclusively from HTTP
+archives (by default, `tarballs.opendev.org`). While simple, this approach has
+several key limitations in modern enterprise environments:
+
+1. **Air-Gapped and Restricted Environments**: Many corporate and production
+   Kubernetes environments completely disallow raw HTTP/HTTPS traffic to the
+   external internet. These environments rely on internal, secured, and
+   audited OCI registries (such as Quay, Harbor, or Artifactory) to store and
+   distribute all software components, including operating system and
+   provisioning images.
```
Rozzii, line 39: IPA downloader is able to pull from custom address. I would not stress the air-gapped use case.

design/oci-ipa-downloader.md
Rozzii: It is clear to me that you @diconico07  have quite in depth understanding of OCI layering and all these things so please excuse my ignorance but could you please explain why my naive approach wouldn't work .

I would have approached this topic by just building a regular OCI container image (multy arch capable ofc) and then instead of tar layers and such I would have just flattened it after IPA downloader downloads it, something like this:
```
# By default, crane picks the platform matching the machine running it
crane export <image>:<tag> image_flat.tar | tar -x -C /tmp/rootfs

# Explicitly pick an architecture/OS
crane export --platform linux/arm64 <image>:<tag> image_flat.tar | tar -x -C /tmp/rootfs
crane export --platform linux/amd64 <image>:<tag> image_flat.tar | tar -x -C /tmp/rootfs

```
Ofc this can be done with other tool combinations but I would have just flattened the image where the root file system of the OCI image is the initramfs of the IPA. Then after I have the /tmp/rootfs or whatever location we extract, then I would just move the kernel out , maybe even directly to /shared/html/images/ironic-python-agent.kernel and then I would just use this to convert the the directory to initramfs:
I would do something like this
```
( cd /tmp/rootfs && find . -mindepth 1 | cpio -o -H newc ) | zstd -q -o /shared/html/images/ironic-python-agent.initramfs
```
ofc the specific command can be further improved I have used a few variants of this initramfs "recreation" workflow.

So basically my question is that what is the reason to spend so much effort avoiding layer flattening when there are multiple tools that can handle it out of the box? it can be done with skopeo and umoci, etc...

Rozzii (commented):
I would like to have oci support in ipa downloader but I have a few comments .

design/oci-ipa-downloader.md
dtantsur: Note that the benefit of using two artifacts without any sort of an envelop is that you'll be able to feed the same URL to Ironic directly without the downloader. Otherwise, we end up supporting two entirely different formats: one that Ironic can use (and IrSO can already pass in) and one that Metal3 understands and publishes.

diconico07: @Rozzii we can definitely do that, I expected the `crane export` to produce something similar to `docker export` so didn't consider it would match the need, for `skopeo`, when I checked it it was a lot slower and had issues when dealing with multi-architecture so I ended switching to a "simpler" approach. I should rephrase this part to say it's okay to have this kind of restrictions rather than enforcing it in the design.

@dtantsur I agree for the benefit of two artifacts, it would still need some kind of envelope (on manifest side I mean) to tell Ironic which one is initramfs and which one is kernel, and last time I checked Ironic wasn't able to take a tag for IPA kernel/initramfs

diconico07: @dtansur, thought again of the two artifacts situation, and I believe it could be beneficial to support a design akin to
```
 {
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.ironic.kernel.initramfs",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar",
      "digest": "sha256:d2a84f4b8b650937ec8f73cd8be2c74add5a911ba64df27458ed8229da804a26",
      "size": 12,
      "annotations": {
        "org.opencontainers.image.title": "ironic-python-agent.kernel"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar",
      "digest": "sha256:d2a84f4b8b650937ec8f73cd8be2c74add5a911ba64df27458ed8229da804a26",
      "size": 12,
      "annotations": {
        "org.opencontainers.image.title": "ironic-python-agent.initramfs"
      }
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2023-08-03T00:21:51Z"
  }
}
```

But ideally we'd want Ironic to be able to handle this kind of manifest as well (and have IPA-downloader to support it for fallback for older Ironic).
We may need a custom annotation instead of the `image.title` one to be more flexible about the filename, but this would make it simpler.
I also thought of using IrSO to do this part (resolve the tag and generate kernel and initramfs digests), but it makes it way more complicated as we would then need to properly handle proxies/authentication/TLS in IrSO.

PR #714, opened 2026-07-16
User guide: PreprovisioningNetworkData now supported by default

### LINKED ISSUE #2839: Extend documentation with details about provisioning without DHCP server
**What steps did you take and what happened:**

I would like to provision baremetal nodes using iDRAC VirtualMedia without DHCP during the provisioning step. So I am using the `preprovisioningNetworkData` field as described in the [documentation](https://book.metal3.io/bmo/advanced_instance_customization.html). This documentation states that we have to generate a custom IPA image with the `simple-init` element in order to support this use-case (or `cloud-init` maybe, the documentation is not very clear), because the default IPA image does not include an initialization tool. However, I am unable to generate a working IPA image supporting this use-case.

**What did you expect to happen:**

I would like to be able to provision nodes without DHCP as described in the documentation and be provided some guidance (preferably)... or the documentation should not indicate that a DHCP-less scenario is possible.

**Anything else you would like to add:**

In fact, I'm wondering if this advanced scenario is/was covered by automated tests. If some one-shot tests were performed, how did you generate the IPA image? Using the ironic python agent builder?

Because in fact, all the things I tried with the ironic python agent builder were broken. I opened an [issue](https://bugs.launchpad.net/ironic-python-agent-builder/+bug/2133509) in the ironic-python-agent-builder issue tracker, but I got not answer (yet). It seems that glean is broken (used in the `simple-init` element), and the `cloud-init` element is explicitly not supported by the ironic-python-agent-builder.

**Environment:**

- Baremetal Operator version: v0.11.1
- Environment (metal3-dev-env or other): the provisioning tools (BMO, Ironic) are running inside Kubernetes on a VM

/kind bug

dtantsur:
Tested in https://github.com/metal3-io/baremetal-operator/pull/3258

Fixes: metal3-io/baremetal-operator#2839

### DIFF
--- a/docs/user-guide/src/bmo/advanced_instance_customization.md
+++ b/docs/user-guide/src/bmo/advanced_instance_customization.md
@@ -13,10 +13,9 @@ for the deploy ramdisk running `ironic-python-agent` (IPA).
 
 Usage of this API requires an IPA ramdisk image with a tool capable of
 interpreting and applying the data such as *cloud-init*, *Glean* or
-alternative. The default community supported ramdisk does not currently contain
-such a tool, but it is possible to build a custom image, for example using
-[ironic-python-agent-builder][ipa_builder] with the [simple-init][simple_init]
-element enabled.
+alternative. The default community supported ramdisk uses [Glean][glean], which
+is a lightweight first-boot tool designed to provide a minimalistic alternative
+to *cloud-init*.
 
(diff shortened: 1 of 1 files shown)


Sunnatillo (commented):
LGTM

Rozzii (commented):
/approve

dtantsur:
We're a bit short on people, so

/approve

PR #725, opened 2026-08-26
Add "Security best practices" section to the user guide

migiFi:
Adds a new top-level "Security best practices" section to the user guide with a landing page that frames secure deployment across the three Metal3 trust boundaries (BMO/control plane, Ironic, and the provisioning network), and adds the corresponding SUMMARY.md entry.

**Why**:

The user guide had no security or hardening section, the only security-adjacent page was the vulnerability-disclosure policy. Hardening guidance existed but was scattered across component repos and never framed as best practice.

**NOTE**:

This is the landing page only. It links out to the current authoritative sources for each topic (TLS configuration, certificate lifecycle, Ironic authentication, TLS-enabled PXE, namespace-scoped BMO); dedicated pages for those will follow in separate PRs.

### DIFF
--- a/docs/user-guide/src/SUMMARY.md
+++ b/docs/user-guide/src/SUMMARY.md
@@ -68,5 +68,6 @@
 - [Try it (for developers)](developer_environment/tryit.md)
 - [API Reference](reference.md)
 - [Version Support](version_support.md)
+- [Security Best Practices](security/introduction.md)
 - [Project Security Policy](security_policy.md)
 - [Glossary](glossary.md)
--- a/docs/user-guide/src/security/introduction.md
+++ b/docs/user-guide/src/security/introduction.md
@@ -1,0 +1,78 @@
+# Security best practices
+
+Metal3 provisions and manages bare-metal machines, which means a Metal3
+deployment spans several trust boundaries at once: the Kubernetes control plane
+that runs [Bare Metal Operator](../bmo/introduction.md) (BMO), the
+[Ironic](../ironic/introduction.md) service that drives the hardware, and the
+provisioning network that carries DHCP, PXE/iPXE, image downloads, and
+out-of-band BMC traffic. Each of these boundaries can expose credentials, host
+firmware, and the operating systems written to disk, so each deserves
+deliberate hardening before a deployment leaves the lab.
+
+The introductory material in this guide, most notably the
+[quick-start](../quick-start.md), favors the shortest path to a working
+deployment. It uses self-signed or minimal TLS, an unauthenticated image
+server, and cluster-wide permissions. Those defaults are fine for evaluation
+but are **not** intended for production. This section collects the
+deployment-time practices that turn an evaluation setup into a hardened one.
+
+## What you are securing
+
+It helps to think about the deployment in terms of three planes, each with its
+own exposure:
+
+- **Control plane (BMO in Kubernetes).** BMO reads `BareMetalHost` resources and
+  the Kubernetes Secrets holding BMC credentials, and talks to the Ironic API.
+  Relevant concerns are how widely BMO's RBAC is scoped and how it
+  authenticates to Ironic.
+- **Provisioning service (Ironic).** Ironic exposes an API and serves boot
+  artifacts and disk images. Relevant concerns are TLS on the API and the
+  image/virtual-media HTTP servers, the cipher/protocol hardening applied to
+  them, and the basic-auth credentials protecting the API.
+- **Provisioning network (PXE/iPXE, DHCP, HTTP, BMC).** This is the path used to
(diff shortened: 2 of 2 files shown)


kashifest (commented):
/approve

kashifest:
/cc @tuminoid @lentzi90 @Rozzii 
/hold 
lets get more eyes on this before landing the PR.

docs/user-guide/src/SUMMARY.md
```diff
@@ -68,5 +68,6 @@
 - [Try it (for developers)](developer_environment/tryit.md)
 - [API Reference](reference.md)
 - [Version Support](version_support.md)
-- [Project Security Policy](security_policy.md)
+- [Security Best Practices](security/introduction.md)
+   - [Project Security Policy](security_policy.md)
```
lentzi90: This looks a bit strange to me. I would keep the security policy at the top level. It is quite a different topic than the security best practices.
(I mean just the indentation in this list, the files can stay where they are)

migiFi: Moved the security policy back to the top level so its not longer nested under Security Best Practices. Left the files where they are.

tuminoid:
I'll take a look next week, sorry for delay.

lentzi90 (commented):
Thanks, this looks good to me!
I'll leave the final lgtm to @tuminoid

docs/user-guide/src/security/introduction.md
```diff
@@ -0,0 +1,76 @@
...
+  [quick-start](../quick-start.md).
+- **Ironic authentication.** The Ironic `auth_strategy` modes (`noauth` versus
+  `http_basic`), where BMO reads the credentials, and the fact that OpenStack
+  Identity (Keystone) is not supported. See the
+  [Bare Metal Operator Authentication guide](https://github.com/metal3-io/baremetal-operator/blob/main/docs/ironic-authentication.md).
+- **TLS-enabled network boot (iPXE).** Running HTTPS iPXE end to end: building
+  an iPXE binary with an embedded trust anchor, how it ties into the httpd TLS
+  port, the tradeoffs, and the limitation that BIOS `undionly.kpxe` stays
+  HTTP/IPv4-only. See the
+  [ironic-image custom iPXE build](https://github.com/metal3-io/ironic-image/blob/main/README.md)
```
Rozzii: This link does not really elaborate about custom iPXE building , I suggest linking this instead https://github.com/metal3-io/utility-images#ipxe-builder-image

https://github.com/metal3-io/ironic-image/blob/main/build-ipxe.sh and https://github.com/metal3-io/ironic-image/blob/main/README.md "describs" the default "built in" iPXE firmware creation

migiFi: Thank you! I switched to the ipxe-builder image for the custom build, and kept build-ipxe.sh/README as references for the default built-in firmware.

docs/user-guide/src/security/introduction.md
```diff
@@ -0,0 +1,79 @@
...
+  `disableCertificateVerification` on a `BareMetalHost`. See
+  [Install Ironic with IrSO](../irso/install-basics.md) and the
+  [quick-start](../quick-start.md).
+- **Ironic authentication.** The Ironic `auth_strategy` modes (`noauth` versus
+  `http_basic`), where BMO reads the credentials, and the fact that OpenStack
+  Identity (Keystone) is not supported. See the
+  [Bare Metal Operator Authentication guide](https://github.com/metal3-io/baremetal-operator/blob/main/docs/ironic-authentication.md).
+- **TLS-enabled network boot (iPXE).** Running HTTPS iPXE end to end: building
+  an iPXE binary with an embedded trust anchor, how it ties into the httpd TLS
+  port, the tradeoffs, and the limitation that BIOS `undionly.kpxe` stays
```
tuminoid: https://github.com/metal3-io/utility-images/blob/19c017d7e8ef2d0e3999ccf74a2e5cafc66ec061/ipxe-builder/buildipxe.sh#L71-L93

This does have ipv6 and TLS options, and AFAIK it does built both binaries to support them. Am I reading this wrong? @migiFi @Rozzii

tuminoid (commented):
One question, otherwise LGTM.

docs/user-guide/src/security/introduction.md
migiFi: Not reading wrong, you are right. I didnt catch that, but yet `buildipxe.sh` builds both undionly.kpxe` and `snponly.efi` with the same IPv6/TLS options, so undionly is not HTTP/IPv4-only.

Dropped that claim and I will cover the behavior in the dedicated iPXE page.

tuminoid (commented):
/lgtm

Thanks for kicking this off!

PR #726, opened 2026-08-26
Add TLS configuration and hardening page to the user guide

migiFi:
Adds a "TLS Configuration and Hardening" page under the user guide's Security section. It consolidates the Ironic TLS tuning knobs that were previously split between the ironic-image README and the IrSO deployment, explains the two TLS surfaces (Ironic API vs. virtual-media HTTP server), documents the protocol/ cipher/curve flags for each, and gives a recommended hardened baseline.
  
Also links the Security landing page's TLS bullet to this new page and adds the
SUMMARY entry.

**Why**:
  
The TLS hardening flags (below) were only documented in the ironic-image README, with nothing in the user guide tying them together or recommending hardened values.

- `IRONIC_SSL_PROTOCOL`
- `IRONIC_TLS_12/13_CIPHERS`
- `IRONIC_TLS_CURVES`
- `IRONIC_TLS_ENFORCE_SERVER_CIPHER_ORDER`
- `IRONIC_VMEDIA_*` (and equivalents) 

**Depends on #725 and should merge after it. /hold until then.**

**NOTE**: until 725 merges, the diff includes its commits. Review only the "Add TLS configuration and hardening page" commit here.

### DIFF
--- a/docs/user-guide/src/security/tls.md
+++ b/docs/user-guide/src/security/tls.md
@@ -1,0 +1,165 @@
+<!-- cSpell:ignore Apache -->
+# TLS configuration and hardening
+
+Transport Layer Security (TLS, [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446))
+encrypts and authenticates the network connections to Ironic's services. Ironic
+in Metal3 terminates TLS on more than one TCP port (see
+[RFC 6335](https://www.rfc-editor.org/rfc/rfc6335) for the meaning of "port").
+The Ironic Standalone Operator (IrSO) controls whether TLS is enabled and which
+certificates are used, while the protocol, cipher and curve tuning lives in the
+ironic-image container environment. This page brings both together: which ports
+exist, how to enable TLS on each, the flags that tune the protocol versions,
+cipher suites and curves, and a recommended hardened baseline.
+
+Running Ironic without TLS is not recommended. Enabling it is covered below and
+in [Install Ironic with IrSO](../irso/install-basics.md); supplying and rotating
+the certificates themselves is a separate concern from the cipher/protocol
+tuning described here.
+
+## The TLS ports
+
+A Metal3 Ironic deployment exposes several externally reachable TLS ports, each
+served by Apache with its own certificate and its own set of tuning flags:
+
+- **Ironic API (TCP `6385`).** The REST API that Bare Metal Operator (BMO) and
+  administrators talk to. Apache terminates TLS on this port and proxies
+  requests to the Ironic API server over a pod-local unix socket
+  (`/shared/ironic.sock`), so there is no internal TCP connection to Ironic. The
+  exception is the `/images` path, which Apache serves directly from local
+  storage rather than proxying — this is how the OS/node images that the Ironic
+  Python Agent (IPA) downloads during provisioning are hosted, over the same
+  port and certificate. Clients of this port are BMO, the IPA running in the
+  ramdisk, and any operator tooling.
+- **Virtual-media image server (TCP `6183` for HTTPS, `6180` for plain HTTP).**
+  The HTTP server that presents boot and configuration images to a host's BMC
+  when provisioning over virtual media. It has its own certificate slot and its
+  own tuning flags, though under IrSO it is populated with the same certificate
+  as the API; its clients are the BMCs themselves, whose TLS support varies
+  widely between vendors and firmware versions.
+
+Because the clients differ, the ports are tuned independently. The API port
+faces software you control (BMO, IPA) and can be hardened aggressively. The
+virtual-media port faces BMC firmware you often do not control, so overly strict
+settings there can break provisioning on older hardware.
+
+Network boot (iPXE) over HTTPS is a third TLS port with its own certificate and
+flags. It is out of scope for this page and covered with the network-boot
+material instead.
+
+## Enabling TLS
+
+With IrSO, TLS is enabled by referencing a certificate secret from the `Ironic`
+resource:
+
+```yaml
+spec:
+  tls:
+    certificateName: ironic-tls
+```
+
+The certificate must include the DNS name derived from the service (for example
+`ironic.test-ironic.svc`) and the IP address the hosts use to reach Ironic, as a
+Subject Alternative Name. See
+[Install Ironic with IrSO](../irso/install-basics.md) for a complete self-signed
+test example, including generating the secret and its required label. In
+production, use a certificate manager such as
+[cert-manager](https://cert-manager.io/) to issue and rotate the certificate.
+
+IrSO also enables TLS on the virtual-media image server by default, reusing the
+same certificate you supply via `spec.tls.certificateName` rather than a
+separately managed one; set `spec.tls.disableVirtualMediaTLS` to turn it off for
+BMCs that cannot use HTTPS.
+
+The hardening flags below are environment variables consumed by the
+ironic-image container's Apache configuration, not Ironic (`ironic.conf`)
+options. IrSO manages the certificates through `spec.tls` but does **not**
+currently expose these protocol, cipher and curve flags as dedicated fields,
+and `spec.extraConfig` only sets `ironic.conf` options, which is a different
+layer. As a result, an IrSO-managed deployment runs with the ironic-image
+defaults listed below unless you inject the environment variables yourself by
+replacing the `httpd` container through the experimental `spec.overrides`
+field. Deployments that build the Ironic pod's environment directly (for
+example the classic script-based install) can set them as additional `IRONIC_*`
+environment variables alongside the others.
+
+## API TLS flags
+
+These environment variables tune the Ironic API port (and the HTTPS image
+server that shares its certificate):
+
+| Variable | Purpose | Default |
+|---|---|---|
+| `IRONIC_SSL_PROTOCOL` | Allowed TLS protocol versions (Apache `SSLProtocol` directive) | `-ALL +TLSv1.2 +TLSv1.3` |
+| `IRONIC_TLS_12_CIPHERS` | Ordered cipher list for TLS 1.2 and below, in OpenSSL format | OpenSSL default |
+| `IRONIC_TLS_13_CIPHERS` | Ordered cipher list for TLS 1.3, in OpenSSL format | OpenSSL default |
+| `IRONIC_TLS_CURVES` | Ordered list of allowed groups/curves, in OpenSSL format | OpenSSL default |
+| `IRONIC_TLS_ENFORCE_SERVER_CIPHER_ORDER` | Make the server, not the client, choose the cipher (TLS 1.2 and below) | `false` |
+
+## Virtual-media TLS flags
+
+These environment variables tune the virtual-media HTTP server. They mirror the
+API flags but apply to the separate virtual-media server (which, under IrSO, uses
+the same certificate as the API):
+
+| Variable | Purpose | Default |
+|---|---|---|
+| `IRONIC_VMEDIA_SSL_PROTOCOL` | Allowed TLS protocol versions (Apache `SSLProtocol` directive) | `ALL` |
+| `IRONIC_VMEDIA_TLS_12_CIPHERS` | Ordered cipher list for TLS 1.2 and below, in OpenSSL format | OpenSSL default |
+| `IRONIC_VMEDIA_TLS_13_CIPHERS` | Ordered cipher list for TLS 1.3, in OpenSSL format | OpenSSL default |
+| `IRONIC_VMEDIA_CURVES` | Ordered list of allowed groups/curves, in OpenSSL format | OpenSSL default |
+| `IRONIC_VMEDIA_TLS_ENFORCE_SERVER_CIPHER_ORDER` | Make the server choose the cipher (TLS 1.2 and below) | `false` |
+
+Note that `IRONIC_VMEDIA_SSL_PROTOCOL` defaults to `ALL`, which permits every
+protocol version the underlying OpenSSL build supports, including legacy ones.
+This is deliberately permissive because BMC TLS stacks are often old, but it
+means the virtual-media port is *not* hardened by default. Tightening it is
+worthwhile, but do so carefully and test against your hardware (see the caveats
+below).
+
+## Recommended hardened baseline
+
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
/hold

Lets hold this for a while until @Rozzii also gets a view. We also need to start from introduction section etc, then Metal3 hardening before we move further.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,135 @@
...
+tuning described here.
+
+## The two TLS surfaces
+
+A Metal3 Ironic deployment exposes two independent TLS surfaces, each with its
+own certificate and its own set of tuning flags:
+
+- **Ironic API.** The REST API that Bare Metal Operator (BMO) and administrators
+  talk to, on TCP `6385`. The same certificate is also used to serve OS and
+  virtual-media images over HTTPS on TCP `6183` (plain HTTP is `6180`). Clients
```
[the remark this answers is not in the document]
migiFi: Fixed, API surface has its own cert and alsio serves node/OS images, virtual media uses a separate cert.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,135 @@
...
+in production.
+
+The hardening flags below are environment variables consumed by the
+ironic-image container. IrSO manages the certificate through
+`spec.tls.certificateName` but does not expose these tuning flags as dedicated
+fields: `spec.extraConfig` sets Ironic configuration options (`ironic.conf`),
+which is a different layer from the Apache-level TLS settings. In deployments
+that configure the Ironic pod's environment directly (for example the classic
+script-based install and its configuration environment), set them as additional
+`IRONIC_*` environment variables alongside the others.
```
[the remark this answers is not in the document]
migiFi: Made explicit that IrSO doesnt expose these flags, they settable where you control the pod env.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,135 @@
...
+
+These environment variables tune the virtual-media HTTP server. They mirror the
+API flags but apply to the separate virtual-media certificate and server:
+
+| Variable | Purpose | Default |
+|---|---|---|
+| `IRONIC_VMEDIA_TLS_12_CIPHERS` | Ordered cipher list for TLS 1.2 and below, in OpenSSL format | OpenSSL default |
+| `IRONIC_VMEDIA_TLS_13_CIPHERS` | Ordered cipher list for TLS 1.3, in OpenSSL format | OpenSSL default |
+| `IRONIC_VMEDIA_CURVES` | Ordered list of allowed groups/curves, in OpenSSL format | OpenSSL default |
+| `IRONIC_VMEDIA_TLS_ENFORCE_SERVER_CIPHER_ORDER` | Make the server choose the cipher (TLS 1.2 and below) | `false` |
```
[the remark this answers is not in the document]
migiFi: Added `IRONIC_VMEDIA_SSL_PROTOCOL` (default `ALL`) to the table and baseline, with a note that vmedia is not hardened by default.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,135 @@
...
+  keep TLS 1.2 suites) while leaving the API surface strict.
+- **TLS 1.3 has no server-side ordering knob.** The
+  `*_ENFORCE_SERVER_CIPHER_ORDER` flags apply to TLS 1.2 and below; TLS 1.3
+  cipher selection is governed by the client per the protocol.
+- **Restricting to TLS 1.3 only** (for example `IRONIC_SSL_PROTOCOL="-ALL
+  +TLSv1.3"`) is stronger but requires every client of that surface to support
+  TLS 1.3. This is more realistic for the API than for the virtual-media server.
+- **Validate the result** with a tool such as `openssl s_client` or a TLS
+  scanner against ports `6385`/`6183` (API) and `6180`/`6183` (virtual media)
+  after applying changes.
```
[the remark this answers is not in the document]
migiFi: Fixed, now scans 6385 and 6183 only, noting 6180 is plain HTTP.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,135 @@
...
+  tls:
+    certificateName: ironic-tls
+```
+
+The certificate must include the DNS name derived from the service (for example
+`ironic.test-ironic.svc`) and the IP address the hosts use to reach Ironic, as a
+Subject Alternative Name. See
+[Install Ironic with IrSO](../irso/install-basics.md) for a complete example,
+including generating the secret and using [cert-manager](https://cert-manager.io/)
+in production.
```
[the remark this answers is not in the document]
migiFi: Reworded: install-basics linked as the self-signed test example, cert-manager recommended for production.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
+<!-- cSpell:ignore Apache -->
+# TLS configuration and hardening
+
+Ironic in Metal3 terminates TLS on more than one network surface. The Ironic
```
Rozzii:
```suggestion
Ironic in Metal3 terminates TLS on more than one port. The Ironic
```

Rozzii: You can also reference  RFC 6335  to make sure everyone is on the same page about "port".

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
+<!-- cSpell:ignore Apache -->
+# TLS configuration and hardening
+
```
Rozzii, line 3: Just mention in one sentence what tls is with a reference to the standard, just to be exact. RFC 8446

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
...
+Running Ironic without TLS is not recommended. Enabling it is covered below and
+in [Install Ironic with IrSO](../irso/install-basics.md); supplying and rotating
+the certificates themselves is a separate concern from the cipher/protocol
+tuning described here.
+
+## The two TLS surfaces
+
+A Metal3 Ironic deployment exposes two externally reachable TLS surfaces, each
+served by Apache with its own certificate and its own set of tuning flags:
```
Rozzii, line 22: Mention here that Ironic talks to the Apache via unix socket so there is no internal networking. TLS traffic is terminated by the Apache server but the traffic is further transmitted via "pod local" unix socket to ironic API server.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
...
+  administrators talk to. The same listener and certificate also serve the
+  OS/node images that the Ironic Python Agent (IPA) downloads during
+  provisioning, so the clients of this surface are BMO, the IPA running in the
+  ramdisk, and any operator tooling.
+- **Virtual-media image server (TCP `6183` for HTTPS, `6180` for plain HTTP).**
+  The HTTP server that presents boot and configuration images to a host's BMC
+  when provisioning over virtual media. It has its own certificate slot and
+  its own tuning flags, though under IrSO it is populated with the same
+  certificate as the API; its clients are the BMCs themselves, whose TLS support
+  varies widely between vendors and firmware versions.
```
Rozzii: There is a third endpoint for TLS enabled iPXE default 8084, same way it has its own apache config file etc.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
+<!-- cSpell:ignore Apache -->
+# TLS configuration and hardening
+
+Ironic in Metal3 terminates TLS on more than one network surface. The Ironic
+Standalone Operator (IrSO) controls whether TLS is enabled and which
```
Rozzii: This reads like IrSO is mandatory to handle these things, I would first explain what is supported by ironic-image then in a separate section explain the config options on higher abstraction level (IrSO) .

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
...
+in [Install Ironic with IrSO](../irso/install-basics.md); supplying and rotating
+the certificates themselves is a separate concern from the cipher/protocol
+tuning described here.
+
+## The two TLS surfaces
+
+A Metal3 Ironic deployment exposes two externally reachable TLS surfaces, each
+served by Apache with its own certificate and its own set of tuning flags:
+
+- **Ironic API (TCP `6385`).** The REST API that Bare Metal Operator (BMO) and
```
Rozzii, line 24: Mention please that although the port is same but for image hosting there is a dedicated "/images" url path that is directly manged by the apache unlike the rest of the endpoint that reach ironic API server.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
...
+  provisioning, so the clients of this surface are BMO, the IPA running in the
+  ramdisk, and any operator tooling.
+- **Virtual-media image server (TCP `6183` for HTTPS, `6180` for plain HTTP).**
+  The HTTP server that presents boot and configuration images to a host's BMC
+  when provisioning over virtual media. It has its own certificate slot and
+  its own tuning flags, though under IrSO it is populated with the same
+  certificate as the API; its clients are the BMCs themselves, whose TLS support
+  varies widely between vendors and firmware versions.
+
+Because the clients differ, the two surfaces are tuned independently. The API
```
Rozzii:
```suggestion
Because the clients differ, the three  ports are tuned independently. The API
```

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
...
+surface faces software you control (BMO, IPA) and can be hardened aggressively.
+The virtual-media surface faces BMC firmware you often do not control, so
+overly strict settings there can break provisioning on older hardware.
+
+Network boot (iPXE) over HTTPS is a third TLS surface with its own certificate
```
Rozzii: Shouldn't be out of scope it is just as important, it exists on the same abastraction level managed by the same apache server.

docs/user-guide/src/security/tls.md
```diff
@@ -0,0 +1,157 @@
...
+  cipher selection is governed by the client per the protocol.
+- **Restricting to TLS 1.3 only** (for example `IRONIC_SSL_PROTOCOL="-ALL
+  +TLSv1.3"`, or `IRONIC_VMEDIA_SSL_PROTOCOL` for virtual media) is stronger but
+  requires every client of that surface to support TLS 1.3. This is realistic
+  for the API but often not for the virtual-media server, given the age of many
+  BMC TLS stacks.
+- **Validate the result** with a tool such as `openssl s_client` or a TLS
+  scanner against port `6385` (Ironic API, which also serves node images) and
+  port `6183` (virtual-media TLS server) after applying changes. Port `6180` is
+  the plain-HTTP virtual-media listener and has no TLS to scan.
```
Rozzii, line 165: IPA can't do onky TLS 1.3. Custom IPA builds could achieve restricted  TLS 1.3 only support mode depending on the host os, and software stack in the IPA but natively IPA can't be forced to use only TLS 1.3.

iPXE as of this moment has 0 support for TLS 1.3, it supports TLS up to 1.2.

Rozzii (commented):
Generally looks good but iPXE has to be included here IMO, also make sure to use proper technical terms instead of generic terms like "surface". IMO security issues are always in the implementation details so phrasing here has to be very specific.

docs/user-guide/src/security/tls.md
migiFi: We had a separate iPXE page planned as a follow-up (the firmware-build and deploy walkthrough), which is why I marked this out of scope. My plan is to document the iPXE TLS port and its flags here, and keep the separate page just for the firmware-building/deploy guide. 

Does that split work, or would you rather have all of iPXE in one place?

docs/user-guide/src/security/tls.md
migiFi: added it

docs/user-guide/src/security/tls.md
migiFi: Added the RFC reference and change surface to port

docs/user-guide/src/security/tls.md
migiFi: Noted Apache terminates TLS and reaches the Ironic API over a pod-local unix socket

docs/user-guide/src/security/tls.md
migiFi: Fixed: API keeps 1.2 (IPA can't be 1.3-only natively), and iPXE has no 1.3 support (max 1.2).

docs/user-guide/src/security/tls.md
migiFi: done, the API bullet now notes /images is serve directly by Apache, while other paths are proxied to the Ironic API server.

docs/user-guide/src/security/tls.md
migiFi: depends if we split the iPXE pages.

Rozzii: IMO this doc talks about TLS enabled ports of the ironic pod in general, in that context we have to include the port used by iPXE.

docs/user-guide/src/security/tls.md
Rozzii: I don't mean that you have to explain how to configure iPXE for ironic, or how to build different iPXE firmwares, but I would like to see the iPXE specifc apache TLS endpoint config mentioned here same as the vmedia and the generic node image endpoints.

docs/user-guide/src/security/tls.md
migiFi: added port 8084, it now sits alongside the API and vmedia ports, with its Apache vhost, cert, and the one protocol flag it has.

docs/user-guide/src/security/tls.md
migiFi: Reworked this so the ironic-image comes first and IrSO is its own section now. Thank you, makes it much clearer now.

docs/user-guide/src/security/tls.md
migiFi: Endpoint config is now documented here. 

Kept only the firmware-build guide out of the scope and dropped the "out of scope" note.

PR #732, opened 2026-09-07
Add Ironic authentication page to the user guide

migiFi:
Adds an "Ironic Authentication" page under the user guide's Security section, consolidating auth guidance that was previously only in the basemetal-operator repo docs.

This PR covers the supported `auth_strategy` modes (`noauth` vs. `http_basic`, and that Keystone is unsupported), the server and client sides (how ironic-image enables `http_basic` and where BMO reads credential via `METAL3_AUTH_ROOT_DIR`), the recommended IrSO path (auto-generated API credentials, `spec.apiCredentialsName`), and the manual `IRONIC_BASIC_AUTH` route.

Also links the Securitty landing page's authentication bullet to the new page and adds the SUMMARY entry.

Depends on #726 and should merge after it. /hold until then.

Until the above PR merges and this is rebase the link check will flag the security/tls.md link as broken, expected, resolved by the rebase.

Welcome any feedback on things I might have missed.

### DIFF
--- a/docs/user-guide/src/security/ironic-authentication.md
+++ b/docs/user-guide/src/security/ironic-authentication.md
@@ -1,0 +1,53 @@
+<!-- cSpell:ignore noauth Keystone -->
+# Ironic authentication
+
+Hosts under Metal3's control contact the Ironic API during inspection and
+provisioning, and they remain connected to the provisioning network while
+running user workloads afterwards. If that API is unauthenticated, anyone with
+access to the provisioning network can use it to manipulate your nodes.
+Requiring authentication on the Ironic API is therefore strongly advised.
+Authentication is independent of TLS (covered in
+[TLS configuration and hardening](./tls.md)); a hardened deployment uses both.
+
+## Authentication strategy
+
+Metal3 authenticates the Ironic API with HTTP Basic Authentication, and Bare
+Metal Operator (BMO) is configured with matching credentials to reach Ironic.
+
+Ironic also has a `noauth` (no-authentication) mode, but it is discouraged
+and is expected to be phased out as part of the Ironic Standalone Operator
+migration, so you should not rely on it. OpenStack Identity (Keystone)
+authentication is not supported.
+
+## How it fits together
+
+Authentication has two sides that must agree: Ironic is configured to
+require credentials, and BMO is configured to present matching ones.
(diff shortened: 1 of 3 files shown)


docs/user-guide/src/security/ironic-authentication.md
```diff
@@ -0,0 +1,65 @@
...
+Requiring authentication on the Ironic API is therefore strongly advised.
+Authentication is independent of TLS (covered in
+[TLS configuration and hardening](./tls.md)); a hardened deployment uses both.
+
+## Supported authentication strategies
+
+Bare Metal Operator (BMO) supports connecting to an Ironic configured with one
+of two `auth_strategy` modes:
+
+- `noauth` — no authentication. This is the default when no auth configuration
```
dtantsur: I would like to deprecate `noauth` as part of IrSO migration, so I'm not sure we should document it as a supported option.

docs/user-guide/src/security/ironic-authentication.md
```diff
@@ -0,0 +1,65 @@
...
+## How it fits together
+
+Authentication has a server side (Ironic) and a client side (BMO):
+
+- **Server (Ironic).** With `http_basic`, Ironic authenticates API callers
+  against an htpasswd file (`auth_strategy = http_basic` and
+  `http_basic_auth_user_file` in `ironic.conf`). In the ironic-image container
+  this is enabled automatically when a credential is supplied — either a
+  pre-computed htpasswd mounted at `/auth/ironic/htpasswd` or the
+  `HTTP_BASIC_HTPASSWD` environment variable.
```
dtantsur: These are internal implementation details, do we actually need them in the user guide? Looks like something for developer docs.

docs/user-guide/src/security/ironic-authentication.md
migiFi: done, reworded so HTTP Basic is the documented strategy; noauth now only noted as discouraged/to be phased out.

docs/user-guide/src/security/ironic-authentication.md
migiFi: Agree on this, I dropped the httpasswd/env/file-layout details, kept the concept, and linked to the BMO auth guide.

PR #733, opened 2026-09-08
Bump the pinned markdownlint-cli2 container image from to 0.14.19

Ahmedaltu:
Bump the pinned markdownlint-cli2 container image from to 0.14.19

Part of https://github.com/metal3-io/project-infra/issues/1531


.markdownlint-cli2.yaml
```diff
@@ -12,3 +12,7 @@ config:
 
 # Don't autofix anything, we're linting here
 fix: false
+
+# CLAUDE.md is a pointer file for AI tooling, not prose documentation
+ignores:
+- "CLAUDE.md"
```
lentzi90, line 18: CLAUDE.md is a link to AGENTS.md. So why should we ignore it but not the source?
Since the linter passed on the PR, I take it there are no issues found in AGENTS.md, right?
If that is the case, let's not ignore it. We can keep it nice and tidy with passing linter.

Ahmedaltu: @lentzi90 

I tested it directly against the pinned 0.14.19 linter without the ignore, and it does fail:

CLAUDE.md:1 MD041/first-line-heading/first-line-h1 First line in a file should be a top-level heading [Context: "AGENTS.md"]
CLAUDE.md:1:9 MD047/single-trailing-newline Files should end with a single newline character

However, AGENT.md passes because it is a real structured doc with a heading and proper content.

lentzi90: Ah it reads the content of the link then instead of following it.
Then please update the comment to explain that CLAUDE.md is just a link to AGENTS.md.

Ahmedaltu: done!

lentzi90 (commented):
/approve

PR #737, opened 2026-09-23
Update node image URLs and versions

peppi-lotta:
Updates the node image source, node image versions, and related component versions used across the user guide and the quick-start hack scripts. The node images are now pulled from the new OCI object storage bucket (`public-metal3-node-image-bucket`) instead of the Nordix artifactory, and the CentOS/Kubernetes/Calico versions are bumped to current releases.


peppi-lotta:
Hmm... test runs successfully in my VM but fails here. Difficult to debug

lentzi90 (commented):
Could have been just a one time issue. It got all the way to deploying calico so we are at least close.
Retitle to trigger the test again
/retitle Update node image URLs and versions

lentzi90 (commented):
Test passed!
/approve

peppi-lotta:
> Test passed! /approve

I had saw test fail to same error 3 times in a row. I added a commit with logging to debug the issue before I noticed the passed test. Test seems to be flaky... I'll revert the logs commit and move it to a separate PR because I think it's a good thing to have anyways. I'll let the test run with the logging just in case the flake happens again. 

3 fails:
https://github.com/metal3-io/metal3-docs/actions/runs/35848104807/job/107139039512
https://github.com/metal3-io/metal3-docs/actions/runs/35844762519/job/107128118464
https://github.com/metal3-io/metal3-docs/actions/runs/35835459087/job/107097714229

